PR #4743, opened 2022-04-18
Tests for PR4740 

navytux:
Add tests for various unicode methods with start/end parameters where None means "default". Suggested by @scoder  at https://github.com/cython/cython/issues/4737#issuecomment-1100174184.


tests/run/unicodemethods.pyx
```diff
@@ -26,16 +26,12 @@ def print_all(l):
...
-    sadas
-    asdas
-    fsdf
-    >>> print_all( split(text) )
+    >>> def test_split():
+    ...     py = text.split()
+    ...     cy = split(text)
+    ...     assert py == cy, (py, cy)
+    ...     return cy
+    >>> print_all( test_split() )
```
scoder, line 34: Instead of returning something, I think we can just move the call to `print_all()` into the test helper function as well.

The same applies to the `print()` calls in other tests. I think the only reason why we call `print()` explicitly is the 'u' prefix of unicode strings in Py2, which prevents a straight doctest output comparison.

scoder (commented):
Cool, thanks!

tests/run/unicodemethods.pyx
navytux: @scoder, thanks for feedback. I believe it is better to keep `print*` outside. Let me explain: `test_X` function verifies method X for consistency in between Py and Cy implementation and allows the test to verify the actual result. Some results are strings, some boolean, some integers. We use `print`/`print_all` only for strings and list of strings kinds of result. We do not use print for boolean and integer results.

So for consistency I suggest to keep all `test_*` functions as close as possible to each other: retrieve py and cy values, assert that they are the same and return that result for inspection. Then in the tests themselves it will be always clear what to expect from `test_X`: it does call `unicode.X` and returns the result.

Do you think that makes sense?

navytux:
( fixed one small overlook in deduplication patch )

scoder:
Thanks @navytux

navytux:
You are welcome, @scoder. And thanks for merging this.

PR #4751, opened 2022-04-20
Add STL `sample` for sampling without replacement.

tillahoffmann:
This PR adds the STL [`sample`](https://en.cppreference.com/w/cpp/algorithm/sample) function to sample without replacement.

### DIFF
--- a/tests/run/cpp_stl_algo_sample.pyx
+++ b/tests/run/cpp_stl_algo_sample.pyx
@@ -1,0 +1,40 @@
+# mode: run
+# tags: cpp, cpp17
+
+from libcpp.algorithm cimport sample
+from libcpp.iterator cimport back_inserter
+from libcpp.random cimport mt19937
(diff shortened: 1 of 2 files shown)


tests/run/cpp_stl_algo_sample.pyx
```diff
@@ -0,0 +1,39 @@
...
+    >>> sample = sample_multiple(10, 7)
+    >>> len(sample), len(set(sample))  # Check sampling without replacement.
+    (7, 7)
+    """
+    cdef:
+        vector[int] x, y
+        int i
+        mt19937 rd = mt19937(1)
+
+    [x.push_back(i) for i in range(population_size)]
```
scoder, line 22: Throwing away the result of a list comprehension is a bit of a funny spelling for a for-loop. I'd rather see the for-loop to avoid making readers think about the effect of this.

PR #4774, opened 2022-05-08
manylinux CI: Update used actions, use Python 3.10

EwoutH:
A bit of maintenance on the manylinux CI:
- Update actions/checkout to v3, actions/setup-python to v3 and actions/upload-artifact to v3. Those version are currently maintained and use NodeJS 16.
- Use Python 3.10 for sdist and wheel building jobs


scoder:
Thanks. This isn't tested in CI, because it's a release workflow. Looks ok, though, so I'll merge it and see how the next release goes. :)

PR #4776, opened 2022-05-09
CI: Updated used actions, allow workflow_dispatch

EwoutH:
A bit of maintenance on the regular CI.
- Updates the used actions to their latest versions (v3)
  _(note that for actions/checkout, the `fetch-depth` is now 1 by default)_
- Allow triggering a run manually from the GitHub UI by adding `workflow_dispatch` (see [GitHub docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch))
- Run pycoverage job with Ubuntu 20.04 and Python 3.10

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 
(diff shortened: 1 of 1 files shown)


scoder:
Thanks!

PR #4794, opened 2022-05-15
dataclasses: enable kw_only dataclasses

Yurzs:
Enables possibility to use kw_only option like in `dataclasesses.dataclass` requested in https://github.com/cython/cython/issues/4303

### DIFF
--- a/Cython/Compiler/Dataclass.py
+++ b/Cython/Compiler/Dataclass.py
@@ -206,7 +206,8 @@ def process_class_get_fields(node):
 def handle_cclass_dataclass(node, dataclass_args, analyse_decs_transform):
     # default argument values from https://docs.python.org/3/library/dataclasses.html
     kwargs = dict(init=True, repr=True, eq=True,
-                  order=False, unsafe_hash=False, frozen=False)
+                  order=False, unsafe_hash=False,
+                  frozen=False, kw_only=False)
     if dataclass_args is not None:
         if dataclass_args[0]:
             error(node.pos, "cython.dataclasses.dataclass takes no positional arguments")
@@ -220,6 +221,8 @@ def handle_cclass_dataclass(node, dataclass_args, analyse_decs_transform):
             kwargs[k] = v
 
     fields = process_class_get_fields(node)
+
+    kw_only = kwargs.pop("kw_only")
 
     dataclass_module = make_dataclasses_module_callnode(node.pos)
 
@@ -250,7 +253,7 @@ def handle_cclass_dataclass(node, dataclass_args, analyse_decs_transform):
     code_lines = []
     placeholders = {}
     extra_stats = []
-    for cl, ph, es in [ generate_init_code(kwargs['init'], node, fields),
+    for cl, ph, es in [ generate_init_code(kwargs['init'], node, fields, kw_only),
                         generate_repr_code(kwargs['repr'], node, fields),
                         generate_eq_code(kwargs['eq'], node, fields),
                         generate_order_code(kwargs['order'], node, fields),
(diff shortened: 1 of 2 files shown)


da-woods:
That looks reasonable. It does at least one test to show it works in tests/run/cdef_class_dataclasses.pyx

tests/run/cdef_class_dataclass.pyx
```diff
@@ -224,6 +224,25 @@ cdef class TestFrozen:
     """
     a: double = 2.0
 
+@dataclasses(kw_only=True)
```
da-woods, line 227:
```suggestion
@dataclass(kw_only=True)
```

You can run the tests in this file locally on your PC with `python runtest.py cdef_class_dataclass` just to confirm that it works.

da-woods:
This looks good to me. I think we might have to rewrite it a bit if/when we support `kw_only` on fields and the `KW_ONLY` placeholder value, but I think this is a worthwhile improvement as is.

Yurzs:
@da-woods should i rebase now or wait?

tests/run/cdef_class_dataclass.pyx
```diff
@@ -224,6 +224,25 @@ cdef class TestFrozen:
...
+    3.0
+    >>> inst.b
+    2
+    >>> fail = TestKwOnly(3, 2)
+    Traceback (most recent call last):
+    TypeError: __init__() takes exactly 0 positional arguments (2 given)
+    >>> fail = TestKwOnly(a=3)
+    Traceback (most recent call last):
+    TypeError: __init__() needs keyword-only argument b
+    """
```
scoder, line 241: Let's also test that default arguments are correctly handled.
```suggestion
    >>> fail = TestKwOnly(a=3)
    Traceback (most recent call last):
    TypeError: __init__() needs keyword-only argument b
    >>> fail = TestKwOnly()
    Traceback (most recent call last):
    TypeError: __init__() needs keyword-only argument b
    """
```

tests/run/cdef_class_dataclass.pyx
```diff
@@ -224,6 +224,25 @@ cdef class TestFrozen:
...
     a: double = 2.0
 
+@dataclass(kw_only=True)
+cdef class TestKwOnly:
+    """
+    >>> inst = TestKwOnly(a=3, b=2)
+    >>> inst.a
+    3.0
+    >>> inst.b
+    2
```
scoder, line 234:
```suggestion
    >>> inst = TestKwOnly(a=3, b=2)
    >>> inst.a
    3.0
    >>> inst.b
    2
    >>> inst = TestKwOnly(b=2)
    >>> inst.a
    2.0
    >>> inst.b
    2
```

Cython/Compiler/Dataclass.py
```diff
@@ -221,6 +222,8 @@ def handle_cclass_dataclass(node, dataclass_args, analyse_decs_transform):
 
     fields = process_class_get_fields(node)
 
+    kw_only = kwargs.pop("kw_only")
```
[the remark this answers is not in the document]
Yurzs: but `_DataclassParams` doesn't have a `kw_only` argument (neither in regular data class)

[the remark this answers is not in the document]
scoder: Hmm. Ok. Sorry, I didn't read the code completely. So, you were only using `kwargs` for the keyword parsing here and it's otherwise passed straight into dataclass params. I reverted my change and added a comment on why it is done this way. It's not obvious from the code as such.

da-woods:
Thanks @Yurzs

PR #4818, opened 2022-05-31
Add mt19937_64 random number generator to libcpp

wjsi:
As `mt19937` already included in #4746, it is better to include `mt19937_64` as well for int64 random variables.


scoder:
Thanks

PR #4857, opened 2022-06-22
Relax adding cython import statement in runtests

matusvalo:


matusvalo:
An attempt whether relaxing adding cython import will break testsuite.

PR #4916, opened 2022-07-22
implement the depfile command-line option for the "cython" tool

eli-schwartz:
Tested with the cythoner from SciPy after editing in the -M argument and updating lots of meson generators to expect a depfile.

/cc @rgommers @ev-br

I do not like the delayed import of `Build.Dependencies.DependencyTree` but it is needed to avoid a circular import, if anyone has suggestions to fix this I'm all ears. (I think this would require more refactoring than I'm comfortable with, but TBH the code is ***massively*** confusing in layout to me.)

This is missing tests but I have no time left today for that (I'll finish this up on Sunday), so, posting for discussion.


rgommers:
This looks pretty clean to me so far; the one code block moved into `write_depfile_for` is a straightforward refactor, and the rest seems about as minimal as it gets for adding a new command-line option.

eli-schwartz:
Tests pass locally.

Cython/Build/Dependencies.py
```diff
@@ -721,6 +721,25 @@ def transitive_merge_helper(self, node, extract, merge, seen, stack, outgoing):
...
+            if fname.startswith(src_base_dir):
+                paths.append(os.path.relpath(fname, cwd))
+            else:
+                paths.append(os.path.abspath(fname))
+
+        depline = os.path.relpath(c_file, cwd) + ": \\\n  "
+        depline += " \\\n  ".join(paths) + "\n"
+
+        with open(c_file+'.dep', 'w') as outfile:
+            outfile.write(depline)
```
scoder, line 742: Excluding the call to `self.all_dependencies()`, I think this can be moved to `Utils.py` where most of the file related utilities live. It's generally a good idea to separate calculation (i.e. figuring out the dependencies) from representation (i.e. writing them down into a dependency file).

tests/build/depfile_package.srctree
```diff
@@ -7,18 +7,21 @@ PYTHON package_test.py
 
 import os.path
 
+def getpath(*args):
```
scoder, line 10: Maybe rename this to `pkgpath()`? `getpath()` seems overly generic.

tests/build/depfile_cython.srctree
```diff
@@ -0,0 +1,61 @@
+"""
```
scoder, line 1: `depfile_package_cython` and `depfile_package_cythonize` now seem better names for the two tests.

tests/build/depfile_cython.srctree
```diff
@@ -0,0 +1,61 @@
...
+PYTHON -c 'import os; os.makedirs("builddir/pkg/sub")'
+CYTHON -M pkg/test.pyx -o builddir/pkg/test.c
+CYTHON -M pkg/sub/test.pyx -o builddir/pkg/sub/test.c
```
scoder, line 4: Why not test both while we're at it?
```suggestion
CYTHON -M pkg/test.pyx -o builddir/pkg/test.c
CYTHON --depfile pkg/sub/test.pyx -o builddir/pkg/sub/test.c
```

Cython/Compiler/Main.py
```diff
@@ -502,6 +502,9 @@ def run_pipeline(source, options, full_module_name=None, context=None):
 
     err, enddata = Pipeline.run_pipeline(pipeline, source)
     context.teardown_errors(err, options, result)
+    if options.depfile:
+        from ..Build.Dependencies import DependencyTree
+        DependencyTree(context).write_depfile_for(result.main_source_file, result.c_file)
```
scoder, line 507: There's a function `create_dependency_tree()` as entry point.

tests/build/depfile_package.srctree
eli-schwartz: Good point, done.

tests/build/depfile_cython.srctree
eli-schwartz: Well, it's basically an argparse test, but fair enough, it cannot possibly hurt and is literally free.

tests/build/depfile_cython.srctree
eli-schwartz: Renamed.

Cython/Compiler/Main.py
eli-schwartz: Okay. I have no idea what the context is at different points, or whether the cached DependencyTree makes a difference, but if you say to use it, I'll use it. :)

tests/build/depfile_package_cython.srctree
```diff
@@ -0,0 +1,61 @@
...
+CYTHON -M pkg/test.pyx -o builddir/pkg/test.c
+CYTHON --depfile pkg/sub/test.pyx -o builddir/pkg/sub/test.c
+PYTHON check.py
+"""
+
+######## check.py ########
+
+import os.path
+
+def getpath(*args):
```
scoder, line 12: Same thing with `pkgpath()`.

Cython/Utils.py
```diff
@@ -626,3 +626,23 @@ def build_hex_version(version_string):
         hexversion = (hexversion << 8) + digit
 
     return '0x%08X' % hexversion
+
+
+def write_depfile(mainfile, source, dependencies):
+    src_base_dir, _ = os.path.split(source)
```
scoder, line 632: I'd also rename `mainfile` to `target_file` to distinguish it from `source`.
```suggestion
def write_depfile(mainfile, source_file, dependencies):
    src_base_dir = os.path.dirname(source_file)
```

tests/build/depfile_package_cython.srctree
eli-schwartz: Oops, fixed.

eli-schwartz:
Thanks!

Is there interest in backporting this to 0.29.x?

ev-br:
Backporting to 0.29.x would need some minor manual tweaks: https://github.com/cython/cython/pull/4576
(optparse/argparse at least)

eli-schwartz:
Alright, I tried a backport.

PR #4949, opened 2022-08-04
[0.29] implement the depfile command-line option for the "cython" tool

eli-schwartz:
Backports #4916

I'm not sure I understand what's up with the second commit here.


scoder:
Looks good. Thanks

PR #4962, opened 2022-08-07
Add C++20's Bit manipulation functions

jhelgert:
- Adds all the new [bit manipulation functions](https://en.cppreference.com/w/cpp/header/bit) inside the `<Bit>` header library coming with C++20.
- Adds a new `cpp20` tag for the tests.


Cython/Includes/libcpp/bit.pxd
```diff
@@ -0,0 +1,31 @@
...
+
+    # integral powers of 2 (gcc >= 10.0, clang >= 12.0)
+    cdef bint has_single_bit[T](T)
+    cdef T bit_ceil[T](T)
+    cdef T bit_floor[T](T)
+    cdef int bit_width[T](T)
+
+    # rotating (gcc >= 9.0, clang >= 9.0)
+    cdef T rotl[T](T, int)
+    cdef T rotr[T](T, int)
```
scoder, line 16: It's often helpful to name arguments in order to support keyword argument calls. Here, users might want to pass the rotation bit count as keyword args (why not).

runtests.py
```diff
@@ -346,6 +346,39 @@ def update_cpp17_extension(ext):
 
     return EXCLUDE_EXT
 
+def update_cpp20_extension(ext):
```
scoder, line 349: I think we have enough copies of this function by now to merit a configurable version, i.e. have one function, pass in all relevant compiler versions, and let it return a closure function that configures the extension build. Then call the outer function three times in `EXT_EXTRAS`, for C++11/17/20, and then some more in the future.

scoder (commented):
Seems good to have. Thanks also for the version comments in bit.pxd.

runtests.py
scoder: I refactored the setup.

Cython/Includes/libcpp/bit.pxd
scoder, line 16:
```suggestion
    cdef T rotl[T](T, int shift)
    cdef T rotr[T](T, int shift)
```

scoder:
Thanks

PR #4977, opened 2022-08-14
ci-run.sh: run msvc build with multiple processes

0dminnimda:
Partial fix for #4963


0dminnimda:
Not sure if there is any speedup.
Maybe a tiny one

0dminnimda:
I looked through all of the cl command line options and it seems that's the most speedup we're gonna get out of them.

0dminnimda:
Well, we can ty to fixate number of processes, but it seems better to leave this choice to the cl

scoder:
Thanks

PR #5024, opened 2022-09-12
✨ Enable OpenSSF Scorecard code scanning alert and badge

joycebrum:
This pull request intend to activate the OpenSSF Scorecard,  a tool to search for security risks. More informations about securitycards can be seen at https://securityscorecards.dev/

Besides listing the security risks and its possible solution, the OpenSSF Scorecard calculates a Score to the project, indicating it security coverage, as can be seen in the badge bellow.

image

This change could help the project to improve its security coverage besides it can be a way of external users to evaluate the security and reliability of the project.

Please feel free to reach me out in case of doubts or concerns.

### DIFF
--- a/.github/workflows/scorecards.yml
+++ b/.github/workflows/scorecards.yml
@@ -1,0 +1,54 @@
+name: Scorecards supply-chain security
+on:
+  # Only the default branch is supported.
+  branch_protection_rule:
+  schedule:
+    - cron: '42 21 * * 0'
+  push:
+    branches: [ "master" ]
+
+# Declare default permissions as read only.
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecards analysis
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
+      # Used to receive a badge.
+      id-token: write
+    
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@a12a3943b4bdde767164f792f33f40b04645d846 # tag=v3.0.0
+        with:
+          persist-credentials: false
+
+      - name: "Run analysis"
+        uses: ossf/scorecard-action@f437b3cc043103683806977b725b99217c67bd46 # tag=v2.0.1
+        with:
+          results_file: results.sarif
+          results_format: sarif
+
+          # Publish the results for public repositories to enable scorecard badges. For more details, see
+          # https://github.com/ossf/scorecard-action#publishing-results. 
+          # For private repositories, `publish_results` will automatically be set to `false`, regardless 
+          # of the value entered here.
+          publish_results: true
+
+      # Upload the results as artifacts (optional). Commenting out will disable uploads of run results in SARIF
+      # format to the repository Actions tab.
+      - name: "Upload artifact"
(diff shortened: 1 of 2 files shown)


joycebrum:
Just to clarify a few things, the OpenSSF is a non-profit foundation dedicated to improving the overall security of the open source community.

Also, the main benefit the [Scorecard GitHub Action](https://github.com/ossf/scorecard) brings is automatic security scanning, which not only points out potential security risks, but also gives you alternatives and insights to solve them. This way you would be able to not just identify but to select the proper solution to each problem. An example of security risk indicated by the scorecard and its possible solution can be seen in the pictures below.

The badge is only meant to show your users you're dedicated to ensuring the security of your project (your project is in the top ~20%). If you decide to integrate it in your project, you will help raise awareness which provides continuous adoption at scale, but also to iterate on continuous feedback, benefiting the community at large.

However, I can perfectly understand if you prefer to keep the scores private. In which case I'll happily remove the badge from the README file.

image
image

scoder:
I'm generally happy to add static analysis and security scans to Cython.

However, scanning only the project sources won't help much. Cython itself isn't a program that runs in security relevant contexts (at least not often). It would be cool if the C code generated in our test runs could be analysed. There is an option in the test runner to skip the cleanup step, thus leaving all code around.

I also don't think that I'd like to have a badge on the project page. We don't currently have any badges there, why this one? If there's an issue, we should just hear about it, and that's it.

scoder:
Forgot to say: thanks for bringing this up, @joycebrum. It does sound like something that could help a lot of OSS projects, including Cython.

joycebrum:
Hi @scoder, I totally understand you and really would be very intersting to verify the C code generated. Although the OpenSSF Scorecard is not able to help with it, it does not just scan the project sources itself, but also check a lot of other criterias, like: CII Best Practices, Security Policy, Vulnerabilities (Uses the [OSV service](https://osv.dev/).), Continuous Testing Criterias, Branch Protection, Code Review, etc.

The OpenSSF Scorecard checks is more about securing the repo's processes than actual code health: having a Scorecard of 10 doesn't mean your code is safe and bug-free, it just means that it's (mostly) safe from malicious sabotage.

In https://securityscorecards.dev/#the-checks you can see a better and complete explanation of the types of checks and its benefits, it would be a very intersting reading on the understanding of what the OpenSSF Scorecard actually does.

I have removed the badge since it is completely optional. 

I also would like to point out that if you want to give a try to the OpenSSF Scorecard and its insights to improve the repository's security and, in the end, you dislike it, it is easy to uninstall/disable (just by removing the scorecards.yml)

scoder:
Ah, thanks for clarifying. I hadn't taken much time to look through the service itself, now it seems that it's a sort of meta-checker, validating that the project adheres to the service's proposed best practices of using other services. Probably still worth a try…

scoder:
I should add that I don't see a need for scoring here. It's nice to get hints what could be done better, but few of these hints are pressing issues. Some are just "ok" or "ok for now".

joycebrum:
Thanks for the feedback @scoder, I'll relay the feedback to the Scorecard team and any other feedback we are always open to hear.

andy778:
If I look on the scorecard report, what I see is things that would be good to work on from that is 
- Dependency-Update-Tool  is 0 and there is quite many python modules here...
- Security-Policy not improving the real security, but is simple and good to add 
- Branch-Protection and score 3, it's better than zero but after the xz incident some weeks ago it gets more important

PR #5032, opened 2022-09-16
Use proper SPDX identifier

Shortfinga:


scoder:
Seems reasonable. Thanks.

PR #5034, opened 2022-09-18
Fix the step for uploading platform-specific wheels to GitHub

SpecLad:
`cibuildwheel` puts the wheels into `wheelhouse`, not `dist`. There also doesn't seem to be any reason to list a bunch of separate per-platform wildcards, so replace them with `*.whl`.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -86,12 +86,7 @@ jobs:
         uses: softprops/action-gh-release@v1
         if: startsWith(github.ref, 'refs/tags/')
         with:
-          files: |
-            dist/*manylinux*.whl
-            dist/*musllinux*.whl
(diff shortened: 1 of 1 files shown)


scoder:
> There also doesn't seem to be any reason to list a bunch of separate per-platform wildcards, so replace them with `*.whl`.

The `auditwheel` processing usually leaves two types of wheels around: the original Linux wheel and the generic manylinux wheel. We only need the latter. Don't know if that's the same for cibuildwheel, but I wouldn't want the useless non-manylinux Linux wheels to be uploaded.

SpecLad:
FWIW, I tested the change with this fake release: <https://github.com/SpecLad/cython/releases/tag/123.456>. You can see there what sorts of wheels are uploaded.

scoder:
Thanks

PR #5080, opened 2022-10-11
[Doc] Improve documentation of disabling annotations in pure mode

matusvalo:


matusvalo:
> I missed that there is also `@cython.annotation_typing()` decorator. It is not documented, so I will add it to this PR.

Done. Ready for review

docs/src/tutorial/pure.rst
```diff
@@ -378,6 +377,19 @@ declare types of variables in a Python 3.6 compatible way as follows:
 
 There is currently no way to express the visibility of object attributes.
 
+Disabling annotations
+^^^^^^^^^^^^^^^^^^^^^
+
+To avoid conflicts with other kinds of annotation
+usages, reading of annotation by Cython can be disabled with the ``@cython.annotation_typing(False)`` decorator:
```
da-woods, line 384:
```suggestion
usages, Cython's use of annotations to specify types can be disabled with the
``annotation_typing`` compiler directive<compiler-directives>. From Cython 3
you can use this as a decorator, as shown in the following example:
```

docs/src/tutorial/pure.rst
```diff
@@ -378,6 +377,19 @@ declare types of variables in a Python 3.6 compatible way as follows:
...
+
+.. literalinclude:: ../../examples/tutorial/pure/disabled_annotations.py
+
+Another way to disable annotations is by using :ref:`compiler directive<compiler-directives>` ``annotation_typing=False``::
+
+    cython -X annotation_typing=False ignored_annotations.py
```
da-woods, line 390:
```suggestion
```

da-woods (commented):
I've rewritten it slightly, but don't feel you have to use my exact wording. The changes I made were:

* I think it's the same compiler directive whether you use it as a decorator, or a `with` statement, or at a file level,
* I think it's worth making clear that the ability to use it as a decorator is a Cython 3 improvement.

matusvalo:
> or a with statement

`cython.annotation_typing()` supports also `with` statement? If yes I prefer to mention it in docs too...

da-woods:
> cython.annotation_typing() supports also with statement? If yes I prefer to mention it in docs too...

Yes it should do. I don't think it needs special mention in the text - most compiler directives can be used as a `with` statement. You could definitely include it in your example though - it'd let you skip typing one or two variables.

docs/examples/tutorial/pure/disabled_annotations.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def annotated_function(a: cython.int, b: cython.int):
+    s: cython.int = a + b
+    with cython.annotation_typing(False):
+        # Cython is ignoring annotations within this code block
+        c: list = []
+        c.append(a)
+        c.append(b)
+        c.append(s)
```
da-woods, line 32: If I were writing this I'd only put the `c: list` in the `with` block. But either way will compile

matusvalo: > If I were writing this I'd only put the `c: list` in the `with` block. But either way will compile

fixed.

matusvalo:
Please have a final look.

da-woods:
Just waiting to make sure that the CI is happy with the code.... looks good otherwise

matusvalo:
> Just waiting to make sure that the CI is happy with the code.... looks good otherwise

btw have you tried auto-merge?
image

matusvalo:
Hmm CI fail seems to be unrelated to this PR...

da-woods:
Thanks - I think it's useful to have this information up front since it Cython won't always do what people want with annotations

PR #5137, opened 2022-11-17
Impose black to new files

gabrieldemarmiesse:
This is related to #5127  , in this version, black is only applied on new files.


scoder:
Closing since it's no longer up to date.

PR #5138, opened 2022-11-17
Apply black to the whole codebase

gabrieldemarmiesse:
related to #5127


gabrieldemarmiesse:
I get errors like this 
```
Error compiling Cython file:
...
          |name| is optional, and may be the name of the file being
          scanned or any other identifying string.
        """
        self.trace = 0

        self.buffer = ""
                      ^

Cython/Plex/Scanners.py:92:22: str objects do not support coercion to unicode, use a unicode string literal instead (u'')
```
after black changes `self.buffer = u''` to `self.buffer = ""`.

I suspect this is because of the directive 
```
# cython: language_level=3str
```
does that mean that we can't apply back because the assumptions of black about strings are not valid in cython with this directive ?

pyproject.toml
```diff
@@ -0,0 +1,22 @@
+[tool.black]
+skip-magic-trailing-comma = true
```
scoder, line 2: Why this option?

gabrieldemarmiesse: Good question,
It's one of the few times where black look at the previous formatting and uses this information. Furthermore, people who haven't read the black documentation trigger the magic comma without knowing about it. We can always remove this option if you think the magic comma is useful, I don't really mind:
https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma

scoder:
Closing since it's no longer up to date.

PR #5186, opened 2022-12-31
Update Cython Debugger

WillAyd:
Wanted to see if we could modernize this a bit - looks like a lot of the documentation still wants users to use Python2.7 alongside this.

Not sure everything here is 100% correct but I think still represents progress over what is there today. You can get a working Cython debugger image and mini instructions from this Docker repo:

https://hub.docker.com/repository/docker/willayd/cython-debug


Cython/Debugger/libcython.py
```diff
@@ -776,8 +776,8 @@ def _break_pyx(self, name):
 
         if (cython_module.filename, lineno) in cython_module.lineno_cy2c:
             c_lineno = cython_module.lineno_cy2c[cython_module.filename, lineno]
-            breakpoint = '%s:%s' % (cython_module.c_filename, c_lineno)
-            gdb.execute('break ' + breakpoint)
+            breakpnt = '%s:%s' % (cython_module.c_filename, c_lineno)
```
WillAyd, line 779: Wanted to avoid shadowing the breakpoint keyword, which didn't exist when this code was first written

Cython/Debugger/libcython.py
```diff
@@ -1172,7 +1172,10 @@ def invoke(self, args, from_tty):
             return
 
         local_cython_vars = cython_function.locals
-        max_name_length = len(max(local_cython_vars, key=len))
+        if local_cython_vars:
```
WillAyd, line 1175: In cases where you ran `cy locals` and there were none to report the call to `max` would fail with `ValueError: max() arg is an empty sequence`

Cython/Debugger/libpython.py
```diff
@@ -628,7 +628,7 @@ class PyCFunctionObjectPtr(PyObjectPtr):
     _typename = 'PyCFunctionObject'
 
     def proxyval(self, visited):
-        m_ml = self.field('m_ml')  # m_ml is a (PyMethodDef*)
+        m_ml = self.field('m_ml') # m_ml is a (PyMethodDef*)
```
WillAyd, line 631: All of the updates in this file are upstreamed from the Python3.7 branch of CPython

docs/src/userguide/debugging.rst
```diff
@@ -9,19 +9,15 @@ Debugging your Cython program
...
-
-    ./configure --with-python=python2
+For optimal results, you should configure gdb from source and use the --with-python
+configuration flag. ::
+
+    ./configure --with-python=python3
     make
     sudo make install
 
-Installing the Cython debugger can be quite tricky. `This installation script and example code <https://gitlab.com/volkerweissmann/cygdb_installation>`_ might be useful.
```
WillAyd, line 24: The original link here has a note that it is broken and unmaintained, so figured harmless enough to remove

docs/src/userguide/debugging.rst
```diff
@@ -9,19 +9,15 @@ Debugging your Cython program
...
-code should be built with Python 3 and the debugger should be run with
-Python 2 (or at least it should be able to find the Python 2 Cython
-installation). Note that in recent versions of Ubuntu, for instance, ``gdb``
-installed with ``apt-get`` is configured with Python 3. On such systems, the
-proper configuration of ``gdb`` can be obtained by downloading the ``gdb``
-source, and then running::
-
-    ./configure --with-python=python2
+For optimal results, you should configure gdb from source and use the --with-python
+configuration flag. ::
```
scoder, line 13: It feels like we could still say a bit more about ways to get this working with recent Python 3 deployments. Most Linux installations these days might actually work out of the box with gdb and Python 3. Users shouldn't have to build anything as basic as gdb from sources.

scoder (commented):
Thanks. I left a few comments.

docs/src/userguide/debugging.rst
WillAyd: Yea I couldn't find good info on this, but as far as I could tell gdb needed to be compiled with a reference to the Python installation you want to use to debug. I wasn't able to use a virtual environment and get it to work

Cython/Debugger/libpython.py
```diff
@@ -1572,8 +1572,8 @@ def is_other_python_frame(self):
         if not caller:
             return False
 
-        if (caller.startswith('cfunction_vectorcall_') or
-            caller == 'cfunction_call'):
+        if caller in ('_PyCFunction_FastCallDict',
+                      '_PyCFunction_FastCallKeywords'):
```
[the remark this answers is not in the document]
WillAyd: Hmm interesting. Actually looking at this again the previous code in the diff matches what is in Python's main branch. What I have here came from the 3.7 branch, so I'm guessing the previous Cython code was upstreamed from a newer version than 3.7

Do you still want all 4 or should we just match what's in Python today?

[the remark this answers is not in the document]
scoder: The more Python versions we can support, the better.

mironleon:
I have been spending some days recently to get cython debugging working for a project, figured this might be a good place to add some suggestions/comments. @WillAyd thanks for your blogpost and docker image, it was a good starting point. 

I am working with the following docker setup:

```docker
FROM ubuntu:22.04

RUN apt-get update && apt-get upgrade -y

# prevent being asked for a timezone
ENV DEBIAN_FRONTEND=noninteractive
# Note that debugging currently only works with python 3.10
ENV PYTHON_VERSION=3.12

# dependencies
RUN apt-get install -y \
    # to get python source and finesse source
    git \
    # for building python
    build-essential lcov pkg-config libbz2-dev libffi-dev libgdbm-dev libgdbm-compat-dev liblzma-dev libncurses5-dev libreadline6-dev libsqlite3-dev libssl-dev lzma lzma-dev tk-dev uuid-dev zlib1g-dev \
    # for gdb
    libgmp-dev wget

# Clone and compile python with debug flag
RUN git clone -b ${PYTHON_VERSION} --depth 1 https://github.com/python/cpython.git /clones/cpython
RUN cd /clones/cpython && ./configure --with-pydebug CFLAGS="-g" && make -s -j$(nproc) && make install

# clone and compile gdb and link to python install
RUN cd /tmp && wget http://mirrors.kernel.org/sourceware/gdb/releases/gdb-13.2.tar.gz && tar -zxf gdb-13.2.tar.gz
RUN cd /tmp/gdb-13.2 && ./configure --with-python=python3
RUN cd /tmp/gdb-13.2 && make all install

# for convenience
RUN ln -s /usr/local/bin/python${PYTHON_VERSION} /usr/local/bin/python

WORKDIR /home
RUN python -m pip install cython
RUN echo 'def foo():\n    message = "Hello world!"\n    print(message)\n' > test.pyx
RUN cython -3 --gdb test.pyx
RUN gcc -g -fPIC -I/usr/local/include/python${PYTHON_VERSION}d -shared test.c -o test.so
```

And following the steps from your docker image:

```bash
cygdb . -- --args python3 -c "import test; test.foo()"
cy break foo
cy run
```
And continuing by stepping into the function and listing source code and printing out variables.

I have tested python 3.10, 3.11, and 3.12. I noticed that the `cy print` command does not work in 3.11 and 3.12 with the same error:

```bash
(gdb) cy list
     1    def foo():
     2        message = "Hello world!"
>    3        print(message)
     4    
(gdb) cy print message
  File "/usr/local/lib/python3.11/site-packages/Cython/Debugger/libpython.py", line 1980, in wrapper
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/Cython/Debugger/libcython.py", line 1112, in invoke
    global_python_dict = self.get_cython_globals_dict()
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/Cython/Debugger/libcython.py", line 383, in get_cython_globals_dict
    for k, v in pyobject_dict.iteritems():
  File "/usr/local/lib/python3.11/site-packages/Cython/Debugger/libpython.py", line 687, in iteritems
    entries, nentries = self._get_entries(keys)
                        ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/Cython/Debugger/libpython.py", line 731, in _get_entries
    dk_size = int(keys['dk_size'])
                  ~~~~^^^^^^^^^^^
gdb.error: There is no member named dk_size.
Python Exception <class 'gdb.error'>: There is no member named dk_size.
Error occurred in Python: There is no member named dk_size.
```

Using cython 0.29.36, `cygdb` fails to show the cython source code on the `cy list` command and after breaking on the function definition line, refuses to step into the function.

Trying to get the setup to work with just packages from the repository:
```docker
FROM ubuntu:22.04

RUN apt-get update && apt-get upgrade -y
# prevent being asked for a timezone
ENV DEBIAN_FRONTEND=noninteractive

RUN apt-get install -y gdb python3-dbg pip

WORKDIR /home
RUN python3 -m pip install cython
RUN echo 'def foo():\n    message = "Hello world!"\n    print(message)\n' > test.pyx
RUN cython -3 --gdb test.pyx
RUN gcc -g -fPIC -I/usr/include/python3.10d -shared test.c -o test.so
```
Results in 

```
ImportError: /home/test.so: undefined symbol: _Py_RefTotal
```

WillAyd:
Thanks @mironleon . I see upstream in the CPython source that that lines where changed to `dk_size = 1<<int(keys['dk_log2_size'])` starting in 3.11

https://github.com/python/cpython/blob/37589d76bbe97b0bf13ffafb8dd6aab361a0209a/Tools/gdb/libpython.py#L798

I am guessing that is the root cause of your issue. I think one of the challenges to this will be figuring out how to provide compat across different Python versions, as I don't _think_ the gdb extension CPython distributes cares all that much about that

WillAyd:
Maybe this current approach works better - I ended up just vendoring one module from CPython for every minor version in the 3.x series. Does that make it easier to use?

runtests.py
```diff
@@ -1823,6 +1823,9 @@ def runTest(self):
         paths = []
         for codedir in source_dirs:
             paths += glob.glob(os.path.join(self.cython_dir, codedir + "/**/*.py"), recursive=True)
+
+        # remove any files vendored from CPython
+        paths = [x for x in paths if not "vendored" in x]
```
WillAyd, line 1828: Not sure if there was an easier way to exclude things from pycodestyle

da-woods:
I'm largely staying out of this PR because I don't know much about how gdb works with Python/Cython. However:

> Maybe this current approach works better - I ended up just vendoring one module from CPython for every minor version in the 3.x series. Does that make it easier to use?

That seems like a reasonable approach to me given the range of versions we have to support. Is there any way to just get the file from the installed Python distribution (in  a sensible, reproducible way). If we could do that it's save maintaining those files. However, if it's impractical then feel free to dismiss it.

scoder:
> Is there any way to just get the file from the installed Python distribution (in  a sensible, reproducible way).

I don't think so. They are not part of a regular Python installation but live in their Tools source folder. We can get them from there, though.

I also think that it's probably easiest to ship version specific (latest) copies rather than maintaining an adaptive joined file.

docs/src/userguide/debugging.rst
WillAyd: OK it may not be as strict as this documentation shows, but it is not as easy as using the system gdb unfortunately. It seems like when calling a system-installed gdb from a virtual environment like conda, gdb loses track of python being called from a local environment and falls back to the system python

```sh
(pandas-dev) willayd@willayd-Lemur-Pro:~/clones/pandas$ which python
/home/willayd/mambaforge/envs/pandas-dev/bin/python
(pandas-dev) willayd@willayd-Lemur-Pro:~/clones/pandas$ cygdb
GNU gdb (Ubuntu 14.0.50.20230907-0ubuntu1) 14.0.50.20230907-git
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word".
There was an error in Python code originating from the file /home/willayd/clones/cython/Cython/Debugger/Cygdb.py
It used the Python interpreter /usr/bin/python
  File "<string>", line 11, in <module>
ModuleNotFoundError: No module named 'Cython'
```

Installing gdb into the virtual environment resolves the issue. Not sure yet if this is a Cython or gdb problem mixing up the python installations

WillAyd: Looking at the gdb source code there is also some compile-time compatability for versions prior to 3.9. This particular instance probably just deserves a note in the documentation, but there is a risk for that happening in the future 

https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=gdb/python/python.c;h=8a36673a3e4c0bef8533d259d8c13118e650df1a;hb=refs/heads/master#l2213

Jake-Moss:
I took this for a spin recently and was able to get it working, although I needed this patch to get it functioning and to manually import `libcython` and `libpython` in gdb due to issues with `cydgb` and my pyenv virtual environments. Though I prefer that workflow anyway. It expects a `activate` script to exist that doesn't.

I would like to be able to provide some more detailed feedback but unfortunately I'm being hit by https://github.com/cython/cython/issues/6004 and cannot build a recent version of Cython from source to work for the code base I'm working on. Stepping back to earlier commits I get hit by failures to import `MatchCaseNodes`.

Regarding the comments on this PR, most are fixed by vendoring the various versions of `libpython.py`
```patch
diff --git a/Cython/Debugger/libpython.py b/Cython/Debugger/libpython.py
index 8738152da..121c166f3 100644
--- a/Cython/Debugger/libpython.py
+++ b/Cython/Debugger/libpython.py
@@ -1,6 +1,8 @@
 #!/usr/bin/python
 
 import sys
+import gdb
+import os
 
 if sys.version_info[0] == 3:
     if sys.version_info[1] == 4:
@@ -17,7 +19,7 @@ if sys.version_info[0] == 3:
         from .vendored.libpython39 import *
     elif sys.version_info[1] == 10:
         from .vendored.libpython310 import *
-    elif sys.version_info[1] == 1:
+    elif sys.version_info[1] == 11:
         from .vendored.libpython311 import *
     elif sys.version_info[1] >= 12:
         from .vendored.libpython312 import *
```

da-woods:
I'm of the opinion that:
* #7016 fixes the virtual env problem as well as is possible.
* This should get us a lot closer to cygdb working again.

What I'd propose to do is merge #7016 and then update this PR and merge this. I'm happy to do the update (I think it's relatively straightforward and just a few small conflicts and picking up the vendored files for 3.13 and 3.14). I'm fairly convinced it gets enough working that we should just take it rather than leaving it sitting around.

da-woods:
I suspect this should just be deferred until later, but the breakpoints on Python functions don't seem to work on recent Python because it doesn't go through `PyEval_EvalFrameEx`. And the functions probably need to be modified to look at the internal `_PyInterpreterFrame` rather than the Python frame objects that are only created on demand.

PR #5196, opened 2023-01-04
Add `annotation_typing` in shadow.py

### LINKED ISSUE #5194: [BUG] module 'cython' has no attribute 'annotation_typing'
### Describe the bug

Hello, I asked this question on StackOverflow and I was referred here as it may be a bug.  I can provide additional information but I think my [SO post](https://stackoverflow.com/questions/75004800/module-cython-has-no-attribute-annotation-typing) covers it succinctly.  

I just recently updated Cython from 0.29.24 to 3.0.0a11 and I was getting compilation errors due to functions that were using Dict type hints. With the previous versions, I never had any issues of this type (no pun intended). I compile my entire python project by building Extensions using cythonize in a setup.py file. I am using Python 3.9.13.

After doing some reading I attempted to solve the issues I was having by using the [compiler directive](https://cython.readthedocs.io/en/latest/src/tutorial/pure.html#disabling-annotations) as follows:

```cython
import cython

@cython.annotation_typing(False)
def my_function_with_type_hints(this_object: Dict) -> Dict:
   pass
```

### Expected behaviour

Compilation succeeded although I have not tested my compiled code yet. I returned to development to test some new code and now I am unable to run code in interpeter mode because Cython is complaining "module 'cython' has no attribute 'annotation_typing'". I confirmed (via pip freeze) that Cython==3.0.0a11 is installed in my environment and that my environment is activated (in vscode).

My questions:

Did I fix the typing issue correctly using the compiler directive or am I not understanding the problem fully?
How should I be using compiler directives in interpreter mode? Do I need another decorator that conditionally includes the cython directive?

### Environment

Python version 3.9.13
Cython version 3.0.0a11

### Additional context

_No response_

0dminnimda:
Closes #5194

### DIFF
--- a/Cython/Shadow.py
+++ b/Cython/Shadow.py
@@ -107,8 +107,8 @@ class _Optimization(object):
 
 cclass = ccall = cfunc = _EmptyDecoratorAndManager()
 
-returns = wraparound = boundscheck = initializedcheck = nonecheck = \
-    embedsignature = cdivision = cdivision_warnings = \
+annotation_typing = returns = wraparound = boundscheck = initializedcheck = \
+    nonecheck = embedsignature = cdivision = cdivision_warnings = \
     always_allows_keywords = profile = linetrace = infer_types = \
     unraisable_tracebacks = freelist = \
         lambda _: _EmptyDecoratorAndManager()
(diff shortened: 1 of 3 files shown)


scoder:
Thanks

PR #5200, opened 2023-01-05
Recognize `nogil()` arguments in pure python mode

### LINKED ISSUE #5113: [BUG] `cython.nogil` does not support function parameters in pure mode
### Describe the bug

When `with cython.nogil(PARAM):` is used, the `PARAM` is ignored and GIL is release in any case - see the example. The example produces following code:
image

On the other hand in the Cython syntax the parameter is not ignored and providing `True` in the example, the lock is not relased:

```cython
def increment(int x):
    with nogil(False):
        x = x + 1
    return x

print(increment(5))
```

image

This behavior contradicts the documentation here:
https://cython.readthedocs.io/en/latest/src/tutorial/pure.html#managing-the-global-interpreter-lock

```cython
import cython

def increment(x: cython.int):
    with cython.nogil(False):
        x = x + 1
    return x

print(increment(5))
```

### Expected behaviour
One of following:

1. Pure version has same behaviour as Cython one
2. Documentation is fixed

### Environment

OS: macOS
Python version 3.9
Cython version master

### Additional context

`cython.gil()` seems to be also (at least partially affected/broken:
* In cython following code generates C code:
```cython
def increment(int x):
    with nogil(False):
        x = x + 1
        with gil(False):
            print(x)
    return x

print(increment(5))
```

> Note: Also works for providing `True` instead of `False` to `gil`/`nogil`

On the other hand following code fails to compile:
```cython
import cython

def increment(x: cython.int):
    with cython.nogil(True):
        x = x + 1
        with cython.gil(True):
            print(x)
    return x

print(increment(5))
```
For both `True`/`False` parametrers in `cython.gil`, `cython.nogil`

On the other hand following code works:
```python
import cython

def increment(x: cython.int):
    with cython.nogil:
        x = x + 1
        with cython.gil:
            print(x)
    return x

print(increment(5))
```

matusvalo:
Fixes #5113.


Cython/Compiler/ParseTreeTransforms.py
```diff
@@ -1365,7 +1368,8 @@ def visit_WithStatNode(self, node):
                     name, value = directive
                     if name in ('nogil', 'gil'):
                         # special case: in pure mode, "with nogil" spells "with cython.nogil"
-                        node = Nodes.GILStatNode(node.pos, state = name, body = node.body)
+                        condition = value if isinstance(value, ExprNodes.PrimaryCmpNode) else None
```
scoder, line 1371: `else None`? But then we'd ignore existing user code. I think we should always honour what's there, as we do with the Cython syntax. Ideally, both should be handled in the same place, or failing that, at least in the same way.

matusvalo: Fixed

Cython/Compiler/ExprNodes.py
```diff
@@ -5991,8 +5991,14 @@ def nogil_check(self, env):
         func_type = self.function_type()
         if func_type.is_pyobject:
             self.gil_error()
-        elif not func_type.is_error and not getattr(func_type, 'nogil', False):
-            self.gil_error()
+        elif not func_type.is_error:
+            if hasattr(func_type, 'nogil'):
+                if isinstance(func_type.nogil, ExprNode) and not func_type.nogil.value:
```
scoder, line 5996: I don't think every `ExprNode` has a `.value`. Isn't this already done differently elsewhere?

scoder (commented):
We somehow lost an error message in the `pure_errors` test.

Cython/Compiler/ParseTreeTransforms.py
```diff
@@ -1188,6 +1188,9 @@ def try_to_parse_directive(self, optname, args, kwds, pos):
                 raise PostParseError(
                     pos, 'The exceptval directive takes 0 or 1 positional arguments and the boolean keyword "check"')
             return ('exceptval', (args[0] if args else None, check))
+        elif optname in ('nogil', 'gil'):
+            # Special Case: with cython.nogil and with cython.gil can contains expression as argument.
```
scoder, line 1192:
```suggestion
            # Special Case: 'with cython.nogil' and 'with cython.gil' can contain an expression as argument.
```

scoder:
@matusvalo, could you try to get this fixed up? It would be nice to have this included in the final release.

matusvalo:
> @matusvalo, could you try to get this fixed up? It would be nice to have this included in the final release.

Yes, I tried to had a look multiple times but I have failed to find enough time for that

matusvalo:
@scoder I had another look in this PR and I did not like it because:
1. The return value of `try_to_parse_directive()` is changed and is not consistent - it returns value in all cases except `gil`/`nogil` when it returns Expression object
2. Point 1. causes issues in the code using this function because the code expects value not expression object. This would need adding additional hacks which I don't like
3. The PR breaks current behavior of `@nogil`/`@gil` decorators. In master we raise an error for `@nogil(1 == 0)` but this PR silently ignores this case.

I have created alternative way how to implement this feature - see https://github.com/cython/cython/pull/5491

Issue #5178, opened 2022-12-18
[BUG] Struct with an array field can't be initialized correctly


GalaxySnail:
### Describe the bug

I found this bug in #5177.

``` cython
cdef struct MyStruct:
    int arr[4]

def main():
    cdef MyStruct s = MyStruct([1, 2, 3, 4])
```

It causes compilation error:

```
$ cython -3 test.pyx

$ gcc -c test.c
test.c: In function '__pyx_pf_4test_main':
test.c:1916:17: error: assignment to expression with array type
 1916 |   __pyx_t_1.arr = __pyx_t_2;
      |                 ^

$ clang -c test.c
test.c:1916:17: error: array type 'int[4]' is not assignable
  __pyx_t_1.arr = __pyx_t_2;
  ~~~~~~~~~~~~~ ^
1 error generated.
```

The generated C code looks like this:

``` c
struct __pyx_t_4test_MyStruct __pyx_v_s;
struct __pyx_t_4test_MyStruct __pyx_t_1;
int __pyx_t_2[4];

/* "test.pyx":5
 * 
 * def main():
 *     cdef MyStruct s = MyStruct([1, 2, 3, 4])             # <<<<<<<<<<<<<<
 */
__pyx_t_2[0] = 1;
__pyx_t_2[1] = 2;
__pyx_t_2[2] = 3;
__pyx_t_2[3] = 4;
__pyx_t_1.arr = __pyx_t_2;
__pyx_v_s = __pyx_t_1;
```

### Expected behaviour

The generated C code should be:

``` c
struct __pyx_t_4test_MyStruct __pyx_v_s;
struct __pyx_t_4test_MyStruct __pyx_t_1;

/* "test.pyx":5
 * 
 * def main():
 *     cdef MyStruct s = MyStruct([1, 2, 3, 4])             # <<<<<<<<<<<<<<
 */
__pyx_t_1.arr[0] = 1;
__pyx_t_1.arr[1] = 2;
__pyx_t_1.arr[2] = 3;
__pyx_t_1.arr[3] = 4;
__pyx_v_s = __pyx_t_1;
```

### Environment

Cython version: `0.29.32`, `3.0.0a11` and `master`

0dminnimda:
My educated guess is that generating
```c
__pyx_t_1.arr[0] = 1;
```
would require more structural change in the compiler. 

And there's a better fix for it:
```c
__pyx_t_1.arr = __pyx_t_2;
```
->
```c
[...]

PR #5201, opened 2023-01-06
Initialize array fields of struct with memcpy

garychia:
Related to #5178
Using `memcpy`, copy the content of an array into the array field of a `struct`.


scoder:
Thanks

PR #5208, opened 2023-01-10
Ignore `.dev` and `.post` segments in `build_hex_version()`

0dminnimda:
Fixes #5207 and covers the #5205

PR #5209, opened 2023-01-10
Enhance `build_hex_version`

0dminnimda:
I [though](https://github.com/cython/cython/pull/5208#issuecomment-1377285773) that at last I could bring some value here, but @scoder it too fast 069794816423948a3dd7cc3365cf969d44fc9326 ^-^
Well, I don't want to throw the refactoring out, so here it is


scoder:
Thanks

PR #5210, opened 2023-01-11
Replace deprecated functions

shadchin:
`Py_SetProgramName` and `PySys_SetArgv` has been deprecated in Python 3.11

Fix #5134 (part) and #4895

PR #5211, opened 2023-01-17
Disable parsing command line arguments

shadchin:
After #5210 parsing command line arguments turned on.

https://docs.python.org/3/c-api/init_config.html#c.PyConfig.parse_argv

### DIFF
--- a/Cython/Utility/Embed.c
+++ b/Cython/Utility/Embed.c
@@ -44,6 +44,8 @@ static int __Pyx_main(int argc, wchar_t **argv)
 
         PyConfig config;
         PyConfig_InitPythonConfig(&config);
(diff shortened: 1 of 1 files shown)


Cython/Utility/Embed.c
```diff
@@ -44,6 +44,8 @@ static int __Pyx_main(int argc, wchar_t **argv)
 
         PyConfig config;
         PyConfig_InitPythonConfig(&config);
+        /* Disable parsing command line arguments */
```
scoder, line 47: We generally exclude comments from the generated C files if they start with `//`. This comment seems worth excluding.
```suggestion
        // Disable parsing command line arguments
```

scoder:
Looks right, thanks.

PR #5242, opened 2023-02-05
Allow initialisation of const variables

matusvalo:
This PR introduces support of initialising const variables with value. Currently only global variables are supported.

### DIFF
--- a/Cython/Compiler/Nodes.py
+++ b/Cython/Compiler/Nodes.py
@@ -6047,6 +6047,20 @@ class SingleAssignmentNode(AssignmentNode):
 
         self.rhs = self.rhs.analyse_types(env)
 
+        if self.lhs.is_name and self.lhs.entry.type.is_const:
+            # const variable can be initialised if:
+            # * we are in module scope
+            # * self.lhs was not assigned before
+            # * self.rhs is constant value
+            # * self.rhs is literal or const variable
+            if env.is_module_scope and self.lhs.entry.init is None and (self.rhs.is_literal or self.rhs.is_name and self.rhs.type.is_const):
+                if self.rhs.is_literal:
+                    self.lhs.entry.init = self.rhs.value
+                else:
+                    self.lhs.entry.init = self.rhs.result()
+            else:
+                error(self.pos, "Assignment to const '%s'" % self.lhs.name)
+
         unrolled_assignment = self.unroll_rhs(env)
         if unrolled_assignment:
             return unrolled_assignment
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -2230,8 +2230,6 @@ class NameNode(AtomicExprNode):
                 entry = self.entry = entry.as_variable
                 self.type = entry.type
 
-        if self.type.is_const:
-            error(self.pos, "Assignment to const '%s'" % self.name)
         if not self.is_lvalue():
             error(self.pos, "Assignment to non-lvalue '%s'" % self.name)
             self.type = PyrexTypes.error_type
@@ -2544,6 +2542,9 @@ class NameNode(AtomicExprNode):
                 # per entry and coupled with it.
                 self.generate_acquire_buffer(rhs, code)
             assigned = False
+            if self.type.is_const:
+                # Const variables are assigned when declared
+                assigned = True
             if self.type.is_pyobject:
                 #print "NameNode.generate_assignment_code: to", self.name ###
                 #print "...from", rhs ###
--- a/tests/errors/const_decl_errors.pyx
+++ b/tests/errors/const_decl_errors.pyx
@@ -2,10 +2,11 @@
 
 cdef const object o
 
-# TODO: This requires making the assignment at declaration time.
-# (We could fake this case by dropping the const here in the C code,
-# as it's not needed for agreeing with external libraries.
 cdef const int x = 10
+x = 20           # nok
+
+cdef const int y
+cdef const int *z
 
 cdef struct S:
     int member
@@ -25,19 +26,32 @@ cdef func(const int a, const int* b, const (int*) c, const S s, int *const d, in
     f = NULL     # ok
     t = &s
 
+cdef func2():
+    global y, z
+    y = 10       # nok
+    z = &y       # ok
+    z[0] = 10    # nok
+
+cdef func3():
+    y = 10       # ok
+    z = &y       # ok
+    z[0] = 10    # ok
+
 cdef volatile object v
 
 
 _ERRORS = """
 3:5: Const/volatile base type cannot be a Python object
-8:5: Assignment to const 'x'
-15:4: Assignment to const 'a'
-16:4: Assignment to const 'c'
-17:5: Assignment to const dereference
-18:5: Assignment to const attribute 'member'
-19:4: Assignment to const 'd'
-22:4: Assignment to const 'e'
-24:5: Assignment to const dereference
-26:4: Assignment to const 't'
-28:5: Const/volatile base type cannot be a Python object
+7:4: Assignment to const 'x'
+17:8: Assignment to const 'a'
+18:8: Assignment to const 'c'
+19:5: Assignment to const dereference
+20:5: Assignment to const attribute 'member'
+21:8: Assignment to const 'd'
+24:8: Assignment to const 'e'
+26:5: Assignment to const dereference
+28:8: Assignment to const 't'
+32:8: Assignment to const 'y'
+34:5: Assignment to const dereference
+41:5: Const/volatile base type cannot be a Python object
 """


0dminnimda:
One question I have is, will constant values be embeded into generated c code? Does cython have such optimisation step?
I am pretty sure that it's not really important since most c compiler will inline that constant anyway if it's gonna be a global variable

tests/run/const_variable.pyx
```diff
@@ -0,0 +1,29 @@
+# mode: run
+
+cdef const int int_constant = 50
+cdef const int int_constant2 = int_constant
+cdef const int int_constant3
+int_constant3 = 5
```
0dminnimda, line 6: Does it make sense to allow setting a constant value after it's declaration?
C does not allow it and it just seems wrong to allow something like this:
```cython
cdef const int X

cdef function():
    # with a lot
    # of code
    # inside

X = 3
```

da-woods:
I think the most important use case (if we want to replace `def`) is for static array lengths. So I'd quite like to make sure that works.

```
cdef const int length = 50

def f():
    cdef double[length] arr
```

So we definitely want to test that this works

-----------------

As an aside, one of the things I like about modern C++ is that it tries to separate compile-time and runtime constants with `const` and `constexpr`. With that said, I don't think this is a good idea here if we're trying to reduce rather than increase the amount of syntax.

tests/run/const_variable.pyx
0dminnimda: The same goes for an uninitialized constants

0dminnimda:
Another common use for DEF is to have a `==` comparison, it won't hurt to have it

tests/run/const_variable.pyx
matusvalo: > Does it make sense to allow setting a constant value after it's declaration?

No it does not but I was not able to restrict it in the code. Currently, (if I am not wrong) it is not possible to check in the code whether is assignment part of declaration or not.

matusvalo:
> I have is, will constant values be embeded into generated c code? Does cython have such optimisation step?

The following cython code
```cython
cdef const int a = 5

def main():
    print(a)

main()
```

 generates following code:

```c
/* #### Code section: module_declarations ### */

/* Module declarations from "deleteme" */
static int const __pyx_v_8deleteme_a = 5;
/* #### Code section: typeinfo ### */
```

So no, this PR is not embedding it into the C code, just initialise static const variable.

> I am pretty sure that it's not really important since most c compiler will inline that constant anyway if it's gonna be a global variable

Exactly this is the goal of the PR:

image

matusvalo:
> I think the most important use case (if we want to replace `DEF`) is for static array lengths.
> 
> ```
> cdef const int length = 50
> 
> def f():
>     cdef double[length] arr
> ```
> 
> So we definitely want to test that this works

Done.

> As an aside, one of the things I like about modern C++ is that it tries to separate compile-time and runtime constants with const and constexpr. With that said, I don't think this is a good idea here if we're trying to reduce rather than increase the amount of syntax.

I understand your point. I am not judging the decision to move from compile-time definition. The goal of this PR is to allow users to declare global constants in the robust way. I am not deep in C++ but If I understand correctly even C++ recommends `const` variables to be used to declare constants. (The rationale is that it is the job of compiler to be smart and inline them)

tests/run/const_variable.pyx
```diff
@@ -0,0 +1,54 @@
...
+cdef const int int_constant3 = 5
+cdef const int *pint_constant = &int_constant
+cdef const int int_sum_constant = 50 + 50
+# FIXME:
+# cdef const int int_sum_constant = 10 + int_constant
+
+cdef const float float_constant = 50.5
+# FIXME:
+# cdef const float float_sum_constant = 50.2 + 50.3
+# cdef const float float_sum_constant = 50.2 + float_constant
```
matusvalo, line 14: @scoder @da-woods the following commented lines are failing now. Should I fix it in this PR (not sure how complex this can be) or separate PR should be created.

da-woods: I think this is potentially a quite difficult problem. It comes down to the "constant at compile time" to "constant at runtime" issue. Needs a little more thought (and also a bit of investigation about exactly how C handles it, because it'll affect the C code we need to generate)

da-woods:
In C there's also a difference with the array definition depending on the scope https://godbolt.org/z/8b6xaPedE. Which I think means we need to be careful if we want to let people do:

```
cdef const int a = 5

ctypedef struct S:
  double attr[a];
```

which I think we do if this is a suitable substitute for `DEF`

matusvalo:
> In C there's also a difference with the array definition depending on the scope https://godbolt.org/z/8b6xaPedE. Which I think means we need to be careful if we want to let people do:
> 
> ```
> cdef const int a = 5
> 
> ctypedef struct S:
>   double attr[a];
> ```
> 
> which I think we do if this is a suitable substitute for `DEF`

Yes this is serious issue. I am not 100% sure what is the best solution here but I think we should not overcomplicate it to the users: you should use const for global variables but ...

da-woods:
> Yes this is serious issue. I am not 100% sure what is the best solution here but I think we should not overcomplicate it for the users: you should use const for global constants but ...

Yeah - I think I agree with you that this should be kept simple from a user point of view. I'm not trying to be awkward here... just make sure we end up with something that works :)

> 
> EDIT: Maybe we can allow const global variables just in C++ (where the issue is not present), but I am not sure if it is good idea and worth it...

That seems like an unnecessary complication.

I wonder if we could convert int consts to enum (https://godbolt.org/z/dbrrKaqrM). It probably needs a bit of thought to do right (we'd need to bit quite careful about what expressions can be used to generate the enum).

tests/run/const_variable.pyx
```diff
@@ -0,0 +1,54 @@
...
+    print(int_constant)
+    print(int_constant2)
+    print(int_sum_constant)
+
+def test_array():
+    """
+    >>> test_array()
+    5
+    """
+    cdef int[int_constant3] int_array
```
da-woods, line 33: The other thing worth pointing out here: if you put it into Godbolt you see it's generating a C99 variable length array (which isn't great because it's not C89 compatible, and is also optional after C11).

On C++ it generates a fixed length array though.

Again, I think the solution involves generating enums, but it's obviously a bit fiddly...

neutrinoceros:
Since this patch was not included in the first 3.0 beta, should we assume it will not make the cut for the final release either ?

0dminnimda:
@neutrinoceros `DEF` is still gonna be around for 3.0 and when it will be removed (3.1 or later) this pr will definitely be merged

da-woods:
> Since this patch was not included in the first 3.0 beta, should we assume it will not make the cut for the final release either ?

I think it's unlikely to be in the final release. There need to be some moderately sized code-generation to make these constants usable in things like C arrays (they currently generate "C variable length arrays" which isn't really what anyone wants).

matusvalo:
I agree with @da-woods . This PR needs revamp. For now marking as draft not to confuse people.

matusvalo:
OK coming back to this PR after long time. In general, I decided that the easiest way how to proceed is to forbid using `const` variables as array dimension parameter. I think even in this limited scope, this PR can be helpful for projects e.g. see https://github.com/scipy/scipy/pull/19170/files (there is some `DEF`s not used as array dimension parameter).

Regarding, solution of array dimension I propose that following cython code:

```cython
cdef const int a = 5
cdef int[a] aa = [1,2,3,4,5]
```

Will generate following C code:

```c
#define __pyx_v_8deleteme_a_value 5
static int const __pyx_v_8deleteme_a = __pyx_v_8deleteme_a_value;
static int __pyx_v_8deleteme_aa[__pyx_v_8deleteme_a_value];
```

Please @da-woods let me know what do you think.

da-woods:
> Which is kind of interesting. For external const variables we create C99 variable length array or am I wrong?

Probably not in that case, but maybe in some others. `CHAR_MAX` will actually be a define, but we're just telling Cython that it's a const char.

Something like your proposal is probably reasonable. Anonymous enums would also work and be less macro-y. But you probably need to exclude extern variables.

matusvalo:
> Something like your proposal is probably reasonable. Anonymous enums would also work and be less macro-y. But you probably need to exclude extern variables.

I have working PoC. The patch is ugly but at least it shows approach. Currently following code:
```cython
cdef const int a = 5

cdef int[a] aa = [1,2,3,4,5]
print(aa)
```

generates following C code:
```c
/*--- Type declarations ---*/
#define __PYX_V_8DELETEME_A 5
// ....
/* Module declarations from "deleteme" */
static int const __pyx_v_8deleteme_a = __PYX_V_8DELETEME_A;
static int __pyx_v_8deleteme_aa[__PYX_V_8DELETEME_A];
```

I am basically constructing macro by stupid uppercasing of C const variable name. Now I am trying to understand how can I manage following situation:

```
cdef extern from *:
    const Py_ssize_t PY_SSIZE_T_MIN

cdef int[PY_SSIZE_T_MIN] aa
```

Here I need to generate following code using plain const variable name (without uppercasing):
```c
static int __pyx_v_8deleteme_aa[PY_SSIZE_T_MIN];
```

https://github.com/cython/cython/pull/5242/files#diff-8789a479ea870f7646251224385afa3ebffe9b6acab136ff55011b31386ddd84R609-R610

But I don't know how can I decide whether the cost variable was declared in in `extern` block (and should not be uppercased) or was declared inside module and should be uppercased in order to use macro. @da-woods do you have any idea how it can be solved?

da-woods:
> @da-woods do you have any idea how it can be solved?

`entry.visibility == 'extern'` I think.

(I haven't had a chance to have a look at the rest yet so can't comment on too much else of it)

Cython/Compiler/Nodes.py
```diff
@@ -606,7 +606,15 @@ def analyse(self, base_type, env, nonempty=0, visibility=None, in_pxd=False):
             self.dimension = self.dimension.analyse_const_expression(env)
             if not self.dimension.type.is_int:
                 error(self.dimension.pos, "Array dimension not integer")
-            size = self.dimension.get_constant_c_result_code()
+            size = None
+            if self.dimension.type.is_const:
+                if self.dimension.entry.defined_in_pxd:
+                    error(self.pos, "Array dimension declared in pxd file")
```
scoder, line 612: This doesn't seem to be tested. Could you explain (or add a comment) why this isn't allowed? It doesn't seem immediately wrong to me if a constant is defined in the module's `.pxd` file. Or is it only about cimported pxd files?

Cython/Compiler/ModuleNode.py
```diff
@@ -1403,10 +1409,14 @@ def generate_cvariable_declarations(self, env, code, definition):
                 code.put(type.cpp_optional_declaration_code(
                     cname, dll_linkage=dll_linkage))
             else:
+
```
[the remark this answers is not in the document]
matusvalo: Yes. The code is in very rough shape. I did not do any clean up for now.

Cython/Compiler/Nodes.py
matusvalo: > This doesn't seem to be tested.

Not yet. The PR is still not 100% ready for review.

matusvalo:
OK. For clarifying about current status of PR. This PR right now is mainly to show an approach what code to generate on C level when user uses `const` variable. The straightforward simple solution is easy to implement. The main point of pain is to support array dimension:

```cython
cdef const int len = 10
cdef int[len] arr = [1,2,3,4,5,6,7,8,9,10]
```
This is problematic in C since in general array dimension can be specified in C using compile-time values - see https://github.com/cython/cython/pull/5242#issuecomment-1419810587 and https://github.com/cython/cython/pull/5242#discussion_r1106321798

Hence, I propose to generate following C code for upper-given example:
```c
# define LEN_CONST_VALUE 10
static const int len = LEN_CONST_VALUE
static int arr[LEN_CONST_VALUE];
```
> [!Note]
> I am using macro for declaring compilation-time value. But there is also alternative in form of enums - see https://github.com/cython/cython/pull/5242#discussion_r1106321798. 

Current PR supports:
* declaring cost variables and using them
* using `const` variables in array dimension using macro
* declaring `const` variables in `cdef extern from ...` block. Here directly const variables is used even when used as array dimension. Hence, if the users declare const variables from header file, **it can cause compilation error when used as array dimension** - see https://github.com/cython/cython/pull/5242#issuecomment-1703934869
* `cimport`ing `const` variables is supported, but using it as array dimension is forbidden because it causes C compiler errors

Cython/Compiler/ModuleNode.py
```diff
@@ -718,6 +718,12 @@ def generate_type_definitions(self, env, modules, vtab_list, vtabslot_list, code
             self.generate_objstruct_predeclaration(entry.type, code)
         vtabslot_entries = set(vtabslot_list)
         ctuple_names = set()
+        for entry in env.var_entries:
+            type = entry.type
+            cname = entry.cname
+            if entry.type.is_const and entry.init is not None:
+                code.putln("#define %s_CONST_VALUE %s" % (entry.cname.upper(), entry.init))
```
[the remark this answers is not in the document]
matusvalo: > Appending something in the end risks collisions

The infrastracture around generating C code is not final. PR mainly serves as PoC of approach

> Do we really need this C macro?

It is used only to support using `const` values as array dimension see: https://github.com/cython/cython/pull/5242#issuecomment-1712574908

Cython/Compiler/Nodes.py
matusvalo: > Could you explain (or add a comment) why this isn't allowed? It doesn't seem immediately wrong to me if a constant is defined in the module's .pxd file.

This error is raised only when `cimport`ed `const` variable is used as array dimension. Array dimensions in C require compile time value, which is not the case of values `cimport`ed.  In `const` variables declared within module this can be circumvented via values declared as macros - see https://github.com/cython/cython/pull/5242#issuecomment-1712574908

matusvalo: > This doesn't seem to be tested.

Test added.

tests/run/cimport_constants.srctree
```diff
@@ -0,0 +1,70 @@
...
+from Cython.Compiler.Errors import CompileError
+from Cython.Build.Dependencies import cythonize
+
+if sys.version_info[0] < 3:
+    from StringIO import StringIO
+else:
+    from io import StringIO
+
+@contextmanager
+def assert_stderr(expected_errors):
```
matusvalo, line 50: This test is based on following code: https://github.com/cython/cython/blob/1078cc635c164b594f4eff8c2fdc401412b8f0e7/tests/run/relative_import_leak.srctree#L16C3-L27

We should build some common infrastructure for it in some near future I suppose.

matusvalo:
@scoder @da-woods could you have a look on the code? Please focus mainly on the approach how I tackled using const as array dimension. I would like to hear if you are fine with the approach of using macros. Of course, the way how macro name is created needs to change (and also place, we need to add a attribute/method to Const Node object I suppose). When you agree how the C code is generated, I will clean up the python code and reorganise it properly.  

The test suite is almost final maybe some cases can be added.

Cython/Compiler/Nodes.py
```diff
@@ -606,7 +606,17 @@ def analyse(self, base_type, env, nonempty=0, visibility=None, in_pxd=False):
             self.dimension = self.dimension.analyse_const_expression(env)
             if not self.dimension.type.is_int:
                 error(self.dimension.pos, "Array dimension not integer")
-            size = self.dimension.get_constant_c_result_code()
+            size = None
+            if self.dimension.type.is_const:
+                if self.dimension.entry.defined_in_pxd:
+                    # Array dimension requires compile-time value.
+                    # cimported const values cannot be used since it is run-time value.
```
matusvalo, line 613:
```suggestion
                    # cimported const value cannot be used since it is run-time value.
```

da-woods:
So I don't *really* like the defines, but I can't see a good alternative if we want to let consts be used as array lengths (which I think we do). (In C23, and recent c++ you can specify the underlying type of a enum, so you could just have one definition of the constants defined as anonymous enum values. But that doesn't work for us unfortunately).

Can we see any corner cases where the "macros get expanded in place each time" could lead to problems? Probably not with genuinely constant expressions, but maybe I'm not being creative enough.

The other option would be to substitute the values in Cython rather than use macros to do it. I think I like that less.

`Symtab.Entry.next_id` is probably the way to avoid name classes here I think

matusvalo:
> So I don't really like the defines, but I can't see a good alternative if we want to let consts be used as array lengths (which I think we do). (In C23, and recent c++ you can specify the underlying type of a enum, so you could just have one definition of the constants defined as anonymous enum values. But that doesn't work for us unfortunately).

Yes I understand your point and that's why I am bringing this up before finishing the PR. Other option is just to forbid users to use const variables in array dimension with exception of const variables declared in `cdef extern from ...` block.

Maybe a thought. Does it make sense to have first version of const variables not supporting array dimensions except const variables declared in `cdef extern from ...` block. We will see how many users will need the array dimension support. What do you think? I would like to hear also @scoder opinion...

matusvalo:
> Can we see any corner cases where the "macros get expanded in place each time" could lead to problems?

This is checked by both cython and C so I am not so much affraid having wrong values in macro. E.g. following example:

```cython
cdef const int a = 'asdf'
```
will error out in cython during compilation:
```
Error compiling Cython file:
...
cdef const int a = 'asdf'
                   ^

deleteme.pyx:1:19: Unicode literals do not support coercion to C types other than Py_UNICODE/Py_UCS4 (for characters) or Py_UNICODE* (for strings).
```

And following cython code:
```cython
cdef const int a = 9999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999
```
will error out in C level:
```
running build_ext
building 'deleteme' extension
creating /Users/matus/dev/cython/test/tmpiy2yr7we/Users
creating /Users/matus/dev/cython/test/tmpiy2yr7we/Users/matus
creating /Users/matus/dev/cython/test/tmpiy2yr7we/Users/matus/dev
creating /Users/matus/dev/cython/test/tmpiy2yr7we/Users/matus/dev/cython
creating /Users/matus/dev/cython/test/tmpiy2yr7we/Users/matus/dev/cython/test
clang -Wno-unused-result -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -I/usr/local/include -I/usr/local/opt/openssl@1.1/include -I/usr/local/opt/sqlite/include -I/Users/matus/dev/cython39/include -I/usr/local/opt/python@3.9/Frameworks/Python.framework/Versions/3.9/include/python3.9 -c /Users/matus/dev/cython/test/deleteme.c -o /Users/matus/dev/cython/test/tmpiy2yr7we/Users/matus/dev/cython/test/deleteme.o
/Users/matus/dev/cython/test/deleteme.c:1682:40: error: integer literal is too large to be represented in any integer type
static int const __pyx_v_8deleteme_a = __PYX_V_8DELETEME_A_CONST_VALUE;
                                       ^
/Users/matus/dev/cython/test/deleteme.c:1431:41: note: expanded from macro '__PYX_V_8DELETEME_A_CONST_VALUE'
#define __PYX_V_8DELETEME_A_CONST_VALUE 9999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999
                                        ^
/Users/matus/dev/cython/test/deleteme.c:1682:40: warning: implicit conversion from 'unsigned long long' to 'int' changes value from 18446744073709551615 to -1 [-Wconstant-conversion]
static int const __pyx_v_8deleteme_a = __PYX_V_8DELETEME_A_CONST_VALUE;
                 ~~~~~~~~~~~~~~~~~~~   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/matus/dev/cython/test/deleteme.c:1431:41: note: expanded from macro '__PYX_V_8DELETEME_A_CONST_VALUE'
#define __PYX_V_8DELETEME_A_CONST_VALUE 9999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999
                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning and 1 error generated.
```

da-woods:
What I was really worried about is if you could engineer a situation where the macro expansion has side effects. E.g. (untested, and not Cython code)

```
static int a = 0;

const int b = ++a

if (b) { ... }

if (b) { ... }
```

This will end up with a different value of `a` if every expansion of `b` is replaced with `++a`.

Right now, I can't see how you would do this in Cython (at least without hacky misleading C code in `cdef extern` blocks). But it's also possible I'm not being creative enough.

matusvalo:
> at least without hacky misleading C code in cdef extern blocks

I think even right now you can break easily cython using `cdef extern` block so I consider using it as "low level approach only on users own responsibility"

> What I was really worried about is if you could engineer a situation where the macro expansion has side effects.

The macro is assigned **only** during initialisation and the value is guarded by cython - see following lines in this PR:

https://github.com/cython/cython/pull/5242/files#diff-8789a479ea870f7646251224385afa3ebffe9b6acab136ff55011b31386ddd84R6178-R6190

Hence, I am not sure whether there is a way of corrupting macro here (but I am not C expert though).

I am more concerned about inconsistency fo cimported const values which cannot be used as array dimension but I think there is nothing that can be done about that.

da-woods:
> I am more concerned about inconsistency fo cimported const values which cannot be used as array dimension but I think there is nothing that can be done about that.

That's probably not *quite* as bad as it seems - I think it mostly will generate valid C++ code, and valid but slightly inefficient C99 code. So it's only really strict c89 where it's likely to be a problem.

scoder:
> you can break easily cython using `cdef extern` block

It's not about letting users break things on purpose or by writing visibly risky code. But we must not make it easy for users to generate brittle C code *by accident*.

When I write `cdef const int x = …` (really, whatever there is on the right side, be it a literal or a complex expression), then I want the value to be determined at exactly this point in the program. It would be very surprising (and very dangerous) to evaluate it at a different time or even re-evaluate it multiple times. And that is difficult to achieve with the mix of Cython translation, C preprocessor and C compiler steps that contribute to the interpretation of the code.

Honestly, if we cannot achieve to build something better than `DEF`, then maybe `DEF` needs to stay for another while.

matusvalo:
At first thank you @da-woods and @scoder for helping me out and your feedback

> But we must not make it easy for users to generate brittle C code by accident.

I agree as as I mentioned earlier, my understanding is that user using this PR can break generated code only by `cdef extern` block which is as I said already potential dangerous. When you use Cython standard syntax, the Cython is ensuring that **only** literal is assigned to the const variable/macro:

> The macro is assigned only during initialisation and the value is guarded by cython - see following lines in this PR:

> https://github.com/cython/cython/pull/5242/files#diff-8789a479ea870f7646251224385afa3ebffe9b6acab136ff55011b31386ddd84R6178-R6190

But honestly I cannot see problem here. But I don't have so much experience with Cython/C codebase as you so please correct me if I am wrong.

If you still insist that macro is wrong way, can we use enums instead of macros as proposed by @da-woods here https://github.com/cython/cython/pull/5242#issuecomment-1420420380?

enums can contain *only* integer value similarly to array dimension so the alternative aproach is to create enums only for constants with integer values? What do you think? Please let me know, I would like to understand if the general approach is good before I spend time with development.

da-woods:
The issue with enums is that before C23 they only work when the type is an exact `int`. In c++ and in C23 onwards they pick a suitable `int` type that will fit as the underlying type of the enum. So that turns into a fairly big limitation

PR #5257, opened 2023-02-25
add missing exception C-APIs

maxbachmann:


0dminnimda:
It's the main that made the tests fail

Cython/Includes/cpython/exc.pxd
```diff
@@ -88,6 +88,11 @@ cdef extern from "Python.h":
     # needs to handle exceptions or by code that needs to save and
     # restore the error indicator temporarily.
 
+    PyObject* PyErr_GetHandledException()
+    void PyErr_SetHandledException(PyObject* exc)
+    PyObject* PyErr_GetRaisedException()
+    void PyErr_SetRaisedException(PyObject* exc)
```
scoder, line 94: I checked the documentation and the implementation. For the last two functions, the documentation doesn't say it, but the implementation allows `NULL` to be returned and passed, so `PyObject*` is correct here (and not `object`).
This is also the case for the first two functions that you added above.

scoder (commented):
Looks all good to me. Don't think we need tests for these, and also, they are really recent Py3.11/12).

PR #5262, opened 2023-02-26
Add libcpp.cmath

jhelgert:
This PR adds the C++ STL [cmath header](https://en.cppreference.com/w/cpp/header/cmath). It's mainly motivated by the new C++17 mathematical special functions (like `assoc_laguerre`, `assoc_legendre`, `beta`, ...) and the C++20 linear interpolation function `lerp`. However, all other functions are already in libc.math, so I'm not really sure if we should have them in both libc.math and libcpp.cmath.


da-woods:
I'm not quite sure why the Python 2.7 tests are failing. Easiest thing is probably to add them to `VER_DEP_MODULES` in runtests.py to skip them there.

tests/run/cpp_stl_cmath_cpp17.pyx
```diff
@@ -0,0 +1,34 @@
...
+
+def test_hypot(double x, double y, double z):
+    """
+    Test C++17 std::hypot function
+    >>> test_hypot(1.0, 2.0, 2.0)
+    3.0
+    >>> test_hypot(3.0, 4.0, 0.0)
+    5.0
+    """
+    return hypot(x, y, z)
```
scoder, line 34: EOL missing
```suggestion
    return hypot(x, y, z)
```

tests/run/cpp_stl_cmath_cpp20.pyx
```diff
@@ -0,0 +1,13 @@
...
+from libcpp.cmath cimport lerp
+
+def test_lerp(double a, double b, double t):
+    """ Test C++20 std::lerp function
+    >>> test_lerp(1.0, 2.0, 0.5)
+    1.5
+    >>> test_lerp(1.0, 4.0, 0.5)
+    2.5
+    """
+    return lerp(a, b, t)
```
scoder, line 13:
```suggestion
    return lerp(a, b, t)
```

tests/macos_cpp_bugs.txt
```diff
@@ -13,3 +13,5 @@ cpp_stl_algo_permutation_ops
 cpp_stl_algo_sorted_ranges_set_ops
 cpp_stl_algo_sorted_ranges_other_ops
 cpp_stl_bit_cpp20
+cpp_stl_cmath_cpp20
+cpp_stl_cmath_cpp17
```
scoder, line 17: If `std::beta()` is the issue here, then could you use a different function in this test? Then we could at least enable the C++17 tests on macOS.

scoder (approved):
Nice addition, just a few comments.

tests/macos_cpp_bugs.txt
jhelgert: I double-checked it on my Mac and it turned out Apple Clang doesn't support the C++17 mathematical special functions yet, see [here](https://en.cppreference.com/w/cpp/compiler_support/17). :/ The only exception is the trivial `hypot` function.

h-vetinari: Not just apple clang, but clang itself. It's not around the corner [either](https://reviews.llvm.org/D142806).

da-woods:
Just rerunning tests - I think the failure was just a github actions glitch and not to do with this PR...

tests/run/cpp_stl_cmath_cpp17.pyx
```diff
@@ -0,0 +1,35 @@
...
+
```
0dminnimda, line 35:
```suggestion
```

0dminnimda, line 35: image

```suggestion
```

0dminnimda:
Huh, it didn't help, interesting

scoder:
I see this crash in the logs:
```
Doctest: cpp_stl_cmath_cpp17.test_legendre ... terminate called after throwing an instance of 'std::domain_error'
  what():  Argument out of range in __poly_legendre_p.
Fatal Python error: Aborted
```

da-woods:
> I see this crash in the logs: [...]

Which suggests at least some of these functions need an `except+` specification...

jhelgert:
> Right. I'm aware that there's a whole bunch of functions declared here, but someone needs to go through them and add an `except +` declaration to all those that can raise C++ exceptions.

I had a look at it and it turned out that the C++ standard allows the implementation to strengthen the exception specification of the cmath functions by adding `noexcept`, see [this SO answer](https://stackoverflow.com/a/57513587/4745529). For instance, [`std::hypot`](https://en.cppreference.com/w/cpp/numeric/math/hypot) is noexcept on Apple Clang, while it throws an exception on gcc and the Standard doesn't mark it as noexcept.

Should we still add the `except +` declaration to all those functions that can raise a C++ exception on *any* Compiler?

da-woods:
So for the `legendre` function specificity

> The function is not required to be defined for |x|>1

https://en.cppreference.com/w/cpp/numeric/special_functions/legendre. Which suggests your tests with `x==2` and `x==3` may not be the best defined.

> Should we still add the except + declaration to all those functions that can raise a C++ exception on any Compiler?

I think so yes (based on looking at the standard/cppreference rather than picking through each compiler's implementation). I think a lot of these functions do have domains where they work and raising an exception outside the domain is pretty reasonable.

jhelgert:
> So for the `legendre` function specificity
> 
> > The function is not required to be defined for |x|>1
> 
> https://en.cppreference.com/w/cpp/numeric/special_functions/legendre. Which suggests your tests with `x==2` and `x==3` may not be the best defined.

Good catch! Thanks for the hint. After declaring all those functions that aren't `noexcept` with `except +`, all test are passing now.

Cython/Includes/libcpp/cmath.pxd
```diff
@@ -0,0 +1,518 @@
...
+    double tan(double x) except +
+    long double tan(long double x) except +
+    float tanf(float x) except +
+    long double tanl(long double x) except +
+    
+    float acosh(float x) except +
+    double acosh(double x) except +
+    long double acosh(long double x) except +
+    float acoshf(float x)
+    long double acoshl(long double x)
```
da-woods, line 50:
```suggestion
    float acoshf(float x) except +
    long double acoshl(long double x) except +
```

da-woods:
Thanks for this

PR #5276, opened 2023-02-27
Make Grammer easier to understand by using .lark

rickmark:


0dminnimda:
This is not lark grammar specifically, but a general EBNF syntax

scoder:
I don't think this is worth merging.

PR #5336, opened 2023-03-26
Allow InterpretCompilerDirectives.visit_AnnotationNode to process nodes having no children

garychia:
Related to #5235
I discovered that the following code did not work with nodes having no child nodes, such as `NameNode`.
https://github.com/cython/cython/blob/0ffa7e91816a00a1694d1466aa91613b6708b244/Cython/Compiler/ParseTreeTransforms.py#L1106-L1111
```python
from cython cimport int as cy_i
cimport cython

def f(a: cython.int, b: cy_i):
    print(cython.typeof(a), cython.typeof(b))
```
(Code from #5235)
It caused the `NameNode` of `cy_i` to be ignored by `InterpretCompilerDirectives` when it was referenced by `node.expr`.


scoder:
Nice catch, thanks.

PR #5341, opened 2023-03-27
fix compilation with language="C++" (wrong capitalization)

### LINKED ISSUE #1226: linking not done using g++ when language specified as "C++" (not "c++")
Hello,

When language="C++" (wrong capitalization) is passed to an extension type, the cython files are indeed compiled using C++ mode. However, the final linking is done using the C compiler, causing the C++ std libs to not be linked in.  It appears that in some places,  `extension.language.lower() == "c++"` is used to determine the language type, but converting to lower case isn't done universally.  

Migrated from http://trac.cython.org/ticket/667

EpigeneMax:
tackle the very few places where `language` was compared to `"c++"` without forcing lower capitalization

extend an existing test rather than creating a new one.

closes #1226

### DIFF
--- a/Cython/Build/Dependencies.py
+++ b/Cython/Build/Dependencies.py
@@ -875,7 +875,7 @@ def create_extension_list(patterns, exclude=None, ctx=None, aliases=None, quiet=
 
                 if file not in m.sources:
                     # Old setuptools unconditionally replaces .pyx with .c/.cpp
-                    target_file = os.path.splitext(file)[0] + ('.cpp' if m.language == 'c++' else '.c')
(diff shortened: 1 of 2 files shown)


scoder:
Thanks. However, as I wrote in the ticket, it's probably a distutils/setuptools issue rather than a Cython issue. Meaning, it's not enough to fix Cython here, support would also have to be added to setuptools.

That said, for Cython's `# distutils: language=…` option, it's probably reasonable to lower-case the value that we pass on into distutils. I think that would be a better fix than lower-casing the value each time we compare it to something.

PR #5349, opened 2023-04-01
Fix reference leak when a for-loop's "else:" branch contains a "break"

scoder:
Closes https://github.com/cython/cython/issues/5347


scoder:
Implemented in https://github.com/cython/cython/commit/4c42949b47eff1bce921fae71ed93ec359dd32a4

PR #5355, opened 2023-04-04
Fix AnnotationWriter for IndexNode with empty TupleNode

dalcinl:
I hit this issue while adding argument annotations like `tuple[()]`. The annotation writer would emit `tuple[]`, which is invalid syntax. Despite the context being related to annotations, I believe the fix goes to the base `ExpressionWriter` class, so I did.

### DIFF
--- a/Cython/CodeWriter.py
+++ b/Cython/CodeWriter.py
@@ -644,7 +644,10 @@ class ExpressionWriter(TreeVisitor):
         self.visit(node.base)
         self.put(u"[")
         if isinstance(node.index, TupleNode):
-            self.emit_sequence(node.index)
+            if node.index.subexpr_nodes():
+                self.emit_sequence(node.index)
+            else:
+                self.put(u"()")
         else:
             self.visit(node.index)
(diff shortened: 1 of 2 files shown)


dalcinl:
Perhaps this fix should be backported to the `0.29.x` branch?

scoder:
This is quite clearly missing from the test suite. Could you please add a test for it?

scoder:
> Perhaps this fix should be backported to the `0.29.x` branch?

Seems reasonable, but I think there were major changes in the annotation code writer. Not sure if it's worth it.

dalcinl:
I'm not quite sure what kind of testing you have in mind. I've added one related to embedding signatures, which was the context where I discovered the issue.

scoder:
> I've added one related to embedding signatures, which was the context where I discovered the issue.

If the test failed before and now succeeds, this is good enough. I just wanted to be sure that the code is exercised *somewhere* in the test suite.

scoder:
To be included in 3.0.0b3 and 0.29.35.

PR #5358, opened 2023-04-04
Fix duplication of `IS_UNSIGNED_IMPL`

### LINKED ISSUE #5356: [BUG] Error at compilation with `overflowcheck(True)`.
### Describe the bug

I get an error at compilation when using `overflowcheck(True)` with 3.0.0b2.
Prior version worked well.

```cython
%%cython --cplus

cimport cython

@cython.overflowcheck(True)
cdef size_t _mul_checked(size_t a, size_t b) except? -1:
    return a * b

def f(size_t[:] a, size_t[:] b):
    return _mul_checked(a.shape[0], b.shape[0])
```
Output:
```
Content of stderr:
/home/eric/.cache/ipython/cython/_cython_magic_7ec89a75959ed7157c2554d3336a3feedfc6cd9d.cpp:2382:8: error: redefinition of ‘struct __PYX_IS_UNSIGNED_IMPL<T>’
 2382 | struct __PYX_IS_UNSIGNED_IMPL {static const bool value = T(0) < T(-1);};
      |        ^~~~~~~~~~~~~~~~~~~~~~
/home/eric/.cache/ipython/cython/_cython_magic_7ec89a75959ed7157c2554d3336a3feedfc6cd9d.cpp:1329:8: note: previous definition of ‘struct __PYX_IS_UNSIGNED_IMPL<T>’
 1329 | struct __PYX_IS_UNSIGNED_IMPL {static const bool value = T(0) < T(-1);};
      |        ^~~~~~~~~~~~~~~~~~~~~~
```

### Expected behaviour

_No response_

### OS

Linux

### Python version

3.10.0

### Cython version

3.0.0b2

### Additional context

_No response_

0dminnimda:
Closes #5356, Fixes #5302

### DIFF
--- a/Cython/Utility/Buffer.c
+++ b/Cython/Utility/Buffer.c
@@ -53,7 +53,7 @@ static void __Pyx_RaiseBufferFallbackError(void) {
 
 /////////////// BufferFormatStructs.proto ///////////////
 //@proto_block: utility_code_proto_before_types
-//@requires: Overflow.c::Common
+//@requires: CMath.c::IsUnsigned
 
 /* Run-time type information about structs used with buffers */
 struct __Pyx_StructField_;
--- a/Cython/Utility/Buffer.c
+++ b/Cython/Utility/Buffer.c
@@ -53,14 +53,7 @@ static void __Pyx_RaiseBufferFallbackError(void) {
 
 /////////////// BufferFormatStructs.proto ///////////////
 //@proto_block: utility_code_proto_before_types
-
-#ifdef __cplusplus
-template<typename T>
-struct __PYX_IS_UNSIGNED_IMPL {static const bool value = T(0) < T(-1);};
-#define __PYX_IS_UNSIGNED(type) __PYX_IS_UNSIGNED_IMPL<type>::value
-#else
-#define __PYX_IS_UNSIGNED(type) (((type) -1) > 0)
-#endif
+//@requires: Overflow.c::Common
 
 /* Run-time type information about structs used with buffers */
 struct __Pyx_StructField_;
--- a/Cython/Utility/Buffer.c
+++ b/Cython/Utility/Buffer.c
@@ -53,7 +53,6 @@ static void __Pyx_RaiseBufferFallbackError(void) {
 
 /////////////// BufferFormatStructs.proto ///////////////
 //@proto_block: utility_code_proto_before_types
-//@requires: CMath.c::IsUnsigned
 
 /* Run-time type information about structs used with buffers */
 struct __Pyx_StructField_;
--- a/Cython/Utility/CMath.c
+++ b/Cython/Utility/CMath.c
@@ -101,9 +101,9 @@ static CYTHON_INLINE %(type)s %(func_name)s(%(type)s b, %(type)s e) {
     #ifdef __cplusplus
         template <typename T>
         struct __PYX_IS_UNSIGNED_IMPL {static const bool value = T(0) < T(-1);};
-        #define __PYX_IS_UNSIGNED(type) __PYX_IS_UNSIGNED_IMPL<type>::value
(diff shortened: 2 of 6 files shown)


0dminnimda:
It is c++ only issue, but I see no reason for why not to have this test checks for c

Cython/Utility/Buffer.c
```diff
@@ -53,14 +53,7 @@ static void __Pyx_RaiseBufferFallbackError(void) {
...
 //@proto_block: utility_code_proto_before_types
-
-#ifdef __cplusplus
-template<typename T>
-struct __PYX_IS_UNSIGNED_IMPL {static const bool value = T(0) < T(-1);};
-#define __PYX_IS_UNSIGNED(type) __PYX_IS_UNSIGNED_IMPL<type>::value
-#else
-#define __PYX_IS_UNSIGNED(type) (((type) -1) > 0)
-#endif
+//@requires: Overflow.c::Common
```
scoder, line 56: I don't think we need the `__Pyx_check_twos_complement()` part in `Common.c` for this. Thus, it might be worth extracting the unsigned checks into a new, separate utility code section and let both existing utility code sections depend on that.

0dminnimda: You're right, I've seen this and thought that it probably won't cause any problems, but I'm not sure about it so this seems like a better option

0dminnimda:
Also pypy-2.7's memoryview doesn't have the buffer protocol, I'd like to keep this test isolated, so no np/cython/python arrays
Should I put it into exceptions or ignore 2.7 testing at all?

mattip:
> pypy-2.7's memoryview doesn't have the buffer protocol

Weird, I wonder where [that error](https://github.com/cython/cython/actions/runs/4611674873/jobs/8151671087?pr=5358#step:5:54501) is coming from. In general, I don't think testing against PyPy2.7 is going to be productive. The PyPy team (me) is unlikely to fix missing features or bugs in the 2.7 cpyext layer (the compatibility layer for the CPython C-API) at this point. Perhaps it should be removed from CI?

0dminnimda:
Ok, I'll just disable pypy2 testing for this testcase

Cython/Utility/CMath.c
```diff
@@ -93,3 +93,20 @@ static CYTHON_INLINE %(type)s %(func_name)s(%(type)s b, %(type)s e) {
     }
     return t;
 }
+
+
+/////////////// IsUnsigned.proto ///////////////
```
scoder, line 98: I think `ModuleSetupCode.c` is a better place for this (`CMath.c` definitely is not), maybe before the `CInitCode` section.

Cython/Utility/CMath.c
```diff
@@ -93,3 +93,20 @@ static CYTHON_INLINE %(type)s %(func_name)s(%(type)s b, %(type)s e) {
...
+
+#ifndef __PYX_IS_UNSIGNED
+    #ifdef __cplusplus
+        template <typename T>
+        struct __PYX_IS_UNSIGNED_IMPL {static const bool value = T(0) < T(-1);};
+        #define __PYX_IS_UNSIGNED(type) __PYX_IS_UNSIGNED_IMPL<type>::value
```
scoder, line 104:
```suggestion
        #define __PYX_IS_UNSIGNED(type) (__PYX_IS_UNSIGNED_IMPL<type>::value)
```

scoder (commented):
Actually, seeing how widely the macros seem to be used, I'd now suggest simply moving it to the end of the `CModulePreamble` section in `ModuleSetupCode.c`, without giving it a separate section title. It seems to be almost as general as many of the other macros in that section. And it doesn't hurt to have it in all C files.

scoder:
Thanks

PR #5369, opened 2023-04-10
Add .readthedocs.yaml

0dminnimda:
Backport #5254

See #5364


0dminnimda:
Well, I guess there is not way around cherry picking

0dminnimda:
Well, I guess there is no way around cherry picking

PR #5371, opened 2023-04-11
Add a "CYTHON_EXTERN_C" macro to allow redefining "__PYX_EXTERN_C" on user side

scoder:
See https://github.com/cython/cython/pull/5366


scoder:
> If this is the preferred way to set it then it might be worth updating https://github.com/cython/cython/blob/c8dd3d0b18510afb9963a8d5c3d9819694a7da50/tests/run/cpp_extern.srctree to reflect that

Thanks, I updated the test.

While doing so, I noticed that the module names in that test (foo, bar, baz) were really badly chosen, because they did not give any hints at what's wrong if one of the three test parts fails. I tried to come up with better names, but it's always better to do that right from the start since it helps in drawing a clean line between test cases.

PR #5394, opened 2023-04-19
Fix GCC -Wsign-conversion warning

dalcinl:
This is a minor code generation fix after changes from @scoder in 781b087c90817a312e3dc6dd55ef174a798b98d8.


scoder:
Thanks, but I don't think the macro really needs changing. I pushed a fix that fixes the usage instead.

dalcinl:
IIRC, `ssize_t` is in POSIX, but not C99. Maybe you should have used `Py_ssize_t`?
https://github.com/python/cpython/blob/f8abfa331421e2c28388c5162f37820495e3c2ee/Include/pyport.h#L123

scoder:
Right. Too much editing back and forth, and I forgot to use the obvious type in the end. Fixed in master.

PR #5417, opened 2023-05-01
Fix usage of _MSC_VER macro

matttyson:
Change #if to #ifdef so we don't get undefined macro warnings on non microsoft compilers.

Fix typo in macro.


scoder:
Thanks

PR #5421, opened 2023-05-02
Silence GCC -Wsign-conversion when using CYTHON_LIMITED_API

dalcinl:


scoder (approved):
Looks good. If CI is all happy, too, feel free to merge this. (GHA and builds logs are currently down, so this may take a while and require some job restarts.)

Note that we don't currently run any file tests in the limited API, but maybe the warnings also show when compiling Cython itself? Would be good to check if they are really gone (and everything is fine) for this PR, once CI is back.

dalcinl:
Running 
```
$ export CFLAGS='-O0 -Wp,-U_FORTIFY_SOURCE -Wconversion -DCYTHON_LIMITED_API=0x030B0000'
$ make
```
no longer generates any `-W[sign-]conversion` warning.

However, the `-DCYTHON_LIMITED_API=0x030B0000` define triggers the following unused warnings.
```

‘__Pyx_GetVtable’ defined but not used [-Werror=unused-function]
‘__Pyx__CallUnboundCMethod0’ defined but not used [-Werror=unused-function]
‘__pyx_doc_6Cython_4Plex_8Scanners_7Scanner___init__’ defined but not used [-Wunused-const-variable=]
```

Addressing these warnings should be a matter of another PR, I don't have time to work on it right now.

Cython/Compiler/Builtin.py
```diff
@@ -233,7 +233,7 @@ def declare_in_type(self, self_type):
     #('sum',       "",     "",      ""),
     #('sorted',    "",     "",      ""),
     #('type',       "O",    "O",     "PyObject_Type"),
-    BuiltinFunction('unichr',     "l",    "O",      "PyUnicode_FromOrdinal", builtin_return_type='unicode'),
+    BuiltinFunction('unichr',     "i",    "O",      "PyUnicode_FromOrdinal", builtin_return_type='unicode'),
```
dalcinl, line 236: @scoder Sorry, but I'm slightly puzzled about this... 

The C function prototype is 
```c
PyAPI_FUNC(PyObject*) PyUnicode_FromOrdinal(int ordinal);
```
however `BuiltinFunction` was getting arg type code `"l"` (`long`)  and not `"i"` (`int`). 
Is my fix correct or am I missing something?

dalcinl: My fix is correct.

scoder:
> triggers the following unused warnings

There are certainly worse problems with the limited API support currently than the inclusion of unused utility functions.

PR #5422, opened 2023-05-02
Support PyBufferProcs with limited C-API under Py3.9+ or Py3.11+

dalcinl:


Cython/Compiler/TypeSlots.py
```diff
@@ -268,6 +268,9 @@ def generate_spec(self, scope, code):
         if value == "0":
             return
         preprocessor_guard = self.preprocessor_guard_code()
+        if not preprocessor_guard:
+            if self.py3 and self.slot_name.startswith('bf_'):
+                preprocessor_guard = "#if defined(Py_%s)" % self.slot_name
```
scoder, line 273: Shouldn't this be part of the slot declarations, rather than special-casing them in this generic function?

dalcinl: I'm not sure how to do it without overcomplicating things.  Note that I removed three lines in another rather generic function that was special casing `if self.slot_name == "tp_as_buffer"`
So far, the `Py_bf_{get|release}buffer` values are the ones that are special. Well, we also have `Py_am_send`. And in the future we may get more. So yes, you could do better.

I can only think of two possible ways to make it better:

* Option 1: We add a `guard=False` argument (like if `ifdef`, `py3`, etc.), and we explicitly pass `guard=True` at slot declaration site. This `guard`  boolean will control whether to guard for the availability of the `Py_{slog_name}` . This will require supporting the extra argument in a few places in `TypeSlots.py`. BTW, feel free to suggest a better name for this boolean argument. Things would look the following way:
```python
MethodSlot(getbufferproc, "bf_getbuffer", "__getbuffer__", method_name_to_slot,
           use_guard = True),
MethodSlot(releasebufferproc, "bf_releasebuffer", "__releasebuffer__", method_name_to_slot,
           use_guard = True),
```

* Option 2: unconditionally guard all `PyType_Slot` entries, despite the fact that most of them would not need a guard. This is totally generic, future-proof, and zero implementation/maintenance effort, at the cost of generating some extra `#ifdef/endif` lines in C sources.

Just pick your preferred option, or suggest an alternative one of your own, and I'll proceed to implement it.

dalcinl: @scoder I really need your input here to make a choice about implementation as per my comment above.

scoder, line 273:
```suggestion
                # The buffer protocol requires Limited API 3.11, so check if the spec slots are available.
                preprocessor_guard = "#if defined(Py_%s)" % self.slot_name
```

scoder (approved):
I think your change is as good as it gets. I just updated the warning messages and added a comment to make things clearer.

scoder:
Thanks!

Cython/Compiler/TypeSlots.py
```diff
@@ -268,6 +268,10 @@ def generate_spec(self, scope, code):
         if value == "0":
             return
         preprocessor_guard = self.preprocessor_guard_code()
+        if not preprocessor_guard:
+            if self.py3 and self.slot_name.startswith('bf_'):
```
dalcinl, line 272: @scoder Although unrelated to buffers, maybe we should extend
```python
    prefix = self.slot_name[:3]
    if self.py3 and (prefix == 'bf_' or prefix == 'am_'):
        ...
```
This will account for the addition of `Py_am_send` to the limited API v3.10.

```diff
$ diff /usr/include/python3.9/typeslots.h  /usr/include/python3.10/typeslots.h 
90a91,94
> #if !defined(Py_LIMITED_API) || Py_LIMITED_API+0 >= 0x030A0000
> /* New in 3.10 */
> #define Py_am_send 81
> #endif
```

dalcinl:
> There's already one for am_send, although not complete yet.

Maybe you rebase the am_send PR and can sneak in my suggestion?

PR #5424, opened 2023-05-05
[0.29.x] Support `except + nogil` statement

### LINKED ISSUE #5423: [BUG] 0.29.X Cython does not support `nogil` keyword after `except +` statement.
### Describe your issue

# Bug description

Cython 3 warns when following function is declared (real example from scipy):
```cython
    int build_weights(ckdtree *self,
                         np.float64_t *node_weights,
                         np.float64_t *weights) nogil except +
```
Warn message is as follows:
```
/home/runner/work/scipy/scipy/scipy/spatial/_ckdtree.pyx:75:53: The keyword 'nogil' should appear at the end of the function signature line. Placing it before 'except' or 'noexcept' will be disallowed in a future version of Cython.
```

When the order of `except +` and `nogil` is reversed:
```cython
    int build_weights(ckdtree *self,
                         np.float64_t *node_weights,
                         np.float64_t *weights) except + nogil
```
Cython 3 is not emiting the warning but Cython 0.29.X is failing with following exception:
```
 Error compiling Cython file:
...
                         np.intp_t start_idx,
                         np.intp_t end_idx,
                         np.float64_t *maxes,
                         np.float64_t *mins,
                         int _median,
                         int _compact) except + nogil
                                                    ^

/home/runner/work/scipy/scipy/scipy/spatial/_ckdtree.pyx:75:53: undeclared name not builtin: nogil
```

# Rationale

Until Cython 3 is not fully released, projects are gradually migrating to Cython3 by supporting both versions. Current bug is blocking this approach since the syntax is preferred in Cython 3 but not working in Cython 0.29.X.

# See also
Discussion in Scipy PR fixing compilation warnings: https://github.com/scipy/scipy/pull/18418#issuecomment-1533746881

# Describe alternatives you've considered.

Other possibility is not fixing this in 0.29.X. Projects can fix the warnings when fully migrate to Cython 3. In this case we need to be sure that support of `nogil except ...` will be removed *after* all users migrate to Cython 3.

matusvalo:
Based on: 77918c57b5

Closes #5423

### DIFF
--- a/Cython/Compiler/Parsing.pxd
+++ b/Cython/Compiler/Parsing.pxd
@@ -163,7 +163,7 @@ cdef p_c_simple_declarator(PyrexScanner s, ctx, bint empty, bint is_type, bint c
                           bint assignable, bint nonempty)
 cdef p_nogil(PyrexScanner s)
 cdef p_with_gil(PyrexScanner s)
-cdef p_exception_value_clause(PyrexScanner s)
+cdef p_exception_value_clause(PyrexScanner s, ctx)
 cpdef p_c_arg_list(PyrexScanner s, ctx = *, bint in_pyfunc = *, bint cmethod_flag = *,
                    bint nonempty_declarators = *, bint kw_only = *, bint annotated = *)
 cdef p_optional_ellipsis(PyrexScanner s)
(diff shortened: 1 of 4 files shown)


scoder:
Now that I see this, we probably allow `except +nogil`, without space. Not necessarily something to fix in this PR, but we should distinguish a user provided exception handling function called `nogil` from the `nogil` modifier (i.e. `except + nogil`). Or at least reject it and require a space.

scoder:
Superseded by https://github.com/cython/cython/commit/fa9b2e92301ce9fc61a802ba11c6183d2688e20c

PR #5431, opened 2023-05-11
Comment updates

goodship1:


scoder (commented):
Hmm. Thanks. I usually try to avoid global style updates because they hurt the (git) history. I'm aware that git can ignore certain commits, but that's only a local user/github setting, not something that you can store in a repo for everyone.

If we accept this change, we might as well run `blue` over the code base next.

Cython/Compiler/Builtin.py
```diff
@@ -1,5 +1,5 @@
 #
-#   Builtin Definitions
+#   Builtin Definitions.
```
scoder, line 2: This is a headline.
```suggestion
#   Builtin Definitions
```

PR #5432, opened 2023-05-12
Prevent calling the dealloc slot of a non-GC base class with GC tracking enabled

scoder:
This shows warnings in CPython (3.12) debug builds and can lead to crashes when GC triggers on an object while deallocating it. This is also an issue in 0.29.x, thus, needs to be backported.

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -1761,28 +1761,27 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
                                    clear_before_decref=True, have_gil=True)
 
         if base_type:
+            base_cname = base_type.typeptr_cname
             if needs_gc:
                 # The base class deallocator probably expects this to be tracked,
                 # so undo the untracking above.
-                if base_type.scope and base_type.scope.needs_gc():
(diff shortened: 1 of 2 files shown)


Cython/Compiler/ModuleNode.py
```diff
@@ -1761,28 +1761,27 @@ def generate_dealloc_function(self, scope, code):
...
+            base_cname = base_type.typeptr_cname
             if needs_gc:
                 # The base class deallocator probably expects this to be tracked,
                 # so undo the untracking above.
-                if base_type.scope and base_type.scope.needs_gc():
-                    code.putln("PyObject_GC_Track(o);")
+                if base_type.scope:
+                    # Assume that we know whether the base class uses GC or not.
+                    if base_type.scope.needs_gc():
+                        code.putln("PyObject_GC_Track(o);")
```
scoder, line 1771: Splitting the condition is not strictly required, but seems reasonable – assuming that we have a complete view of the base type. We may consider keeping the conditions together in 0.29.x, at the cost of a bit more runtime checking, but a higher chance of not missing a possibly required `PyObject_GC_Track()` call. Can't say if there really are such cases, though…

PR #5437, opened 2023-05-15
PyPy < 3.10 needs tp_print

### LINKED ISSUE #5436: PyTypeObject fields on PyPy are not consistent with CPython
### Describe your issue

I have been trying to get cython to handle the differences in `PyTypeObject` between the different PyPy versions, but there are still `-Wmissing-field-initializers` warnings. The various attempts were #4509, #4517. Currently the problem is the `tp_print` field, which was removed for CPython3.9. It is still there in PyPy3.9. PyPy3.10 removed both the `tp_print` field and the `tp_pypy_flags` field, so finally the struct is similar to the CPython3.10 one.

I thought I could sidestep the problem in #5435 by using designated initializers. That PR is still not passing CI, and @scoder recommended merging it only for 3.1 since it requires C99 compilers. 

<details><summary>status of C99 designated initializers in various compilers</summary>

FWIW, gcc has [supported designated initializers since v3.0](https://gcc.gnu.org/c99status.html) (about half-way down the page) and MSVC supports them since VS 2019 16.1](https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170). Clang does [not list a version number](https://clang.llvm.org/c_status.html) for support of the feature. 

</details>

So I am left with adding yet another preprocessor macro around the `tp_print` field in `PyTypeObject`. Maybe I should collapse the multiple duplicate long conditional expressions into a `CYTHON_HAS_TP_PRINT` macro.

mattip:
Fixes #5436. PyPy < 3.10 still has a `tp_print` field, where CPython removed it for 3.9

### DIFF
--- a/Cython/Utility/ModuleSetupCode.c
+++ b/Cython/Utility/ModuleSetupCode.c
@@ -557,17 +557,9 @@
 #endif
 
 #if CYTHON_COMPILING_IN_PYPY == 1
-  #if PY_VERSION_HEX >= 0x030800b4 && PY_VERSION_HEX < 0x030A0000
-    #define CYTHON_NEED_TP_PRINT 1
-  #else
-    #define CYTHON_NEED_TP_PRINT 0
-  #endif
+  #define __PYX_NEED_TP_PRINT_SLOT  (PY_VERSION_HEX >= 0x030800b4 && PY_VERSION_HEX < 0x030A0000)
 #else
-  #if PY_VERSION_HEX >= 0x030800b4 && PY_VERSION_HEX < 0x03090000
-    #define CYTHON_NEED_TP_PRINT 1
-  #else
-    #define CYTHON_NEED_TP_PRINT 0
-  #endif
+  #define __PYX_NEED_TP_PRINT_SLOT  (PY_VERSION_HEX >= 0x030800b4 && PY_VERSION_HEX < 0x03090000)
 #endif
 // reinterpret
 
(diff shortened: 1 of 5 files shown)


Cython/Utility/ModuleSetupCode.c
```diff
@@ -556,6 +556,19 @@
...
+  #else
+    #define CYTHON_NEED_TP_PRINT 0
+  #endif
+#else
+  #if PY_VERSION_HEX >= 0x030800b4 && PY_VERSION_HEX < 0x03090000
+    #define CYTHON_NEED_TP_PRINT 1
+  #else
+    #define CYTHON_NEED_TP_PRINT 0
+  #endif
+#endif
```
scoder, line 571: Either, we consider this a feature-flag, in which case it should go into the CPYTHON/PYPY/… target define blocks further up (and be definable/overridable by users), or we consider this something internal, then it shouldn't start with `CYTHON_` but with `__PYX_`.

I'm not sure if it's worth making this visible to users. The slot is unlikely to reappear in the future and for historical Python versions, we should be able to control it ourselves correctly. Thus, `__PYX_` seems better to me.

You can also collapse the conditional defines to
```c
#define __PYX_NEED_TP_PRINT_SLOT  (PY_VERSION_HEX >= 0x030800b4 && PY_VERSION_HEX < 0x03090000)
```

scoder (commented):
A dedicated macro seems right by now.

Cython/Utility/ModuleSetupCode.c
mattip: done, thanks

scoder (approved):
Thanks. Looks good to me now. Let's wait for CI to be happy, too.

scoder:
Thanks

PR #5374, opened 2023-04-11
Move FAQ from wiki to userguide

h-vetinari:
Towards https://github.com/cython/cython/issues/4867.

This is mostly still a suggestion, but before I spend more time on it, I wanted to check what people think of the idea. As I mentioned in the issue, I think the wiki on github is orders of magnitude less visible than the checked-in & online docs, and it would be helpful to move the FAQ over. Its remoteness has let it bit-rot quite a bit already.

To make reviewing of this somewhat feasible, I tried to distinguish between small commits where I change something, and large mechanical commits that don't need to be reviewed (assuming you trust that I didn't mess up a copy & paste).

This still needs _a lot_ more editing, clean-ups, cross-referencing the existing docs, outright deletion, etc.

Obviously my choice of sections and how to group (and order) the questions is highly subjective and I'm very open for feedback on this. I just wanted to make _a_ choice. Mostly I've gone with absolutely trivial questions under Basics, questions under "How do I ...?" sorted _roughly_ in ascending order of difficulty/length/obsolteness, and then the rest.

I haven't tried sorting through possible formatting issues either, but in any case, [this](https://github.com/h-vetinari/cython/blob/faq/docs/src/userguide/faq.rst) is how github would render the content.

PR #5461, opened 2023-06-01
fomatting and linking the FAQ

mattip:
Follow on to #5374 in order to
- Reformat the page and get rid of the warnings and errors, hopefully without changing the content
- Add a top heading
- Link the FAQ to the userguide index

### DIFF
--- a/docs/src/userguide/index.rst
+++ b/docs/src/userguide/index.rst
@@ -28,6 +28,7 @@ Contents:
    numpy_pythran
    nogil
    troubleshooting
+   FAQ <faq>
 
 Indices and tables
 ------------------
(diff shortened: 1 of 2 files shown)


h-vetinari:
I avoided introducing linebreaks because that felt quite intrusive compared to the original, even though I agree it's a good idea. In the context of this PR, it kinda overwhelms the other changes, but I think improvement on that file is welcome even so. 👍

docs/src/userguide/index.rst
```diff
@@ -28,6 +28,7 @@ Contents:
    numpy_pythran
    nogil
    troubleshooting
+   FAQ <faq>
```
scoder, line 31: I don't think we need to specify the title here. It just works everywhere else.
```suggestion
   faq
```

scoder (approved):
It's a new file, I don't mind the reformatting.

scoder:
Thanks

PR #5463, opened 2023-06-03
Adjust error handling regex in Annotate.py to support html escaped characters

### LINKED ISSUE #5462: Cython incorrectly highlights error handling code in html page
### Describe your issue

Consider following code:

```cython
cdef int bar(int a, int b) nogil:
    a + b

with nogil:
    bar(3, 5)

bar(3, 5)

cdef void bar2(int a, int b) nogil:
    a + b

with nogil:
    bar2(3, 5)

bar2(3, 5)
```

Cython incorrectly highlights error handling code which contains characters escaped in HTML (`&` in the case of example below):

image

## Expected behaviour

Code is correctly highlighted:
image

matusvalo:
Fixes #5462

### DIFF
--- a/Cython/Compiler/Annotate.py
+++ b/Cython/Compiler/Annotate.py
@@ -315,7 +315,7 @@ _parse_code = re.compile((
     br'(?P<py_macro_api>Py[A-Z][a-z]+_[A-Z][A-Z_]+)|'
     br'(?P<py_c_api>Py[A-Z][a-z]+_[A-Z][a-z][A-Za-z_]*)'
     br')(?=\()|'       # look-ahead to exclude subsequent '(' from replacement
-    br'(?P<error_goto>(?:(?<=;) *if [^;]* +)?__PYX_ERR\([^)]+\))'
+    br'(?P<error_goto>(?:(?<=;) *if \(.*\) +)?__PYX_ERR\([^)]+\))'
 ).decode('ascii')).sub
 
 


matusvalo:
Now having second thought this code expects to have only one single if which is not enough I suppose. I am marking this as draft until I figure out something more robust...

matusvalo:
> Now having second thought this code expects to have only one single `if` statement which is not enough I suppose. I am marking this as draft until I figure out something more robust...

OK I managed to do that.

Cython/Compiler/Annotate.py
```diff
@@ -315,7 +315,7 @@ def annotate(match):
     br'(?P<py_macro_api>Py[A-Z][a-z]+_[A-Z][A-Z_]+)|'
     br'(?P<py_c_api>Py[A-Z][a-z]+_[A-Z][a-z][A-Za-z_]*)'
     br')(?=\()|'       # look-ahead to exclude subsequent '(' from replacement
-    br'(?P<error_goto>(?:(?<=;) *if [^;]* +)?__PYX_ERR\([^)]+\))'
+    br'(?P<error_goto>(?:(?<=;) *if \(((?!if).)*\) +)?__PYX_ERR\([^)]+\))'
```
matusvalo, line 318: Instead of trying to match all text not having `;` character (`[^;]*`), I am trying to match `if (<if condition>)` statement. To ensure that `<if condition>` does not match another `if`-statement, I am not allowing `if condition` to contain `if` substring. Error handler is pretty static, so I think this is safe. In worst case we can extend `if` forbidden substring to e.g. `if (`.

matusvalo:
image
https://xkcd.com/208/

Cython/Compiler/Annotate.py
scoder, line 318: We shouldn't start using positional groups here. And we should make sure that we really find `if …` and not `iffy` etc.
```suggestion
    br'(?P<error_goto>(?:(?<=;) *if \((?:(?!if[ (]).)*\) +)?__PYX_ERR\([^)]+\))'
```
Or could we somehow include HTML escapes explicitly? Are there more than `&amp;` and `&quot;` that can occur here?

scoder (commented):
I always hate making regexes more complex because you never know when their performance starts blowing up. Can this lead to expensive backtracking?

matusvalo:
>I always hate making regexes more complex because you never know when their performance starts blowing up. 

Me too. Moreover, they are hard to read + they can easily match/not-match in not expected way... Much cleaner way is to match regex first and after do html escape operation but I failed to find any easy way to do that.

> Can this lead to expensive backtracking?

Not sure to be honest...

Cython/Compiler/Annotate.py
matusvalo: > We shouldn't start using positional groups here. And we should make sure that we really find if … and not iffy etc.

I agree.

> Or could we somehow include HTML escapes explicitly? Are there more than `&amp;` and `&quot;` that can occur here?

I was thinking about that but I found it rather difficult to express that `&amp;` is allowed but `;` is not. They share the same character `;` which causes it is difficult to express it. Hence, I implemented it by disallowing `if` statement. Maybe it is possible, but I am lacking necessary regex skills to implement that.

Cython/Compiler/Annotate.py
```diff
@@ -315,7 +315,7 @@ def annotate(match):
     br'(?P<py_macro_api>Py[A-Z][a-z]+_[A-Z][A-Z_]+)|'
     br'(?P<py_c_api>Py[A-Z][a-z]+_[A-Z][a-z][A-Za-z_]*)'
     br')(?=\()|'       # look-ahead to exclude subsequent '(' from replacement
-    br'(?P<error_goto>(?:(?<=;) *if [^;]* +)?__PYX_ERR\([^)]+\))'
+    br'(?P<error_goto>(?:(?<=;) *if \((?:(?!if[ (]).)*\) +)?__PYX_ERR\([^)]+\))'
```
scoder, line 318: Hmm – so, if a nested `if (…)` appears, the first closing bracket belongs to the inner `if`, which is not followed by a ` __PYX_ERR`, so it needs to unread the closing bracket and accept it as `.` (any) character instead, and then continue to parse for another closing bracket. That's at least a bit of backtracking involved.

Atomic groups are new in Py3.11 and not spelled `(?!` but `(?>`.
https://docs.python.org/3/library/re.html#regular-expression-syntax

Issue #5467, opened 2023-06-04
Python 3.12 wheels for Cython 3.0 beta (on PyPI)


EwoutH:
### Describe your issue

Currently there are no Python 3.12 (`cp312`) wheels released either on [GitHub](https://github.com/cython/cython/releases) or [PyPI](https://pypi.org/project/Cython/). It would be very useful for these wheels to be build and published, to start early testing for many of the root dependencies in the Python package ecosystem. This could either be done for the existing [3.0.0b3](https://github.com/cython/cython/releases/tag/3.0.0b3) release, or the next one. Especially projects like NumPy could benefit a lot from early Python 3.12 wheels (see also https://github.com/numpy/numpy/pull/23824).

scoder:
These wheels are not needed. All environments that don't have a dedicated wheel use the pure Python wheel instead. We will eventually provide wheels. Currently, they'd only crash due to https://github.com/python/cpython/issues/104614

EwoutH:
Thanks for getting back!

The linked PR was just an example, probably a bad one though. I can remember Cython Python 3.11 wheels being published almost directly after CPython 3.11 hit its beta release. Especially for CI setups it can be quite beneficial to not have to build wheels from scratch.

scoder:
We only provide binary wheels to speed up the Cython *runs* on user side (especially CI setups on Linux). The pip installation always uses a wheel, binary or generic Python, so it's always fast and never builds a wheel locally.

scoder:
Just to be clear: Cython itself does not have to be compiled in order to compile user code with it. It runs just fine as plain Python code, just a little slower. Usually, the user side build times are dominated by the C compiler runs, not by Cython's execution (or installation).

EwoutH:
Thanks for the clarification, I learned something!

Could we keep this issue open until Python 3.12 wheels are released?

PR #5468, opened 2023-06-04
wheel CI: Update cibuildwheel to v2.13.0

EwoutH:
This commit updates cibuildwheel to v2.13.0 in the GitHub Actions wheel.yml CI workflow. cibuildwheel v2.13 adds CPython 3.12 support, under the prerelease flag CIBW_PRERELEASE_PYTHONS, among other this.

A small first step in solving https://github.com/cython/cython/issues/5467.

> ### [v2.13.0](https://github.com/pypa/cibuildwheel/releases/tag/v2.13.0)
> ✨ Adds CPython 3.12 support, under the prerelease flag [CIBW_PRERELEASE_PYTHONS](https://cibuildwheel.readthedocs.io/en/stable/options/#prerelease-pythons). This version of cibuildwheel uses 3.12.0b1.
> 
> While CPython is in beta, the ABI can change, so your wheels might not be compatible with the final release. For this reason, we don't recommend distributing wheels until RC1, at which point 3.12 will be available in cibuildwheel without the flag.
> 
> ✨ Adds the ability to pass arguments to the container engine when the container is created, using the [CIBW_CONTAINER_ENGINE](https://cibuildwheel.readthedocs.io/en/stable/options/#container-engine) option.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -74,7 +74,7 @@ jobs:
           platforms: all
       
       - name: Build wheels
-        uses: pypa/cibuildwheel@v2.11.4
+        uses: pypa/cibuildwheel@v2.13.0
         env:
           # TODO: Build Cython with the compile-all flag?
           CIBW_BUILD: ${{ matrix.python }}-${{ matrix.buildplat[1] }}


scoder:
Newer is usually better, so let's try this.

PR #5490, opened 2023-06-16
Update pypy in CI to released version

mattip:
PyPy released a new version, so CI does not need to use a nightly.


scoder:
Thanks

PR #5499, opened 2023-06-23
Upgrade cygdb from optparse to argparse

WillAyd:
This isn't exactly the same usage but I think is more idiomatic

### DIFF
--- a/Cython/Debugger/Cygdb.py
+++ b/Cython/Debugger/Cygdb.py
@@ -17,7 +17,7 @@ import glob
 import tempfile
 import textwrap
 import subprocess
-import optparse
+import argparse
 import logging
 
 logger = logging.getLogger(__name__)
@@ -32,8 +32,8 @@ def make_command_file(path_to_debug_info, prefix_code='',
         debug_files = glob.glob(pattern)
 
         if not debug_files:
-            sys.exit('%s.\nNo debug files were found in %s. Aborting.' % (
-                                   usage, os.path.abspath(path_to_debug_info)))
+            sys.exit('No cython debug files were found in %s. Aborting.' % (
+                                   os.path.abspath(path_to_debug_info)))
 
     fd, tempfilename = tempfile.mkstemp()
     f = os.fdopen(fd, 'w')
@@ -104,67 +104,62 @@ def make_command_file(path_to_debug_info, prefix_code='',
 
     return tempfilename
 
-usage = "Usage: cygdb [options] [PATH [-- GDB_ARGUMENTS]]"
 
-def main(path_to_debug_info=None, gdb_argv=None, no_import=False):
+def main():
     """
     Start the Cython debugger. This tells gdb to import the Cython and Python
     extensions (libcython.py and libpython.py) and it enables gdb's pending
     breakpoints.
-
-    path_to_debug_info is the path to the Cython build directory
-    gdb_argv is the list of options to gdb
-    no_import tells cygdb whether it should import debug information
     """
-    parser = optparse.OptionParser(usage=usage)
-    parser.add_option("--gdb-executable",
+    parser = argparse.ArgumentParser(
+        prog="cygdb",
+        description="Cython debugger",
+    )
+    parser.add_argument("gdb_argv", nargs="*",
+                        help="arguments to forward to gdb; specified after --")
+    parser.add_argument("--build_dir", default=os.curdir,
+                        help="directory containing cython_build/ files")
+    parser.add_argument("--no_import", default=False,
+                        help="whether cygdb should import debug information")
+    parser.add_argument("--gdb-executable",
         dest="gdb", default='gdb',
         help="gdb executable to use [default: gdb]")
-    parser.add_option("--verbose", "-v",
+    parser.add_argument("--verbose", "-v",
         dest="verbosity", action="count", default=0,
         help="Verbose mode. Multiple -v options increase the verbosity")
-    parser.add_option("--skip-interpreter",
+    parser.add_argument("--skip-interpreter",
                       dest="skip_interpreter", default=False, action="store_true",
                       help="Do not automatically point GDB to the same interpreter "
                            "used to generate debugging information")
 
-    (options, args) = parser.parse_args()
-    if path_to_debug_info is None:
-        if len(args) > 1:
-            path_to_debug_info = args[0]
-        else:
-            path_to_debug_info = os.curdir
-
-    if gdb_argv is None:
-        gdb_argv = args[1:]
+    parsed = parser.parse_args()
+    path_to_debug_info = parsed.build_dir
+    gdb_argv = parsed.gdb_argv
+    no_import = parsed.no_import
 
     if path_to_debug_info == '--':
         no_import = True
 
     logging_level = logging.WARN
-    if options.verbosity == 1:
+    if parsed.verbosity == 1:
         logging_level = logging.INFO
-    if options.verbosity >= 2:
+    if parsed.verbosity >= 2:
         logging_level = logging.DEBUG
     logging.basicConfig(level=logging_level)
 
     skip_interpreter = options.skip_interpreter
 
-    logger.info("verbosity = %r", options.verbosity)
-    logger.debug("options = %r; args = %r", options, args)
-    logger.debug("Done parsing command-line options. path_to_debug_info = %r, gdb_argv = %r",
-        path_to_debug_info, gdb_argv)
-
+    logger.debug("args = %r", parsed)
     tempfilename = make_command_file(path_to_debug_info,
                                      no_import=no_import,
                                      skip_interpreter=skip_interpreter)
     logger.info("Launching %s with command file: %s and gdb_argv: %s",
-        options.gdb, tempfilename, gdb_argv)
+        parsed.gdb, tempfilename, gdb_argv)
     with open(tempfilename) as tempfile:
         logger.debug('Command file (%s) contains: """\n%s"""', tempfilename, tempfile.read())
-        logger.info("Spawning %s...", options.gdb)
-        p = subprocess.Popen([options.gdb, '-command', tempfilename] + gdb_argv)
-        logger.info("Spawned %s (pid %d)", options.gdb, p.pid)
+        logger.info("Spawning %s...", parsed.gdb)
+        p = subprocess.Popen([parsed.gdb, '-command', tempfilename] + gdb_argv)
+        logger.info("Spawned %s (pid %d)", parsed.gdb, p.pid)
         while True:
             try:
                 logger.debug("Waiting for gdb (pid %d) to exit...", p.pid)


Cython/Debugger/Cygdb.py
```diff
@@ -104,67 +104,62 @@ def make_command_file(path_to_debug_info, prefix_code='',
...
     logging_level = logging.WARN
-    if options.verbosity == 1:
+    if parsed.verbosity == 1:
         logging_level = logging.INFO
-    if options.verbosity >= 2:
+    if parsed.verbosity >= 2:
         logging_level = logging.DEBUG
     logging.basicConfig(level=logging_level)
 
     skip_interpreter = options.skip_interpreter
```
matusvalo, line 150:
```suggestion
    skip_interpreter = parsed.skip_interpreter
```

Cython/Debugger/Cygdb.py
```diff
@@ -104,67 +104,62 @@ def make_command_file(path_to_debug_info, prefix_code='',
 
     return tempfilename
 
-usage = "Usage: cygdb [options] [PATH [-- GDB_ARGUMENTS]]"
 
-def main(path_to_debug_info=None, gdb_argv=None, no_import=False):
+def main():
```
matusvalo, line 108: I am for this change since this is `main()` function but not sure whether it won't break somewhere outside cython. Maybe I would go with it but I would like to see someone else opinion...

matusvalo (changes requested):

Cython/Debugger/Cygdb.py
```diff
@@ -104,67 +104,62 @@ def make_command_file(path_to_debug_info, prefix_code='',
...
-            path_to_debug_info = os.curdir
-
-    if gdb_argv is None:
-        gdb_argv = args[1:]
+    parsed = parser.parse_args()
+    path_to_debug_info = parsed.build_dir
+    gdb_argv = parsed.gdb_argv
+    no_import = parsed.no_import
 
     if path_to_debug_info == '--':
```
matusvalo, line 140: Is it possible that parameter `build_dir` will contain `--`. I suppose `no_import` variable is controlled by `--no-import` argument, so this can be removed.

Cython/Debugger/Cygdb.py
matusvalo: Moreover, the arguments are changed. Again, I like new version but it can break users so I think another opinion is needed.

Cython/Debugger/Cygdb.py
```diff
@@ -104,67 +104,62 @@ def make_command_file(path_to_debug_info, prefix_code='',
...
-    (options, args) = parser.parse_args()
-    if path_to_debug_info is None:
-        if len(args) > 1:
-            path_to_debug_info = args[0]
-        else:
-            path_to_debug_info = os.curdir
-
-    if gdb_argv is None:
-        gdb_argv = args[1:]
+    parsed = parser.parse_args()
```
ferdnyc, line 135: TBH — and this is _heavy_ bikeshedding, admittedly — I really dislike `parsed` as a name for the results of `ArgumentParser.parse_args()`.

`parsed` represents the **state** of the results, not the **contents**, which is focusing on the wrong thing. The fact that the contents were _parsed_ is irrelevant compared to the question of **what the contents _are_**. But the name doesn't even _hint_ at that.

`options` is still pretty common when using `argparse`, same as it was with `optparse`. But if you don't like `options`... well, if you call a parser's `parse_args()` method, it seems pretty tautological what that would produce: `args`.

(You even label them exactly that way on line 152.):

```python3
logger.debug("args = %r", parsed)
```

So even your own code is subconsciously agreeing that `parsed` should be named `args`. :wink:

WillAyd: Sure no strong attachment to `parsed`. The downside to `args` is that it conflicts with the pdb args command, so you have to jump through more hoops to inspect the object when debugging. Happy to revert back to options

Cython/Debugger/Cygdb.py
WillAyd: Yea I agree it makes more sense to just control that from the command line arguments

WillAyd:
I don't think any of the test failures are related to this change (?)

Cython/Debugger/Cygdb.py
```diff
@@ -104,42 +104,38 @@ def make_command_file(path_to_debug_info, prefix_code='',
...
     """
-    parser = optparse.OptionParser(usage=usage)
-    parser.add_option("--gdb-executable",
+    parser = argparse.ArgumentParser(
+        prog="cygdb",
+        description="Cython debugger",
+    )
+    parser.add_argument("gdb_argv", nargs="*",
+                        help="arguments to forward to gdb; specified after --")
+    parser.add_argument("--build_dir", default=os.curdir,
```
matusvalo, line 120: For better name consistency:
```suggestion
    parser.add_argument("--build-dir", default=os.curdir,
```

Cython/Debugger/Cygdb.py
```diff
@@ -104,42 +104,38 @@ def make_command_file(path_to_debug_info, prefix_code='',
...
+                        help="directory containing cython_build/ files")
+    parser.add_argument("--no_import", default=False,
```
matusvalo, line 122: I am thinking about this argument. Do we need it? I think that because `--no_import` should be used only and only when `--build_dir` is not specified and vise versa. 

I propose just to set `no_import` to `True` when `--build-dir` is not specified.

This will avoid code managing the case when both arguments are specified (PR currently does not manages it). Moreover, current master version has same semantics.

Cython/Debugger/Cygdb.py
ferdnyc: :+1: Thanks, `options` makes sense to me.

It struck me after posting my original comment: _another_ reason to avoid `parsed` is that having `parsed` and `parser` in the same scope (differing by only the last letter) feels like a recipe for long-term maintainability issues. Someone down the road could _easily_ confuse the two.

matusvalo:
Except one small issue, I did only cosmetics. There is still one missing point - documentation update: 

https://cython.readthedocs.io/en/latest/src/userguide/debugging.html#running-the-debugger

I would like to ask also for review at least one cython maintainer @da-woods, @scoder could you please review the PR?

da-woods:
I think my view is "let's not change too much more before Cython 3; let's pick this PR up after Cython 3". I'm broadly in favour in principle though but hopefully not now.

matusvalo:
> "let's not change too much more before Cython 3; let's pick this PR up after Cython 3"

I agree. I am not going to merge this release **before** Cython 3.0

Cython/Debugger/Cygdb.py
```diff
@@ -104,42 +104,36 @@ def make_command_file(path_to_debug_info, prefix_code='',
 
     return tempfilename
 
-usage = "Usage: cygdb [options] [PATH [-- GDB_ARGUMENTS]]"
 
-def main(path_to_debug_info=None, gdb_argv=None, no_import=False):
+def main():
```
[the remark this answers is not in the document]
WillAyd: These were removed because they are superfluous. `path_to_debug_info` is replaced by `--build_dir`, `gdb_argv` is replaced by ``gdb_argv` and `no_import` is inferred from the other options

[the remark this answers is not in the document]
WillAyd: I also don't think these were used at all? If you look at `bin/cygdb.py` is simply invokes this module without arguments:

```python
import sys

from Cython.Debugger import Cygdb as cygdb

if __name__ == '__main__':
    cygdb.main()
```

Cython/Debugger/Cygdb.py
```diff
@@ -104,42 +104,36 @@ def make_command_file(path_to_debug_info, prefix_code='',
...
-    parser = optparse.OptionParser(usage=usage)
-    parser.add_option("--gdb-executable",
+    parser = argparse.ArgumentParser(
+        prog="cygdb",
+        description="Cython debugger",
+    )
+    parser.add_argument("gdb_argv", nargs="*",
+                        help="Arguments to forward to gdb; specified after --")
+    parser.add_argument("--build-dir", dest="build_dir", default=None,
+                        help="Directory containing cython_build/ files")
```
scoder, line 121: Does this argument give the directory where the `cython_build/*` *files* live (i.e. `some/path/cython_build/`) or the directory where the `cython_build/` *directory* lives? What is a `cython_build` directory anyway? Was the `cython_debug` directory meant, maybe? I'm really confused by the name of the argument and its description.

scoder (commented):
IMHO, this can go into 3.1, but I'm having difficulties understanding the new `--build-dir` option.

Cython/Debugger/Cygdb.py
WillAyd: It essentially replaces the old argument of `path_to_debug_info` - would you prefer to revert back to that name?

scoder: Since it's a generated directory, which should usually go into the `build/` directory, `--build-dir` seems appropriate. The question is then whether we really write the `debug_info` directory into the `build/` directory, and it seems that we don't but rather write it into the current directory instead. If that's the build directory, that's fine, otherwise, it'd probably be the project root.

Should we move the `debug_info` directory into `build/`, unless explicitly set otherwise? There's probably a way to extract the build directory from something like `setuptools` if that's used, but if we're called via command line, that won't be available/interesting/correct.

scoder:
I think this is good enough to merge, just to get it in. The debate around `--build-dir` is mostly a matter of backwards compatibility. Maybe we could keep the old option as an alternative name?

Cython/Debugger/Cygdb.py
WillAyd: If anyone knows how to query for that information it is probably @rgommers 

Ralf - do you know if there's a way to query the build (frontend? backend?) to know what the "builddir" setting is?

rgommers: (disclaimer: I'm answering with probably a bit of context lacking here)

> Ralf - do you know if there's a way to query the build (frontend? backend?) to know what the "builddir" setting is?

My understanding is that `cygdb` is invoked as a standalone tool, sometime well after a build was completed. At that point there is no build frontend or build backend to query. Also, `cygdb` can be executed anywhere, rather than it having to be run from a project root, so there's no good way to even guess.

The current invocation (from [the cygdb docs](https://cython.readthedocs.io/en/latest/src/userguide/debugging.html#running-the-debugger)) will be something like:
```
cygdb path/to/builddir -- --args python-dbg mainscript.py
```
and this PR replaces the absolute argument so `path/to/builddir` would become `--build-dir path/to/builddir`? Either way I don't quite see a way around the user having to provide the location of the build dir.



> Should we move the `debug_info` directory into `build/`, unless explicitly set otherwise?

I don't think so, since the `cython` executable is typically invoked from within the build directory already - at least with Meson/CMake/ninja - so if you'd make this change then it would land in `build/build/`.

scoder:
> this PR replaces the absolute argument so `path/to/builddir` would become `--build-dir path/to/builddir`

Yes. And it seems a change that is nice, but not strictly required, and potentially very annoying for users who have to employ different (pre/post 3.1) Cython versions for different projects. They'd end up typing the wrong command syntax all the time.

I'd be happier to accept this if users could continue to type `cygdb build/dir/path` in addition to `cygdb --build-dir build/dir/path`.

scoder:
> I'd be happier to accept this if users could continue to type `cygdb build/dir/path` in addition to `cygdb --build-dir build/dir/path`.

… maybe with a visible warning being printed, but I think it would help users of older Cython versions.

Cython/Debugger/Cygdb.py
```diff
@@ -135,6 +135,12 @@ def main():
     gdb_argv = options.gdb_argv
     no_import = path_to_debug_info is None
 
+    if options.build_dir is None and gdb_argv and os.path.exists(gdb_argv[0]):
+        import warnings
+        gdb_argv = options.gdb_argv[1:]
+        path_to_debug_info = options.gdb_argv[0]
+        warnings.warn(f'Using deprecated positional parameter PATH. Use "--build-dir {path_to_debug_info}" argument instead.')
```
scoder, line 142: `PATH` suggests the environment variable to me, which isn't meant here.
```suggestion
        warnings.warn(f'Using deprecated positional parameter to find build directory. Use "--build-dir {path_to_debug_info}" argument instead.')
```

scoder:
> when user wants to pass actual path as gdb parameter which I find unlikely

You could be more specific about the parameter and only consider it if it's a directory path (or symlink to a directory). Not sure if that's worth it, but it might reduce the chance of misinterpretation further.

PR #5503, opened 2023-06-28
Fix display error on Cython3.0 document on troubleshooting

runzhi214:
This addition letter "s" cause error on displaying format.

Here is what it originally looks like:

image

This is what you really mean:

image

### DIFF
--- a/docs/src/userguide/troubleshooting.rst
+++ b/docs/src/userguide/troubleshooting.rst
@@ -86,7 +86,7 @@ essentially disables all type-inference. Therefore it doesn't know the type of `
 Writing into extension types
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-``AttributeError``s can also happen when writing into a ``cdef class``, commonly in ``__init__``::
+``AttributeError`` can also happen when writing into a ``cdef class``, commonly in ``__init__``::
 
     cdef class Company:
         def __init__(self, staff):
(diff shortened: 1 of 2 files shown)


matusvalo (approved):

scoder:
I think this was intentional. The name is AttributeError, but it's used in plural form here, thus the added "s" after the formatted name.

mattip:
The current text [is rendering](https://cython.readthedocs.io/en/latest/src/userguide/troubleshooting.html#writing-into-extension-types) with the double quotes, also see the screen shot above. I think a different solution would be to add a space between the double quotes and the `s`

da-woods:
> I think this was intentional. The name is AttributeError, but it's used in plural form here, thus the added "s" after the formatted name.

Yes, that would have been my intent, although the way it renders now is definitely wrong.

> I think a different solution would be to add a space between the double quotes and the `s`

I'd follow @mattip's suggestion and do this. It's what I look to have done elsewhere in the document.

Thanks for pointing the mistake out @runzhi214

docs/src/userguide/troubleshooting.rst
```diff
@@ -86,7 +86,7 @@ essentially disables all type-inference. Therefore it doesn't know the type of `
 Writing into extension types
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-``AttributeError``s can also happen when writing into a ``cdef class``, commonly in ``__init__``::
+``AttributeError`` can also happen when writing into a ``cdef class``, commonly in ``__init__``::
```
da-woods, line 89:
```suggestion
``AttributeErrors`` can also happen when writing into a ``cdef class``, commonly in ``__init__``::
```

runzhi214:
Either way is fine by me.

runzhi214:
I have just found another one: on the document of embedding Cython module in C/C ++ applications where one backquote is missed.

image

After the change it would be 

image

da-woods:
Thanks

PR #5522, opened 2023-07-11
Document @cython.with_gil

matusvalo:
Document new decorator implemented in #5491

### DIFF
--- a/docs/src/userguide/nogil.rst
+++ b/docs/src/userguide/nogil.rst
@@ -42,7 +42,7 @@ You can mark a whole function (either a Cython function or an :ref:`external fun
 .. tabs::
 
     .. group-tab:: Pure Python
-    
+
         .. code-block:: python
 
             @cython.nogil
@@ -52,9 +52,9 @@ You can mark a whole function (either a Cython function or an :ref:`external fun
             ...
 
     .. group-tab:: Cython
-    
+
         .. code-block:: cython
-    
+
             cdef void some_func() noexcept nogil:
                 ....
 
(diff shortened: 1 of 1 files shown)


docs/src/userguide/nogil.rst
```diff
@@ -77,37 +77,63 @@ To actually release the GIL you can use context managers
...
+        .. code-block:: python
+
+            @cython.with_gil
+            @cython.cfunc
+            def some_func() -> cython.int
+                ...
+
+            with cython.nogil:
+                ...          # some code that runs without the GIL
+                some_func()  # calling some_func() automatically acquire the GIL
```
scoder, line 120:
```suggestion
                some_func()  # some_func() will internally acquire the GIL
```

docs/src/userguide/nogil.rst
```diff
@@ -77,37 +77,63 @@ To actually release the GIL you can use context managers
...
+                ...          # some code that runs without the GIL
+            some_func()      # GIL is already held hence the function is called without acquiring the GIL
```
scoder, line 122:
```suggestion
            some_func()      # GIL is already held hence the function does not need to acquire the GIL
```

docs/src/userguide/nogil.rst
```diff
@@ -77,37 +77,63 @@ To actually release the GIL you can use context managers
...
+        .. code-block:: cython
+
+            cdef int some_func() with gil:
+                ...
+
+            with nogil:
+                ...          # some code that runs without the GIL
+                some_func()  # calling some_func() automatically acquire the GIL
+                ...          # some code that runs without the GIL
+            some_func()      # GIL is already held hence nothing happen
```
scoder, line 135: I think we should use the same comments in both variants. And as for the wording, it's the function that acquires the GIL (internally), not the call (or the caller).
```suggestion
            with nogil:
                ...          # some code that runs without the GIL
                some_func()  # some_func() will internally acquire the GIL
                ...          # some code that runs without the GIL
            some_func()      # GIL is already held hence the function does not need to acquire the GIL
```

scoder (commented):
Thanks. I just improved the wording a little.

PR #5572, opened 2023-07-28
Only write the version to stderr, if it's a separate file descriptor from stdout

eli-schwartz:
At least one project tries to detect the Cython version by redirecting stderr to stdout and capturing it. This is done in pure POSIX shell, so it probably seemed like the simple and obvious solution for a less capable programming language given that no output at all was expected on stdout.

But the result is that the version number appears twice, and then gets misparsed and ends up triggering bad assumptions in the code running cython.

It turns out that it's pretty easy to just print once, though. Detect when stdout and stderr are redirected to the same location, and only print once.

See #5504
Fixes https://bugs.gentoo.org/911333


scoder:
Thanks

PR #5627, opened 2023-08-15
Fix typo and provide deprecation shim for delimeters kwarg

yarikoptic:
might indeed be an overkill, then discard the last commit ;)


scoder:
Thanks. I think I found a simpler implementation, though. Also, a quick search on GitHub returned no usages of this argument outside of Cython itself, so we're hopefully safe.

yarikoptic:
whatever works!  but I think that for a "proper" deprecation cycle, that "simpler" one is incomplete since it has compatibility layer only for the `__init__` call but absent for any other function like `lex` where if code used `delimeters` kwargs -- now it would blow instead of issuing a warning.

Issue #5644, opened 2023-08-25
[BUG] A default parameter value for a fused type mixing complex and real generates invalid C code.


WarrenWeckesser:
### Describe the bug

The SciPy issue https://github.com/scipy/scipy/issues/19132 describes the problem.  In particular [my comment](https://github.com/scipy/scipy/issues/19132#issuecomment-1693689574) explains the bad code that is generated by Cython 3.0.1

@tylerjereddy did a git bisect and found that the bug was introduced in 6776c96f66fc799c62be4d6d4022903a5e36cbb2.

### Code to reproduce the behaviour:

See the snippets from the linked SciPy issue.

### OS

Linux and Mac (probably others)

### Python version

3.11.4 (and probably others)

### Cython version

3.0.1

PR #5653, opened 2023-08-26
Fix invalid C code when assigning int->fused complex with CYTHON_CCOMPLEX=0

scoder:
Fix invalid C code with -DCYTHON_CCOMPLEX=0 when assigning a simple (integer) default value to a fused type generated complex argument (which ends up as a struct).

Closes https://github.com/cython/cython/issues/5644


WarrenWeckesser:
Thanks for the quick fix!

PR #5655, opened 2023-08-26
Fix loop var declaration

vyasr:
https://github.com/cython/cython/commit/584078a371ef412554a2cd0b96fcc20fb3b9c476 looks like it introduced a minor typo.


vyasr:
CC @scoder

scoder:
Thanks

PR #5657, opened 2023-08-26
Fix issue 5609

### LINKED ISSUE #5609: [BUG] Cython doesn't handle mixed imports and cimports of enums cleanly
### Describe the bug

I would like to use Cython to make a C++ scoped enum accessible in Cython and Python code. I'm using a `cpdef enum class` declaration to accomplish this in one pxd file. In another file I would like to define a cpdef function that accepts this enum as a parameter. The cpdef function should use the C++ enum directly when invoked from typed Cython code, while it should use the Python wrapper enum when called in pure Python code. In order to support this, the enum needs to be both imported and cimported.

What I observe is that imports in the second module are confused when there is both a pyx and a pxd file, leading to conflicts that are not easily resolved.

This MWE requires four separate files:
```cython
# declarations.pxd
# distutils: language = c++
cdef extern from *:
    """
    enum class my_enum : int
    {
        a = 1,
        b = 0
    };
    """
    cpdef enum class my_enum(int):
        a
        b
```
```cython
# declarations.pyx
# distutils: language = c++
```
```cython
# definitions.pxd
# distutils: language = c++

# Doesn't work
# from declarations cimport my_enum
# cpdef void f(my_enum x)

# Works
cimport declarations
cpdef void f(declarations.my_enum x)
```

```cython
# definitions.pyx
# distutils: language = c++

from declarations import my_enum
from declarations cimport my_enum


cpdef void f(my_enum x):
    pass
```

If I run `cythonize -3 -i declarations.pyx && cythonize -3 -i -f definitions.pyx` with these files, everything compiles successfully. However, if I comment out the two lines below "Works" in definitions.pxd and instead include the two lines below "Doesn't work", I get the following error:
```
------------------------------------------------------------                                                                                                                                                                        [39/421]
...
# distutils: language = c++

from declarations import my_enum
                         ^

definitions.pyx:3:25: Assignment to non-lvalue 'my_enum'
```

Note that the error is coming from the definitions.pyx file, not the pxd file. What appears to be happening is that the line `from declarations cimport my_enum` in definitions.pxd makes Cython treat `my_enum` as a strongly typed C object, and therefore in Python it forbids the import because `my_enum` is no longer a valid PyObject name. 

The workaround above isn't too bad, but it does mean that `my_enum` is not cimportable from definitions, only pure Python importable. It can be cimported from declarations in this example, but in my production use case declarations is an internal module not intended for public consumption and I'm looking to export the enum into the public API (trying to solve the same problem as intended in #5602). I could equivalently remove the import from definitions.pyx and only include the cimport, which would create the inverse problem where `my_enum` is only visible in Cython, not Python. 

### Expected behaviour

I expect to be able to cimport the module into the pxd file with the same name as in the pyx file without any name collisions.

### OS

Linux

### Python version

3.11.4

### Cython version

3.0.0

### Additional context

An aside: declarations.pyx is an empty file that's required to trigger compilation of the Python version of `my_enum`. It makes sense why this file is necessary, but it's a bit of a footgun that simply changing `my_enum` from a `cdef enum class` to a `cpdef` triggers the requirement that this class exist. It would be nice if the Cython compiler could detect that a pxd file contains a cpdef and errors if a corresponding pyx file doesn't exists. That's orthogonal to the current issue though, and not something new in Cython, just a bit extra surprising in this case IMHO.

vyasr:
This PR resolves #5609. I'm not entirely confident that the fix won't have any other effects, but it does allow Cython to properly identify that the name should be assignable in Python.


vyasr:
@scoder or @da-woods I could use a close eye on this one, I'm not fully confident in the correctness of my solution. Hopefully CI will reveal any unexpected side effects though.

scoder:
I commented in the ticket. This approach does not look right to me.

scoder:
Closing since it's probably not the right solution.

PR #5667, opened 2023-08-29
Fix a null comparison compiler warning visible with gcc 12

rgommers:
Closes gh-5663

The older workaround was added in commit 12bbe5f3d2. Since `strchr` may return `NULL`, I don't see a reason here not to check for that the regular way. This fixes the warnings I'm seeing in the SciPy build.

### DIFF
--- a/Cython/Utility/ImportExport.c
+++ b/Cython/Utility/ImportExport.c
@@ -216,8 +216,7 @@ static PyObject *__Pyx_Import(PyObject *name, PyObject *from_list, int level) {
     {
         #if PY_MAJOR_VERSION >= 3
         if (level == -1) {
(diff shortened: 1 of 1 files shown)


scoder:
Looks good. Let's see if gcc, g++ and clang are all happy with this.

matusvalo (approved):
For me LGTM. I tried to compile scipy from bug report and it seems that no additional warnings are raised.

matusvalo:
> For me LGTM. I tried to compile scipy from bug report and it seems that no additional warnings are raised.

But I did not check GCC since, I don't have installed one. I will believe here @rgommers .

scoder:
I didn't see any warnings in the logs, so let's merge this. Thanks.

PR #5671, opened 2023-08-31
Add reference_wrapper

vyasr:
Adds libcpp bindings for `std::reference_wrapper`


scoder:
Thanks. Could you add a little test that exercises the new declaration? Maybe in `cpp_stl_cpp11.pyx`? Just to see that it compiles, and maybe also to have an example of how you would use the declaration in Cython code.

vyasr:
Good idea, done.

PR #5682, opened 2023-09-04
Simplify the handling of immediate (and highly unlikely) failures when entering a Python function wrapper

scoder:
Returning immediately instead of going through the error label (and creating a traceback) avoids a potentially unused label (due to preprocessor guards).

Closes https://github.com/cython/cython/issues/5681


scoder:
Triggering a fresh build due to a test selection problem in master.

PR #5706, opened 2023-09-15
Avoid undefined behavior in overflow checks

WillAyd:
ref https://github.com/cython/cython/issues/5705#issuecomment-1721865551

These were detected downstream when trying to use UBSAN with Cython. Not sure if there is a way to test that in current CI


scoder:
Thanks. This is code that was specifically written to minimise or avoid branching, so we'd want to be very careful to not add unnecessarily to that.

WillAyd:
I take it that's just a matter of performance right?

WillAyd:
Guessing for gcc / clang we can look at the overflow builtins. Doesn't appear MSVC has that yet https://developercommunity.visualstudio.com/t/10326281

WillAyd:
Hmm actually I think this file should already be using the bulitins when they are available. Going to close for now - have to do a bit more research as to why the builtins weren't being used to avoid the UB

PR #5712, opened 2023-09-19
cythonize 0.29.x: use force=None as default instead of force=False

haampie:
I completely overlooked this in https://github.com/cython/cython/pull/5307,
that backport was useless because the signature of `cythonize` is

```python
def cythonize(..., force=False, ...)
```

in 0.29.x instead of

```python
def cytonize(..., force=None, ...)
```

like it is on `3.x`, meaning that setting `CYTHON_FORCE_REGEN=1` had no effect.


haampie:
@scoder does cython still do 0.29 releases? If so, this would be helpful to have

adamjstewart (approved):

PR #5753, opened 2023-10-07
Add setuptools as dependency for python > 3.11

### LINKED ISSUE #5751: [BUG] distutils used in cythonize (fails in 3.12)
### Describe the bug

The `cythonize` command imports distutils at the top of its script, which is no longer available in Python 3.12. 

Simply running:

```bash
cythonize --help
```

With Python 3.12 results in 

```
  File "/home/joncrall/.pyenv/versions/3.12.0/envs/pyenv3.12.0/bin/cythonize", line 5, in <module>
    from Cython.Build.Cythonize import main
  File "/home/joncrall/.pyenv/versions/3.12.0/envs/pyenv3.12.0/lib/python3.12/site-packages/Cython/Build/Cythonize.py", line 6, in <module>
ModuleNotFoundError: No module named 'distutils'
```

### Expected behaviour

distutils is deprecated, so I'm not sure what should be done.

### OS

Linux

### Python version

3.12

### Cython version

3.0.3

### Additional context

I'm working on adding 3.12 support to line-profiler in this MR: https://github.com/pyutils/line_profiler/pull/246

I'm attempting to do some standalone debugging by simply calling the cythonize command:

```
    # Standalone compile instructions for developers 
    # Assuming the cwd is the repo root.
    cythonize --annotate --inplace \
        ./line_profiler/_line_profiler.pyx \
        ./line_profiler/timers.c \
        ./line_profiler/unset_trace.c
```

But obviously because the script can't even import itself it errors immediately.

matusvalo:
distutils is removed from python >= 3.12. Setuptools contains backward compatible distutils interface.

Fixes #5751

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,7 @@
+[options]
+install_requires =
+    setuptools;python_version>'3.11'
+
 [flake8]
 max-complexity = 10
 


matusvalo:
Having another thought about it this dependency can affect users who does not use setuptools but other build system instead. Another option is to check whether setuptools is installed and ask user to install it in commands which needs it.

setup.cfg
```diff
@@ -1,3 +1,7 @@
+[options]
+install_requires =
+    setuptools;python_version>'3.11'
```
dnicolodi, line 3: I don't like adding a dependency on `setuptools`, but most likely you want `>=3.12` here.

haampie (changes requested):
Just adding a dependency on setuptools is not a solution.

Setuptools uses this `distutils-precedence.pth` "trick" to hijack `sys.path`. That only runs on startup if setuptools is installed as a system site-package.

It **doesn't** run if setuptools is located through `PYTHONPATH`.

If you want to make it more robust, you have to use the alternative `sys.path` hijack that setuptools does, and that is by importing setuptools before importing distutils.

But if you do that, you can just as well make a proper fix and stop using distutils altogether.

matusvalo:
Closing this PR in preference of https://github.com/cython/cython/pull/5754

PR #5754, opened 2023-10-08
Add setuptools as optional dependency

matusvalo:
This PR is alternative to #5753. I think this is superior because:
1. projects which uses not-setuptools build system will not have unneeded dependency
2. this is standard way how to manage dependencies which are optional - setuptools is needed only for building .c files.

Fixes https://github.com/cython/cython/issues/5751

### DIFF
--- a/Cython/Build/__init__.py
+++ b/Cython/Build/__init__.py
@@ -3,8 +3,17 @@ from .Dependencies import cythonize
 import sys
 if sys.version_info < (3, 7):
     from .Distutils import build_ext
-del sys
 
+if sys.version_info > (3, 11):
+    try:
+        import setuptools
+    except ImportError:
+        raise ImportError(
+            "Missing optional dependency 'setuptools'. "
+            "Use pip or conda to install setuptools."
+        )
+
+del sys
 
 def __getattr__(name):
     if name == 'build_ext':
--- a/docs/src/quickstart/install.rst
+++ b/docs/src/quickstart/install.rst
@@ -44,7 +44,11 @@ according to the system used:
 
 The simplest way of installing Cython is by using ``pip``::
 
-  pip install Cython
+  pip install Cython[setuptools]
+
+.. Note::
+   When cython is used only to generate .c files, setuptools optional dependency can be omited.
+   This is useful mainly when alternative build system like meson is used.
 
 On platforms that are covered by one of the binary wheel packages provided on PyPI,
 this will install an accelerated wheel which contains some Cython compiled modules.
@@ -58,7 +62,7 @@ The newest Cython release can always be downloaded from
 https://cython.org/.  Unpack the tarball or zip file, enter the
 directory, and then run::
 
-  pip install .
+  pip install .[setuptools]
 
 
 For one-time installations from a Cython source checkout, it is substantially
@@ -67,11 +71,11 @@ of Cython with something like
 
 ::
 
-    pip install .  --install-option="--no-cython-compile"
+    pip install .[setuptools] --install-option="--no-cython-compile"
 
     or
 
-    NO_CYTHON_COMPILE=true  pip install .
+    NO_CYTHON_COMPILE=true  pip install .[setuptools]
 
 
 .. [Anaconda] https://docs.anaconda.com/anaconda/
--- a/pyximport/pyxbuild.py
+++ b/pyximport/pyxbuild.py
@@ -6,6 +6,17 @@ out_fname = pyx_to_dll("foo.pyx")
 import os
 import sys
 
+if sys.version_info > (3, 11):
+    try:
+        import setuptools
+    except ImportError:
+        raise ImportError(
+            "Missing optional dependency 'setuptools'. "
+            "Use pip or conda to install setuptools."
+        )
+
+del sys
+
 from distutils.errors import DistutilsArgError, DistutilsError, CCompilerError
 from distutils.extension import Extension
 from distutils.util import grok_environment_error
--- a/Cython/Distutils/__init__.py
+++ b/Cython/Distutils/__init__.py
@@ -1,2 +1,14 @@
+import sys
+if sys.version_info > (3, 11):
+    try:
+        import setuptools
+    except ImportError:
+        raise ImportError(
+            "Missing optional dependency 'setuptools'. "
+            "Use pip or conda to install setuptools."
+        )
+
+del sys
+
 from Cython.Distutils.build_ext import build_ext
 from Cython.Distutils.extension import Extension
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,7 @@
+[options.extras_require]
+setuptools =
+    setuptools
+
 [flake8]
 max-complexity = 10
 


Cython/Build/__init__.py
```diff
@@ -3,8 +3,17 @@
 import sys
 if sys.version_info < (3, 7):
     from .Distutils import build_ext
-del sys
 
+if sys.version_info > (3, 11):
+    try:
+        import setuptools
+    except ImportError:
+        raise ImportError(
```
matusvalo, line 11: This handling is repeating all the times and maybe it deserves to be moved to function. But not sure where this function put.

docs/src/quickstart/install.rst
```diff
@@ -44,7 +44,11 @@ according to the system used:
 
 The simplest way of installing Cython is by using ``pip``::
 
-  pip install Cython
+  pip install Cython[setuptools]
```
matusvalo, line 47: I was thinking that I would add only a note about setuptools dependency instead of putting it directly to example but I suppose this is what 99% of users will do. Also I intentionally not mentioned python3.12 (which is version of python when user needs to use setuptools), but I wanted to keep it simple and straightforward. Moreover it is better to start users teaching to use setuptools instead of distutils.

dnicolodi:
Can you elaborate on on which circumstances you need setuptools to run Cython? I have projects that use Cython on Python 3.12 without installing setuptools.

matusvalo:
> Can you elaborate on on which circumstances you need setuptools to run Cython?

setuptools is needed when one wants to build shared library. For compiling to .c file setuptools is not needed

> I think the easiest solution is to have setuptoiols as an optional dependency for a cythonize variant. Something like:

This is already done in the PR with the same aproach you mentioned in an example.

> I really don't want to have my projects depends (transitively) on setuptools after all the work that went into getting rid of it.

This PR is not adding hard dependency rather than optional one. In this PR when you install cython via `pip install cython` setuptools is **not** installed.

dnicolodi:
> setuptools is needed when one wants to build shared library. For compiling to .c file setuptools is not needed

I don't think this is correct.

> This is already done in the PR with the same approach you mentioned in the example.

Except that the variant is called `setuptools` and not `cythonize`. I think `cythonize` is a much better choice as the fact that `cythonize` requires `setuptools` is an implementation details that may change.

> This PR is not adding hard dependency rather than optional one. In this PR when you install cython via `pip install cython` setuptools is **not** installed.

This PR raises `ImportError` when `setuptools` is not present to a code path that is traversed by code that does not require `setuptools`, making `setuptools` a requirement when before it was not.

Cython/Build/__init__.py
```diff
@@ -3,8 +3,17 @@
 import sys
 if sys.version_info < (3, 7):
     from .Distutils import build_ext
-del sys
 
+if sys.version_info > (3, 11):
```
dnicolodi, line 7: I don't think this version check is correct.

dnicolodi:
Furthermore, `setuptools` is required only on Python >= 3.12, thus the optional dependency should probably be
```
[options.extras_require]
cythonize =
    setuptools;python_version>='3.12'
```

Cython/Build/__init__.py
matusvalo: Hm I am not sure why? Could you be more specific?

dnicolodi:
```
>>> (3, 11, 1) > (3, 11)
True
```

matusvalo:
> > setuptools is needed when one wants to build shared library. For compiling to .c file setuptools is not needed

> I don't think this is correct.

If not please correct me.

> Except that the variant is called setuptools and not cythonize. I think cythonize is a much better choice

Yes I agree but I think this is more about how we name thinks not that implementation/approach is **wrong**. If there will be consensus among core devs on some name I have no problem changing it.

> This PR raises ImportError when setuptools is not present to a code path that is traversed by code that does not require setuptools, making setuptools a requirement when before it was not.

which code path? Could you be please more specific? `cython` command is working without setuptools:
```
(cython312) matus@MacBook-Pro-2240:~/dev/cython/test$ python -V
Python 3.12.0
(cython312) matus@MacBook-Pro-2240:~/dev/cython/test$ pip freeze
-e git+https://github.com/matusvalo/cython.git@b0552058e178c0adab4d42fbdcf900284bbb95a7#egg=Cython
(cython312) matus@MacBook-Pro-2240:~/dev/cython/test$ cython -h
usage: cython [-h] [-V] [-l] [-I INCLUDE_PATH] [-o OUTPUT_FILE] [-t] [-f] [-v] [-p] [--cleanup GENERATE_CLEANUP_CODE] [-w WORKING_PATH] [--gdb] [--gdb-outdir GDB_OUTDIR] [-D] [-a] [--annotate-fullc] [--annotate-coverage ANNOTATE_COVERAGE_XML] [--line-directives]
              [-+] [--embed] [-2] [-3] [--3str] [--lenient] [--capi-reexport-cincludes] [--fast-fail] [-Werror] [-Wextra] [-X NAME=VALUE,...] [-E NAME=VALUE,...] [--module-name MODULE_NAME] [-M]
              [sources ...]

Cython (https://cython.org/) is a compiler for code written in the Cython language. Cython is based on Pyrex by Greg Ewing.
```
`cythonize`, `pyximport` and other tools raises an `ImportError` with message to install `setuptools`.

if you **do** know about any usage of Cython not using distutils/setuptools which fails in this PR please let me know.

dnicolodi:
> which code path? Could you be please more specific?

You added an `import setuptools` to a source file that dind't contain it or `import distutils` before. The source file in question is imported, for example, when `cython` is run with the `-M` argument.

Cython/Build/__init__.py
matusvalo: Now I see. Thanks!

matusvalo:
> You added an import setuptools to a source file that dind't contain it or import distutils before. The source file in question is imported, for example, when cython is run with the -M argument.

Yes I missed that. I wanted to avoid check that directly in every file. Hence we can either:

* check every distutils import
* check at main functions called by commands

Thank you @dnicolodi for your review.

dnicolodi:
AFAIK, the only user visible interfaces that require `distutils` are `cythonize` and `pyximport`. If you really want to add this check, please add it to the main entry points for these facilities, ie the `cythonize` main function and to the `pyximport` module. Do not sprinkle it everywhere.

matusvalo:
> AFAIK, the only user visible interfaces that require distutils are cythonize and pyximport

I was thinking about that see comment about but there is also `cython.inline` and Jupiter notebook integration.

haampie:
Repeating my comment from a now closed PR:

Just adding a dependency on `setuptools` is not a solution for making `distutils` importable

Setuptools uses this `distutils-precedence.pth` "trick" to hijack `sys.path`. That only runs on startup if setuptools is installed as a system site-package.

It **doesn't** run if setuptools is located through `PYTHONPATH`.

If you want to make it more robust, you have to use the alternative `sys.path` hijack, by importing setuptools before importing distutils.

But if you do that, you can just as well make a proper fix and stop using distutils altogether.

matusvalo:
> and stop using distutils altogether.

Not sure if we don't break someone who is not using setuptools and has python < 3.12. To be backward compatible, I am afraid we need to do this kind of dance:

https://github.com/cython/cython/blob/459f7c6236b7507b664703e6a20aafafe22f8a21/Cython/Distutils/build_ext.py#L16-L23

Cython/Build/__init__.py
scoder: It'll probably be removed at some point, and it's simple enough. I'm fine with repeating it.

scoder:
> add it to the main entry points for these facilities, ie the `cythonize` main function and to the `pyximport` module.

Agreed for `pyximport`, which definitely depends on `distutils`. But `cythonize` has several modes of operation, only some of which actually build an extension module using distutils/setuptools. It would be great if we could also avoid keeping `-M` and the plain `.c` file generation from importing distutils/setuptools and avoid failing if it's not available.

scoder:
> the variant is called `setuptools` and not `cythonize`. I think `cythonize` is a much better choice

I disagree because `cythonize` is ambiguous. The `cythonize()` function shouldn't depend on setuptools, and several operations of the `cythonize` command could also work without it. We may well end up with two or more build implementations that our tools use, and setuptools would probably remain one of them for quite a while. That makes it appear quite reasonable to install `Cython[setuptools]` if you want your build to use setuptools.

dnicolodi:
@scoder I was thinking about the `cythonize` command line tool, not the function. However, there are more interfaces that depend on `distutils` that I initially thought about, therefore using `cythonize` as a variant does not seem appropriate. However, using `setuptools` as a variant name does not seem right either. In this case it is simpler to just install setuptools alongside cython:
```
python -m pip install cython[setuptools]
```
is not any simpler than
```
python -m pip install cython setuptools
```

scoder:
> it is simpler to just install setuptools alongside cython

The difference is that listing it in `setup.cfg` makes it available as package metadata, thus listing `setuptools` explicitly as an optional dependency. Otherwise, you'd have to find out yourself that Cython can make use of `setuptools`.

dnicolodi:
I don't think there is a way to discover optional dependencies that is not through documentation. AFAIK, variants are not listed anywhere on PyPI or via any installation tool. Am I missing something?

dnicolodi:
Technically, setuptools is required only with Python 3.12 and later. That is one reason why I was proposing to name the variant something else than `setuptools` and have it depend on `setuptools` conditionally to the Python version, so that installation instructions do not need to be diversified per Python version. https://github.com/cython/cython/pull/5754#issuecomment-1752050969

scoder:
> I don't think there is a way to discover optional dependencies that is not through documentation. AFAIK, variants are not listed anywhere on PyPI or via any installation tool. Am I missing something?

So, you are suggesting that we should not expose this metadata in any of our wheel distributions because PyPI is currently missing a very common feature of package indices and because you haven't used any tools yet that evaluate optional package dependencies?

I think we should expose it. And allow users to decide whether they want to use it or not. As you say, it makes almost no difference in usage, so we can just as well make it explicit.

Cython/Distutils/extension.py
```diff
@@ -95,10 +99,6 @@ def __init__(self, name, sources,
         self.no_c_in_traceback = no_c_in_traceback
         self.cython_compile_time_env = cython_compile_time_env
 
-# class Extension
-
-read_setup_file = _Extension.read_setup_file
-
```
dnicolodi, line 98: This is API. I don't think it should be removed.

matusvalo: Problem is that in setuptools I was not able to find equivalent function.

matusvalo: Let's do it that way.

pyximport/pyxbuild.py
```diff
@@ -6,9 +6,18 @@
...
-from distutils.util import grok_environment_error
+try:
+    from distutils.errors import DistutilsArgError, DistutilsError, CCompilerError
+    from distutils.extension import Extension
+    from distutils.util import grok_environment_error
+except ImportError:
+    try:
+        from setuptools.errors import DistutilsArgError, DistutilsError, CCompilerError
+        from setuptools.extension import Extension
+        from setuptools._distutils.util import grok_environment_error
```
matusvalo, line 17: setuptools does not contains `grok_environment_error `

Cython/Build/IpythonMagic.py
```diff
@@ -52,15 +52,36 @@
...
+try:
+    import distutils.log as logging
+    from distutils.log import set_threshold
+    from distutils.core import Distribution, Extension
+    from distutils.command.build_ext import build_ext
+    import distutils.errors as _setuptools_errors
+except ImportError:
+    import logging
+
+    def set_threshold(level):
```
scoder, line 65: Without the qualifying module prefix, we'd better give this an explicit name (also for the import).
```suggestion
    def set_logging_threshold(level):
```

Cython/Build/IpythonMagic.py
```diff
@@ -52,15 +52,36 @@
...
+        raise ImportError(
+            "'distutils' cannot be imported. Please install setuptools."
+        )
+
 import textwrap
 
 IO_ENCODING = sys.getfilesystemencoding()
 IS_PY2 = sys.version_info[0] < 3
 
 import hashlib
```
scoder, line 84: It seems worth moving these stray stdlib imports to the top while we're at it.

docs/src/quickstart/install.rst
```diff
@@ -44,7 +44,11 @@ according to the system used:
...
+
+.. Note::
+   When cython is used only to generate .c files, setuptools optional dependency can be omitted.
+   This is useful mainly when alternative build system like meson is used.
```
scoder, line 51:
```suggestion
   If cython is used only to generate .c files, the optional dependency on setuptools
   can be omitted (i.e. just ``pip install Cython``).
   This is useful mainly when an alternative build system like meson is used.
```

pyximport/pyxbuild.py
scoder: We should probably get rid of that usage eventually. Not necessarily now.

scoder (commented):
I wonder, instead of changing all of our imports, can't we just try-import setuptools and then rely on distutils being there magically? Probably something for Cython 3.1 rather than 3.0.x…

haampie:
> I wonder, instead of changing all of our imports, can't we just try-import setuptools and then rely on distutils being there magically? Probably something for Cython 3.1 rather than 3.0.x…

Yeah, that's a valid way to deal with it, also works when setuptools is in `PYTHONPATH`.

nitzmahone:
We also struggled with the best way to handle this issue while fixing up Python 3.12 support for [CFFI](https://github.com/python-cffi/cffi)... Glad to see you've reached the same main conclusion that we did: a hard dep on `setuptools` is almost certainly the wrong approach. At least for CFFI, there are common situations (eg `cryptography`) that have a *runtime* `cffi` dep, but only need setuptools/distutils at build-time. Forcing everything that uses `cffi` to take an implicit setuptools dep would've been easy and convenient, but arguably creates a lot more problems than it solves.

CFFI's `setuptools` / `distutils` imports were limited enough that we ended up rolling our own internal shim module that basically exposes the various imports we need on a single module, but provides more prescriptive error messaging for users/devs that are missing a dep beyond `ModuleNotFoundError: no module named 'setuptools'`. I didn't love it, but it was at least simple...

We'd debated the `extras_require` as well. It would allow any given Cython version to (optionally) specify a supported/tested `setuptools` range that would be opaque to consumers just depending on eg `Cython[setuptools]>=3.1`- the same technique could work for both runtime and build-time deps, and at least `pip` seems to just warn if you ask for an extra spec that doesn't exist (eg if the extra was dropped for some reason in the future). We ended up not doing it for the moment, but knowing that at least modern versions of `pip` just warn on a missing extra instead of failing, I'm warming to the idea for CFFI as well. 

It's a minor point, but there's also the build-time UX around PEP517 build requirements- requiring folks to do `requires=['Cython[setuptools]']` in their `pyproject.toml` isn't so bad, but still suboptimal over just declaring Cython and letting it figure out if setuptools is needed from there. I can't remember off the top of my head if Cython provides its own native PEP517 build backend or does any shimming of the `setuptools` backend (some of which would be required to dynamically augment the build-time deps to completely hide setuptools), or if all the magic still occurs down at the legacy setuptools/distutils command level. I've been playing with a `setup.py`-free rewrite of PyYAML's libyaml extension build with custom in-tree backend shims over both setuptools and Cython >= 3.0, and while I can make it work, it feels "icky" and brittle in some ways. OTOH, I understand Cython not wanting to provide a full PEP517 backend as long as most of the setuptools integration is still happening down at the distutils/command layer (I feel the same way right now about trying to implement a native PEP517 backend for CFFI).

PR #5761, opened 2023-10-10
Use sysconfig module instead of distutils.sysconfig for python >= 3.9

matusvalo:
Related to #5751

### DIFF
--- a/Cython/Build/BuildExecutable.py
+++ b/Cython/Build/BuildExecutable.py
@@ -13,11 +13,10 @@ DEBUG = True
 
 import sys
 import os
-if sys.version_info < (3, 12):
+if sys.version_info < (3, 2):
     from distutils.sysconfig import get_python_inc
     from distutils.sysconfig import get_config_var
 else:
-    # Python 3.12 no longer has distutils
     from sysconfig import get_path
     from sysconfig import get_config_var
     get_python_inc = lambda: get_path('include')
--- a/Cython/Build/BuildExecutable.py
+++ b/Cython/Build/BuildExecutable.py
@@ -13,17 +13,18 @@ DEBUG = True
 
 import sys
 import os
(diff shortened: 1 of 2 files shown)


dnicolodi:
Up to Python 3.9 `sysconfig` in the standard library and from `distutils` did return some different value on some platforms. Without carefully checking it is not really safe to switch from one to the other.

dnicolodi:
I think 3.8.7 is the the minimum version on which `sysconfig` from the standard library can be trusted. Rounding up to 3.9 should do not have any adverse effect. From Python 3.11 onward, importing `distutils` (when `setuptools` is not installed) emits a warning, so it would be great to gate the switch at a lower version number.

Cython/Build/BuildExecutable.py
```diff
@@ -13,13 +13,20 @@
...
 import os
-from distutils import sysconfig
+if sys.version_info < (3, 9):
+    from distutils.sysconfig import get_python_inc
+    from distutils.sysconfig import get_config_var as sysconfig_get_config_var
+else:
+    # sysconfig can be trusted from cpython >= 3.8.7
+    from sysconfig import get_path
+    from sysconfig import get_config_var as sysconfig_get_config_var
+    get_python_inc = lambda: get_path('include')
```
scoder, line 23: I suggest turning this around. Since the old implementation (<3.9) will go away at some point, I'd change the code to use only the new style, and adapt the old imports to match what we now use. That way, we can just remove the old imports at some point, rather than having to change the implementation again in the future.

matusvalo: fixed

matusvalo:
Marking as 3.0.X but not sure if it should not go to 3.1. Definitely, I think it should not go to 3.0.4

scoder:
Thanks

Issue #5790, opened 2023-11-01
[BUG] Python 3.13 C++17: error: standard attributes in middle of decl-specifiers (CYTHON_UNUSED)


hroncok:
### Describe the bug

Hello.
When using Python 3.13.0a1 and Cython 3.0.5 to build grpc 1.48.4 with [a Cython 3 patch](https://src.fedoraproject.org/rpms/grpc/raw/8bf11b8c20d2c54dcf7b88777d7590237b6b0555/f/0001-Specify-noexcept-for-cdef-functions.patch). I see a GCC 12+ C++17 error:

```
src/python/grpcio/grpc/_cython/cygrpc.cpp: At global scope:
src/python/grpcio/grpc/_cython/cygrpc.cpp:1360:31: error: standard attributes in middle of decl-specifiers
 1360 |         #define CYTHON_UNUSED [[maybe_unused]]
      |                               ^
src/python/grpcio/grpc/_cython/cygrpc.cpp:5501:12: note: in expansion of macro ‘CYTHON_UNUSED’
 5501 |     static CYTHON_UNUSED PyObject *__Pyx_KwargsAsDict_FASTCALL(PyObject *kwnames, PyObject *const *kwvalues);
      |            ^~~~~~~~~~~~~
src/python/grpcio/grpc/_cython/cygrpc.cpp:1360:31: note: standard attributes must precede the decl-specifiers to apply to the declaration, or follow them to apply to the type
 1360 |         #define CYTHON_UNUSED [[maybe_unused]]
      |                               ^
src/python/grpcio/grpc/_cython/cygrpc.cpp:5501:12: note: in expansion of macro ‘CYTHON_UNUSED’
 5501 |     static CYTHON_UNUSED PyObject *__Pyx_KwargsAsDict_FASTCALL(PyObject *kwnames, PyObject *const *kwvalues);
      |            ^~~~~~~~~~~~~
src/python/grpcio/grpc/_cython/cygrpc.cpp:1360:31: warning: attribute ignored [-Wattributes]
 1360 |         #define CYTHON_UNUSED [[maybe_unused]]
      |                               ^
src/python/grpcio/grpc/_cython/cygrpc.cpp:5501:12: note: in expansion of macro ‘CYTHON_UNUSED’
 5501 |     static CYTHON_UNUSED PyObject *__Pyx_KwargsAsDict_FASTCALL(PyObject *kwnames, PyObject *const *kwvalues);
      |            ^~~~~~~~~~~~~
[...]

PR #5791, opened 2023-11-01
Fix C++17 error: standard attributes in middle of decl-specifiers (CYTHON_UNUSED)

hroncok:
This is only relevant on Python 3.13+

Fixes https://github.com/cython/cython/issues/5790

Still testing this.


hroncok:
I belive there should be a test for this, so it won't regress. However, I don't currently have the capacity to add it.

scoder:
> I belive there should be a test for this

We could probably make sure that it gets triggered by one of the C++ tests. But it's a bugfix either way. Let's merge it.

scoder:
Thanks

PR #5852, opened 2023-11-20
PEP 703: Rename `PY_NOGIL` to `Py_GIL_DISABLED`

hugovk:
Follow on from https://github.com/cython/cython/pull/4914.

The Python Steering Council has chosen `Py_GIL_DISABLED` instead of `Py_NOGIL`:

* https://github.com/python/steering-council/issues/214#issuecomment-1817942256

Updated in CPython:

* https://github.com/python/cpython/pull/111864
* https://github.com/python/cpython/issues/111863

Let's update to match.

cc @colesbury

### DIFF
--- a/Cython/Utility/ModuleSetupCode.c
+++ b/Cython/Utility/ModuleSetupCode.c
@@ -243,7 +243,7 @@
     #define CYTHON_UPDATE_DESCRIPTOR_DOC 0
   #endif
 
-#elif defined(Py_GIL_DISABLED)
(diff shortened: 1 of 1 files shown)


colesbury:
I would prefer 'defined(Py_GIL_DISABLED) || defined(Py_NOGIL)' until the 3.13 because the nogil fork is still in use, including in the scikit-learn continuous build.

hugovk:
Sure, updated!

scoder:
3.0.x commit is https://github.com/cython/cython/commit/f258b674d3117141ceae53226e7fe8530f51385c

PR #5878, opened 2023-11-28
Modernise documentation

matusvalo:
* `from __future__ import print_function` was removed
* `%` operator replaced with f-string


scoder:
Looks good to me, too. Thanks!

PR #5948, opened 2024-01-15
Debugger tab completion and test_libcython_in_gdb support

kentslaney:
Followup to #5924 (@da-woods). The exit status of `Cython/Debugger/Tests/test_lib_cython_in_gdb.py` is currently ignored, but needed patching to run at all, at least in my docker environment.


Cython/Debugger/Tests/test_libcython_in_gdb.py
```diff
@@ -60,7 +60,7 @@ class DebugTestCase(unittest.TestCase):
     """
 
     def __init__(self, name):
-        super().__init__(name)
+        super(DebugTestCase, self).__init__(name)
```
scoder, line 63: Why was this needed?

Cython/Debugger/Tests/test_libcython_in_gdb.py
```diff
@@ -175,6 +175,12 @@ def test_break_lineno(self):
         self.lineno_equals(nextline)
         assert step_result.rstrip().endswith(nextline)
 
+    def test_break_completion(self):
+        completer = libcython.cy.break_.complete
+        assert 'spam' in completer("codefile.SomeClass.s", "s")
+        assert 'spam' in completer("codefile.SomeClass.", None)
+        assert 'pam' in completer("codefile.s", None)
```
scoder, line 182: It's difficult to debug failures if the output is just "assertion failed".
```suggestion
        self.assertIn('spam', completer("codefile.SomeClass.s", "s"))
        self.assertIn('spam', completer("codefile.SomeClass.", None))
        self.assertIn('pam', completer("codefile.s", None))
```

Cython/Debugger/Cygdb.py
```diff
@@ -89,9 +89,7 @@ def make_command_file(path_to_debug_info, prefix_code='',
                 f.write(textwrap.dedent('''\
                     python
                     import sys
-                    try:
-                        gdb.lookup_type('PyModuleObject')
-                    except RuntimeError:
+                    if not hasattr(gdb.selected_inferior().progspace, "symbol_file"):
```
scoder, line 92: This seems worth a comment. (Also for the other occurrence.)

Cython/Debugger/Tests/test_libcython_in_gdb.py
kentslaney: The class [uses](https://github.com/cython/cython/blob/b31623d5c049063849c7fd4cdf5e72be2ddcf4a7/Cython/Debugger/Tests/test_libcython_in_gdb.py#L55) a [six](https://six.readthedocs.io) based [method](https://github.com/cython/cython/blob/0e1026cb7c767450a88ab822e8eda2b451dee3c6/Cython/Utils.py#L587) to add a metaclass in a Python 2 and 3 compatible way. This results in the class inheriting only from the metaclass that created it. [Here is a StackOverflow question on the topic](https://stackoverflow.com/q/58475561/3476782).

Cython/Debugger/Tests/test_libcython_in_gdb.py
kentslaney: I'm fine with that, I just used assert since that was what was used in the rest of the file. I was hoping for a more end to end test of tab completion but settled on this since it would require digging up a good amount of GDB source. Should I add more detailed explanations? Nothing I can find in the GDB API docs describe what the behavior should be.

kentslaney: Actually, I'm going to revert this pending consistency with the rest of the tests

Cython/Debugger/Cygdb.py
kentslaney: In my testing environment, this part of the symbol table hadn't been loaded by the time this was executed. For reference, [`PyModuleObject` is defined here](https://github.com/python/cpython/blob/fd49e226700e2483a452c3c92da6f15d822ae054/Include/internal/pycore_moduleobject.h#L25). Not sure if it got moved since this was written or GDB started loading symbol tables more lazily. Regardless, it seems better practice to check that GDB has a symbol table corresponding to the executable rather than checking for the existence of a specific type, even if it is in the core definitions.

Does the original setup work on your setup? Was there something more specific you wanted a comment about? Here's documentation on [inferiors and progspaces](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Inferiors-Connections-and-Programs.html) and on [the `symbol_file` attribute specifically](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Progspaces-In-Python.html#index-Progspace_002esymbol_005ffile)

scoder: Then it's just a matter of wording. The words used in the expression are not very obvious (to me).
```suggestion
                    # Check if the Python executable provides a symbol table.
                    if not hasattr(gdb.selected_inferior().progspace, "symbol_file"):
```

Cython/Debugger/Tests/test_libcython_in_gdb.py
scoder: Ok, then let's get rid of six for this instead. The master branch is Py3-only.

Cython/Debugger/Tests/test_libcython_in_gdb.py
scoder: It's a mix currently that is worth cleaning up at some point. But we shouldn't add to the mess from the wrong end.

Cython/Debugger/Tests/test_libcython_in_gdb.py
```diff
@@ -505,9 +511,7 @@ def _debug(*messages):
 
 
 def run_unittest_in_module(modulename):
-    try:
-        gdb.lookup_type('PyModuleObject')
-    except RuntimeError:
+    if not hasattr(gdb.selected_inferior().progspace, "symbol_file"):
```
scoder, line 514:
```suggestion
    # Check if the Python executable provides a symbol table.
    if not hasattr(gdb.selected_inferior().progspace, "symbol_file"):
```

Cython/Debugger/Tests/TestLibCython.py
```diff
@@ -195,7 +195,7 @@ def excepthook(type, value, tb):
             end
             ''')
 
-        code = textwrap.dedent('''\
+        code = textwrap.dedent('''
```
kentslaney, line 198: `../Cygdb.py:99` doesn't include a tailing line break, so I fixed it here since it seemed like the solution with the minimal change to behavior. Based on the willingness to remove dependency on `add_metaclass`, I'm going to switch.

scoder:
Thanks

PR #5950, opened 2024-01-16
cygdb step reliability

kentslaney:
Cygdb's step function is specified as stepping into any Cython call or any C function called directly from Cython. This means it's possible for the existing `get_c_lineno` to return a line number in an included C file that was called directly by Cython. This patch fixes, for example, this [minimal repo](https://github.com/pearcemc/cython-example/tree/8a468acba5aabdd41a432efb86e9e24172942004).

I've left in the `repr` implementations I used to debug the problem.

### DIFF
--- a/Cython/Debugger/Tests/test_libcython_in_gdb.py
+++ b/Cython/Debugger/Tests/test_libcython_in_gdb.py
@@ -388,7 +388,7 @@ class TestUpDown(DebugTestCase):
 
         result = gdb.execute('cy up', to_string=True)
         assert 'spam()' in result
-        assert 'os.path.join("foo", "bar")' in result
+        assert 'some_c_function()' not in result
 
 
 class TestExec(DebugTestCase):
--- a/Cython/Debugger/libcython.py
+++ b/Cython/Debugger/libcython.py
@@ -117,6 +117,25 @@ def gdb_function_value_to_unicode(function):
 # Classes that represent the debug information
 # Don't rename the parameters of these classes, they come directly from the XML
 
+def simple_repr(self, renamed={}, skip=[], state=True):
+    import inspect
+    args = tuple(inspect.signature(self.__init__).parameters)
+    args = tuple(renamed[i] if i in renamed else i for i in args)
+    local = tuple(filter(lambda x: x not in args, vars(self).keys())) \
+        if state else ()
+
+    def equals (prefix, attrs):
+        for i in (j for j in attrs if j not in skip):
+            attr = repr(getattr(self, i))
+            yield prefix + i + " = " + attr.replace("\n", "\n\t\t")
+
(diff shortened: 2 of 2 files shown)


kentslaney:
Speaking of which, why are C functions called directly from Cython considered relevant when stepping? It seems like a Cython step should mean stepping to another line of Cython rather than C. Wouldn't using `step` directly after `cy step` achieve the same thing as a second `cy step`?

kentslaney:
For the sake of making the part of the PR that resolves errors more reviewable, I've split the debugging convenience methods into a separate pull request: #5958

kentslaney:
I'm not sure if anyone else has this problem too, but I can't get `test_libcython_in_gdb.py` to use python compiled with debugging symbols unless I use a docker environment that only has that version. I made a [branch with the Dockerfile](https://github.com/kentslaney/cython/tree/dockerfile) and also [uploaded the resulting image](https://hub.docker.com/repository/docker/kentslaney/cython-dbg/).

I haven't gotten around to reproducing the test case in `codefile`, but [this script illustrates the problem](https://gist.githubusercontent.com/kentslaney/d3c68aa1fe04d61ece1c89ea83f5c851/raw/7dc10a1e662b7161bde099ad0b862581cd01178a/cython-step-example.sh) (file link in case you prefer `curl ... | bash`):
```bash
git clone https://github.com/pearcemc/cython-example.git example
cd example
sed -i 's/print \(.*\)$/print(\1)/' src/foo.pyx
sed -i 's/\(cythonize.*\))$/\1, gdb_debug=True)/' setup.py
cd test
cygdb .. -- \
    -ex "cy break src.metafoo.MetaFoo.special_print" \
    -ex run \
    -ex "cy step" \
    --args python test_metafoo.py
```

kentslaney:
I'm getting some permissions errors in the docker container now. They can be resolved by stopping the docker (systemctl) service and changing the user in `config.v2.json` to root, then restarting docker. The bashrc won't work so the venv in the cython repo will have to be activated manually. I might fix the image if I can't find a minimal reproduction of the issue. For the record, the error is
```
  File "/home/docker/cython/Cython/Debugger/libpython.py", line 1953, in wrapper
  File "/home/docker/cython/Cython/Debugger/libcython.py", line 926, in invoke
    self.step(stepinto=self.stepinto)
  File "/home/docker/cython/Cython/Debugger/libpython.py", line 2390, in step
    lineno = self.lang_info.lineno(newframe)
  File "/home/docker/cython/Cython/Debugger/libcython.py", line 874, in lineno
    return self.get_cython_lineno(frame)[1]
TypeError: 'int' object is not subscriptable
Python Exception <class 'TypeError'>: 'int' object is not subscriptable
Error occurred in Python: 'int' object is not subscriptable
```

PR #5958, opened 2024-01-22
Add repr methods to GDB classes for debugging purposes

kentslaney:
I found these convenient for debugging, so I figured I'd offer them to the repo.

That being said, they were written for functionality rather than readability and violate [some people's style recommendations](https://docs.python-guide.org/writing/style/#explicit-code).


Cython/Debugger/libcython.py
```diff
@@ -117,6 +117,25 @@ def wrapper(self, string, *args, **kwargs):
 # Classes that represent the debug information
 # Don't rename the parameters of these classes, they come directly from the XML
 
+def simple_repr(self, renamed={}, skip=[], state=True):
+    import inspect
+    args = tuple(inspect.signature(self.__init__).parameters)
+    args = tuple(renamed[i] if i in renamed else i for i in args)
+    local = tuple(filter(lambda x: x not in args, vars(self).keys())) \
+        if state else ()
```
scoder, line 125: I'm not sure if we want ordering here in order to get reproducible results?
```suggestion
    local = ()
    if state:
        local = tuple(arg_name for arg_name in vars(self).keys() if arg_name not in args)
```

Cython/Debugger/libcython.py
```diff
@@ -117,6 +117,25 @@ def wrapper(self, string, *args, **kwargs):
...
+    def equals(prefix, attrs):
+        for i in (j for j in attrs if j not in skip):
+            attr = repr(getattr(self, i))
+            yield prefix + i + " = " + attr.replace("\n", "\n\t\t")
+
+    return "".join((
+            self.__class__.__name__,
+            "(", ",".join(equals("\n\t\t", args)), "\n\t)",
+            *equals("\nself.", local)
+        ))
```
scoder, line 136: This doesn't feel like a tuple, more like a list.
```suggestion
    return "".join([
            self.__class__.__name__,
            "(", ",".join(equals("\n\t\t", args)), "\n\t)",
            *equals("\nself.", local)
        ])
```

scoder (commented):
I added comments on style issues. Could you please add docstrings so that it's clearer what the purpose of these functions is and when/how they'd be used? Tests would also be nice.

Cython/Debugger/libcython.py
```diff
@@ -117,6 +117,61 @@ def wrapper(self, string, *args, **kwargs):
 # Classes that represent the debug information
 # Don't rename the parameters of these classes, they come directly from the XML
 
+def simple_repr(self, renamed={}, skip=[], state=True):
```
scoder, line 120: Let's avoid mutable default arguments.
```suggestion
def simple_repr(self, renamed=None, skip=(), state=True):
```

Cython/Debugger/libcython.py
```diff
@@ -117,6 +117,61 @@ def wrapper(self, string, *args, **kwargs):
...
+                Computed properties are already skipped.
+                Only changes behavior if `state` argument is True (default).
+
+    state       Boolean representing whether properties outside the
+                initialization parameters should be printed (self.prop = ...).
+                Using `False` may make the class more amenable to recursive repr
+    """
+    import inspect
+    init_arg_names = tuple(inspect.signature(self.__init__).parameters)
+    init_attrs = tuple(renamed.get(arg, arg) for arg in init_arg_names)
```
scoder, line 154:
```suggestion
    init_attrs = tuple(renamed.get(arg, arg) for arg in init_arg_names) if renamed else init_arg_names
```

Cython/Debugger/libcython.py
```diff
@@ -129,6 +184,9 @@ def __init__(self, module_name, filename, c_filename):
         self.lineno_c2cy = {}
         self.functions = {}
 
+    def __repr__(self):
+        return simple_repr(self, {"module_name": "name"}, state=False)
```
scoder, line 188:
```suggestion
        return simple_repr(self, renamed={"module_name": "name"}, state=False)
```

Cython/Debugger/libcython.py
```diff
@@ -165,6 +226,45 @@ def __init__(self,
...
+        value = getattr(frame, attribute)
+        if callable(value):
+            try:
+                value = value()
+            except Exception as e:
+                pass
+
+        if type(value) in [gdb.Symtab_and_line, gdb.Symbol, gdb.Symtab]:
+            # strip last line since it will get added on at the end of the loop
+            value = frame_repr(value).strip("\n").replace("\n", "\n\t")
```
scoder, line 260:
```suggestion
            # strip last line since it will get added on at the end of the loop
            value = frame_repr(value).rstrip("\n").replace("\n", "\n\t")
```

Cython/Debugger/libcython.py
```diff
@@ -165,6 +226,45 @@ def __init__(self,
...
+        res += attribute + ": "
+        if type(value) == int:
```
scoder, line 262: This is an unusual spelling. You'd normally expect `is` for a type check like this rather than `==`, but why not just use the normal `isinstance()`?
```suggestion
        if isinstance(value, int):
```

Cython/Debugger/libcython.py
```diff
@@ -165,6 +226,45 @@ def __init__(self,
...
+    res = str(frame) + "\n"
+    for attribute in sorted(dir(frame)):
+        if attribute.startswith("__"):
+            continue
+        value = getattr(frame, attribute)
+        if callable(value):
+            try:
+                value = value()
+            except Exception as e:
+                pass
```
scoder, line 256: Calling arbitrary methods on frame objects seems unexpected and risky if the goal is something as obviously read-only as printing out their state. I read the docstring and your reasoning, but it doesn't really convince me that this is a good idea. Frame objects have been subject to major changes recently and in the past, so we should try to make little assumptions about their behaviour.

I'd rather see a whitelist of safe and relevant properties to be printed. This is only for printing, after all. If users want more information, they can query it.

Cython/Debugger/Tests/test_libcython_in_gdb.py
```diff
@@ -137,6 +139,26 @@ def test_CythonFunction(self):
...
+    def test_simple_repr(self):
+        test_class = libcython.CythonModule
+        num_args = len(inspect.signature(test_class).parameters)
+        lorem_ipsum = random.Random(0)
+        filler_args = (lorem_ipsum.randbytes(8) for _ in range(num_args))
+        instance = test_class(*filler_args)
+        recreated = eval("libcython." + repr(instance))
+        self.assertEqual(pickle.dumps(instance), pickle.dumps(recreated))
+
+
```
scoder, line 153:
```suggestion
```

Cython/Debugger/Tests/test_libcython_in_gdb.py
```diff
@@ -137,6 +139,26 @@ def test_CythonFunction(self):
...
+    def test_frame_repr(self):
+        # check that frame_repr's function expansion is idempotent
+        beginline = 'import os'
+
+        self.break_and_run(beginline)
+        frame = gdb.selected_frame()
+        self.assertEqual(libcython.frame_repr(frame),
+                         libcython.frame_repr(frame))
```
scoder, line 162:
```suggestion


```

Cython/Debugger/Tests/test_libcython_in_gdb.py
```diff
@@ -137,6 +139,26 @@ def test_CythonFunction(self):
         self.assertEqual(self.spam_func.lineno, expected_lineno)
         self.assertEqual(sorted(self.spam_func.locals), list('abcd'))
```
scoder, line 141:
```suggestion


```

kentslaney:
I haven't abandoned the thread, my backlog is just taking longer than expected

Cython/Debugger/libcython.py
```diff
@@ -117,6 +117,62 @@ def wrapper(self, string, *args, **kwargs):
...
+        instance_only = vars(self).keys()
+        state_only = (attr for attr in instance_only if attr not in init_attrs)
+        state_repr = tuple(sorted(state_only))
+
+    def equals(prefix, attrs, args=None):
+        for attr, arg in zip(attrs, args or attrs):
+            if attr in skip:
+                continue
+            param = repr(getattr(self, attr))
+            yield prefix + arg + " = " + param.replace("\n", "\n\t\t")
```
scoder, line 167:
```suggestion
            yield f'{prefix}{arg} = {param.replace("\n", "\n\t\t")}'
```

Cython/Debugger/libcython.py
```diff
@@ -117,6 +117,62 @@ def wrapper(self, string, *args, **kwargs):
...
+    init_arg_names = tuple(inspect.signature(self.__init__).parameters)
+    init_attrs = tuple(renamed.get(arg, arg) for arg in init_arg_names) \
+            if renamed else init_arg_names
+    state_repr = ()
+    if state:
+        instance_only = vars(self).keys()
+        state_only = (attr for attr in instance_only if attr not in init_attrs)
+        state_repr = tuple(sorted(state_only))
+
+    def equals(prefix, attrs, args=None):
```
scoder, line 162: `equals` reads like you're making a comparison here. I think `names_and_values` is a better name for this function.

Cython/Debugger/libcython.py
```diff
@@ -117,6 +117,62 @@ def wrapper(self, string, *args, **kwargs):
...
+                Only changes behavior if `state` argument is True (default).
+
+    state       Boolean representing whether properties outside the
+                initialization parameters should be printed (self.prop = ...).
+                Using `False` may make the class more amenable to recursive repr
+    """
+    import inspect
+    init_arg_names = tuple(inspect.signature(self.__init__).parameters)
+    init_attrs = tuple(renamed.get(arg, arg) for arg in init_arg_names) \
+            if renamed else init_arg_names
```
scoder, line 155: There isn't really a reason for this to be a tuple, is there?
```suggestion
    init_attrs = [renamed.get(arg, arg) for arg in init_arg_names] if renamed else init_arg_names
```

Cython/Debugger/libcython.py
```diff
@@ -117,6 +117,62 @@ def wrapper(self, string, *args, **kwargs):
...
+    state_repr = ()
+    if state:
+        instance_only = vars(self).keys()
+        state_only = (attr for attr in instance_only if attr not in init_attrs)
+        state_repr = tuple(sorted(state_only))
```
scoder, line 160: I'd expect the number of attributes to be low enough to just sort all and select afterwards.
```suggestion
        instance_attrs = sorted(vars(self).keys())
        state_repr = [attr for attr in instance_attrs if attr not in init_attrs]
```

Cython/Debugger/libcython.py
```diff
@@ -165,6 +227,65 @@ def __init__(self,
...
+    corresponding source code. This prints a tree of instance properties,
+    expanding the values for Symtab_and_line, Symbol, and Symtab.
+
+    Most of these properties require computation to determine, meaning much of
+    relevant info is behind a monad, a subset of which are evaluated.
+
+    Arguments
+    frame       The GDB.Frame instance to be represented as a string
+    """
+    res = str(frame) + "\n"
```
scoder, line 271:
```suggestion
    res = f"{frame}\n"
```

Cython/Debugger/libcython.py
```diff
@@ -165,6 +227,65 @@ def __init__(self,
...
+            value = value()
+
+        if type(value) in [gdb.Symtab_and_line, gdb.Symbol, gdb.Symtab]:
+            # strip last line since it will get added on at the end of the loop
+            value = frame_repr(value).rstrip("\n").replace("\n", "\n\t")
+        res += attribute + ": "
+        if isinstance(value, int) and attribute != "line":
+            res += hex(value) + "\n"
+        else:
+            res += str(value) + "\n"
```
scoder, line 286:
```suggestion
        res += f"{attribute}: " + (f"{value:x}\n" if isinstance(value, int) and attribute != "line" else f"{value}\n")
```

Cython/Debugger/libcython.py
```diff
@@ -117,6 +117,62 @@ def wrapper(self, string, *args, **kwargs):
 # Classes that represent the debug information
 # Don't rename the parameters of these classes, they come directly from the XML
 
+def simple_repr(self, renamed=None, skip=(), state=True):
```
scoder, line 120: The `skip` argument seems unused and can be removed.
```suggestion
def simple_repr(self, renamed=None, state=True):
```

Cython/Debugger/libcython.py
```diff
@@ -117,6 +117,62 @@ def wrapper(self, string, *args, **kwargs):
...
+    Function arguments:
+    self        Instance to be represented
+
+    renamed     Dictionary of initialization arguments that are stored under a
+                different property name in the form { argument: property }
+
+    skip        List of instance variables not to be listed as properties.
+                Computed properties are already skipped.
+                Only changes behavior if `state` argument is True (default).
+
```
scoder, line 147:
```suggestion
```

Cython/Debugger/libcython.py
```diff
@@ -117,6 +117,62 @@ def wrapper(self, string, *args, **kwargs):
...
+    state_repr = ()
+    if state:
+        instance_only = vars(self).keys()
+        state_only = (attr for attr in instance_only if attr not in init_attrs)
+        state_repr = tuple(sorted(state_only))
+
+    def equals(prefix, attrs, args=None):
+        for attr, arg in zip(attrs, args or attrs):
+            if attr in skip:
+                continue
```
scoder, line 165:
```suggestion
```

scoder (commented):
Looks good overall, mostly style changes. We can use f-strings in Cython 3.1.

scoder (approved):
Sorry for the long silence. This looks good to me. Thanks.

Cython/Debugger/libcython.py
```diff
@@ -165,6 +220,63 @@ def __init__(self,
...
+    gdb.Frame.architecture,
+    gdb.Frame.type,
+    gdb.Frame.pc,
+    gdb.Frame.block,
+    gdb.Frame.function,
+    gdb.Frame.older,
+    gdb.Frame.newer,
+    gdb.Frame.find_sal,
+    gdb.Frame.select,
+    gdb.Frame.static_link,
```
avm19, line 236: `gdb.Frame.static_link` was added in GDB 14.1, which is not yet widely available in package managers. This line makes cygdb incompatible with early versions.

Frankly, I don't know why and how this PR passed CI tests, which seem to use GDB 12.

kentslaney: Just to double check, is this the only line that needs to be removed to fix the build with GDB 12? I spent a bit trying to set up a docker container with a GDB 12.1 manual install but it was taking more troubleshooting than I wanted.

avm19: I am unable to check this right now, also troubleshooting. But generally speaking, do you have to evaluate qualnames on import? If you don't, then hard-coding qualnames should work. And if you do, than you could iterate through `dir(gdb.Frame)` or `gdb.Frame.__dict__` and get the methods. This way you could keep `gdb.Frame.static_link` for newer GDB versions, while maintaining compatibility with older versions.

Also, I do not understand the `# fail fast if the API changes` part. I might be missing some important point, but if the "fail fast" is necessary, could it be moved from runtime to unit tests?

kentslaney: Iterating over `dir(gdb.Frame)` was how this PR worked initially, but it got changed as a result of [this comment](https://github.com/cython/cython/pull/5958#discussion_r1487460084). I don't remember why I thought it was important to fail fast for API changes. Hard-coding the qualnames is probably fine. I'll try to get my development environment running again.

PR #5972, opened 2024-02-03
Don't miss a required PyObject_GC_Track() in dealloc

### LINKED ISSUE #5971: [BUG] null-deref inside _PyObject_GC_UNTRACK from extention class dealloc method.
### Describe the bug

```
SCARINESS: 10 (null-deref)
    #0 0x53c9a5 in _PyObject_GC_UNTRACK(_object*) (.__uniq.71160431842585912442991593982683040670) python/3.10/Include/internal/pycore_object.h:123
    #1 0x53c9a5 in BaseException_dealloc(PyBaseExceptionObject*) (.__uniq.71160431842585912442991593982683040670) python/3.10/Objects/exceptions.c:92
    #2 0x7fb1f0bcfea7 in __pyx_tp_dealloc_6thrift_3py3_10exceptions_GeneratedError(_object*) exceptions.cpp:20592 *cython*
    #3 0x7fb1f4369b4f in __pyx_tp_dealloc_7testing_5types_UnusedError(_object*) types.cpp:165249  *cython* 
    #4 0x3d3658 in _Py_Dealloc python/3.10/Objects/object.c:2301
```

I believe this is related to this commit https://github.com/cython/cython/commit/663c924462adec202c9f469fa346bda1b2264dfa from #5432 

The generated .cpp code between the old version of cython and 3.0.8 is missing these lines
```
#if CYTHON_USE_TYPE_SLOTS
if (PyType_IS_GC(Py_TYPE(o)->tp_base))
#endif
PyObject_GC_Track(o);
 ```
 
 The missing GC_Track would explain why GC is exploding  

```cython
# We have no custom __dealloc__ or __del__
# Exception is the builtin python Exception
# inside the exceptions.pxd we have this def for Exception
cdef extern from "Python.h":
    ctypedef extern class builtins.Exception[object PyBaseExceptionObject]:
        pass


# inside the exceptions.pyx
cdef class Error(Exception):
   # Has no slots
    ...

cdef class GeneratedError(Error):
  # has some attribute slots
   ...
```

### Expected behaviour

no crash

### OS

Linux

### Python version

3.10.9

### Cython version

3.0.8

### Additional context

_No response_

fried:
Fixes #5971 

This works to solve me.  But I'm not sure if there is some mechanism to figure this situation in the compiler 

class DoesntNeedGC(BuiltinThatNeedsGC): pass

class NeedsGC(DoesntNeedGC): pass

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -1761,21 +1761,24 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
 
         if base_type:
             base_cname = base_type.typeptr_cname
-            if needs_gc:
-                # The base class deallocator probably expects this to be tracked,
-                # so undo the untracking above.
-                # Assume that we do not know whether the base class uses GC or not.
-                if base_type.scope and base_type.scope.needs_gc():
-                    code.putln("PyObject_GC_Track(o);")
-                else:
-                    code.putln("if (PyType_IS_GC(%s)) PyObject_GC_Track(o);" % base_cname)
-
             tp_dealloc = TypeSlots.get_base_slot_function(scope, tp_slot)
             if tp_dealloc is not None:
+                if needs_gc and base_type.scope and base_type.scope.needs_gc():
+                    # We know that the base class uses GC, so probably expects it to be tracked.
+                    # Undo the untracking above.
+                    code.putln("PyObject_GC_Track(o);")
                 code.putln("%s(o);" % tp_dealloc)
             elif base_type.is_builtin_type:
+                if needs_gc and base_type.scope and base_type.scope.needs_gc():
+                    # We know that the base class uses GC, so probably expects it to be tracked.
+                    # Undo the untracking above.
+                    code.putln("PyObject_GC_Track(o);")
                 code.putln("__Pyx_PyType_GetSlot(%s, tp_dealloc, destructor)(o);" % base_cname)
             else:
+                if needs_gc:
+                    # We don't know if the base class uses GC or not, so must find out at runtime
+                    # whether we should undo the untracking above or not.
+                    code.putln("if (PyType_IS_GC(%s)) PyObject_GC_Track(o);" % base_cname)
                 # This is an externally defined type.  Calling through the
                 # cimported base type pointer directly interacts badly with
                 # the module cleanup, which may already have cleared it.
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -1764,11 +1764,13 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
             if needs_gc:
                 # The base class deallocator probably expects this to be tracked,
                 # so undo the untracking above.
-                # Assume that we do not know whether the base class uses GC or not.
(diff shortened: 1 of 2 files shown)


Cython/Compiler/ModuleNode.py
```diff
@@ -1764,10 +1764,9 @@ def generate_dealloc_function(self, scope, code):
             if needs_gc:
                 # The base class deallocator probably expects this to be tracked,
                 # so undo the untracking above.
-                if base_type.scope:
-                    # Assume that we know whether the base class uses GC or not.
-                    if base_type.scope.needs_gc():
-                        code.putln("PyObject_GC_Track(o);")
+                # Assume that we do not know weather the base class uses GC or not.
```
da-woods, line 1767:
```suggestion
                # Assume that we do not know whether the base class uses GC or not.
```

da-woods (commented):
Could you add the example from your issue to https://github.com/cython/cython/blob/master/tests/run/exttype_gc.pyx so that we have a test and we don't revert it again?

scoder:
Thanks. Your failing test uses an extern class, so I think it should be enough to do this for extern classes only, i.e. assume that we may not know the entire scope correctly if the class is extern, and do not rely on the "needs GC" detection in this case.

Cython/Compiler/ModuleNode.py
```diff
@@ -1764,10 +1764,9 @@ def generate_dealloc_function(self, scope, code):
...
                 # so undo the untracking above.
-                if base_type.scope:
-                    # Assume that we know whether the base class uses GC or not.
-                    if base_type.scope.needs_gc():
-                        code.putln("PyObject_GC_Track(o);")
+                # Assume that we do not know whether the base class uses GC or not.
+                if base_type.scope and base_type.scope.needs_gc():
+                    code.putln("PyObject_GC_Track(o);")
                 else:
                     code.putln("if (PyType_IS_GC(%s)) PyObject_GC_Track(o);" % base_cname)
```
da-woods, line 1771:
```suggestion
                # extern types are unknown so we always have to check for gc
                if not base_type.scope or (base.type.is_external and not base_type.is_builtin_type):
                    code.putln("if (PyType_IS_GC(%s)) PyObject_GC_Track(o);" % base_cname)
                else:
                    # Assume that we know whether the base class uses GC or not.
                    if base_type.scope.needs_gc():
                        code.putln("PyObject_GC_Track(o);")
```

I think @scoder's suggestion was something like this, to special case extern types and treat everything else as not needing GC.

I've reordered the if/else because the logic was clearer to me that way, but you may not want to do this.

fried: Let me import the change and see if it works in our codebase

scoder:
… and you reverted it, because … ?

fried:
It didn’t work the segfault came back

da-woods:
My mistake I think - the `base_type` that we were checking in my change is `Error` (i.e. our intermediate type rather than the extern type that is the ultimate base class).

Try the commit below - this bases the logic on the correct base class:

https://github.com/da-woods/cython/commit/9121158278939dcb10747bc68d9db0525a998f1e

da-woods:
@scoder if you're looking at a 3.0.9 release in the near future, it'd be good to have this I think

PR #6003, opened 2024-02-16
Use Python 3.13a4 PyCFunctionFastWithKeywords

vstinner:
Python 3.13a4 adds a public PyCFunctionFastWithKeywords type and removes the private _PyCFunctionFastWithKeywords type:

* https://github.com/python/cpython/commit/9e3729bbd77fb9dcaea6a06ac760160136d80b79
* https://github.com/python/cpython/pull/114627
* https://github.com/capi-workgroup/decisions/issues/11

### DIFF
--- a/Cython/Utility/ModuleSetupCode.c
+++ b/Cython/Utility/ModuleSetupCode.c
@@ -907,8 +907,13 @@ class __Pyx_FakeReference {
   typedef PyObject *(*__Pyx_PyCFunctionFastWithKeywords) (PyObject *self, PyObject *const *args,
                                                           Py_ssize_t nargs, PyObject *kwnames);
 #else
-  #define __Pyx_PyCFunctionFast _PyCFunctionFast
-  #define __Pyx_PyCFunctionFastWithKeywords _PyCFunctionFastWithKeywords
+  #if PY_VERSION_HEX >= 0x030d00A4
+  #  define __Pyx_PyCFunctionFast PyCFunctionFast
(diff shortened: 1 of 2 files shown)


Cython/Utility/ModuleSetupCode.c
```diff
@@ -908,7 +908,11 @@ class __Pyx_FakeReference {
                                                           Py_ssize_t nargs, PyObject *kwnames);
 #else
   #define __Pyx_PyCFunctionFast _PyCFunctionFast
```
davidhewitt, line 910: Should the define using `_PyCFunctionFast` also be updated in the same way?

Cython/Utility/ModuleSetupCode.c
```diff
@@ -908,7 +908,11 @@ class __Pyx_FakeReference {
...
-  #define __Pyx_PyCFunctionFastWithKeywords _PyCFunctionFastWithKeywords
+  #if PY_VERSION_HEX >= 0x030D00A4
```
scoder, line 911: 'D' is not easy to distinguish from '0', so we spell the hexver with a lower-case 'd'.
```suggestion
  #if PY_VERSION_HEX >= 0x030d00A4
```

Cython/Utility/ModuleSetupCode.c
vstinner: Oh right, fixed.

vstinner:
@scoder @davidhewitt: Please review the PR updated to address your reviews.

vstinner:
Thanks for the quick merge!

PR #6037, opened 2024-02-26
Added missing `replace` functions to C++ `std::string` definition

oh-ok:
Hi, I noticed the other day that the `string.replace(...)` functions were missing in libcpp.string. Assuming there was no reason they are missing, I've added them to the pxd file.


scoder:
Thanks

PR #6041, opened 2024-02-28
Add fixing of warning in 3.0.9 to CHANGES.rst

matusvalo:
Changlog for https://github.com/cython/cython/commit/ead1ac69413e13d57a23a5a336a6896850a5c172

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -122,6 +122,8 @@ Features added
 Bugs fixed
 ----------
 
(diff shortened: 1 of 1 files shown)


scoder:
Yeah, I had left it out because I didn't consider it relevant enough to be mentioned in the changelog, but I admit that it was on the edge. I'm fine with adding it.

matusvalo:
I considered it as valid since other fixes of C warnings are present in the changelog.

PR #6061, opened 2024-03-06
Add some more limited API support

vyasr:
This PR adds some additional support for using Python's limited API in more parts of Cython.

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -10255,12 +10255,6 @@ class CodeObjectNode(ExprNode):
         num_posonly_args = func.num_posonly_args  # Py3.8+ only
         kwonly_argcount = func.num_kwonly_args
         nlocals = len(self.varnames)
-        # TODO: Not clear if something changed in Python make this typing change
-        # necessary, but I am seeing compiler warnings about narrowing. Looking in
-        # Include/cpython/code.h (in the cpython source), co_flags has always been a
-        # signed int but the flags are just #defined as bit flags, so the type (signed
-        # or unsigned) should be contextually determined. I wonder if this is a function
-        # of my compiler.
         flags = '(unsigned int)(%s)' % '|'.join(flags) or '0'
         first_lineno = self.pos[1]
 
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -10255,6 +10255,13 @@ class CodeObjectNode(ExprNode):
         num_posonly_args = func.num_posonly_args  # Py3.8+ only
         kwonly_argcount = func.num_kwonly_args
         nlocals = len(self.varnames)
+        # TODO: Not clear if something changed in Python make this typing change
+        # necessary, but I am seeing compiler warnings about narrowing. Looking in
+        # Include/cpython/code.h (in the cpython source), co_flags has always been a
+        # signed int but the flags are just #defined as bit flags, so the type (signed
+        # or unsigned) should be contextually determined. I wonder if this is a function
+        # of my compiler.
+        flags = [f"((unsigned int) {flag})" for flag in flags]
         flags = '|'.join(flags) or '0'
         first_lineno = self.pos[1]
 
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -11500,11 +11500,19 @@ class CythonArrayNode(ExprNode):
         ))
         code.put_gotref(shapes_temp, py_object_type)
 
+        code.putln("#if CYTHON_COMPILING_IN_LIMITED_API")
+        code.putln('%s = __pyx_array_new(%s, %s, PyBytes_AsString(%s), "%s", (char *) %s); %s' % (
+            self.result(),
+            shapes_temp, itemsize, format_temp, self.mode, self.operand.result(),
+            code.error_goto_if_null(self.result(), self.pos),
+        ))
+        code.putln("#else")
         code.putln('%s = __pyx_array_new(%s, %s, PyBytes_AS_STRING(%s), "%s", (char *) %s); %s' % (
             self.result(),
             shapes_temp, itemsize, format_temp, self.mode, self.operand.result(),
             code.error_goto_if_null(self.result(), self.pos),
         ))
+        code.putln("#endif")
         self.generate_gotref(code)
 
         def dispose(temp):
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -10261,8 +10261,7 @@ class CodeObjectNode(ExprNode):
         # signed int but the flags are just #defined as bit flags, so the type (signed
         # or unsigned) should be contextually determined. I wonder if this is a function
         # of my compiler.
-        flags = [f"((unsigned int) {flag})" for flag in flags]
-        flags = '|'.join(flags) or '0'
+        flags = '(unsigned int)(%s)' % '|'.join(flags) or '0'
         first_lineno = self.pos[1]
 
         # See "generate_codeobject_constants()" in Code.py.
@@ -11533,7 +11532,7 @@ class CythonArrayNode(ExprNode):
 
         dispose(shapes_temp)
         dispose(format_temp)
-        dispose(format_ptr_temp)
+        code.funcstate.release_temp(format_ptr_temp)
 
     @classmethod
     def from_carray(cls, src_node, env):
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -30,7 +30,7 @@ from . import Naming
 from . import Nodes
 from .Nodes import Node, SingleAssignmentNode
 from . import PyrexTypes
-from .PyrexTypes import py_object_type, typecast, error_type, \
+from .PyrexTypes import c_char_ptr_type, py_object_type, typecast, error_type, \
     unspecified_type
 from . import TypeSlots
 from .Builtin import (
@@ -11473,6 +11473,7 @@ class CythonArrayNode(ExprNode):
 
         shapes_temp = code.funcstate.allocate_temp(py_object_type, True)
         format_temp = code.funcstate.allocate_temp(py_object_type, True)
+        format_ptr_temp = code.funcstate.allocate_temp(c_char_ptr_type, True)
 
         itemsize = "sizeof(%s)" % dtype.empty_declaration_code()
         type_info = Buffer.get_type_information_cname(code, dtype)
@@ -11500,19 +11501,23 @@ class CythonArrayNode(ExprNode):
         ))
         code.put_gotref(shapes_temp, py_object_type)
 
+
         code.putln("#if CYTHON_COMPILING_IN_LIMITED_API")
-        code.putln('%s = __pyx_array_new(%s, %s, PyBytes_AsString(%s), "%s", (char *) %s); %s' % (
-            self.result(),
-            shapes_temp, itemsize, format_temp, self.mode, self.operand.result(),
-            code.error_goto_if_null(self.result(), self.pos),
+        code.putln('%s = PyBytes_AsString(%s); %s' % (
+            format_ptr_temp, format_temp,
+            code.error_goto_if_null(format_ptr_temp, self.pos),
         ))
         code.putln("#else")
-        code.putln('%s = __pyx_array_new(%s, %s, PyBytes_AS_STRING(%s), "%s", (char *) %s); %s' % (
-            self.result(),
-            shapes_temp, itemsize, format_temp, self.mode, self.operand.result(),
-            code.error_goto_if_null(self.result(), self.pos),
+        code.putln('%s = PyBytes_AS_STRING(%s);' % (
+            format_ptr_temp, format_temp,
         ))
         code.putln("#endif")
+
+        code.putln('%s = __pyx_array_new(%s, %s, %s, "%s", (char *) %s); %s' % (
+            self.result(),
+            shapes_temp, itemsize, format_ptr_temp, self.mode, self.operand.result(),
+            code.error_goto_if_null(self.result(), self.pos),
+        ))
         self.generate_gotref(code)
 
         def dispose(temp):
@@ -11521,6 +11526,7 @@ class CythonArrayNode(ExprNode):
 
         dispose(shapes_temp)
         dispose(format_temp)
+        dispose(format_ptr_temp)
 
     @classmethod
     def from_carray(cls, src_node, env):
--- a/Cython/Utility/CppConvert.pyx
+++ b/Cython/Utility/CppConvert.pyx
@@ -69,7 +69,7 @@ cdef extern from *:
 cdef extern from "Python.h":
     void Py_INCREF(object)
     list PyList_New(Py_ssize_t size)
-    void PyList_SetItem(object list, Py_ssize_t i, object o)
+    int __Pyx_PyList_SET_ITEM(object list, Py_ssize_t i, object o) except -1
     const Py_ssize_t PY_SSIZE_T_MAX
 
 @cname("{{cname}}")
@@ -86,7 +86,7 @@ cdef object {{cname}}(const vector[X]& v):
     for i in range(v_size_signed):
         item = v[i]
         Py_INCREF(item)
-        PyList_SetItem(o, i, item)
+        __Pyx_PyList_SET_ITEM(o, i, item)
 
     return o
 
@@ -121,7 +121,7 @@ cdef extern from *:
 cdef extern from "Python.h":
     void Py_INCREF(object)
     list PyList_New(Py_ssize_t size)
-    void PyList_SetItem(object list, Py_ssize_t i, object o)
+    void __Pyx_PyList_SET_ITEM(object list, Py_ssize_t i, object o)
     cdef Py_ssize_t PY_SSIZE_T_MAX
 
 @cname("{{cname}}")
@@ -138,7 +138,7 @@ cdef object {{cname}}(const cpp_list[X]& v):
     while iter != v.end():
         item = cython.operator.dereference(iter)
         Py_INCREF(item)
-        PyList_SetItem(o, i, item)
+        __Pyx_PyList_SET_ITEM(o, i, item)
         cython.operator.preincrement(iter)
         i += 1
 
--- a/Cython/Utility/CppConvert.pyx
+++ b/Cython/Utility/CppConvert.pyx
@@ -67,7 +67,7 @@ cdef extern from *:
         T& operator[](size_t)
 
 cdef extern from "Python.h":
-    void Py_IncRef(object)
+    void Py_INCREF(object)
     list PyList_New(Py_ssize_t size)
     void PyList_SetItem(object list, Py_ssize_t i, object o)
     const Py_ssize_t PY_SSIZE_T_MAX
@@ -85,7 +85,7 @@ cdef object {{cname}}(const vector[X]& v):
 
     for i in range(v_size_signed):
         item = v[i]
-        Py_IncRef(item)
+        Py_INCREF(item)
         PyList_SetItem(o, i, item)
 
     return o
@@ -119,7 +119,7 @@ cdef extern from *:
         size_t size()
 
 cdef extern from "Python.h":
-    void Py_IncRef(object)
+    void Py_INCREF(object)
     list PyList_New(Py_ssize_t size)
     void PyList_SetItem(object list, Py_ssize_t i, object o)
     cdef Py_ssize_t PY_SSIZE_T_MAX
@@ -137,7 +137,7 @@ cdef object {{cname}}(const cpp_list[X]& v):
 
     while iter != v.end():
         item = cython.operator.dereference(iter)
-        Py_IncRef(item)
+        Py_INCREF(item)
         PyList_SetItem(o, i, item)
         cython.operator.preincrement(iter)
         i += 1
--- a/Cython/Utility/CppConvert.pyx
+++ b/Cython/Utility/CppConvert.pyx
@@ -67,9 +67,9 @@ cdef extern from *:
         T& operator[](size_t)
 
 cdef extern from "Python.h":
-    void Py_INCREF(object)
+    void Py_IncRef(object)
     list PyList_New(Py_ssize_t size)
-    void PyList_SET_ITEM(object list, Py_ssize_t i, object o)
+    void PyList_SetItem(object list, Py_ssize_t i, object o)
     const Py_ssize_t PY_SSIZE_T_MAX
 
 @cname("{{cname}}")
@@ -85,8 +85,8 @@ cdef object {{cname}}(const vector[X]& v):
 
     for i in range(v_size_signed):
         item = v[i]
-        Py_INCREF(item)
-        PyList_SET_ITEM(o, i, item)
+        Py_IncRef(item)
+        PyList_SetItem(o, i, item)
 
     return o
 
@@ -119,9 +119,9 @@ cdef extern from *:
         size_t size()
 
 cdef extern from "Python.h":
-    void Py_INCREF(object)
+    void Py_IncRef(object)
     list PyList_New(Py_ssize_t size)
-    void PyList_SET_ITEM(object list, Py_ssize_t i, object o)
+    void PyList_SetItem(object list, Py_ssize_t i, object o)
     cdef Py_ssize_t PY_SSIZE_T_MAX
 
 @cname("{{cname}}")
@@ -137,8 +137,8 @@ cdef object {{cname}}(const cpp_list[X]& v):
 
     while iter != v.end():
         item = cython.operator.dereference(iter)
-        Py_INCREF(item)
-        PyList_SET_ITEM(o, i, item)
+        Py_IncRef(item)
+        PyList_SetItem(o, i, item)
         cython.operator.preincrement(iter)
         i += 1
 
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -1440,11 +1440,7 @@ __pyx_FusedFunction_getitem(__pyx_FusedFunctionObject *self, PyObject *idx)
             Py_DECREF(item);
 #endif
             if (unlikely(!string)) goto __pyx_err;
-#if CYTHON_COMPILING_IN_LIMITED_API
-            PyList_SetItem(list, i, string);
-#else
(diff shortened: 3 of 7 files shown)


Cython/Utility/CppConvert.pyx
vyasr, line 1: Since this file is Cython the C macros aren't available to create compile-time decision-points. I assumed that they wouldn't be performance-critical and switched all macros over to use functions from the limited API instead. Let me know if you prefer a different approach.

Cython/Utility/ImportExport.c
vyasr, line 1: Before reviewing this file I'd recommend looking at my changes in ModuleSetupCode.c

Cython/Utility/ModuleSetupCode.c
```diff
@@ -1767,6 +1767,20 @@ static CYTHON_INLINE int __Pyx_Is_Little_Endian(void)
...
+#define _ModuleSetupXDecRef(obj) Py_DecRef(obj);
+#define _ModuleSetupIncRef(obj) Py_IncRef(obj);
+#define _ModuleSetupXIncRef(obj) Py_IncRef(obj);
+#else
+#define _ModuleSetupDecRef(obj) Py_DECREF(obj);
+#define _ModuleSetupXDecRef(obj) Py_XDECREF(obj);
+#define _ModuleSetupIncRef(obj) Py_INCREF(obj);
+#define _ModuleSetupXIncRef(obj) Py_XINCREF(obj);
+#endif
+#endif
```
vyasr, line 1782: I've basically copy-pasted equivalent macros throughout this PR at the moment because I wasn't sure what order different snippets would get included into the file to know which macros would be defined when. Ideally these kinds of macros could be centralized somewhere. Is this module the best place to put something like that? Could this be patched into another `#else` block of the `#if CYTHON_REFNANNY` if-else below?

Cython/Utility/CythonFunction.c
```diff
@@ -1530,7 +1538,11 @@ __pyx_FusedFunction_call(PyObject *func, PyObject *args, PyObject *kw)
 #else
             PyObject *item = __Pyx_PySequence_ITEM(args, i);  if (unlikely(!item)) goto bad;
 #endif
-            PyTuple_SET_ITEM(new_args, i + 1, item);
+#if CYTHON_COMPILING_IN_LIMITED_API
+        PyTuple_SetItem(new_args, i + 1, item);
+#else
+        PyTuple_SET_ITEM(new_args, i + 1, item);
+#endif
```
vyasr, line 1545: I ask the same question about incref/decref below where it's much more important due to frequency of use, but asking here more generally: instead of having these conditionals everywhere would it make sense to define a common set of conditional macros somewhere that can be used throughout the code base and would be correctly selected for the limited API or not at compile time?

da-woods (commented):
`Py_INCREF` and `Py_DECREF`: - now that I look these aren't specified as being in the limited api.  Although practically it looks like they are actually included so we can get away with using them.

I think the main advantage of using the functions rather than the macros is that it handles the upcoming "immortal object" Python changes better?  Using the functions will always do the right thing, while using the macros will attempt to refcount immortal objects with earlier `Py_LIMITED_API` versions set (but that doesn't cause a huge problem).

I think it's a problem if we're inconsistent - if we replaced a decref but not an incref then we'd slowly lose references to immortal objects and they'd become non immortal. That suggests to me that any change should be done *everywhere* in one go.

Agree that macros in ModuleSetupCode is the way to do it - I'd change the naming scheme though

Cython/Utility/ImportExport.c
```diff
@@ -3,16 +3,31 @@
 static PyObject *__Pyx_ImportDottedModule(PyObject *name, PyObject *parts_tuple); /*proto*/
 static PyObject *__Pyx_ImportDottedModule_WalkParts(PyObject *module, PyObject *name, PyObject *parts_tuple); /*proto*/
 
+
 /////////////// ImportDottedModule ///////////////
 //@requires: Import
 
+#ifndef _ImportDecRef
+#if CYTHON_COMPILING_IN_LIMITED_API
+#define _ImportDecRef(obj) Py_DecRef(obj);
```
da-woods, line 12: I'd just use the macros in ModuleSetupCode.c - they're available absolutely everywhere so you don't need specific macros here.

Cython/Utility/ImportExport.c
```diff
@@ -3,16 +3,31 @@
...
+#endif
+
 static PyObject *__Pyx__ImportDottedModule_Error(PyObject *name, PyObject *parts_tuple, Py_ssize_t count) {
     PyObject *partial_name = NULL, *slice = NULL, *sep = NULL;
     Py_ssize_t size;
     if (unlikely(PyErr_Occurred())) {
         PyErr_Clear();
     }
-#if CYTHON_ASSUME_SAFE_SIZE
+#if CYTHON_ASSUME_SAFE_SIZE && !CYTHON_COMPILING_IN_LIMITED_API
```
da-woods, line 30: Not needed - `CYTHON_ASSUME_SAFE_SIZE` is implied by LIMITED_API

Cython/Utility/ImportExport.c
```diff
@@ -57,15 +72,15 @@ static PyObject *__Pyx__ImportDottedModule_Lookup(PyObject *name) {
 
 static PyObject *__Pyx_ImportDottedModule_WalkParts(PyObject *module, PyObject *name, PyObject *parts_tuple) {
     Py_ssize_t i, nparts;
-#if CYTHON_ASSUME_SAFE_SIZE
+#if CYTHON_ASSUME_SAFE_SIZE && !CYTHON_COMPILING_IN_LIMITED_API
```
da-woods, line 75: Also not needed

Cython/Utility/ImportExport.c
```diff
@@ -57,15 +72,15 @@ static PyObject *__Pyx__ImportDottedModule_Lookup(PyObject *name) {
...
     nparts = PyTuple_GET_SIZE(parts_tuple);
 #else
     nparts = PyTuple_Size(parts_tuple);
     if (nparts < 0) return NULL;
 #endif
     for (i=1; i < nparts && module; i++) {
         PyObject *part, *submodule;
-#if CYTHON_ASSUME_SAFE_MACROS && !CYTHON_AVOID_BORROWED_REFS
+#if CYTHON_ASSUME_SAFE_MACROS && !CYTHON_AVOID_BORROWED_REFS && !CYTHON_COMPILING_IN_LIMITED_API
```
da-woods, line 83: Also not needed

Cython/Utility/ImportExport.c
```diff
@@ -336,21 +407,32 @@ static int ${import_star}(PyObject* m) {
     if (__Pyx_import_all_from(locals, m) < 0) goto bad;
     list = PyDict_Items(locals);        if (!list) goto bad;
 
+#if CYTHON_COMPILING_IN_LIMITED_API
+    for(i=0; i<PyList_Size(list); i++) {
+        name = PyTuple_Get_Item(PyList_Get_Item(list, i), 0);
+        item = PyTuple_Get_Item(PyList_Get_Item(list, i), 1);
```
da-woods, line 413: These will need some error checking I think

Cython/Utility/ImportExport.c
```diff
@@ -336,21 +407,32 @@ static int ${import_star}(PyObject* m) {
...
+#else
     for(i=0; i<PyList_GET_SIZE(list); i++) {
         name = PyTuple_GET_ITEM(PyList_GET_ITEM(list, i), 0);
         item = PyTuple_GET_ITEM(PyList_GET_ITEM(list, i), 1);
+#endif
+
         utf8_name = PyUnicode_AsUTF8String(name);
         if (!utf8_name) goto bad;
+#if CYTHON_COMPILING_IN_LIMITED_API
+        s = PyBytes_AsString(utf8_name);
```
da-woods, line 423: Also needs error checking

Cython/Utility/ModuleSetupCode.c
da-woods: > Ideally these kinds of macros could be centralized somewhere. Is this module the best place to put something like that?

Yes - ModuleSetupCode.c contains the bit of utility code that's always included, and is where the centralized stuff goes.

As I said in another comment - if we change it then I think think we should change it absolutely everywhere at once (with a big find/replace).

The naming scheme for Cython helper macros is `__Pyx_something` so they should stick to that. e.g. `__Pyx_Py_INCREF`

scoder: I think `__Pyx_INCREF()` is enough.

Cython/Compiler/ExprNodes.py
```diff
@@ -11441,11 +11441,19 @@ def generate_result_code(self, code):
         ))
         code.put_gotref(shapes_temp, py_object_type)
 
+        code.putln("#if CYTHON_COMPILING_IN_LIMITED_API")
+        code.putln('%s = __pyx_array_new(%s, %s, PyBytes_AsString(%s), "%s", (char *) %s); %s' % (
```
scoder, line 11445: Lacks error handling. I'd rather extract *only* the `char*` reading and keep everything else as before, and then only add error handling for this call in the Limited API case.

scoder (commented):
I agree that the "macro to function" changes should be undone. @da-woods described better ways to do this. I'm not even sure it needs to be done for incref/decref. Is that documented anywhere?

da-woods:
> I'd guess that if such a change comes in CPython, they couldn't afford to change only `Py_IncRef()` but not `Py_INCREF()`. I've never seen code use `Py_IncRef()`.

> I'm not even sure it needs to be done for incref/decref. Is that documented anywhere?

I think the relevant information is in https://peps.python.org/pep-0683/#stable-abi.

As I understand it: using `Py_INCREF/Py_DECREF` won't break anything, but if it's compiled against `Py_LIMITED_API` <3.12 then it will change the reference count of immortal objects (when run on Python versions >3.12). That's because the implementation of `Py_INCREF` and `Py_DECREF` is inlined, so it can't easily be changed while maintaining compatibility.

Changing the reference count of immortal objects means we lose the performance advantages of immortal objects.

Therefore:
* We don't have to make this change to make Cython+limited API work,
* We probably should make the change though, because it will make it work better.
* The change is only beneficial in the limited API - outside the limited API `Py_INCREF` is always the most up-to-date version.
* If we make the change, I think we have to make it everywhere at once (otherwise we risk being inconsistent)

My view is that we should do it in one PR that changes the whole of Cython (and has no other changes in it).

Cython/Compiler/ExprNodes.py
vyasr: I've updated this per what I think you were looking for, but please let me know if I've missed something.

Cython/Utility/ImportExport.c
vyasr: I think I've added this correctly now, assuming that all error checking is safe to jump to the `bad` label.

vyasr:
OK, I think I've addressed all your requests. This PR is _much_ slimmer now without the incref/decref changes, which I agree from my experiments do not actually seem to be required to compile under the limited API. Happy to see those changes made wholesale in a separate PR.

Cython/Utility/ImportExport.c
```diff
@@ -334,12 +335,23 @@ static int ${import_star}(PyObject* m) {
     if (__Pyx_import_all_from(locals, m) < 0) goto bad;
     list = PyDict_Items(locals);        if (!list) goto bad;
 
+#if CYTHON_COMPILING_IN_LIMITED_API
+    for(i=0; i<PyList_Size(list); i++) {
+        name = PyTuple_Get_Item(PyList_Get_Item(list, i), 0); if (!name) goto bad;
+        item = PyTuple_Get_Item(PyList_Get_Item(list, i), 1); if (!item) goto bad;
```
vyasr, line 341: Should we add `__Pyx_Py[Tuple|List]_GET_ITEM` to ModuleSetupCode.c? I imagine this won't be the last time we need this kind of conditional otherwise.

Cython/Utility/ImportExport.c
```diff
@@ -334,12 +335,23 @@ static int ${import_star}(PyObject* m) {
...
     for(i=0; i<PyList_GET_SIZE(list); i++) {
         name = PyTuple_GET_ITEM(PyList_GET_ITEM(list, i), 0);
         item = PyTuple_GET_ITEM(PyList_GET_ITEM(list, i), 1);
+#endif
+
         utf8_name = PyUnicode_AsUTF8String(name);
         if (!utf8_name) goto bad;
+#if CYTHON_COMPILING_IN_LIMITED_API
+        s = PyBytes_AsString(utf8_name); if (!s) goto bad;
+#else
```
vyasr, line 352: Same question as above, should we define a new macro in ModuleSetupCode.c?

da-woods: That's in TypeConversion.c (in a section that's also unconditionally always included) as `__Pyx_PyBytes_AsString`

Cython/Utility/ImportExport.c
da-woods: Yes - it looks like we already have `SET_ITEM` so I think that'd make sense.

Cython/Utility/ModuleSetupCode.c
vyasr: Going to resolve this since we've opened a separate issue to address the incref/decref changes.

vyasr:
To make things easier to test while also reducing churn, I've rebased the changes in this branch onto the new `limited-api-preview` branch (#6082) and retargeted the PR. Once we're happy with the state of things I'm happy to rebase again on master, or we can just merge into the preview branch, whichever you all prefer.

In the meantime, I uncovered one odd limited API bug in #6093 that I'm uncertain how to resolve.

Cython/Compiler/ExprNodes.py
```diff
@@ -10255,6 +10255,13 @@ def generate_codeobj(self, code):
...
         kwonly_argcount = func.num_kwonly_args
         nlocals = len(self.varnames)
+        # TODO: Not clear if something changed in Python make this typing change
+        # necessary, but I am seeing compiler warnings about narrowing. Looking in
+        # Include/cpython/code.h (in the cpython source), co_flags has always been a
+        # signed int but the flags are just #defined as bit flags, so the type (signed
+        # or unsigned) should be contextually determined. I wonder if this is a function
+        # of my compiler.
+        flags = [f"((unsigned int) {flag})" for flag in flags]
         flags = '|'.join(flags) or '0'
```
da-woods, line 10265: I think I can see where this is coming from (https://github.com/cython/cython/blob/4e0eee43210d6b7822859f3001202910888644af/Cython/Compiler/Code.py#L1920).

I suspect it does need a cast but it might be better later and just done once:

```suggestion
        flags = '(unsigned int)(%s)' % '|'.join(flags) or '0'
```

vyasr:
For now all the changes I've made here are "tested" in the sense that I have local code that leverages them, so I can at least be certain of their correctness. Do you want me to start making broader changes to globally replace things like `*_GET_ITEM` with the pyx versions? I won't be able to test in the same way, but I guess at this point that's still progress and we can start turning on tests when all of your other PRs are merged?

Cython/Compiler/ExprNodes.py
```diff
@@ -11513,6 +11533,7 @@ def dispose(temp):
 
         dispose(shapes_temp)
         dispose(format_temp)
+        dispose(format_ptr_temp)
```
da-woods, line 11536:
```suggestion
        code.funcstate.release_temp(format_ptr_temp)
```

The `format_ptr_temp` isn't a `PyObject*` so doesn't want the `code.put_decref_clear(temp, py_object_type)` part of `dispose`

Cython/Utility/ImportExport.c
```diff
@@ -329,17 +330,20 @@ static int ${import_star}(PyObject* m) {
...
-        item = PyTuple_GET_ITEM(PyList_GET_ITEM(list, i), 1);
+    for(i=0; i<PyList_Size(list); i++) {
+        import_obj = __Pyx_PyList_GET_ITEM(list, i); if (!import_obj) goto bad;
+        name = __Pyx_PyTuple_GET_ITEM(import_obj, 0); if (!name) goto bad;
+        item = __Pyx_PyTuple_GET_ITEM(import_obj, 1); if (!item) goto bad;
+
         utf8_name = PyUnicode_AsUTF8String(name);
         if (!utf8_name) goto bad;
-        s = PyBytes_AS_STRING(utf8_name);
+        s = __Pyx_PyBytes_AsWritableString(utf8_name); if (!s) goto bad;
```
da-woods, line 346:
```suggestion
        s = __Pyx_PyBytes_AsString(utf8_name); if (!s) goto bad;
```

I think you've used `AsWriteableString` to match the `char*` type.

Assuming https://github.com/cython/cython/pull/6101 doesn't break anything unexpected, I'll change it to `const char*` and we should just use `AsAtring` (since it isn't something that gets written to)

da-woods (commented):
> Do you want me to start making broader changes to globally replace things like *_GET_ITEM with the pyx versions? I won't be able to test in the same way, but I guess at this point that's still progress and we can start turning on tests when all of your other PRs are merged. It's up to you, I can keep this PR narrowly scoped to whatever makes my code compile or expand it to cover all similar uses that I find in some subset or all of Cython.

I'd rather keep this PR narrowly scoped I think. If we can get it working I'm basically happy with this.

I'd also rather not do completely untested replacements. There's definitely bits where they aren't needed because the code is in `ifdef`s already for example.

One way of doing (locally) tested replacements is to do:

```
CFLAGS="-DCYTHON_LIMITED_API=0x03070000 -DPy_LIMITED_API=0x03070000" python3 runtests.py some_specific_test
```

That will run an individual test file under the limited API. There's a bit of work needed to start turning them on on the CI, but that should be a way of looking at some locally testable failures. That's how I'd proceed anyway (I'd look at "run" tests and not bother with anything ".srctree")

Cython/Utility/CppConvert.pyx
```diff
@@ -76,7 +76,7 @@ cdef object {{cname}}(const vector[X]& v):
     for i in range(v_size_signed):
```
da-woods, line 76: This'll get broken again by https://github.com/cython/cython/pull/6077 fairly soon. Not your problem necessarily though (I can go fix it up after)

da-woods: Fix will be https://github.com/cython/cython/pull/6077

vyasr: Thanks!

Cython/Utility/ImportExport.c
vyasr: Yup that's right, I originally had AsString but changed it to match the current usage of it. Happy to change it if that works.

Cython/Compiler/ExprNodes.py
vyasr: Ah great thanks for the fix.

vyasr:
OK, at this point I think I've gotten almost all of the changes in that I want to make in this PR. I'm compiling a suite of projects and almost all of them build successfully now (along with corresponding changes in those repos). The main outstanding issue is that a couple of them make use of [array.pxd](https://github.com/cython/cython/blob/master/Cython/Includes/cpython/array.pxd), which I'm guessing will be a slightly heavier lift to fix up since it's not just going to be simple find-and-replace changes. I'm happy to work on it, but I'm guessing that'll be better done in a separate PR. WDYT @da-woods?

Assuming that you agreed, how should I target this PR? Do you want it going into the `limited-api-preview` branch, or should I rebase on master? I assume `limited-api-preview` is just an integration branch for testing and all the individual PRs used to make that branch will need to be merged separately anyway, so we probably don't want to merge this into that as well, right?

vyasr:
> I'd rather keep this PR narrowly scoped I think. If we can get it working I'm basically happy with this.
> 
> I'd also rather not do completely untested replacements. There's definitely bits where they aren't needed because the code is in `ifdef`s already for example.

That sounds good. I'm happy to do some of that in follow-ups to keep individual PRs more tightly scoped and easier to finish up and merge. I'm guessing getting reviews on your open PRs is a higher priority though since those are the blockers for even being able to start testing this more systematically, correct?

da-woods:
> The main outstanding issue is that a couple of them make use of [array.pxd](https://github.com/cython/cython/blob/master/Cython/Includes/cpython/array.pxd), which I'm guessing will be a slightly heavier lift to fix up since it's not just going to be simple find-and-replace changes. I'm happy to work on it, but I'm guessing that'll be better done in a separate PR. WDYT @da-woods?

I had a bit of a look at `array.pxd` when working on https://github.com/cython/cython/pull/5871. I'm not actually convinced it's solvable because the array data structure isn't even exposed in the Limited API. It's definitely one for a separate PR anyway.

> Assuming that you agreed, how should I target this PR? Do you want it going into the limited-api-preview branch, or should I rebase on master? I assume limited-api-preview is just an integration branch for testing and all the individual PRs used to make that branch will need to be merged separately anyway, so we probably don't want to merge this into that as well, right?

Yes - this PR should ultimately be targeted at `master`.

>  I'm happy to do some of that in follow-ups to keep individual PRs more tightly scoped and easier to finish up and merge. I'm guessing getting reviews on your open PRs is a higher priority though since those are the blockers for even being able to start testing this more systematically, correct?

I think so. My plan after my PRs was merged was to start turning on the `tests/run` tests in the Limited API. Which doesn't *absolutely* need my PRs, but I doubt many of the tests will work without them.

With that said, if you produce some individual PRs that are tightly scoped then I can review and merge them (while I'm not sure I should review my own work).

------------------------------------------------

I'll try to have a look into the crashing tests for this PR in the next few days. There's nothing I can see just by looking at the code unfortunately.

da-woods:
> OK, at this point I think I've gotten almost all of the changes in that I want to make in this PR. I'm compiling a suite of projects and almost all of them build successfully now (along with corresponding changes in those repos)

That's encouraging anyway.

Cython/Utility/CythonFunction.c
```diff
@@ -1440,7 +1440,7 @@ __pyx_FusedFunction_getitem(__pyx_FusedFunctionObject *self, PyObject *idx)
             Py_DECREF(item);
 #endif
             if (unlikely(!string)) goto __pyx_err;
-            PyList_SET_ITEM(list, i, string);
+	    if (__Pyx_PyList_SET_ITEM(list, i, string) < 0) goto __pyx_err;
```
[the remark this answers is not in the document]
vyasr: This one looks correct, I think.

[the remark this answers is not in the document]
da-woods: Sorry yes that one is.

vyasr:
> Spotted the issue - `Tuple/List`

Thanks, I was planning to look into these failures this afternoon!

> I had a bit of a look at array.pxd when working on https://github.com/cython/cython/pull/5871. I'm not actually convinced it's solvable

Neither am I, although I'm far less knowledgeable than you regarding the Python C API. I was certainly debating reimplementing those bits of our code using something else to get to a place where I can test out the limited API because I don't know if it'll really be possible to migrate that part painlessly.

> With that said, if you produce some individual PRs that are tightly scoped then I can review and merge them (while I'm not sure I should review my own work).

For sure. That said, once this PR goes in I'm happy to start looking at your PRs. My review won't count for much, but I am at least more familiar than I was a few weeks ago with these parts of Cython and can provide some basic sanity checking. I won't be a substitute for a proper review from a maintainer of course, but hopefully I can be helpful.

da-woods:
> I was certainly debating reimplementing those bits of our code using something else to get to a place where I can test out the limited API because I don't know if it'll really be possible to migrate that part painlessly.

You could also isolate them in a small Cython module compiled with the regular API. I don't know how well that fits with your long-term goal with using the Limited API but it would work for testing.

vyasr:
I have temporarily cherry-picked #6101 onto this branch so that I can see if tests will pass. If they do, I'll rebase these changes on master, remove that commit (since it'll be redundant) and open this PR up for review.

vyasr:
Tests are passing now :tada: I can rebase if you're also happy and think this is ready to be finalized (after whatever remaining review there is). If you think there's more to do here then I'll hold off rebasing until all the changes we want are on the branch so that I can continue easily testing locally.

da-woods:
No - seems like a good time to rebase I think

Cython/Utility/ImportExport.c
```diff
@@ -329,17 +330,20 @@ static int ${import_star}(PyObject* m) {
...
+    PyObject *import_obj;
 
     locals = PyDict_New();              if (!locals) goto bad;
     if (__Pyx_import_all_from(locals, m) < 0) goto bad;
     list = PyDict_Items(locals);        if (!list) goto bad;
 
-    for(i=0; i<PyList_GET_SIZE(list); i++) {
-        name = PyTuple_GET_ITEM(PyList_GET_ITEM(list, i), 0);
-        item = PyTuple_GET_ITEM(PyList_GET_ITEM(list, i), 1);
+    for(i=0; i<PyList_Size(list); i++) {
```
da-woods, line 339: `__Pyx_PyList_GET_SIZE` makes this selection at compile-time. In principle it can also fail so if probably wants a check `#if !CYTHON_ASSUME_SAFE_SIZE`.

It's a bit debatable how much we really have to have a check given the only real failure is "not a list" and `PyDict_Items` is specified as returning a list. I can't quite remember what we decided on unfortunately.

Cython/Utility/ImportExport.c
```diff
@@ -329,17 +330,20 @@ static int ${import_star}(PyObject* m) {
     PyObject *utf8_name = 0;
     PyObject *name;
     PyObject *item;
+    PyObject *import_obj;
```
da-woods, line 333:
```suggestion
    PyObject *import_obj;
    Py_ssize_t size;
```

Cython/Utility/ImportExport.c
da-woods, line 339:
```suggestion
    size = __Pyx_PyList_GET_SIZE(list);
    # !CYTHON_ASSUME_SAFE_SIZE
    if (size < 0) goto bad;
    #endif
    for(i=0; i<size; i++) {
```

da-woods (commented):
Just going to apply my size suggestion since I'm basically happy with this so will merge when that's checked

da-woods:
* cython.array
* C++ type conversion
* Fused functions
* Import dotted

vyasr:
I'm sorry, I got very caught up in some projects at work and completely lost track of this. Thanks for getting it over the finish line! I'll continue testing with the latest master merged into the limited API preview branch when I can to continue validating. I also plan to have a look at the rest of your open limited API PRs to see if I can provide any useful feedback there when I get a chance.

PR #6074, opened 2024-03-14
enabled the cache-like feature to cyton.py

Blanco10022:
i enabed and documented the cache like feature in the setup.py  since the option.py seem to already have set it to true

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,8 +26,13 @@ from Cython.Build import cythonize
 # List of Cython source files to compile
 source_files = ['your_file.pyx']
 
-# Set the cache option to True
-options = {'cache': True}
+# Check if the cache option is enabled via command-line argument or environment variable
+enable_cache = '--enable-cache' in sys.argv
+if not enable_cache:
+    enable_cache = os.environ.get('CYTHON_ENABLE_CACHE', '').lower() in ['true', '1']
+
+# Set the cache option accordingly
+options = {'cache': enable_cache}
 
 # Specify other setup parameters as needed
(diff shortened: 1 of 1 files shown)


scoder:
This is a weird place to make such a change. What exactly is the intention? What do you want to see covered by this option?

Blanco10022:
The intention behind enabling the cache option is to improve the efficiency of the build process when working with Cython modules. By enabling caching, Cython will reuse previously generated C files if the source files haven't changed since the last compilation. This can significantly reduce build times, especially in large projects or when iterating frequently on code changes.
Enabling the cache option doesn't change the behavior of the generated code or introduce any new features; it simply optimizes the build process by avoiding unnecessary recompilation of unchanged source files. This can be particularly beneficial in continuous integration workflows or when collaborating on projects with multiple developers, where build times can impact productivity.

Futhermore , I added the cache option in the cythonize() function call in the setup.py file. This function handles compiling Cython source files into C extensions during the build process. I chose this spot because it's where we configure Cython compilation settings, making it the logical place to enable caching for better build performance.

Let me know if you need more info or have other suggestions!

scoder:
> I added the cache option in the cythonize() function call in the setup.py file. This function handles compiling Cython source files into C extensions during the build process.

Yes, but only for building Cython itself. It makes no difference for building user code.

Blanco10022:
ok i  understand , i will look over it again and apply modifications

Blanco10022:
i jusst remodified

PR #6079, opened 2024-03-15
Fix warnings for docs during build procedure

stefanos82:
I read https://www.sphinx-doc.org/en/master/usage/domains/c.html and said to myself `don't tell me it's the whitespace mandatory...!`; seems like it is!

### DIFF
--- a/docs/src/tutorial/clibraries.rst
+++ b/docs/src/tutorial/clibraries.rst
@@ -36,14 +36,14 @@ Defining external declarations
 You can download CAlg `here <https://codeload.github.com/fragglet/c-algorithms/zip/master>`_.
 
 The C API of the queue implementation, which is defined in the header
-file :file:`c-algorithms/src/queue.h`, essentially looks like this:
+file :file: `c-algorithms/src/queue.h`, essentially looks like this:
 
 .. literalinclude:: ../../examples/tutorial/clibraries/c-algorithms/src/queue.h
     :language: C
     :caption: queue.h
 
 To get started, the first step is to redefine the C API in a ``.pxd``
-file, say, :file:`cqueue.pxd`:
+file, say, :file: `cqueue.pxd`:
 
 .. literalinclude:: ../../examples/tutorial/clibraries/cqueue.pxd
     :caption: cqueue.pxd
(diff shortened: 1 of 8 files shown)


stefanos82:
@da-woods based on our conversation in #6071 and the original report at #4217 I have fixed the warnings.

Can you double-check that everything is OK so it can get merged?

da-woods:
Thanks - I'll just have a quick look at the generated docs from the CI before merging it to double check.

Does this fix your problems with the code examples or just the warnings? (I'm just curious - I'm happy to merge something for the warnings)

stefanos82:
> Thanks - I'll just have a quick look at the generated docs from the CI before merging it to double check.
> 
> Does this fix your problems with the code examples or just the warnings? (I'm just curious - I'm happy to merge something for the warnings)

It fixes the `make html` build, but not the `make singlehtml`; the problem remains with the aforementioned build option.

Also, I have discovered a number of problems with other build options, such as `make latexpdf`; it produces the full `tutorial.pdf` correctly, but the `reference.pdf` generates only 11 pages for some reason.

Another problem is with `make texinfo`; it produces the necessary files in `build` directory, but when you decide to generate a PDF out of `texinfo` files, the generated PDF is messed up and the images do not respect the margin sizes, they go full sized and go out of margin limitations; also, the links go out of margin dimensions too.

da-woods:
This isn't looking right I'm afraid

image

image

image

image

stefanos82:
OK, let me double-check everything because this indeed looks weird!

stefanos82:
image

In other words, without my changes, it should have worked as expected!

I have no idea why it does not work.

I'm closing this PR as it's completely unnecessary and actually break things.

da-woods:
I cherry-picked the really easy stuff out of this in 7484c0102 (just the orphan files and the length of the underline).

We may as well have those changes. Thanks.

stefanos82:
By the way, https://readthedocs.org/ directs to https://about.readthedocs.com/?ref=readthedocs.org and has a pricing too!

image

Am I missing something here?

da-woods:
I think that's for commercial use - it's free (with ads) or low-cost (without ads) for open source. I don't control Cython's readthedocs account so I don't really know though

PR #6091, opened 2024-03-16
Add compilation cache to cython command

matusvalo:
Adds `--cache` argument to cython command. It is built on top of #6090 and should be merged after it.

Related to #6070


Cython/Build/Tests/TestCyCache.py
```diff
@@ -7,6 +7,7 @@
 import unittest
```
matusvalo, line 7: Not sure whether this TestCyCache module should not be placed somewhere else since it is not related to Build package only anymore...

Cython/Compiler/Main.py
```diff
@@ -653,8 +672,6 @@ def compile(source, options = None, full_module_name = None, **kwds):
     """
     options = CompilationOptions(defaults = options, **kwds)
     if isinstance(source, str):
-        if not options.timestamps:
-            return compile_single(source, options, full_module_name)
```
matusvalo, line 657: This change is causing test_grammar.py tests failing. Not sure what is best to do here. Should be compile_single extended to support cache (context creation needs to be added then)?

Cython/Build/Tests/TestCyCache.py
matusvalo: Or cache module will be moved to `Build` directory - see: https://github.com/cython/cython/pull/6090#discussion_r1527455984

matusvalo: Now when `Cache.py` is in `Build`directory this comment is not valid anymore I think...

matusvalo:
Rebased from master containing refactor of Cache code. Now PR is clean and ready for review.

Cython/Compiler/Main.py
```diff
@@ -623,9 +623,24 @@ def compile_multiple(sources, options):
     cwd = os.getcwd()
     for source in sources:
         if source not in processed:
+            output_filename = get_output_filename(source, cwd, options)
             if context is None:
                 context = Context.from_options(options)
-            output_filename = get_output_filename(source, cwd, options)
+            if options.cache:
```
scoder, line 629: Shouldn't the caching happen *after* the out of date check?

Cython/Compiler/Main.py
scoder: The cache is persistent, so it shouldn't depend on the number of modules that users compile whether it's being used or not.

scoder (commented):
I would guess that it could be cleaner if the caching (not necessarily the Cache object creation) was done just around the call to `run_pipeline()`. Maybe add a wrapper function `run_cached_pipeline`?

Cython/Compiler/Main.py
matusvalo: Reverted back to original logic

Cython/Compiler/Main.py
matusvalo: fixed.

Cython/Compiler/Main.py
```diff
@@ -623,9 +623,24 @@ def compile_multiple(sources, options):
...
+                from ..Build.Dependencies import create_dependency_tree
+                from ..Build.Cache import Cache
+                cache = Cache(options.cache)
+                dependencies = create_dependency_tree(context)
+                fingerprint = cache.transitive_fingerprint(
+                        source, dependencies, options, 'c++' if options.cplus else None, np_pythran=options.np_pythran)
```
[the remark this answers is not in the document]
matusvalo: fixed.

matusvalo:
> Maybe add a wrapper function run_cached_pipeline?

implemented.

Cython/Compiler/Main.py
```diff
@@ -472,32 +472,66 @@ def create_default_resultobj(compilation_source, options):
...
+        options.configure_language_defaults(source_ext[1:])  # py/pyx
+
+        source = setup_source_object(source, source_ext, full_module_name, options, context)
+        # Set up result object
+        return create_default_resultobj(source, options)
+    else:
+        result = run_pipeline(source, options, full_module_name, context)
+        if cache and fingerprint:
+            cache.store_to_cache(output_filename, fingerprint, result)
+    return result
```
scoder, line 520: Let's make it clear that we consider the "build the module" case the normal one:
```suggestion
        return create_default_resultobj(source, options)

    result = run_pipeline(source, options, full_module_name, context)
    if cache and fingerprint:
        cache.store_to_cache(output_filename, fingerprint, result)
    return result
```

Cython/Compiler/Main.py
```diff
@@ -472,32 +472,66 @@ def create_default_resultobj(compilation_source, options):
...
+    return CompilationSource(source_desc, full_module_name, cwd)
+
+def run_cached_pipeline(source, options, full_module_name=None, context=None, cache=None, fingerprint=None):
+
+    if context is None:
+        context = Context.from_options(options)
+
+    if cache:
+        if not fingerprint:
+            fingerprint = get_fingerprint(cache, source, options)
```
scoder, line 499: Let's make some arguments non-optional to reduce the overall redundancy and complexity.

I think for a function called `run_cached_pipeline`, we should assume that caching is enabled, so `cache` is not optional. We should also consider whether `fingerprint` really needs to be optional. If we need to pass it in in some cases, why not always pass it in?

I also don't see why `context` should be optional. It just adds more places where we create the context in exactly the same way. Just pass it into both pipeline functions, always.

And, once the context is always created externally, we can also use it to build the `full_module_name` and make that argument non-optional as well.

I would hope that this allows us to distribute the responsibilities more cleanly. For example, I wonder why the Cache isn't responsible to fingerprinting? Couldn't it somehow remember the result for a all source files, so that we don't have to do that externally?

scoder (commented):
Sorry, I had lost this from sight.

Cython/Build/Dependencies.py
```diff
@@ -1212,14 +1214,9 @@ def cythonize_one(pyx_file, c_file, fingerprint, cache, quiet, options=None,
...
-        if cached:
-            if not quiet:
-                print("%sFound compiled %s in cache" % (progress, pyx_file))
-            cache.load_from_cache(c_file, cached)
-            return
-    if not quiet:
+    if cache and fingerprint and cache.lookup_cache(c_file, fingerprint) and not quiet:
+        print("%sFound compiled %s in cache" % (progress, pyx_file))
+    elif not quiet:
         print("%sCythonizing %s" % (progress, Utils.decode_filename(pyx_file)))
```
scoder, line 1220: `quiet` is the main thing to test here, so let's move it first.
```suggestion
    if not quiet:
        if cache and fingerprint and cache.lookup_cache(c_file, fingerprint):
            print(f"{progress}Found compiled {pyx_file} in cache")
        else:
            print(f"{progress}Cythonizing {Utils.decode_filename(pyx_file)}")
```

Cython/Compiler/Main.py
```diff
@@ -599,23 +622,45 @@ def add(self, source, result):
...
     recursion.
     """
-    return run_pipeline(source, options, full_module_name)
 
+    if context is None:
+        context = Context.from_options(options)
 
-def compile_multiple(sources, options):
+    if cache:
+        fingerprint = fingerprint if fingerprint else get_fingerprint(cache, source, options)
```
scoder, line 652: A little too many fingerprints for my taste.
```suggestion
        fingerprint = fingerprint or get_fingerprint(cache, source, options)
```

Cython/Compiler/Main.py
```diff
@@ -660,12 +702,27 @@ def compile(source, options = None, full_module_name = None, **kwds):
     checking is requested, a CompilationResult is returned, otherwise a
     CompilationResultSet is returned.
     """
+    from ..Build.Cache import Cache
```
scoder, line 705: This should go under `if options.cache` below because we don't need it otherwise.

Cython/Compiler/Main.py
```diff
@@ -660,12 +702,27 @@ def compile(source, options = None, full_module_name = None, **kwds):
...
+    # * options.cache is True (the default path to the cache base dir is used)
+    # * options.cache is the explicit path to the cache base dir
+    # * annotations are not generated
+    if options.cache:
+        cache_path = None if options.cache is True else options.cache
+        cache = Cache(cache_path) if not (options.annotate or Options.annotate) else None
+    else:
+        if (options.annotate or Options.annotate) and options.verbose:
+            sys.stderr.write(f'Cache is ignored when annotations are enabled.\n')
+        cache = None
```
scoder, line 719: I don't think the comment matches the implementation. The cache is not enabled automatically when annotations are not generated, but the comment seems to suggest that.

I think enabling the cache should require `options.cache` to be true (or a path), right?

```suggestion
    # cache is enabled when:
    # * options.cache is True (the default path to the cache base dir is used)
    # * options.cache is the explicit path to the cache base dir
    # unless annotations are generated
    cache = None
    if options.cache:
        if options.annotate or Options.annotate:
            if options.verbose:
                sys.stderr.write('Cache is ignored when annotations are enabled.\n')
        else:
            cache_path = None if options.cache is True else options.cache
            cache = Cache(cache_path)
```

scoder (commented):
Fine to merge if the changes I propose are ok with you.

Cython/Compiler/Main.py
scoder:
```suggestion
```

PR #6178, opened 2024-05-03
Avoid thread unsafe borrowed references under free-threading

lysnikolaou:


Cython/Compiler/ExprNodes.py
```diff
@@ -3119,19 +3119,46 @@ def generate_next_sequence_item(self, test_name, result_name, code):
...
+                    self.counter_cname,
+                    result_name,
+                    self.counter_cname,
+                    inc_dec,
+                    # use the error label to avoid C compiler warnings if we only use it below
+                    code.error_goto_if_neg('0', self.pos)
+                    ))
+            code.putln("#elif CYTHON_ASSUME_SAFE_MACROS && !CYTHON_AVOID_BORROWED_REFS")
+            code.putln(
+                "%s = Py%s_GetItemRef(%s, %s); %s%s; %s" % (
```
da-woods, line 3138: I think this'll need a `__Pyx_GOTREF(%s)` (in place of `__Pyx_INCREF` in the others) to pass our tests.

That's a no-op in a regular build, but when running with our "refnanny" tool it helps track the new reference.

Cython/Compiler/ExprNodes.py
```diff
@@ -8409,16 +8436,36 @@ def generate_special_parallel_unpacking_code(self, code, rhs, use_loop):
         if len(sequence_types) == 2:
             code.putln("if (likely(Py%s_CheckExact(sequence))) {" % sequence_types[0])
         for i, item in enumerate(self.unpacked_items):
-            code.putln("%s = Py%s_GET_ITEM(sequence, %d); " % (
-                item.result(), sequence_types[0], i))
+            if sequence_types[0] == "List":
+                code.putln("#if !CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS")
```
da-woods, line 8440: We usually try to put special cases first just because it's easier if we need to add other special cases.

da-woods (commented):
Linking https://github.com/cython/cython/issues/6164 because it's essentially this. I think I'm fine with the new macro rather than using two levels of the existing one though.

When the details of this are decided, can you add it to the documentation (https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#c-macro-defines) since we've now tried to document these macros.

I wonder if a bunch of duplication could be avoided by defining in the module setup code:

```
#if CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS && PY_VERSION_HEX >= 0x030D0000  // adjust version hex here!
#define __Pyx_PyList_GetItemRef(o, i) Py_GetItemRef(o, i)
#else
#endif
```

your current version has a lot of duplicated code which could be compacted quite a bit.

Cython/Utility/ObjectHandling.c
```diff
@@ -1421,10 +1458,18 @@ static CYTHON_INLINE PyObject *__Pyx__GetModuleGlobalName(PyObject *name)
         return result;
     }
 #else
+#if !CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS
     result = PyDict_GetItem($moddict_cname, name);
+#else
+    PyDict_GetItemRef($moddict_cname, name, &result);
```
scoder, line 1464: This function is new in Py3.13a1 and needs a version guard.
See https://github.com/python/cpython/commit/41ca16455188db806bfc7037058e8ecff2755e6c

Cython/Utility/ObjectHandling.c
```diff
@@ -443,7 +442,29 @@ static CYTHON_INLINE PyObject *__Pyx_GetItemInt_{{type}}_Fast(PyObject *o, Py_ss
...
+    Py_ssize_t wrapped_i = i;
+    if (wraparound & unlikely(i < 0)) {
+        wrapped_i += PyList_GET_SIZE(o);
+    }
+    if ((!boundscheck) || likely(__Pyx_is_valid_index(wrapped_i, PyList_GET_SIZE(o)))) {
+#if !CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS
+        PyObject *r = PyList_GET_ITEM(o, wrapped_i);
+        Py_INCREF(r);
+#else
+        PyObject *r = PyList_GetItemRef(o, wrapped_i);
```
scoder, line 459: This function is new in Py3.13a4 and needs a version guard.
See https://github.com/python/cpython/commit/d0f1307580a69372611d27b04bbf2551dc85a1ef

scoder: That said, rather than repeating the "use GET_ITEM+INCREF or GetItemRef" decision in various places, it seems worth having an inline function that does the equivalent of "GET_ITEM+INCREF" efficiently in different environments.

scoder (changes requested):

lysnikolaou:
> I wonder if a bunch of duplication could be avoided by defining in the module setup code:

That's a good idea! Thanks @da-woods for the pointer. It also takes care of the version guard @scoder mentioned in his review comments.

scoder:
Note that __Pyx_NewRef is an unsafe macro that uses its argument twice. Thus my preference for an inline function.

That said, maybe __Pyx_NewRef should actually be an inline function these days.

lysnikolaou:
I've pushed a commit with the changes suggested. My feeling is a more serious cleanup of all of this is needed, which I can do in a follow-up PR.

Cython/Utility/ModuleSetupCode.c
```diff
@@ -2190,17 +2214,25 @@ static PyObject* __Pyx_PyCode_New(
     }
 
     #if CYTHON_COMPILING_IN_LIMITED_API
+    #if CYTHON_AVOID_UNSAFE_BORROWED_REFERENCES && PY_VERSION_HEX > 0x030d00b1
+    PyDict_GetItemRef(tuple_dedup_map, varnames_tuple, &varnames_tuple_dedup);
```
da-woods, line 2218: These aren't quite the same from an exception handling point-of-view.

Cython/Utility/ModuleSetupCode.c
```diff
@@ -2190,17 +2214,25 @@ static PyObject* __Pyx_PyCode_New(
     }
 
     #if CYTHON_COMPILING_IN_LIMITED_API
+    #if CYTHON_AVOID_UNSAFE_BORROWED_REFERENCES && PY_VERSION_HEX > 0x030d00b1
```
da-woods, line 2217:
```suggestion
    #if CYTHON_AVOID_UNSAFE_BORROWED_REFERENCES && __PYX_LIMITED_VERSION_HEX > 0x030d00b1
```

We should base it on the version of the Limited API it's compiled with rather than the version of Python it's compiled with.

Cython/Utility/ModuleSetupCode.c
lysnikolaou: That's true. `PyDict_GetItemRef` raises an exception on error. What would you prefer we do in this case? One option would be:

```cpp
#if CYTHON_AVOID_UNSAFE_BORROWED_REFERENCES && __PYX_LIMITED_VERSION_HEX > 0x030d00b1
int res = PyDict_GetItemRef(tuple_dedup_map, varnames_tuple, &varnames_tuple_dedup);
if (res == -1) goto done;
if (res == 0) {
#else
varnames_tuple_dedup = PyDict_GetItem(tuple_dedup_map, varnames_tuple);
if (!varnames_tuple_dedup) {
#endif
```

Cython/Utility/CythonFunction.c
```diff
@@ -1216,15 +1216,17 @@ static int __Pyx_CyFunction_InitClassCell(PyObject *cyfunctions, PyObject *class
 
     for (i = 0; i < count; i++) {
         __pyx_CyFunctionObject *m = (__pyx_CyFunctionObject *)
-#if CYTHON_ASSUME_SAFE_MACROS && !CYTHON_AVOID_BORROWED_REFS
+#if CYTHON_ASSUME_SAFE_MACROS && !CYTHON_AVOID_BORROWED_REFS && !CYTHON_AVOID_UNSAFE_BORROWED_REFERENCES
```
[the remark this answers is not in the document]
lysnikolaou: Would you prefer I revert this change or should we keep it with a comment?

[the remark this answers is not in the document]
da-woods: It's probably fine to leave I think.

Cython/Utility/TypeConversion.c
```diff
@@ -88,7 +88,28 @@ static CYTHON_INLINE PyObject* __Pyx_PyUnicode_FromString(const char*);
 #define __Pyx_PyUnicode_FromOrdinal(o)       PyUnicode_FromOrdinal((int)o)
 #define __Pyx_PyUnicode_AsUnicode            PyUnicode_AsUnicode
 
-#define __Pyx_NewRef(obj) (Py_INCREF(obj), obj)
+#if CYTHON_COMPILING_IN_CPYTHON && PY_VERSION_HEX >= 0x030a00b1
```
da-woods, line 91: Don't change this but we only worry about getting the exact version hex for pre-releases versions of older Python versions - we aren't supporting 3.10 betas at this stage so I'd have used `0x030A0000` and called it good enough (and saved a bit of time looking up exactly when it was introduced).

lysnikolaou: Totally agree. I'll change it since this was probably a copy paste from somewhere else and I didn't look too closely to see what exact prerelease introduced this.

Cython/Utility/ObjectHandling.c
```diff
@@ -1421,10 +1438,10 @@ static CYTHON_INLINE PyObject *__Pyx__GetModuleGlobalName(PyObject *name)
         return result;
     }
 #else
-    result = PyDict_GetItem($moddict_cname, name);
+    __Pyx_PyDict_GetItemRef(result, $moddict_cname, name);
```
da-woods, line 1441:
```suggestion
    if (unlikely(__Pyx_PyDict_GetItemRef(result, $moddict_cname, name) == -1)) PyErr_Clear();
```

It looks like this function is just ignoring errors so I'll keep doing that here.

da-woods (commented):
A few comments but I think this mostly looks reasonable to me.

I'm going to leave this and wait for @scoder to look at it I think

Cython/Utility/ObjectHandling.c
```diff
@@ -1405,10 +1422,10 @@ static CYTHON_INLINE PyObject *__Pyx__GetModuleGlobalName(PyObject *name)
 #if !CYTHON_AVOID_BORROWED_REFS
 #if CYTHON_COMPILING_IN_CPYTHON && PY_VERSION_HEX < 0x030d0000
     // Identifier names are always interned and have a pre-calculated hash value.
-    result = _PyDict_GetItem_KnownHash($moddict_cname, name, ((PyASCIIObject *) name)->hash);
+    __Pyx__PyDict_GetItemRef_KnownHash(result, $moddict_cname, name, ((PyASCIIObject *) name)->hash);
```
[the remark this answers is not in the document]
lysnikolaou: `_PyDict_GetItem_KnownHast` was added back in 3.13.0a2 in https://github.com/python/cpython/pull/112115, so we can remove the version guard if we want. In that case, we'll probably have to keep this in?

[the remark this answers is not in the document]
da-woods: Yes keeping it and removing the version-guard makes sense to me.

Cython/Utility/ObjectHandling.c
```diff
@@ -424,17 +424,35 @@ static PyObject *__Pyx_GetItemInt_Generic(PyObject *o, PyObject* j) {
...
+    return __Pyx_GetItemInt_Generic(o, PyInt_FromSsize_t(i));
+#else
+    return PySequence_GetItem(o, i);
+#endif
+}
+
+static CYTHON_INLINE PyObject *__Pyx_GetItemInt_List_Fast(PyObject *o, Py_ssize_t i,
+                                                              CYTHON_NCP_UNUSED int wraparound,
+                                                              CYTHON_NCP_UNUSED int boundscheck) {
+#if CYTHON_ASSUME_SAFE_MACROS && CYTHON_ASSUME_SAFE_SIZE && !CYTHON_AVOID_BORROWED_REFS && !CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS
```
ngoldbaum, line 449: Could you avoid the duplication by keeping the templating that was here before by conditionally inserting `CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS` based on the value of `type`? I'm no tempita expert though so take that worth a grain of salt.

docs/src/userguide/source_files_and_compilation.rst
```diff
@@ -1266,7 +1266,13 @@ hidden by default since most users will be uninterested in changing them.
             a reference to objects it manipulates.  Most useful for
             non-reference-counted implementations of Python, like PyPy
             (where it is enabled by default).
-            
+
+        ``CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS``
+            Avoid using APIs that return "borrowed references" and instead use
+            the equivalent APIs that return "strong references". Most useful for
+            the free-threaded build of CPython, where incrementing the reference
+            count of borrowed references might introduce thread safety issues.
```
ngoldbaum, line 1274:
```suggestion
            Avoid using APIs that return unsafe "borrowed references" and instead use
            the equivalent APIs that return "strong references". Most useful for
            the free-threaded build of CPython, where incrementing the reference
            count of borrowed references to items in mutable containers might 
            introduce thread safety issues. Borrowed refernces to items in immutable
            containers are still allowed with this setting.
```

Cython/Utility/ObjectHandling.c
lysnikolaou: Hmm, I also don't know enough about tempita to know whether this is possible. I'll look into it.

docs/src/userguide/source_files_and_compilation.rst
lysnikolaou:
```suggestion
            Avoid using APIs that return unsafe "borrowed references" and instead use
            the equivalent APIs that return "strong references". Most useful for
            the free-threaded build of CPython, where incrementing the reference
            count of borrowed references to items in mutable containers might 
            introduce thread safety issues. Borrowed references to items in immutable
            containers are still allowed with this setting.
```

lysnikolaou: There was a typo in the last sentence, that's why I did a new suggestion. Thanks for the improvement!

Cython/Utility/ModuleSetupCode.c
```diff
@@ -1042,6 +1056,23 @@ static CYTHON_INLINE PyObject * __Pyx_PyDict_GetItemStrWithError(PyObject *dict,
   #define __Pyx_SET_SIZE(obj, size) Py_SIZE(obj) = (size)
 #endif
 
+#if CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS && PY_VERSION_HEX >= 0x030d00b1
+#define __Pyx_PyList_GetItemRef(o, i) PyList_GetItemRef(o, i)
+static CYTHON_INLINE int __Pyx_PyDict_GetItemRef(PyObject *dict, PyObject *key, PyObject **result) {
+  return PyDict_GetItemRef(dict, key, result);
+}
```
scoder, line 1063: Looks like these could both be macros.
```suggestion
#define __Pyx_PyList_GetItemRef(o, i)  PyList_GetItemRef(o, i)
#define __Pyx_PyDict_GetItemRef(dict, key, result)  PyDict_GetItemRef(dict, key, result)
```

Cython/Utility/ModuleSetupCode.c
```diff
@@ -1042,6 +1056,23 @@ static CYTHON_INLINE PyObject * __Pyx_PyDict_GetItemStrWithError(PyObject *dict,
...
+#else
+#define __Pyx_PyList_GetItemRef(o, i) __Pyx_NewRef(PyList_GET_ITEM(o, i))
```
scoder, line 1065: Since we're not doing any error checking here, this should depend on `CYTHON_ASSUME_SAFE_MACROS` and actually also `!CYTHON_AVOID_BORROWED_REFS` since we're preferring a borrowed reference over a strong returned reference here.

Cython/Utility/ModuleSetupCode.c
```diff
@@ -2190,17 +2221,25 @@ static PyObject* __Pyx_PyCode_New(
...
+    PyDict_SetDefaultRef(tuple_dedup_map, varnames_tuple, varnames_tuple, &varnames_tuple_dedup);
+    #else
     varnames_tuple_dedup = PyDict_SetDefault(tuple_dedup_map, varnames_tuple, varnames_tuple);
+    #endif
     if (unlikely(!varnames_tuple_dedup)) goto done;
     #endif
 
-    #if CYTHON_AVOID_BORROWED_REFS
+    #if CYTHON_AVOID_BORROWED_REFS && !CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS
     Py_INCREF(varnames_tuple_dedup);
```
scoder, line 2243: This is not correct since we've applied a version check above. In older Pythons, the reference will not be increfed.

I'd move the error check + incref into the `#else` block above (and duplicate the error check in the `#if` block).

scoder (commented):
Haven't finished my review yet, but here are a few comments.

Cython/Utility/ModuleSetupCode.c
```diff
@@ -308,6 +315,13 @@
   #ifndef CYTHON_AVOID_BORROWED_REFS
     #define CYTHON_AVOID_BORROWED_REFS 0
   #endif
+  // CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS - Avoid borrowed references that are not thread-safe in the free-threaded build of CPython.
+  #if CYTHON_COMPILING_IN_CPYTHON_NOGIL
```
da-woods, line 319:
```suggestion
  #if CYTHON_COMPILING_IN_CPYTHON_FREETHREADING
```

We changed the macro just because `nogil` was looking ambiguous with Cython's `nogil` feature in other places

lysnikolaou:
Is this good to go? Any more feedback before it's okay to be merged?

Also, on another note, could we maybe do an "intermediate" release since Cython currently does not build with the free-threaded build and people seem to need it? If not, is there a plan for an upcoming release?

lysnikolaou:
Another ping on this. Please let me know in case pinging is not okay and I'll stop.

Cython/Compiler/ExprNodes.py
```diff
@@ -8374,16 +8387,26 @@ def generate_special_parallel_unpacking_code(self, code, rhs, use_loop):
...
-                code.putln("%s = Py%s_GET_ITEM(sequence, %d); " % (
-                    item.result(), sequence_types[1], i))
+                if sequence_types[1] == "List":
+                    code.putln("%s = __Pyx_PyList_GetItemRef(sequence, %d); " % (
+                        item.result(), i))
+                    code.put_xgotref(item.result(), item.ctype())
+                else:  # Tuple
+                    code.putln("%s = PyTuple_GET_ITEM(sequence, %d); " % (
+                        item.result(), i))
+                    code.put_incref(item.result(), item.ctype())
```
scoder, line 8408: I think the space in the end of the (original) template string probably intended (or was left over from) allowing to merge the lines, so it's either `code.put("… ")` with trailing space or `code.putln("…")` without it. I don't mind either way but would probably lean towards using separate lines since that keeps the difference between gotref and incref visible. I.e. remove the trailing space here and above.

Cython/Compiler/ExprNodes.py
```diff
@@ -3119,19 +3119,32 @@ def generate_next_sequence_item(self, test_name, result_name, code):
...
+                "%s = PyTuple_GET_ITEM(%s, %s); __Pyx_INCREF(%s); %s%s; %s" % (
+                    result_name,
+                    self.py_result(),
+                    self.counter_cname,
+                    result_name,
+                    self.counter_cname,
+                    inc_dec,
+                    # use the error label to avoid C compiler warnings if we only use it below
+                    code.error_goto_if_neg('0', self.pos)
+                    ))
```
scoder, line 3147: It would be easier to spot the differences between both cases if you limited the conditional code to just those, without repeating the entire code generation. I'd decide between the two macro names and then use them to do the code generation outside of the if-else blocks.

Cython/Utility/ModuleSetupCode.c
scoder: This looks unresolved.

Cython/Utility/ModuleSetupCode.c
```diff
@@ -2196,19 +2225,33 @@ static PyObject* __Pyx_PyCode_New(
...
     if (!varnames_tuple_dedup) {
         if (unlikely(PyDict_SetItem(tuple_dedup_map, varnames_tuple, varnames_tuple) < 0)) goto done;
         varnames_tuple_dedup = varnames_tuple;
     }
-    #else
+
+    #else // !CYTHON_COMPILING_IN_LIMITED_API
+
+    #if CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS && PY_VERSION_HEX >= 0x030d00b1
+    if(unlikely(PyDict_SetDefaultRef(tuple_dedup_map, varnames_tuple, varnames_tuple, &varnames_tuple_dedup) < 0)) goto done;
```
scoder, line 2245:
```suggestion
    if (unlikely(PyDict_SetDefaultRef(tuple_dedup_map, varnames_tuple, varnames_tuple, &varnames_tuple_dedup) < 0)) goto done;
```

Cython/Utility/ObjectHandling.c
scoder:
```suggestion
#if CYTHON_ASSUME_SAFE_MACROS && CYTHON_ASSUME_SAFE_SIZE && !CYTHON_AVOID_BORROWED_REFS {{if type == 'List' }}&& !CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS{{endif}}
```

Cython/Utility/TypeConversion.c
```diff
@@ -83,7 +83,28 @@ static CYTHON_INLINE PyObject* __Pyx_PyUnicode_FromString(const char*);
...
+static CYTHON_INLINE PyObject *__Pyx_XNewRef(PyObject *obj) {
+    return Py_XNewRef(obj);
+}
+#else
+static CYTHON_INLINE PyObject *__Pyx_XNewRef(PyObject *obj) {
+    Py_XINCREF(obj);
+    return obj;
+}
+#endif
+
```
scoder, line 107: It's always better to limit conditional code differences to the a minimum to make them easy to compare and read. Also, this looks like a case where we could allow other Python implementations to hit the newer code path.
```suggestion
static CYTHON_INLINE PyObject *__Pyx_NewRef(PyObject *obj) {
  #if CYTHON_COMPILING_IN_CPYTHON && PY_VERSION_HEX >= 0x030a0000  ||  defined(Py_NewRef)
    return Py_NewRef(obj);
  #else
    Py_INCREF(obj);
    return obj;
  #endif
}

static CYTHON_INLINE PyObject *__Pyx_XNewRef(PyObject *obj) {
  #if CYTHON_COMPILING_IN_CPYTHON && PY_VERSION_HEX >= 0x030a0000  ||  defined(Py_XNewRef)
    return Py_XNewRef(obj);
  #else
    Py_XINCREF(obj);
    return obj;
  #endif
}

```

Cython/Utility/ObjectHandling.c
```diff
@@ -528,10 +543,14 @@ static CYTHON_INLINE int __Pyx_SetItemInt_Fast(PyObject *o, Py_ssize_t i, PyObje
...
         if ((!boundscheck) || likely(__Pyx_is_valid_index(n, PyList_GET_SIZE(o)))) {
+#if !CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS
             PyObject* old = PyList_GET_ITEM(o, n);
-            Py_INCREF(v);
             PyList_SET_ITEM(o, n, v);
             Py_DECREF(old);
+#else
+            PyList_SetItem(o, n, v);
+#endif
+            Py_INCREF(v);
```
scoder, line 553: Why did you move the incref to the end here? If `old == v`, I think we'd risk losing the reference.

Cython/Utility/ModuleSetupCode.c
```diff
@@ -2230,7 +2273,7 @@ static PyObject* __Pyx_PyCode_New(
     );
 
 done:
-    #if CYTHON_AVOID_BORROWED_REFS
+    #if CYTHON_AVOID_BORROWED_REFS || CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS
     Py_XDECREF(varnames_tuple_dedup);
```
scoder, line 2277: This seems incorrect, given the version specific condition above. If `CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS == 1` but `PY_[LIMITED_]VERSION_HEX < 3.13b1`, we'd kill a reference here.

scoder (commented):
Sorry for the long silence. More comments below.

Cython/Utility/ObjectHandling.c
```diff
@@ -438,7 +438,7 @@ static CYTHON_INLINE PyObject *__Pyx_GetItemInt_{{type}}_Fast(PyObject *o, Py_ss
         Py_INCREF(r);
         return r;
     }
-    return __Pyx_GetItemInt_Generic(o, PyLong_FromSsize_t(i));
+    return __Pyx_GetItemInt_Generic(o, PyInt_FromSsize_t(i));
```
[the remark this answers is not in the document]
lysnikolaou: Yeah, I'm not sure how that got in there.

Cython/Utility/ObjectHandling.c
```diff
@@ -1427,7 +1429,7 @@ static CYTHON_INLINE PyObject *__Pyx__GetModuleGlobalName(PyObject *name)
     PyObject *result;
 // FIXME: clean up the macro guard order here: limited API first, then borrowed refs, then cpython
 #if !CYTHON_AVOID_BORROWED_REFS
-#if CYTHON_COMPILING_IN_CPYTHON && PY_VERSION_HEX < 0x030d0000
+#if CYTHON_COMPILING_IN_CPYTHON && !CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS
     // Identifier names are always interned and have a pre-calculated hash value.
     result = _PyDict_GetItem_KnownHash($moddict_cname, name, ((PyASCIIObject *) name)->hash);
```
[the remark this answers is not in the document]
lysnikolaou: Do we still want it even though `_KnownHash` has been restored since 3.13.0a2?

Cython/Utility/ModuleSetupCode.c
```diff
@@ -2196,20 +2225,42 @@ static PyObject* __Pyx_PyCode_New(
...
     #if CYTHON_AVOID_BORROWED_REFS
     Py_INCREF(varnames_tuple_dedup);
+    #endif // CYTHON_AVOID_BORROWED_REFS
+    #endif // CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS && PY_VERSION_HEX >= 0x030d00b1
+
+    #if CYTHON_AVOID_BORROWED_REFS || (CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS && PY_VERSION_HEX >= 0x030d00b1)
+    Py_XDECREF(varnames_tuple_dedup);
     #endif
 
+    #endif // CYTHON_COMPILING_IN_LIMITED_API
```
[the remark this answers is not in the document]
lysnikolaou: Agreed. If it can't be used in threads and since the dict is thread-local the complications do not make sense. I'll revert all the changes.

Cython/Utility/ModuleSetupCode.c
lysnikolaou: Understood. Thanks for the context!

lysnikolaou:
Thanks for your patience here @scoder! I've addressed the latest round of feedback.

Cython/Utility/ObjectHandling.c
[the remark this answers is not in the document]
scoder: Hmm, right, I had forgotten that it was restored. The change is fine then.

Cython/Compiler/ExprNodes.py
```diff
@@ -3119,19 +3119,18 @@ def generate_next_sequence_item(self, test_name, result_name, code):
...
-                result_name,
-                self.counter_cname,
-                inc_dec,
-                # use the error label to avoid C compiler warnings if we only use it below
-                code.error_goto_if_neg('0', self.pos)
-                ))
+            f"{result_name} = {getitem}({self.py_result()}, {self.counter_cname}); "
+            f"{incref}({result_name}); "
+            f"{self.counter_cname}{inc_dec}; "
+            f"{code.error_goto_if_neg('0', self.pos)}")
```
scoder, line 3133: Let's keep the comment.
```suggestion
            # Use the error label to avoid C compiler warnings if we only use it below.
            f"{code.error_goto_if_neg('0', self.pos)}")
```

Cython/Utility/ModuleSetupCode.c
```diff
@@ -1037,6 +1051,38 @@ static CYTHON_INLINE PyObject * __Pyx_PyDict_GetItemStrWithError(PyObject *dict,
...
 
+#if CYTHON_AVOID_THREAD_UNSAFE_BORROWED_REFS && PY_VERSION_HEX >= 0x030d00b1
+#define __Pyx_PyList_GetItemRef(o, i) PyList_GetItemRef(o, i)
+#define __Pyx_PyDict_GetItemRef(dict, key, result) PyDict_GetItemRef(dict, key, result)
+#elif !CYTHON_AVOID_BORROWED_REFS
+
+#if CYTHON_ASSUME_SAFE_MACROS
+#define __Pyx_PyList_GetItemRef(o, i) __Pyx_NewRef(PyList_GET_ITEM(o, i))
+#else
+#define __Pyx_PyList_GetItemRef(o, i) __Pyx_NewRef(PyList_GetItem(o, i))
```
scoder, line 1062: `__Pyx_NewRef()` doesn't do any error checking, so if `PyList_GetItem()` fails (for whatever reason, e.g. because it's PyPy and accidentally needs to do memory allocation that fails), then this would probably crash.
```suggestion
#define __Pyx_PyList_GetItemRef(o, i) PySequence_GetItem(o, i)
```

lysnikolaou:
Everything seems to have been addressed here.

Cython/Compiler/ExprNodes.py
```diff
@@ -8374,16 +8374,22 @@ def generate_special_parallel_unpacking_code(self, code, rhs, use_loop):
         if len(sequence_types) == 2:
             code.putln("if (likely(Py%s_CheckExact(sequence))) {" % sequence_types[0])
         for i, item in enumerate(self.unpacked_items):
-            code.putln("%s = Py%s_GET_ITEM(sequence, %d); " % (
-                item.result(), sequence_types[0], i))
+            if sequence_types[0] == "List":
+                code.putln(f"{item.result()} = __Pyx_PyList_GetItemRef(sequence, {i});")
```
scoder, line 8378: We need to do error handling here, in case `__Pyx_PyList_GetItemRef()` returns NULL.

Cython/Compiler/ExprNodes.py
```diff
@@ -8374,16 +8374,22 @@ def generate_special_parallel_unpacking_code(self, code, rhs, use_loop):
...
+            else:  # Tuple
+                code.putln(f"{item.result()} = PyTuple_GET_ITEM(sequence, {i});")
+                code.put_incref(item.result(), item.ctype())
         if len(sequence_types) == 2:
             code.putln("} else {")
             for i, item in enumerate(self.unpacked_items):
-                code.putln("%s = Py%s_GET_ITEM(sequence, %d); " % (
-                    item.result(), sequence_types[1], i))
+                if sequence_types[1] == "List":
+                    code.putln(f"{item.result()} = __Pyx_PyList_GetItemRef(sequence, {i});")
```
scoder, line 8387: We need to do error handling here, in case `__Pyx_PyList_GetItemRef()` returns NULL.

Cython/Compiler/ExprNodes.py
```diff
@@ -3119,19 +3119,19 @@ def generate_next_sequence_item(self, test_name, result_name, code):
...
-                self.counter_cname,
-                inc_dec,
-                # use the error label to avoid C compiler warnings if we only use it below
-                code.error_goto_if_neg('0', self.pos)
-                ))
+            f"{result_name} = {getitem}({self.py_result()}, {self.counter_cname}); "
+            f"{incref}({result_name}); "
+            f"{self.counter_cname}{inc_dec}; "
+            # Use the error label to avoid C compiler warnings if we only use it below.
+            f"{code.error_goto_if_neg('0', self.pos)}")
```
scoder, line 3134: This error handling code is actually needed now. We should probably use `error_goto_if_null` in the case of a list.

scoder (commented):
I had missed the fact that we're now using functions that can fail, so we need to handle errors.

scoder (approved):
Looks good now. Thanks!

lysnikolaou:
Thanks a lot for all the reviews and the help @scoder!

scoder:
Closes https://github.com/cython/cython/issues/6164

da-woods:
I'm going to revise this a little for 2 reasons:
* I want to add a directive to turn it off so user's can use faster but thread-unsafe code,
* wraparound isn't thread safe (because we can't trust that the size hasn't changed)

da-woods:
> wraparound isn't thread safe (because we can't trust that the size hasn't changed)

This may not be a huge issue actually. CPython doesn't look to add locks to ensure this. So I think it'll be caught by boundschecking

rgommers:
Agreed with the last comment, not an issue. Size changes only if the user explicitly resizes or mutates in some other way (or insert/pop for lists)- and that's the kind of thing that should be managed by the user. It's inherently unsafe to modify a data structure in one thread while accessing it from another thread.

PR #6209, opened 2024-05-21
[3.0.X] Fix crash when function contains fused extension type

matusvalo:
backport of #6204


da-woods (approved):
Think this change is fine. I wouldn't backport any of the cleanup around `[1 for func...]` but I don't think you were planning to do that anyway.

matusvalo:
python 3.5 tests seems to be failing but they are already failing in 3.0.X branch so I am merging PR.

PR #6224, opened 2024-05-31
fix: `cythonize -i src/` command in python3.12

trim21:
`distutils` has been removed in python 3.12 so it will failed.

close https://github.com/cython/cython/issues/5751


trim21:
sorry, didn't notice that I should also install setuptools.

PR #6321, opened 2024-08-03
wheel CI: Update cibuildwheel to v2.20.0

EwoutH:
This commit updates [cibuildwheel](https://github.com/pypa/cibuildwheel) to the latest version ([v2.20.0](https://github.com/pypa/cibuildwheel/releases/tag/v2.20.0)) in the GitHub Actions `wheel.yml` CI workflow.

cibuildwheel 2.20.0 uses the ABI stable Python 3.13.0rc1 and build Python 3.13 wheels by default.

I think it would be really beneficial to automate these updates in some fashion, since they need to happen quite often and can be easily forgotten.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -65,7 +65,7 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install cibuildwheel
         # Nb. keep cibuildwheel version pin consistent with job below
-        run: pipx install cibuildwheel==2.19.2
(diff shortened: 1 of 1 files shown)


scoder:
Thanks

EwoutH:
Thanks for merging! What would be a quick and feasible way to get new wheels on PyPI? Would a backport to 3.0.x be possible, or is the 3.1.0 release close?

PR #6329, opened 2024-08-06
Set module initialization trace context to the correct file

### LINKED ISSUE #2882: Tracing/coverage of include files reports wrong file during module init
As noted in https://github.com/cython/cython/issues/1461#issuecomment-468894250, module init code in include files reports line traces for the file that contains the main module function, not for the include file. This is because the frame and code object for the module init function is only created and configured once at the beginning of the function, and points to the module file.

### LINKED ISSUE #6295: [BUG] Coverage for the signature line of inline cdef/cpdef function in pxd file got attributed to the wrong file
### Describe the bug

For an inline function defined in a pxd file, the coverage of the initial line, which is generated during the initialization of the module, attributes it to the pyx file that uses the pxd file, rather than to the pxd file. (See example below).

`mod.pxd`
```cython
# cython: language_level=3

cpdef inline function_pxd():
    pass
```

`mod.pyx`
```cython
# cython: language_level=3
























cpdef function_pyx():
    pass
```
 The empty lines here are to make sure there isn't any overlapping valid line numbers between the pyx and the pxd file.

Compiled using `cython -X linetrace=True mod.pyx; gcc -ggdb3 -O0 -DCYTHON_TRACE_NOGIL=1 -DCYTHON_TRACE=1 mod.c -shared -fPIC -o mod.so -I/usr/include/python3.12` and generated a coverage report with the following script, (with cython plugin loaded)

```python
import mod

mod.function_pyx()
mod.function_pxd()
```

The coverage sqlite db file shows a lines bit of `\x0a\x00\x00\x04` for the pyx file (which decodes to 1, 3, 26) and empty for the pxd file. Note that line 3 is completely empty for the pyx and it is actually the cpdef line for the pxd function.

The reason behind this is that in the generate C code, there's

```c
  __Pyx_TraceCall("__Pyx_PyMODINIT_FUNC PyInit_mod(void)", __pyx_f[1], 1, 0, __PYX_ERR(1, 1, __pyx_L1_error));

  /* "mod.pxd":3
 * # cython: language_level=3
 * 
 * cpdef inline function_pxd():             # <<<<<<<<<<<<<<
 *     pass
 */
  __Pyx_TraceLine(3,0,__PYX_ERR(0, 3, __pyx_L1_error))

  __pyx_t_2 = __Pyx_CyFunction_New(&__pyx_mdef_3mod_1function_pxd, 0, __pyx_n_s_function_pxd, NULL, __pyx_n_s_mod, __pyx_d, ((PyObject *)__pyx_codeobj_)); if (unlikely(!__pyx_t_2)) __PYX_ERR(0, 3, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_2);
  if (PyDict_SetItem(__pyx_d, __pyx_n_s_function_pxd, __pyx_t_2) < 0) __PYX_ERR(0, 3, __pyx_L1_error)

  /* "mod.pyx":26
 * 
 * 
 * cpdef function_pyx():             # <<<<<<<<<<<<<<
 *     pass
 */
  __Pyx_TraceLine(26,0,__PYX_ERR(1, 26, __pyx_L1_error))
```

where the file name table is defined as,
```c
static const char *__pyx_f[] = {
  "mod.pxd",
  "mod.pyx",
};
```

We can see that the call context was set up (by `__Pyx_TraceCall`) to be part of the pyx file (i.e. `__pyx_f[1]`) and the pxd file line number is used in this context without any change of file name context. Maybe a "dummy" init function should be generated for each of the pxd files that's included/used?

### Expected behaviour

Line 3 should be reported as being covered in the pxd file and not pyx file.

### OS

Linux

### Python version

3.12.3

### Cython version

3.0.10

### Additional context

This effect also show up for pxd files cimported into another pxd/pyx file.

yuyichao:
This is almost certainly not the best/cleanest way to achieve the goal but this works for my test cases.

Fix #2882
Fix #6295


scoder:
Thanks. As you noted, this isn't ready for merging as is, but it does seem reasonable to split the tracing in the module init function into file scopes.

Note that there is a major rewrite of the tracing code on the way in
https://github.com/cython/cython/pull/6144/
This uses the new `sys.monitoring` infrastructure. Whatever we do to fix this issue would have to fit in there.

The `sys.monitoring` API bases everything on code objects (which have no byte code in Cython but do contain the code meta data). There might be ways to juggle with multiple COs in the module init function.

Issue #6186, opened 2024-05-06
Coverage measurement of Cython code in a meson build


oscarbenjamin:
### Describe your issue

There is a related `spin` issue: https://github.com/scientific-python/spin/issues/182.

I have created a repo to demonstrate how a `spin/meson` build of a Cython project looks and what is needed to get coverage of Cython code working. The README explains the problems.

There are two related things that prevent Cython's coverage plugin from working in a meson-based project:

The paths recorded in the generated C files in a meson build look like
```c
/* #### Code section: filename_table ### */

static const char *__pyx_f[] = {
  "thing.pyx",
  "<stringsource>",
};
```
When doing an inplace build with setuptools they instead look like:
```c
/* #### Code section: filename_table ### */

static const char *__pyx_f[] = {
  "src/stuff/thing.pyx",
  "<stringsource>",
};
```
I'm not sure if this is just something that should be changed in `spin/meson` when doing a coverage-enabled build or something. Possibly this is just the way that `cython` is invoked. In any case Cython's coverage plugin breaks if the path to the source file is not included here.

The other problem is that Cython's coverage plugin assumes that all files are in-tree like:
```console
$ tree src
src
└── stuff
    ├── __init__.py
    ├── meson.build
    ├── __pycache__
    │   ├── __init__.cpython-311.pyc
    │   └── test_thing.cpython-311-pytest-8.2.0.pyc
    ├── test_thing.py
    ├── thing.c
    ├── thing.cpython-311-x86_64-linux-gnu.so
    └── thing.pyx

2 directories, 8 files
```
This is not how meson ever lays things out. Instead in a `spin/meson` build:

1. The source tree is left pristine (no `.c` or `.so` files).
2. Intermediate built files (e.g. `.c`) are in the `build` directory.
3. Final output files (`.so` and `.py`) are assembled in the `build-install` directory.

It is the `build-install` directory that is actually used for running tests and generates the coverage data.

When running tests with a coverage-enabled Cython build Cython's coverage plugin fails to find the files that are needed and then coverage reports warnings like:
```
.../site-packages/coverage/report_core.py:115: CoverageWarning: Couldn't parse '/home/oscar/current/active/cython_coverage_demo/thing.pyx': No source for code: '/home/oscar/current/active/cython_coverage_demo/thing.pyx'. (couldnt-parse)
  coverage._warn(msg, slug="couldnt-parse")
```

I can get coverage working if I patch Cython like this:
```diff
[...]

PR #6341, opened 2024-08-15
Output import-relative paths in generated C code.

oscarbenjamin:
When cython is run on a file that is not in the current working directory it outputs filepaths that are either absolute or are basenames. It is not good to output absolute paths in the generated C code and outputing only basenames messes up coverage measurement.

Partial fix for gh-6186. In combination with https://github.com/flintlib/python-flint/pull/188 this makes it possible to measure coverage of Cython for for python-flint which uses meson as its build system.

The two places changed here were inconsistent. In ModuleNodes.py the code detects an absolute path and then outputs `basename(path)` instead. The path in question will usually not be absolute but it will be absolute if the source file being compiled is not in CWD and so in that case it is the basename that is emitted to the generated C code. The case in ExprNodes.py outputs the path without checking but again that will be an absolute path if the source file is not iin CWD.

Here in both cases we instead output the "description" which is the path relative to import root. If the project has a `src` directory then the path output for `src/foo/bar.pyx` (module `foo.bar`) would be `foo/bar.pyx` i.e. it is the path relative to the `src` directory or more generally the path of the source file relative to `sys.path`. In both cases we only do this if we would otherwise output an absolute path which in turn is a case that arises from compiling a `.pyx` file in another directory sort of like:
```
cd build && cython ../src/foo/bar.pyx
```
This change is needed to make coverage measurement work in a meson build. In https://github.com/flintlib/python-flint/pull/188 I have almost entirely rewritten Cython's coverage plugin: 
https://github.com/cython/cython/blob/master/Cython/Coverage.py.
This PR for Cython and that one for python-flint together make it so that I can measure coverage in python-flint's meson build. (This is the last outstanding item for us to drop the setuptools build.)

The only significant thing I have kept in the rewritten plugin is the function for parsing the .c files. The rest of the code is just for discovering the source files and implementing the coverage plugin interface. A lot of that is meson-specific because it is a lot easier to implement those parts if you make use of the information provided by the build system (much of the complexity in Cython's plugin comes from just not knowing where exactly the .c and .pyx etc files are).

I would like to upstream that plugin somehow but I'm not sure that Cython itself is the right place for it because a lot of it is meson specific. I also think that changing anything about Cython's coverage plugin looks quite hairy because there are a lot of heuristics in there to make different (setuptools-based) situations just about work and I expect that any change will break someone.

For now though, this PR which fixes the paths that Cython stores in the .c files is the minimum that is needed to get coverage working for a meson build. This is needed because there's only so much a plugin can do if the paths in the trace events are not right. At least under the conditions that I am running Cython the change in this PR ensures that all the paths provided from coverage to the plugin are exactly of a predictable form that the plugin can handle without heuristics.

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -10213,7 +10213,7 @@ class CodeObjectNode(ExprNode):
         if os.path.isabs(file_path):
             file_path = func.pos[0].get_description()
         # Always use / as separator
-        file_path = str(pathlib.Path(file_path).as_posix())
+        file_path = pathlib.Path(file_path).as_posix()
         file_path = StringEncoding.bytes_literal(file_path.encode('utf-8'), 'utf8')
         file_path_result = code.get_py_string_const(file_path, identifier=False, is_str=True)
 
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -10210,7 +10210,7 @@ class CodeObjectNode(ExprNode):
         # FIXME: better way to get the module file path at module init time? Encoding to use?
         file_path = func.pos[0].get_filenametable_entry()
         if os.path.isabs(file_path):
-            file_path = func.pos[0].get_description()
+            file_path = func.pos[0].get_description().replace('\\', '/')
         file_path = StringEncoding.bytes_literal(file_path.encode('utf-8'), 'utf8')
         file_path_result = code.get_py_string_const(file_path, identifier=False, is_str=True)
 
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -10210,7 +10210,9 @@ class CodeObjectNode(ExprNode):
         # FIXME: better way to get the module file path at module init time? Encoding to use?
         file_path = func.pos[0].get_filenametable_entry()
         if os.path.isabs(file_path):
-            file_path = func.pos[0].get_description().replace('\\', '/')
+            file_path = func.pos[0].get_description()
+        # Always use / as separator
+        file_path = file_path.replace('\\', '/')
         file_path = StringEncoding.bytes_literal(file_path.encode('utf-8'), 'utf8')
         file_path_result = code.get_py_string_const(file_path, identifier=False, is_str=True)
 
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -19,6 +19,7 @@ import re
 import sys
 import copy
 import os.path
+import pathlib
 import operator
 
 from .Errors import (
@@ -10212,7 +10213,7 @@ class CodeObjectNode(ExprNode):
         if os.path.isabs(file_path):
             file_path = func.pos[0].get_description()
         # Always use / as separator
-        file_path = file_path.replace('\\', '/')
+        file_path = str(pathlib.Path(file_path).as_posix())
         file_path = StringEncoding.bytes_literal(file_path.encode('utf-8'), 'utf8')
         file_path_result = code.get_py_string_const(file_path, identifier=False, is_str=True)
 
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -10208,7 +10208,10 @@ class CodeObjectNode(ExprNode):
         func_name_result = code.get_py_string_const(
             func.name, identifier=True, is_str=False, unicode_value=func.name)
         # FIXME: better way to get the module file path at module init time? Encoding to use?
-        file_path = StringEncoding.bytes_literal(func.pos[0].get_filenametable_entry().encode('utf8'), 'utf8')
+        file_path = func.pos[0].get_filenametable_entry()
+        if os.path.isabs(file_path):
+            file_path = func.pos[0].get_description()
+        file_path = StringEncoding.bytes_literal(file_path.encode('utf-8'), 'utf8')
         file_path_result = code.get_py_string_const(file_path, identifier=False, is_str=True)
 
         # '(CO_OPTIMIZED | CO_NEWLOCALS)' makes CPython create a new dict for "frame.f_locals".
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -12,6 +12,7 @@ from collections import defaultdict
 import json
 import operator
 import os
+import pathlib
 import re
 import sys
 
@@ -967,7 +968,7 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
                     # never include absolute paths
                     file_path = source_desc.get_description()
                 # Always use / as separator
-                file_path = file_path.replace('\\', '/')
+                file_path = str(pathlib.Path(file_path).as_posix())
                 escaped_filename = as_encoded_filename(file_path)
                 code.putln('%s,' % escaped_filename.as_c_string_literal())
         else:
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -964,7 +964,7 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
             for source_desc in code.globalstate.filename_list:
                 file_path = source_desc.get_filenametable_entry()
                 if isabs(file_path):
-                    file_path = basename(file_path)  # never include absolute paths
+                    file_path = source_desc.get_description()  # never include absolute paths
                 escaped_filename = file_path.replace("\\", "\\\\").replace('"', r'\"')
                 escaped_filename = as_encoded_filename(escaped_filename)
                 code.putln('%s,' % escaped_filename.as_c_string_literal())
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -968,7 +968,7 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
                     # never include absolute paths
                     file_path = source_desc.get_description()
                 # Always use / as separator
-                file_path = str(pathlib.Path(file_path).as_posix())
+                file_path = pathlib.Path(file_path).as_posix()
                 escaped_filename = as_encoded_filename(file_path)
                 code.putln('%s,' % escaped_filename.as_c_string_literal())
         else:
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -965,7 +965,9 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
                 file_path = source_desc.get_filenametable_entry()
                 if isabs(file_path):
                     # never include absolute paths
-                    file_path = source_desc.get_description().replace('\\', '/')
+                    file_path = source_desc.get_description()
+                # Always use / as separator
+                file_path = file_path.replace('\\', '/')
                 escaped_filename = as_encoded_filename(file_path)
                 code.putln('%s,' % escaped_filename.as_c_string_literal())
         else:
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -964,9 +964,9 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
             for source_desc in code.globalstate.filename_list:
                 file_path = source_desc.get_filenametable_entry()
                 if isabs(file_path):
-                    file_path = source_desc.get_description()  # never include absolute paths
-                escaped_filename = file_path.replace("\\", "\\\\").replace('"', r'\"')
-                escaped_filename = as_encoded_filename(escaped_filename)
+                    # never include absolute paths
+                    file_path = source_desc.get_description().replace('\\', '/')
+                escaped_filename = as_encoded_filename(file_path)
                 code.putln('%s,' % escaped_filename.as_c_string_literal())
         else:
             # Some C compilers don't like an empty array
--- a/tests/build/run_cython_not_in_cwd.srctree
+++ b/tests/build/run_cython_not_in_cwd.srctree
@@ -1,18 +1,19 @@
 
 PYTHON check_paths_from_cython.py
 
-######## src/foo/__init__.py ########
+######## src/pkg/__init__.py ########
 
-# foo is a package and foo.bar is the module to build
+# __init__.py is needed because cython uses this to determine the
+# root of the package structure.
 
-######## src/foo/bar.pyx ########
+######## src/pkg/mod.pyx ########
 
 def func(x):
     return 2*x
 
-######## build/src/foo/placeholder.txt ########
+######## build/src/pkg/empty_build_output_directory.txt ########
 
-# Build outputs go in the build directory and cython is run from here.
+# Build outputs go in the "build/" directory and cython is run from there.
 
 ######## check_paths_from_cython.py ########
 
@@ -22,35 +23,24 @@ import subprocess
 python = sys.executable
 
 def cython(cmdstr, cwd=None):
-    argv = [sys.executable, '-m', 'cython'] + cmdstr.split()
+    argv = [python, '-m', 'cython'] + cmdstr.split()
     subprocess.run(argv, check=True, cwd=cwd)
 
-def grep(fname, pattern):
+def grep_paths(fname, pattern):
     with open(fname) as fin:
-        return [line for line in fin if pattern in line]
+        for line in fin:
+            if pattern in line:
+                [part] = [p for p in line.split() if pattern in p]
+                yield part.rstrip(',\n;:01234566789').strip('"')
 
-cython('-3 src/foo/bar.pyx -o build/src/foo/bar1.pyx.c')
-cython('-3 ../src/foo/bar.pyx -o src/foo/bar2.pyx.c', cwd='build')
+cython('-3 src/pkg/mod.pyx -o build/src/pkg/mod1.pyx.c')
+cython('-3 ../src/pkg/mod.pyx -o src/pkg/mod2.pyx.c', cwd='build')
 
-res1 = grep('build/src/foo/bar1.pyx.c', 'bar.pyx')
-res2 = grep('build/src/foo/bar2.pyx.c', 'bar.pyx')
+res1 = grep_paths('build/src/pkg/mod1.pyx.c', 'mod.pyx')
+res2 = grep_paths('build/src/pkg/mod2.pyx.c', 'mod.pyx')
 
-assert res1 == [
-    # Running cython on a file in CWD uses some paths relative to CWD.
-    '  "src/foo/bar.pyx",\n',
-    'static const char __pyx_k_src_foo_bar_pyx[] = "src/foo/bar.pyx";\n',
-    '/* "foo/bar.pyx":2\n',
-    '  /* "foo/bar.pyx":3\n',
-    '  /* "foo/bar.pyx":2\n',
-    '  /* "foo/bar.pyx":2\n'
-], '\n' + ''.join(res1)
+# Running cython on a file in CWD uses some paths relative to CWD.
+assert set(res1) == {"src/pkg/mod.pyx", "pkg/mod.pyx"}, str(res1)
 
-assert res2 == [
-    # When not in CWD all paths are relative to sys.path.
-    '  "foo/bar.pyx",\n',
-    'static const char __pyx_k_foo_bar_pyx[] = "foo/bar.pyx";\n',
-    '/* "foo/bar.pyx":2\n',
-    '  /* "foo/bar.pyx":3\n',
-    '  /* "foo/bar.pyx":2\n',
-    '  /* "foo/bar.pyx":2\n',
-], '\n' + ''.join(res2)
+# When not in CWD all paths are relative to sys.path.
+assert set(res2) == {"pkg/mod.pyx"}, str(res2)
--- a/tests/build/run_cython_not_in_cwd.srctree
+++ b/tests/build/run_cython_not_in_cwd.srctree
@@ -36,11 +36,11 @@ def grep_paths(fname, pattern):
 cython('-3 src/pkg/mod.pyx -o build/src/pkg/mod1.pyx.c')
 cython('-3 ../src/pkg/mod.pyx -o src/pkg/mod2.pyx.c', cwd='build')
 
-res1 = grep_paths('build/src/pkg/mod1.pyx.c', 'mod.pyx')
-res2 = grep_paths('build/src/pkg/mod2.pyx.c', 'mod.pyx')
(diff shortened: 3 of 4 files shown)


jakirkham:
Is there a way to test this? Just thinking how we can avoid accidental regression

oscarbenjamin:
> Is there a way to test this?

Which aspect would you want to test?

It is clear from the diff that this only affects situations in which the previous output would have been an absolute path (not good) or a basename (also not good). In other situations the output is unchanged.

jakirkham:
Well this references fixes two coverage use cases (possibly more). Is there a way to test something akin to those use cases (without needing to pull in those dependencies per se)?

scoder:
You can write one or more `.srctree` (sort-of multi-file in one file) tests and include 1) a Python script that parses and validates the location comments from the generated C files, and 2) a coverage test that operates in the (build) directory situation that this tries to solve. There are several examples in the tests/run/ directory, including coverage tests.

oscarbenjamin:
I'll look into making a test. The test suite failed immediately when I ran it under Python 3.12:
```
$ python runtests.py -vv
  File "/home/oscar/current/active/cython/runtests.py", line 78, in <module>
    from distutils.command.build_ext import build_ext as _build_ext
ModuleNotFoundError: No module named 'distutils'
```

oscarbenjamin:
> a coverage test that operates in the (build) directory situation that this tries to solve.

Note that this PR does not make Cython's coverage plugin work under the conditions described. It makes it possible to write a plugin that works but it does not make the necessary changes to Cython's plugin.

The associated python-flint PR (https://github.com/flintlib/python-flint/pull/188) uses a rewrite of Cython's plugin that does work in combination with the change to Cython that is in this PR. I think that it would be good do get the needed changes into Cython's coverage plugin but it involves including some meson-specific code.

I don't think that it would be good to extend the heuristics in Cython's current plugin to the extent that they would be able to handle the meson build setup without having any meson-specific code. Rather what is needed is some way for the user to configure the plugin to say "I'm using meson" so that the plugin can then use meson to find all of the .c and .pyx files rather then searching heuristically.

tests/build/run_cython_not_in_cwd.srctree
```diff
@@ -0,0 +1,56 @@
...
+
+assert res2 == [
+    # When not in CWD all paths are relative to sys.path.
+    '  "foo/bar.pyx",\n',
+    'static const char __pyx_k_foo_bar_pyx[] = "foo/bar.pyx";\n',
+    '/* "foo/bar.pyx":2\n',
+    '  /* "foo/bar.pyx":3\n',
+    '  /* "foo/bar.pyx":2\n',
+    '  /* "foo/bar.pyx":2\n',
+], str(res2)
```
oscarbenjamin, line 56: The first case here (`res1`) is the typical setuptools case. This PR is intended not to change what happens in that case when cython is run from within cwd. Cython's coverage plugin already works in that case.

The second case here is what happens when building with meson or cmake. The behaviour on master would be to output a mixture of basename `bar.pyx` and absolute `/home/.../bar.pyx` paths. The change here is that all paths are `foo/bar.pyx` which is relative to sys.path in the sense that the built module would be imported as `foo.bar` rather than `src.foo.bar`.

The test case shows that actually the first case (`res1`) is a bit inconsistent. I actually think that it would be better if the second case `res2` was always the output generated by Cython regardless of CWD and that in fact Cython should not use CWD for anything besides parsing relative paths at the CLI. Changing that first case `res1` to be like `res2` would likely break some things including possibly Cython's current coverage plugin when using setuptools.

Using import-relative paths (`foo/bar.pyx`) is the most robust choice given that at cythonisation time there is no way to know where the built module will end up (e.g. site-packages, a different machine etc) or what the CWD will be when the module is actually imported. All we can really say is that the module will end up being on `sys.path` under a directory `foo` so it makes sense to use `foo/bar.pyx` as the canonical path that is baked into the source and used for trace events. The coverage plugin is then responsible for knowing that in the source tree `foo/bar.pyx` is a path relative to `src` so that from project root the relative path would be `src/foo/bar.pyx`.

This PR only changes the second case `res2` which on master is broken at least as far measuring coverage goes. Hopefully then that is only an improvement without breaking something.

oscarbenjamin:
From the CI run in my fork I have seen that the test failed on Windows because of `\` vs `/`. 

That leads to another question that I can't easily answer right now though:

Do the paths generated on Windows use `\` at all?

Note that at least some paths are already unixy paths e.g. the CI failure looks like:
```
AssertionError: [
    '  "src\\\\\\\\foo\\\\\\\\bar.pyx",\n',
    'static const char __pyx_k_src_foo_bar_pyx[] = "src\\\\foo\\\\bar.pyx";\n',
    '/* "foo/bar.pyx":2\n',
    '/* "foo/bar.pyx":2\n',
    '  /* "foo/bar.pyx":3\n',
    '  /* "foo/bar.pyx":2\n',
    '  /* "foo/bar.pyx":2\n'
]
```
I'm not sure if this PR caused a change here to the first two by using `get_description()`. I've changed it to use `get_description().replace('\\', '/')` so that unixy paths are always output for all paths but that might have already been the case before.

We do need to double the slashes for a string in the C code and Python has doubled them in the log. The first string in the test failure output still has too many slashes though. That is a bug where the string gets escaped twice. I'll push the fix here:
```diff
diff --git a/Cython/Compiler/ModuleNode.py b/Cython/Compiler/ModuleNode.py
index 2695e2559..c9732bb6d 100644
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -965,8 +965,7 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
                 file_path = source_desc.get_filenametable_entry()
                 if isabs(file_path):
                     file_path = source_desc.get_description()  # never include absolute paths
-                escaped_filename = file_path.replace("\\", "\\\\").replace('"', r'\"')
-                escaped_filename = as_encoded_filename(escaped_filename)
+                escaped_filename = as_encoded_filename(file_path)
                 code.putln('%s,' % escaped_filename.as_c_string_literal())
         else:
             # Some C compilers don't like an empty array
```

oscarbenjamin:
Although I don't have direct access to a Windows machine I did try pushing the test case added here up to my fork to test against the master branch. The Windows job fails with these lines having been generated in the C code:
```
  "src\\\\foo\\\\bar.pyx",
static const char __pyx_k_src_foo_bar_pyx[] = "src\\foo\\bar.pyx";
/* "foo/bar.pyx":2
  /* "foo/bar.pyx":3
  /* "foo/bar.pyx":2
  /* "foo/bar.pyx":2
```
So you can see that the first path has double-escaped slashes on master. This also shows that the behaviour on master is to output `\` in some of the paths on Windows but `/` in others.

With the changes as they stand in this PR the double-escaping problem is fixed but although I have tried to use `/` in all paths somehow on Windows it still uses `\` for those two paths. So the output on Windows would be like:
```
  "src\\foo\\bar.pyx",
static const char __pyx_k_src_foo_bar_pyx[] = "src\\foo\\bar.pyx";
/* "foo/bar.pyx":2
  /* "foo/bar.pyx":3
  /* "foo/bar.pyx":2
  /* "foo/bar.pyx":2
```
I'm not sure if there is something somewhere that post-processes the strings to force the use of `\` on Windows.

What should be the expected result here?

I would have thought that Cython should be expected to produce C code deterministically in a platform-independent way because e.g. you might run cython on Linux but then compile the C code on Windows.

If that is the case then it would seem that the desired result would likely be to output unix paths everywhere including on Windows. This is what I have tried to do here but the test added here seems to fail in Windows CI because the generated paths are still using `\`. I don't have easy access to Windows to debug it.

oscarbenjamin:
Okay, so after the last commit it is all fine with all CI passing including on Windows. Someone would need to approve the workflow to see that here though.

There are three changes here:

1. Don't double escape the path.
2. Always use `/` in paths even on Windows.
3. If the source file is not in CWD then all paths are import-relative (`foo/bar.pyx`) rather than basename or abs-path.

The only thing I am unsure about is the change to use `/` paths on Windows. I am not sure what the generally expected behaviour should be but I could easily revert that and just adjust the test to allow `\\` in the paths.

Cython/Compiler/ExprNodes.py
```diff
@@ -10208,7 +10208,12 @@ def generate_codeobj(self, code):
         func_name_result = code.get_py_string_const(
             func.name, identifier=True, is_str=False, unicode_value=func.name)
         # FIXME: better way to get the module file path at module init time? Encoding to use?
-        file_path = StringEncoding.bytes_literal(func.pos[0].get_filenametable_entry().encode('utf8'), 'utf8')
+        file_path = func.pos[0].get_filenametable_entry()
+        if os.path.isabs(file_path):
+            file_path = func.pos[0].get_description()
+        # Always use / as separator
+        file_path = file_path.replace('\\', '/')
```
scoder, line 10215: The [`.as_posix()`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix) method in `pathlib` seems a better choice here because this should only be done on Windows. A Unix path can contain `\` as character:
```bash
$ touch abcdefg\\xyz
$ ls abcdefg\\xyz
'abcdefg\xyz'
```
OTOH, Windows directory names can actually contain `/`, and `.as_posix()` doesn't try to escape that either.

It seems slightly fragile to "normalise" path names like this, but I guess users would try to avoid them – at least after running into them once.

Cython/Compiler/ModuleNode.py
```diff
@@ -964,9 +964,11 @@ def generate_filename_table(self, code):
...
                 if isabs(file_path):
-                    file_path = basename(file_path)  # never include absolute paths
-                escaped_filename = file_path.replace("\\", "\\\\").replace('"', r'\"')
-                escaped_filename = as_encoded_filename(escaped_filename)
+                    # never include absolute paths
+                    file_path = source_desc.get_description()
+                # Always use / as separator
+                file_path = file_path.replace('\\', '/')
+                escaped_filename = as_encoded_filename(file_path)
                 code.putln('%s,' % escaped_filename.as_c_string_literal())
```
scoder, line 972: I'm not sure whether we should only make paths relative on output or rather earlier on input. We do put some effort into making paths absolute and comparable, though, so maybe absolute paths are still better for finding and processing files.

If we only do this on output, then we should move the "make relative" logic into `source_desc.get_filenametable_entry()` (see `Scanning.py`). It should consistently return the same file path usable for putting it in the filename table (which is what we do here), without further adaptation (except for C output escaping).

tests/build/run_cython_not_in_cwd.srctree
```diff
@@ -0,0 +1,56 @@
...
+######## check_paths_from_cython.py ########
+
+import sys
+import subprocess
+
+python = sys.executable
+
+def cython(cmdstr, cwd=None):
+    argv = [sys.executable, '-m', 'cython'] + cmdstr.split()
+    subprocess.run(argv, check=True, cwd=cwd)
```
scoder, line 26: You can also use something like `CYTHON src/pkg/mod.pyx` (or the same with `CYTHONIZE`) as command at the top to do the build for you before you run the (then simpler) script.

tests/build/run_cython_not_in_cwd.srctree
```diff
@@ -0,0 +1,56 @@
...
+######## src/foo/__init__.py ########
+
+# foo is a package and foo.bar is the module to build
+
+######## src/foo/bar.pyx ########
+
+def func(x):
+    return 2*x
+
+######## build/src/foo/placeholder.txt ########
```
scoder, line 13:
```suggestion
######## build/src/pkg/empty_build_output_directory.txt ########
```

tests/build/run_cython_not_in_cwd.srctree
```diff
@@ -0,0 +1,56 @@
+
+PYTHON check_paths_from_cython.py
+
+######## src/foo/__init__.py ########
+
+# foo is a package and foo.bar is the module to build
```
scoder, line 6: This becomes a useless comment if you write e.g. `pkg` instead of `foo` to give the test setup itself some semantics. `foo` and `bar` are semantically the same, but here, the package name and the module name are different things. Why not name them after what they are?

tests/build/run_cython_not_in_cwd.srctree
```diff
@@ -0,0 +1,56 @@
...
+
+######## src/foo/bar.pyx ########
+
+def func(x):
+    return 2*x
+
+######## build/src/foo/placeholder.txt ########
+
+# Build outputs go in the build directory and cython is run from here.
```
scoder, line 15:
```suggestion
# Build outputs go in the "build/" directory and cython is run from here.
```

tests/build/run_cython_not_in_cwd.srctree
```diff
@@ -0,0 +1,56 @@
...
+
+assert res1 == [
+    # Running cython on a file in CWD uses some paths relative to CWD.
+    '  "src/foo/bar.pyx",\n',
+    'static const char __pyx_k_src_foo_bar_pyx[] = "src/foo/bar.pyx";\n',
+    '/* "foo/bar.pyx":2\n',
+    '  /* "foo/bar.pyx":3\n',
+    '  /* "foo/bar.pyx":2\n',
+    '  /* "foo/bar.pyx":2\n'
+], '\n' + ''.join(res1)
```
scoder, line 46: Putting a complete list of occurrences into a test may let the test break if we make non-trivial changes to the code generation elsewhere. I'd rather extract the paths, deduplicate them, and make sure that there are no other path variants in there.

You could use a regex like `r"pkg[/\\]mod[.]pyx"` to make sure it hits any path occurrence, Windows or Posix.

tests/build/run_cython_not_in_cwd.srctree
oscarbenjamin: How would I set the working directory in that case?

The purpose of the test here is to run the same command from two different working directories.

I realise that cython has a `--working` CLI option but I didn't want to write the test with the assumption that setting that option would be equivalent to actually running cython from a different working directory.

tests/build/run_cython_not_in_cwd.srctree
oscarbenjamin: If you run this on test on master then for the out of cwd case the output is like:
```
  "bar.pyx",
static const char __pyx_k_home_oscar_current_active_cytho[] = "/home/oscar/current/active/cython/TEST_TMP/build/run_cython_not_in_cwd2/src/foo/bar.pyx";
/* "foo/bar.pyx":2
  /* "foo/bar.pyx":3
  /* "foo/bar.pyx":2
  /* "foo/bar.pyx":2
```
If you also remove the `__init__.py` then the output is like this in both cases:
```
  "src/foo/bar.pyx",
static const char __pyx_k_src_foo_bar_pyx[] = "src/foo/bar.pyx";
/* "bar.pyx":2
  /* "bar.pyx":3
  /* "bar.pyx":2
  /* "bar.pyx":2
```
We want to extract the full path so that we can confirm that it is not an absolute path. We also need to pick up the lone basename bar.pyx as well though because we don't want that either.

I'll adjusted the test to extract a set of paths but I prefer it as it is now because it shows you directly what you would see if you grepped the .c file yourself.

I realise this kind of test is fragile but there need to be some tests that are fragile so that you can see if something like this changes. The new test is more permissive of changes to exactly which paths are output in different places.

Cython/Compiler/ExprNodes.py
```diff
@@ -10208,7 +10209,12 @@ def generate_codeobj(self, code):
         func_name_result = code.get_py_string_const(
             func.name, identifier=True, is_str=False, unicode_value=func.name)
         # FIXME: better way to get the module file path at module init time? Encoding to use?
-        file_path = StringEncoding.bytes_literal(func.pos[0].get_filenametable_entry().encode('utf8'), 'utf8')
+        file_path = func.pos[0].get_filenametable_entry()
+        if os.path.isabs(file_path):
+            file_path = func.pos[0].get_description()
+        # Always use / as separator
+        file_path = str(pathlib.Path(file_path).as_posix())
```
scoder, line 10216: `.as_posix()` deliberately returns a `str`.
```suggestion
        file_path = pathlib.Path(file_path).as_posix()
```

Cython/Compiler/ModuleNode.py
```diff
@@ -964,9 +965,11 @@ def generate_filename_table(self, code):
             for source_desc in code.globalstate.filename_list:
                 file_path = source_desc.get_filenametable_entry()
                 if isabs(file_path):
-                    file_path = basename(file_path)  # never include absolute paths
-                escaped_filename = file_path.replace("\\", "\\\\").replace('"', r'\"')
-                escaped_filename = as_encoded_filename(escaped_filename)
+                    # never include absolute paths
+                    file_path = source_desc.get_description()
+                # Always use / as separator
+                file_path = str(pathlib.Path(file_path).as_posix())
```
scoder, line 971:
```suggestion
                file_path = pathlib.Path(file_path).as_posix()
```

tests/build/run_cython_not_in_cwd.srctree
scoder: Ah, right. It seems something we might want to do in more than one test, so we could allow `cd` as a command (or maybe `CD`) and set the argument as the working directory for the following commands.

Could be a later cleanup PR, though.

tests/build/run_cython_not_in_cwd.srctree
```diff
@@ -22,35 +23,24 @@ import subprocess
...
+                yield part.rstrip(',\n;:01234566789').strip('"')
+
+cython('-3 src/pkg/mod.pyx -o build/src/pkg/mod1.pyx.c')
+cython('-3 ../src/pkg/mod.pyx -o src/pkg/mod2.pyx.c', cwd='build')
+
+res1 = grep_paths('build/src/pkg/mod1.pyx.c', 'mod.pyx')
+res2 = grep_paths('build/src/pkg/mod2.pyx.c', 'mod.pyx')
+
+# Running cython on a file in CWD uses some paths relative to CWD.
+assert set(res1) == {"src/pkg/mod.pyx", "pkg/mod.pyx"}, str(res1)
```
scoder, line 43: The way you wrote this, `str(res1)` will print something like `<generator  …>` on failure. You should collect the generator result into a `set()` right when calling it.

tests/build/run_cython_not_in_cwd.srctree
oscarbenjamin: Ideally if the top part was just a shell script I would have written it like this:
```
cython src/pkg/mod.pyx -o build/src/pkg/mod1.pyx.c
cd build && cython ../src/pkg/mod.pyx -o src/pkg/mod1.pyx.c
python check.py
```
I suppose a simple version could be like:
```
CYTHON src/pkg/mod.pyx -o build/src/pkg/mod1.pyx.c
CD build
CYTHON ../src/pkg/mod.pyx -o src/pkg/mod1.pyx.c
CD ..
PYTHON check.py
```
Managing the changed directory in the test runner could be fiddly though. I would say it is not worth it unless there are many tests that need this.

scoder: Here's an untested proposal:
```patch
diff --git a/runtests.py b/runtests.py
index 3b8abe9aa..5473e4f1a 100755
--- a/runtests.py
+++ b/runtests.py
@@ -2028,6 +2028,9 @@ class EndToEndTest(unittest.TestCase):
                     envvar = None
                 env.pop(envvar, None)
                 continue
+            elif command[0] == 'CD':
+                os.chdir(command[1] if len(command) > 1 else self.workdir)
+                continue
             time_category = 'etoe-build' if (
                 'setup.py' in command or 'cythonize.py' in command or 'cython.py' in command) else 'etoe-run'
             with self.stats.time('%s(%d)' % (self.name, command_no), 'c', time_category):
```

oscarbenjamin: Using chdir is almost always a bad idea. Instead the runner should maintain some notion of a working directory and then pass that as the `cwd` parameter to every single subprocess invocation. That is why I said that it could be fiddly.

scoder: It's how things currently work, though. The EndToEnd tests have `chdir(workdir)` in their `setUp()` method and reset it in `tearDown()`.

Longer patch, also untested:
```patch
diff --git a/runtests.py b/runtests.py
index 3b8abe9aa..22aef7447 100755
--- a/runtests.py
+++ b/runtests.py
@@ -1996,8 +1996,6 @@ class EndToEndTest(unittest.TestCase):
     def setUp(self):
         from Cython.TestUtils import unpack_source_tree
         _, self.commands = unpack_source_tree(self.treefile, self.workdir, self.cython_root)
-        self.old_dir = os.getcwd()
-        os.chdir(self.workdir)
 
     def tearDown(self):
         if self.cleanup_workdir:
@@ -2008,7 +2006,6 @@ class EndToEndTest(unittest.TestCase):
                     time.sleep(0.1)
                 else:
                     break
-        os.chdir(self.old_dir)
 
     def runTest(self):
         self.success = False
@@ -2020,6 +2017,7 @@ class EndToEndTest(unittest.TestCase):
         cmd = []
         out = []
         err = []
+        workdir = self.workdir
         for command_no, command in enumerate(self.commands, 1):
             if command[0] == "UNSET":
                 try:
@@ -2028,15 +2026,21 @@ class EndToEndTest(unittest.TestCase):
                     envvar = None
                 env.pop(envvar, None)
                 continue
+            elif command[0] == "CD":
+                if len(command) == 1:
+                    workdir = self.workdir
+                else:
+                    workdir = os.path.join(workdir, command[1])
+                continue
             time_category = 'etoe-build' if (
                 'setup.py' in command or 'cythonize.py' in command or 'cython.py' in command) else 'etoe-run'
             with self.stats.time('%s(%d)' % (self.name, command_no), 'c', time_category):
                 if self.capture:
-                    p = subprocess.Popen(command, stderr=subprocess.PIPE, stdout=subprocess.PIPE, env=env)
+                    p = subprocess.Popen(command, stderr=subprocess.PIPE, stdout=subprocess.PIPE, env=env, cwd=workdir)
                     _out, _err = p.communicate()
                     res = p.returncode
                 else:
-                    p = subprocess.call(command, env=env)
+                    p = subprocess.call(command, env=env, cwd=workdir)
                     _out, _err = b'', b''
                     res = p
             cmd.append(command)
```

oscarbenjamin: That being said I don't want to mess around with the runner too much in this PR. It already uses chdir so why not use it more I guess.

I have just pushed this implementation of `CD`.

oscarbenjamin: Our messages crossed paths. I've pushed that longer version here now.

The CI won't run until someone approves it though.

oscarbenjamin:
I understand it takes time to review these things but to be clear I believe I have addressed all review comments above.

Please let me know if I have misunderstood and anyone is actually waiting for me to make further changes...

scoder:
Looks good. Thanks.

oscarbenjamin:
Thanks!

PR #6349, opened 2024-08-22
Change C code encoding from Latin-1 to UTF-8

scoder:
I don't think there's a reason to produce plain ASCII in comments, as long as the actual C code (including string literals) that we generate is ASCII.


robertwb:
The only risk I see is that we lose the validation that non-comment lines are plain ASCII. Being able to embed file paths may be worth this downside though.

PR #6360, opened 2024-08-25
Fixes for cdef subclass of int/float/complex user defined types

dalcinl:
* Lookup complex type in current scope.
* Remove special handling for `c_long_type` as `long` is Python 2.

@scoder's objections https://github.com/cython/cython/commit/a2120861a9c3f87b246c9ade4bed68b84ce95dba#r145771848 to my changes in #6346 proved very accurate. Indeed, I trusted type analysis too much. As per Stefan's request, I'm PRing the fix.

The following type-analysis code doesn't not do the right thing. If the scope is a C context (which it is in a cdef class?), I believe it goes too quickly to check whether the name can be parsed as int/float/complex basic C type, rather than looking first in the scope to see if name is already user defined.

```python
# Cython/Compiler/ExprNodes.py
class NameNode(AtomicExprNode):
    ...
        type = None
        if self.cython_attribute:
            type = PyrexTypes.parse_basic_type(self.cython_attribute)
        elif env.in_c_type_context: # <-- HERE is the problem
            type = PyrexTypes.parse_basic_type(self.name)
        if type:
            return type
    ...
```
In any case, after the changes in this PR things will work as expected when subclassing a builtin vs user-defined  `complex` type, but maybe the type analysis issue may bite in other scenarios.
Unfortunately, I do not consider myself expert enough in type analysis as to even suggest a fix, much less provide an implementation.

### DIFF
--- a/Cython/Compiler/Nodes.py
+++ b/Cython/Compiler/Nodes.py
@@ -5482,10 +5482,10 @@ class CClassDefNode(ClassDefNode):
             base_type = base.analyse_as_type(env)
 
             # If we accidentally picked the C type of the same name, use the Python rather than the C variant.
-            if base_type in (PyrexTypes.c_int_type, PyrexTypes.c_long_type, PyrexTypes.c_float_type):
+            if base_type in (PyrexTypes.c_int_type, PyrexTypes.c_float_type):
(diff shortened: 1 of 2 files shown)


scoder:
Thanks @dalcinl

PR #6362, opened 2024-08-26
Fix access to builtins dictionary with Py_LIMITED_API

dalcinl:
The first commit is the important one (including a fix for limited API), while the second are related optimizations.

### DIFF
--- a/Cython/Utility/ObjectHandling.c
+++ b/Cython/Utility/ObjectHandling.c
@@ -1424,10 +1424,13 @@ static int __Pyx_SetNewInClass(PyObject *ns, PyObject *name, PyObject *value) {
 #if !CYTHON_COMPILING_IN_LIMITED_API
         staticnew = PyStaticMethod_New(value);
 #else
-        PyObject *builtins, *staticmethod;
+        PyObject *builtins, *staticmethod_str, *staticmethod;
         builtins = PyEval_GetBuiltins(); // borrowed
         if (!builtins) return -1;
-        staticmethod = PyObject_GetAttrString(builtins, "staticmethod");
+        staticmethod_str = PyUnicode_FromStringAndSize("staticmethod", 12);
+        if (!staticmethod_str) return -1;
+        staticmethod = PyObject_GetItem(builtins, staticmethod_str);
+        Py_DECREF(staticmethod_str);
         if (!staticmethod) return -1;
         staticnew = PyObject_CallFunctionObjArgs(staticmethod, value, NULL);
         Py_DECREF(staticmethod);


Cython/Utility/Builtins.c
```diff
@@ -127,9 +127,9 @@ static PyObject* __Pyx_PyExec3(PyObject* o, PyObject* globals, PyObject* locals)
         // For the limited API we just defer to the actual builtin
         // (after setting up globals and locals) - there's too much we can't do otherwise
         PyObject *builtins, *exec;
-        builtins = PyEval_GetBuiltins();
+        builtins = PyEval_GetBuiltins();  // borrowed
         if (!builtins) return NULL;
-        exec = PyDict_GetItemString(builtins, "exec");
+        exec = PyDict_GetItem(builtins, PYIDENT("exec"));  // borrowed
```
scoder, line 132: If you move the comments into a separate line instead of at the end of a code line, they'll be removed from the generated C files.

Cython/Utility/Exceptions.c
```diff
@@ -22,17 +22,13 @@ if (likely(__Pyx_init_assertions_enabled() == 0)); else
...
+    PyObject *builtins, *debug;
     int flag;
-    builtins = PyEval_GetBuiltins();
+    builtins = PyEval_GetBuiltins();  // borrowed
     if (!builtins) goto bad;
-    debug_str = PyUnicode_FromStringAndSize("__debug__", 9);
-    if (!debug_str) goto bad;
-    debug = PyObject_GetItem(builtins, debug_str);
-    Py_DECREF(debug_str);
+    debug = PyDict_GetItem(builtins, PYIDENT("__debug__"));  // borrowed
```
scoder, line 29: I spelled this out intentionally at the time because I didn't want to add the `"__debug__"` string constant to modules (and Python versions) that don't use it. This is one-time initialisation code, not worth keeping a live name constant reference around.

Also, I probably used `PyObject_GetItem()` instead of `PyDict_GetItem()` to avoid the borrowed reference. There's no need for a borrowed reference here either. They are generally risky things and this is Limited API code, so we don't know which implementation it will run with.

scoder (commented):
The bug fix is fine. The change in the assertion initialisation code is unnecessary. And I'd rather try to avoid borrowed references in Limited API code.

dalcinl:
PR updated with a single commit and only touching what is needed. My attempts to uniformize coding style usually lead to objections; therefore, from now on, I'll refrain from such changes.

scoder:
> My attempts to uniformize coding style usually lead to objections; therefore, from now on, I'll refrain from such changes.

It's an old code base (>20 years by now), so it's usually a good thing to update code that we touch anyway to newer styles and infrastructure (such as f-strings). I consider it *less* helpful to do larger style updates (because they make reading the blame history harder). And occasionally, like in this case, there's a reason why things were written the way they are. If so, I should learn to leave a comment in the code more often than not. You were really just unlucky here.

dalcinl:
@scoder This one should also be ready for merge. Note that now I'm using  `PyObject_GetItem()` following your advice related to avoiding borrowed refs in limited API code.

scoder:
Thanks

Issue #6364, opened 2024-08-26
Missing Python 3.13 info from the `classifiers` in setup.py


rathaROG:
### Describe the bug

Not a bug! Just missing info.

### Python version

3.13

### Cython version

3.0.11

### Additional context

Hello, 

I know Cython 3.0.11 supports Python 3.13, but this info is missing from the `classifiers` in setup.py, and at first [glance on PyPI](https://pypi.org/project/Cython/), I thought it didn't support Python 3.13 yet...

https://github.com/cython/cython/blob/b355423c7c6a38e2f7d2e8dc9d711acd533f8e5f/setup.py#L327

Please don't forget this in the next release! Thanks! 

Bests,

PR #6365, opened 2024-08-26
Add Python 3.13 info to the `classifiers` in setup.py

rathaROG:
Fix https://github.com/cython/cython/issues/6364


scoder:
Thanks

rathaROG:
> Thanks

It's the least I can do

Issue #6366, opened 2024-08-26
[BUG] Assertion error from `encode_single_position`,


oscarbenjamin:
### Describe the bug

Currently python-flint has a CI job that tests against Cython master branch to measure coverage (gh-6341). That job was passing but has recently (today?) started failing.

### Code to reproduce the behaviour:

I don't have a simple reproducer but here is the error message:
```console
[34/111] Compiling Cython source /home/runner/work/python-flint/python-flint/src/flint/types/nmod.pyx
FAILED: src/flint/types/nmod.cpython-312-x86_64-linux-gnu.so.p/src/flint/types/nmod.pyx.c 
cython -M --fast-fail -3 -X embedsignature=True -X emit_code_comments=True -X linetrace=True /home/runner/work/python-flint/python-flint/src/flint/types/nmod.pyx -o src/flint/types/nmod.cpython-312-x86_64-linux-gnu.so.p/src/flint/types/nmod.pyx.c
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.12.5/x64/bin/cython", line 8, in <module>
    sys.exit(setuptools_main())
             ^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.5/x64/lib/python3.12/site-packages/Cython/Compiler/Main.py", line [74](https://github.com/flintlib/python-flint/actions/runs/10567756039/job/29278457452?pr=179#step:8:75)9, in setuptools_main
    return main(command_line = 1)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.5/x64/lib/python3.12/site-packages/Cython/Compiler/Main.py", line 772, in main
    result = compile(sources, options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.5/x64/lib/python3.12/site-packages/Cython/Compiler/Main.py", line 668, in compile
    return compile_multiple(source, options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/hostedtoolcache/Python/3.12.5/x64/lib/python3.12/site-packages/Cython/Compiler/Main.py", line 642, in compile_multiple
    result = run_pipeline(source, options,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6369, opened 2024-08-28
Give cdef functions the same position and first line number as def and @cfunc functions

scoder:
I.e. the one of the outermost (first) decorator, not necessarily the function signature line.
This affects both the error reporting (column before instead of after a `cdef`, for example) as well as line numbers in profiling/tracing.

Closes https://github.com/cython/cython/issues/6366


scoder:
I left out further changes to `cdef` statements that are not functions, but I consider this a net improvement.

PR #6395, opened 2024-09-14
Fix case when `typing.Union[...]` contains multiple nested types

matusvalo:
Related to #6333


scoder:
There's no strict need for the indentation in the else case, but apart from that, this looks good to me.

Issue #6400, opened 2024-09-18
[BUG] Cython-generated variable name is a C++ keyword 


tvalentyn:
### Describe the bug

I noticed the following Cython-translated excerpt from running Cython over fastavro sources.

```

/* "fastavro/_schema.pyx":561
 *
 *
 * cdef _inject_schema(outer_schema, inner_schema, namespace="", is_injected=False):             # <<<<<<<<<<<<<<
 *     # Once injected, we can stop checking to see if we need to inject since it
 *     # should only be done once at most
 */
struct __pyx_opt_args_8fastavro_7_schema__inject_schema {
  int __pyx_n;
  PyObject *namespace;
  PyObject *is_injected;
};
```

Because in C++ `namespace` is a reserved keyword,  the translated code doesn't compile in my environment with default compiler setting, and I have to apply some workarounds (https://github.com/fastavro/fastavro/commit/dd026c6e584b3ed3e731002e815500f49dd554c2).

Per discussion on https://groups.google.com/g/cython-users/c/qIJzaBurOWg, it should be possible to select better variable names here.

scoder:
Fixed in https://github.com/cython/cython/commit/6aeff0e9d76d130b4c300f90863d505f5bea3ef3

PR #6401, opened 2024-09-19
Also avoid C++ keywords as identifiers.

robertwb:
We currently rename identifiers (such as struct fields) that are C keywords, we should do the same for the larger set of C++ keywords.

At this point in the code we do not have the context of whether we're compiling in C++ or not, but it should be safe to escape these (private) names a bit more agressively even when it is not strictly necessary.

### DIFF
--- a/Cython/Compiler/Symtab.py
+++ b/Cython/Compiler/Symtab.py
@@ -30,12 +30,26 @@ iso_c99_keywords = {
     '_Bool', '_Complex'', _Imaginary', 'inline', 'restrict',
 }
 
+iso_cpp_keywords = {
+    'and', 'and_eq', 'asm', 'auto', 'bitand', 'bitor', 'bool', 'break', 'case',
+    'catch', 'char', 'class', 'compl', 'const', 'const_cast', 'continue',
+    'default', 'delete', 'do', 'double', 'dynamic_cast', 'else', 'enum',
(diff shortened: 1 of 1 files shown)


robertwb:
This should fix #6400.

scoder:
> At this point in the code we do not have the context of whether we're compiling in C++ or not

We call the `c_safe_identifier()` function exclusively from inside Scope classes and those always know their module scope, which in turn knows `self.context.cpp`. We could have a `cpp_safe_identifier()` and call either of the two depending on the context. Or, move `c_safe_identifier` into `ModuleScope` as a method and use the appropriate set of keywords. That way, we could avoid escaping names unnecessarily in C code.

scoder:
> Scope classes and those always know their module scope

… hmm, what I wrote here wasn't correct, actually. Specifically struct scopes are independent of a module. Whether or not that is a good thing, it is how it currently is.

I implemented a slightly improved change in https://github.com/scoder/cython/commit/b8b5a2412fc35fe3905d16199de34ea2be9cfd8a, moving the set of reserved words from `Symtab.py` to `Naming.py` and merging both sets into one (at init time). I also took both sets from the C/C++23 standards. Does that match your expectations?

robertwb:
That looks fine to me.

scoder:
Superseded by https://github.com/cython/cython/commit/6aeff0e9d76d130b4c300f90863d505f5bea3ef3

PR #6412, opened 2024-09-27
Use `memoryview` type in `cpython.memoryview`

jakirkham:
Convert these from `object` to `memoryview`. This will allow these to return the `memoryview` typed object and thus bypass additional checks and benefit from associated optimizations.

Left alone the input arguments of...:
* `PyMemoryView_Check`
* `PyMemoryView_GET_BUFFER`
* `PyMemoryView_GET_BASE`

The first one should handle any type.

The latter two say the don't check type currently. So presumably we don't want to add our own checks on top of those.

That said, perhaps there is a better way to accomplish what the latter two functions offer through `memoryview` attributes?

### DIFF
--- a/Cython/Includes/cpython/memoryview.pxd
+++ b/Cython/Includes/cpython/memoryview.pxd
@@ -6,27 +6,27 @@ cdef extern from "Python.h":
     # A memoryview object exposes the C level buffer interface as a Python
     # object which can then be passed around like any other object
 
-    object PyMemoryView_FromObject(object obj)
+    memoryview PyMemoryView_FromObject(object obj)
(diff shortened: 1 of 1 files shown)


scoder (approved):
`object` is almost always the right declaration for Python *input* arguments of C-API functions. There is no use in letting Cython generate additional type checks at call time that may or may not be needed by the function. Especially macros are designed to work without costly checks.

jakirkham:
Thanks Stefan! 🙏

Yep that makes sense to me. Please let me know if you need anything else 🙂

jakirkham:
Thanks all! 🙏

PR #6452, opened 2024-10-25
Improve Bazel support

mering:
- Prefer rules from `rules_python` as the native Bazel rules are going to be removed soon (Bazel 8?)
- Rename a target as it caused problems on OSX but add an alias such that existing references (e.g. in `deps` attributes) continue to work


mering:
Tagging @robertwb as you authored the existing Bazel support.

robertwb (approved):

PR #6455, opened 2024-10-25
Fix #6454

### LINKED ISSUE #6454: [BUG] Passing ctuple literal to function with fused type results in error
### Describe the bug

Passing a ctuple literal or constructor to a function with a fused type results in the error "Invalid use of fused types, type cannot be specialized". Notably, this only occurs with literals and literal constructors.

```cython
ctypedef fused ab:
    (int, int, int)
    (int, int, int, int)

cdef int test(ab a):
    return 1

print(test((1, 2, 3)))
```

### Expected behaviour

_No response_

### OS

_No response_

### Python version

3.8

### Cython version

dev

### Additional context

_No response_

bbb23exposed:
Fixes #6454

### DIFF
--- a/tests/compile/ctuple_6454.pyx
+++ b/tests/compile/ctuple_6454.pyx
@@ -1,0 +1,10 @@
+# mode: compile
+
+ctypedef fused ab:
+    (int, int, int)
+    (int, int, int, int)
+
+cdef int test(ab a):
+    return 1
+
+print(test((1, 2, 3)))
--- a/Cython/Compiler/PyrexTypes.py
+++ b/Cython/Compiler/PyrexTypes.py
@@ -4656,6 +4656,9 @@ class CTupleType(CType):
         new_entry = self.entry.scope.declare_tuple_type(self.entry.pos, components)
         return new_entry.type
 
+    def assignable_from(self, src_type):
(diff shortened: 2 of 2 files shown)


bbb23exposed:
> I think we are missing negative scenarios for:
> 
> 1. not matching number of types in ctuple
> 2. one type of ctuple does not match
> 
> Tests can be put here: tests/errors/fused_types.pyx. Otherwise LGTM for me.

These just fall back to c-to-py-to-c conversions.

tests/compile/ctuple_6454.pyx
```diff
@@ -0,0 +1,16 @@
...
+cdef int test(ab x):
+    return 1
+
+ctypedef (float, float) c
+ctypedef (int, int) d
+
+cdef d test2(c x):
+    return x
+
+print(test((1, 2, 3)))
```
scoder, line 16:
```suggestion
print(test((1, 2, 3)))
```

tests/compile/ctuple_6454.pyx
```diff
@@ -0,0 +1,16 @@
+# mode: compile
```
scoder, line 1: I'd prefer a runnable test that actually verifies the correct assignment mapping.

scoder (commented):
Looks mostly good.

PR #6464, opened 2024-10-27
Add comparison code for structs and ctuples

### LINKED ISSUE #6438: [BUG] Comparing ctuple to constant ctuple results in error
### Describe the bug

Comparing a ctuple to a constant ctuple results in a cython compilation error even if every type is explicitly casted.

```cython
ctypedef unsigned short node_input_size_t

ctypedef (void*, node_input_size_t) slot_pointer

cdef void add_connection(slot_pointer path):
    print(path == <slot_pointer>(<void*>NULL, <node_input_size_t>-1))
```

### Expected behaviour

_No response_

### OS

_No response_

### Python version

3.8

### Cython version

3.0.10

### Additional context

_No response_

bbb23exposed:
Fixes #6438 and #6456

Requires #6455


Cython/Compiler/ExprNodes.py
```diff
@@ -13589,7 +13588,46 @@ def generate_operation_code(self, code, result_code,
...
+
+            self.operand1_temp = code.funcstate.allocate_temp(operand1.type, manage_ref=False)
+            self.operand2_temp = code.funcstate.allocate_temp(operand1.type, manage_ref=False)
+            code.putln(f"{self.operand1_temp} = {operand1.result()};")
+            coercion_node = operand2.coerce_to(operand1.type, code.funcstate.scope)
+            coercion_node.generate_result_code(code)
+            code.putln(f"{self.operand2_temp} = {coercion_node.result()};")
+            coercion_node.generate_disposal_code(code)
+            coercion_node.free_temps(code)
+            code.putln(f"{result_code} = {self.generate_struct_code(self.operand1_temp, self.operand2_temp, operand1.type, operand2.type, op, code)};")
```
scoder, line 13611: Long function calls (or lengthy calculations) are not easy to read and follow inside of f-string values. Same below.
```suggestion
            struct_code = self.generate_struct_code(self.operand1_temp, self.operand2_temp, operand1.type, operand2.type, op, code)
            code.putln(f"{result_code} = {struct_code};")
```

Cython/Compiler/ExprNodes.py
```diff
@@ -13589,7 +13588,46 @@ def generate_operation_code(self, code, result_code,
...
+                error(self.pos, "cannot compare struct to non-struct")
+
+            if (operand1.type.is_struct_or_union and operand1.type.kind == 'union') or (operand2.type.is_struct_or_union and operand2.type.kind == 'union'):
+                error(self.pos, "cannot compare unions")
+
+            if not operand1.type.assignable_from(operand2.type):
+                error(self.pos, "cannot compare structs of different types")
+
+            self.operand1_temp = code.funcstate.allocate_temp(operand1.type, manage_ref=False)
+            self.operand2_temp = code.funcstate.allocate_temp(operand1.type, manage_ref=False)
```
scoder, line 13604: I think these can just be local variables. Same in the elif section below.

Cython/Compiler/ExprNodes.py
```diff
@@ -13589,7 +13588,46 @@ def generate_operation_code(self, code, result_code,
...
+            code.funcstate.release_temp(self.operand1_temp)
+            code.funcstate.release_temp(self.operand2_temp)
+        elif (operand1.type.is_ctuple and operand2.type.is_ctuple) or (op in ("in", "not_in") and operand2.type.is_ctuple and any([operand1.type.assignable_from(type_) for type_ in operand2.type.components])):
+            self.operand1_temp = code.funcstate.allocate_temp(operand1.type, manage_ref=False)
+            self.operand2_temp = code.funcstate.allocate_temp(operand2.type, manage_ref=False)
+            code.putln(f"{self.operand1_temp} = {operand1.result()};")
+            if op in ("in", "not_in"):
+                code.putln(f"{self.operand2_temp} = {operand2.result()};")
+            else:
+                coercion_node = operand2.coerce_to(operand1.type, code.funcstate.scope)
```
scoder, line 13622: Coercions should not happen during code generation. They are a type-analysis thing.

Cython/Compiler/ExprNodes.py
```diff
@@ -13619,6 +13657,59 @@ def generate_operation_code(self, code, result_code,
             else:
                 code.putln(statement)
 
+    def generate_struct_code(self, operand1_name, operand2_name, operand1_type, operand2_type, op, code):
+        return '(' + (' && ' if op == '==' else ' || ').join([self.generate_ctype_code(operand1_name + '.' + member.name, operand2_name + '.' + member.name, member.type, member.type, op, code) for member in operand2_type.scope.var_entries]) + ')'
```
scoder, line 13661: Please cut long expressions like this into separate parts, explaining their meaning with variable names.

Cython/Compiler/ExprNodes.py
```diff
@@ -13619,6 +13657,59 @@ def generate_operation_code(self, code, result_code,
...
+
+        temps = list()
+
+        nodes = dict()
+        for itr in range(len(operand2_type.components)):
+            if op in ('in', 'not_in') and not operand1_type.assignable_from(operand2_type.components[itr]):
+                continue
+
+            node = IndexNode(self.pos, base=RawCNameExprNode(self.pos, cname=operand2_name, type=operand2_type), index=IntNode(self.pos, value=str(itr), constant_result=itr, type=PyrexTypes.c_py_ssize_t_type))
+            node.analyse_types(code.funcstate.scope)
```
scoder, line 13675: Again, this is way too late in the translation process. This should be done much earlier, either during type analysis (usually) or in a later optimisation phase (if that's what it is).

Cython/Compiler/Pipeline.py
```diff
@@ -194,7 +194,6 @@ def create_pipeline(context, mode, exclude_classes=()):
         AutoCpdefFunctionDefinitions(context),
         RemoveUnreachableCode(context),
         ConstantFolding(),
-        FlattenInListTransform(),
```
scoder, line 197: This is intentionally done early. Why do you want to move it?

scoder (commented):
It would really help me reviewing your PR if you provided informative commit comments to explain why you made a specific change. "Update file.py" is something I can see from the diff. "Changed X to achieve Y" is something I cannot easily infer from the change.

da-woods:
I've not had time to look at the implementation I'm afraid, but:

We definitely don't want to apply this to extern structs where:
1. We may not have visibility over all the contents of the struct,
2. In c++ mode there may already be an operator== and thus people may already be relying on the existing behaviour

(Sorry if you have covered this already, but it didn't look like it on a very quick skim)

Cython/Compiler/ExprNodes.py
```diff
@@ -13722,25 +13819,28 @@ def analyse_types(self, env):
...
         elif self.find_special_bool_compare_function(env, self.operand1):
-            if not self.operand1.type.is_pyobject:
+            if not type1.is_pyobject:
                 self.operand1 = self.operand1.coerce_to_pyobject(env)
             common_type = None  # if coercion needed, the method call above has already done it
             self.is_pycmp = False  # result is bint
         else:
             common_type = self.find_common_type(env, self.operator, self.operand1)
-            self.is_pycmp = common_type.is_pyobject
+            self.is_pycmp = common_type is not None and common_type.is_pyobject
```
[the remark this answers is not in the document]
bbb23exposed: It was already checked in some places and not in others. I made it consistent.

Also, yes, if you do `x in <ctuple>` you get None.

Cython/Compiler/ExprNodes.py
```diff
@@ -13589,7 +13588,46 @@ def generate_operation_code(self, code, result_code,
                 operand1.type.unary_op('eq'),
                 operand1.result(),
                 operand2.result()))
-
+        elif operand1.type.is_struct_or_union or operand2.type.is_struct_or_union:
+            assert op in ('==', '!=')
+
+            if not operand1.type.is_struct_or_union or not operand2.type.is_struct_or_union:
+                error(self.pos, "cannot compare struct to non-struct")
```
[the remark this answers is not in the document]
bbb23exposed: Okay, I'll try to implement that.

Cython/Compiler/ExprNodes.py
```diff
@@ -13589,7 +13588,46 @@ def generate_operation_code(self, code, result_code,
...
+
+            if (operand1.type.is_struct_or_union and operand1.type.kind == 'union') or (operand2.type.is_struct_or_union and operand2.type.kind == 'union'):
+                error(self.pos, "cannot compare unions")
+
+            if not operand1.type.assignable_from(operand2.type):
+                error(self.pos, "cannot compare structs of different types")
```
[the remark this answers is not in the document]
bbb23exposed: You're right. I was think the equivalent to ctuples where comparing (int, int, int) to (long, long, long) is a use case. I'll change it to ==.

Cython/Compiler/Pipeline.py
bbb23exposed: Because ctuple `in` and `not in` require type information as they can only compare compatible types, to match with python behavior.

Cython/Compiler/ExprNodes.py
bbb23exposed: Okay, I'll work on moving this to that stage.

bbb23exposed:
> It would really help me reviewing your PR if you provided informative commit comments to explain why you made a specific change. "Update file.py" is something I can see from the diff. "Change X to achieve Y" is something I cannot easily infer from the change.

I'm very new to Cython development so I often attempt multiple things before I settle on a solution.

bbb23exposed:
> I've not had time to look at the implementation I'm afraid, but:
> 
> We definitely don't want to apply this to extern structs where:
> 
> 1. We may not have visibility over all the contents of the struct,

When is this the case?

da-woods:
> > 1. We may not have visibility over all the contents of the struct,
> 
> When is this the case?

Because you don't have to tell Cython about everything in an extern struct. For example if you have a header file with

```c
struct SomeStruct {
    int x;
};
```

it's fine to declare it to Cython as

```cython
cdef extern from "header.h":
    struct SomeStruct:
        pass
```

and just use in as an opaque type - you can pass it around to C functions, but not actually access its internals.

Alternatively you can just tell it about the fields that you plan to access from Cython.

Cython/Compiler/ExprNodes.py
bbb23exposed: Done, this was left over from copied code that used these in a `clear_temps()` function.

Cython/Compiler/ExprNodes.py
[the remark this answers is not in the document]
bbb23exposed: Done.

Cython/Compiler/ExprNodes.py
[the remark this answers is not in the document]
bbb23exposed: Done.

Cython/Compiler/ExprNodes.py
bbb23exposed: Done.

Cython/Compiler/ExprNodes.py
bbb23exposed: Done.

Cython/Compiler/ExprNodes.py
bbb23exposed: Done.

bbb23exposed:
> ```cython
> cdef extern from "header.h":
>     struct SomeStruct:
>         pass
> ```

This seems to be broken in Cython already. For example:
```cython
cdef extern from "header.h":
    struct v:
        pass

cdef bint test(v a, v b):
    return a.x
```

..will result in broken C code.

bbb23exposed:
> 2. In c++ mode there may already be an operator== and thus people may already be relying on the existing behaviour

Okay, it now avoids generate_struct_code if either struct has in_cpp set.

da-woods:
> This seems to be broken in Cython already. For example:
> 
> ```cython
> cdef extern from "header.h":
>     struct v:
>         pass
> 
> cdef bint test(v a, v b):
>     return a.x
> ```
> 
> ..will result in broken C code.

No... that's not the point. You can't use attributes that Cython doesn't know about. However a lot of the time you don't want to use the attributes - you just want to pass the struct from an extern C function to another extern C function.

bbb23exposed:
> > This seems to be broken in Cython already. For example:
> > ```cython
> > cdef extern from "header.h":
> >     struct v:
> >         pass
> > 
> > cdef bint test(v a, v b):
> >     return a.x
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ..will result in broken C code.
> 
> No... that's not the point. You can't use attributes that Cython doesn't know about. However a lot of the time you don't want to use the attributes - you just want to pass the struct from an extern C function to another extern C function.

The point is that in this case, Cython automatically casts the struct to an empty dict instead throwing an error.

Cython/Compiler/ExprNodes.py
```diff
@@ -14101,6 +14156,33 @@ def generate_result_code(self, code):
...
+        self.arg = arg
+
+    def generate_result_code(self, code):
+        self.recursive_assign(self.arg.result(), self.temp_code, self.arg.type, self.type, code)
+
+    def recursive_assign(self, src, dest, src_type, dest_type, code):
+        if dest_type.is_ctuple and dest_type != src_type:
+            for index, (member_src_type, member_dest_type) in enumerate(zip(src_type.components, dest_type.components)):
+                member_src = src + '.f' + str(index)
+                member_dest = dest + '.f' + str(index)
```
scoder, line 14179: This might be something that the type should be responsible for rather than an arbitrary external syntax node.

Cython/Compiler/ExprNodes.py
```diff
@@ -9532,11 +9526,12 @@ def calculate_constant_result(self):
         self.constant_result = (
             self.key.constant_result, self.value.constant_result)
 
-    def analyse_types(self, env):
+    def analyse_types(self, is_dict, env):
```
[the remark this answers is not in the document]
bbb23exposed: > `analyse_types()` is a base class API method. You cannot just change its signature arbitrarily inside of the syntax tree. It's ok to unpack coerced values later, if we decide that there is a better way to process them. That's better than distributing full global knowledge everywhere first.

The problem is the values are IntNodes, and instead of getting wrapped, they just get `type` changed.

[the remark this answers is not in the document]
bbb23exposed: I suppose we could fix this by just using CoerceToPyTypeNode directly.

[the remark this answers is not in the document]
bbb23exposed: Should be fixed now.

Cython/Compiler/Optimize.py
```diff
@@ -1417,18 +1417,23 @@ def visit_PrimaryCmpNode(self, node):
...
-                                          ExprNodes.SetNode)):
+                                          ExprNodes.SetNode)) or node.operand2.type.is_ctuple):
             return node
 
         lhs = node.operand1
-        args = node.operand2.args
+
+        if node.operand2.type.is_ctuple and not isinstance(node.operand2, ExprNodes.TupleNode):
+            rhs = UtilNodes.ResultRefNode(node.operand2)
+            args = [ExprNodes.IndexNode(node.pos, base=rhs, index=ExprNodes.IntNode(node.pos, value=str(index), constant_result=index, type=PyrexTypes.c_py_ssize_t_type)).analyse_types(self.current_env()).coerce_to(node.operand1.type, self.current_env()) for index, type_ in enumerate(rhs.type.components) if node.operand1.type.assignable_from(type_) or ((node.operand1.type.is_pyobject and PyrexTypes.py_object_type.assignable_from(type_)) or (type_.is_pyobject and PyrexTypes.py_object_type.assignable_from(node.operand1.type)))]
```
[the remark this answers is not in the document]
bbb23exposed: Done

Cython/Compiler/ExprNodes.py
bbb23exposed: Done

Cython/Compiler/Optimize.py
```diff
@@ -1417,18 +1417,33 @@ def visit_PrimaryCmpNode(self, node):
...
+                        base=rhs,
+                        index=ExprNodes.IntNode(node.pos,
+                            value=str(index),
+                            constant_result=index,
+                            type=PyrexTypes.c_py_ssize_t_type)
+                        ).analyse_types(self.current_env()).coerce_to(node.operand1.type, self.current_env())
+                    for index, type_ in enumerate(rhs.type.components)
+                        if node.operand1.type.assignable_from(type_) or \
+                            ((node.operand1.type.is_pyobject and PyrexTypes.py_object_type.assignable_from(type_)) or \
+                             (type_.is_pyobject and PyrexTypes.py_object_type.assignable_from(node.operand1.type)))]
```
scoder, line 1439: If you need the type information here, why not do this in `PrimaryCmpNode.analyse_types()`? This code is just spelling out how we need to implement `in` tests on ctuples correctly. There are cases where they don't currently work at all, e.g. for ctuples of pointers which cannot be converted to Python tuples. Thus, it's not actually an optimisation but rather a language feature implementation. It does not need to happen in the (optimising) `FlattenInListTransform` and `analyse_types()` is a good place for this kind of syntax tree replacement.

Cython/Compiler/ExprNodes.py
```diff
@@ -9497,7 +9491,10 @@ def generate_evaluation_code(self, code):
                 member = struct_scope.lookup_here(item.key.value)
                 assert member is not None, f"struct member {item.key.value} not found, error was not handled during coercion"
                 key_cname = member.cname
-                value_cname = item.value.result()
+                value = item.value
+                if isinstance(value, CoerceToPyTypeNode):
+                    value = value.arg
```
scoder, line 9496: The code generation phase is too late to unpack this. It needs to be done earlier in a transform. When we get to the code generation, the tree needs to be correct.

Cython/Compiler/ExprNodes.py
```diff
@@ -14101,6 +14160,33 @@ def generate_result_code(self, code):
         ))
 
 
+class CTupleCastNode(CoercionNode):
```
scoder, line 14163: Why is this not a `TypecastNode`?

PR #6482, opened 2024-11-05
pypy has no 'internal/*.h' include files

mattip:
We have started to work on PyPy for python3.11, and I came across this code. So far, we are sticking with the pre-3.11 include layout, if things change I will submit another PR.


scoder:
Thanks

PR #6487, opened 2024-11-08
Document `__pyx_capi__`

### LINKED ISSUE #6484: Document `__pyx_capi__`
### Describe your issue

`__pyx_capi__` seems to be used in other projects, so it became defacto API of cython[1][2]. Currently it is only mentioned in FAQ [3]. It seems to be important for the clients (see https://github.com/cython/cython/issues/3077#issuecomment-2456181198) and there is no definition of stability of this API. So I propose:

1. Document `__pyx_capi__` since it is already an API used by other projects
2. Define the rules of the stability - e.g. we guarantee that it will be changed with deprecation period and migration strategy.

[1] https://github.com/ashwinvis/cython_capi
[2] https://github.com/numba/numba/blob/0f363d1b2dd19f2aa1a8cec5f0a99c3dd95512f8/numba/_helperlib.c#L570
[3] https://cython.readthedocs.io/en/latest/src/userguide/faq.html#how-do-i-compile-cython-with-subpackages

matusvalo:
Fixes #6484

### DIFF
--- a/docs/src/userguide/sharing_declarations.rst
+++ b/docs/src/userguide/sharing_declarations.rst
@@ -275,6 +275,61 @@ Some things to note about this example:
   time, so to access :func:`shrubbing.standard_shrubbery` we also need to
   ``import shrubbing``.
 
+``__pyx_capi__`` module attribute
+=================================
+Every Cython extension providing API via .pxd file contains a special attribute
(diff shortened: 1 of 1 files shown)


scoder:
See https://github.com/cython/cython/issues/6484#issuecomment-3376763325

PR #6520, opened 2024-11-25
Accelerate divmod() by calling C ldiv() for C long type integers

unnecessary-username:
Inherited from #6073 , this PR expand the utilization of C functions for divmod() to C long type by calling the C ldiv() function.


Cython/Utility/Builtins.c
```diff
@@ -344,6 +344,51 @@ static CYTHON_INLINE PyObject* __Pyx_divmod_int(int a, int b) {
...
 
 
+//////////////////// divmod_long.proto //////////////////
+
+static CYTHON_INLINE PyObject* __Pyx_divmod_long(long a, long b); /*proto*/
+
+
+//////////////////// divmod_long //////////////////
+
+static CYTHON_INLINE PyObject* __Pyx_divmod_long(long a, long b) {
```
scoder, line 354: Since the code is practically the same as above, except for the types, I think we can use Tempita template generation to avoid the duplication. Look for `TempitaUtilityCode`. You can then easy add `long long` as well.

unnecessary-username: I will have a look on the Tempita template later.
Before submitting this PR, I was more hesitate on how to declare `long long` in the `Cython/Compiler/Builtin.py` file.

user202729:
This is a bad idea. **Calling `div()` is slower than doing a `/` and a `%`, and let the compiler optimizes it away.**

Stack Overflow: https://stackoverflow.com/questions/4565272/why-use-div-or-ldiv-in-c-c

Godbolt: https://godbolt.org/z/Td4jzf9P4

I suggest the other pull request to be reverted too.

scoder:
The idea here is that three Python object allocations (or, eventually, none) are faster than five. Using Python's C-API for a divmod operation requires the input arguments to be converted to Python integers, then call into CPython, which returns a tuple of two integer results. Doing the calculation directly avoids the two allocations of the input arguments and can potentially avoid converting the output arguments as well, if we can determine that they are (or can be) assigned separately as C integers.

So, yes, we can happily use `/` and `%` instead of the `*div()` family of functions, but the rest isn't done without reason.

user202729:
Alright, sound good to me.

Be careful of the case `INT_MIN / (-1)` overflows. Maybe at least make a test case for it.

tests/run/builtin_divmod.py
```diff
@@ -6,23 +6,23 @@ def divmod_regular(a,b):
...
     >>> divmod_regular(10,5)
     (2, 0)
-    >>> divmod_regular(9191,4096)
-    (2, 999)
-    >>> divmod_regular(10000,10010)
-    (0, 10000)
+    >>> divmod_regular(4399,2048)
+    (2, 303)
+    >>> divmod_regular(88321773,98539211)
+    (0, 88321773)
```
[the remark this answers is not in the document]
unnecessary-username: Finally managed to review back this PR before the new year. I have just submitted a commit to address this. Thanks and happy new year!

Cython/Utility/Builtins.c
```diff
@@ -344,6 +344,51 @@ static CYTHON_INLINE PyObject* __Pyx_divmod_int(int a, int b) {
...
+    PyObject *result_tuple = NULL, *pyvalue = NULL;
+    // Python and C/C++ use different algorithms in calculating quotients and remainders.
+    // This results in different answers between Python and C/C++
+    // when the dividend is negative and the divisor is positive and vice versa.
+    long q, r;
+    if ((a < 0 && b > 0) || (a > 0 && b < 0)) {
```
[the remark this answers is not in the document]
unnecessary-username: Finally managed to review back this PR before the new year. I have just submitted a commit to rearrange the conditions. Thanks and happy new year!

Cython/Utility/Builtins.c
```diff
@@ -312,17 +312,21 @@ static CYTHON_INLINE PyObject* __Pyx_divmod_int(int a, int b) {
...
-    if ((a < 0 && b > 0) || (a > 0 && b < 0)) {
+    if (unlikely(b == 0)) {
+        PyErr_SetString(PyExc_ZeroDivisionError, "integer division or modulo by zero");
+        return NULL;
+    }
+    else if (a == 0) {
+        q = 0;
+        r = 0;
+    }
+    else if ((a > 0) != (b > 0)) {
```
scoder, line 323: `b == 0` was probably tested last because it's an exceptional case, to make sure that we jump in the fast path immediately. Can't say how important `a == 0` is. Probably not something to invest time on in the normal case.

scoder: OTOH, your ordering seems perfectly reasonable and fast to me. I'm fine with it.

PR #6545, opened 2024-12-08
coverage plugin base on pyx itself

trim21:
This PR add a pyx file tracker and reporter can work without c or cpp files.

Current we parse pyx code from generated c files which is unnecessary, we can just parse them from pyx file itself.

This maybe the final piece of https://github.com/cython/cython/issues/6186

This is an immature implement and still need some fix, but I want to create a PR first to have some discussion.


scoder:
The advantage of parsing the generated C file is that it only contains source code that actually lead to executable code. That makes decisions about expected and missing code lines easier. It also simplifies include file tracking, pxd file injections, etc., because it only has a single file to look at for the whole module code.

Usually, the generated files can be expected to be available when running tests, because that tends to happen shortly after a (fresh) build. Basically, if the source files are available, why not the generated files?

That said, I'll take a look later today, if I get to it.

scoder:
Generally speaking, this seems incomplete and already adds a lot of complexity. If we can avoid that, I'd rather take a simpler path.

trim21:
Yes, it's still incomplete. I only tested it on single pyx file and not a non-executed lines are excluded.

PR #6592, opened 2024-12-29
PyPy: fix overstrict regex, allow CYTHON_UPDATE_DESCRIPTOR_DOC

mattip:
- Fix over-strict regex in error message check that started failing in tests fro PyPy3.11 which is pre-release. From a discussion in pypy/pypy#5160
  > CPython is unfortunately super inconsistent about when to use the the qualname in error messages, and when to use only the method name

- We went back and forth with enabling `CYTHON_UPDATE_DESCRIPTOR_DOC`, I think this time I fixed it (as of py7.3.17).


mattip:
Thanks!

PR #6630, opened 2025-01-24
Fix LIMITED_VERSION version checks

### LINKED ISSUE #6629: What value are users supposed to give to Py_LIMITED_API?
### Describe your issue

I was testing adding `-DPy_LIMITED_API=0x03090000` to `CFLAGS` to target Python 3.9. Afterwards, I ran the `abi3audit` tool, and noticed something suspicious about a symbol not being used as I expected, so I had to dig deeper. Long story short, I found the following code in Cython utility code:

```c
/* Cython/Utility/ModuleSetupCode.c */
...
// PEP-573: PyCFunction holds reference to defining class (PyCMethodObject)
#if __PYX_LIMITED_VERSION_HEX < 0x030900B1
  #define __Pyx_PyType_FromModuleAndSpec(m, s, b)  ((void)m, PyType_FromSpecWithBases(s, b))
  typedef PyObject *(*__Pyx_PyCMethod)(PyObject *, PyTypeObject *, PyObject *const *, size_t, PyObject *);
#else
  #define __Pyx_PyType_FromModuleAndSpec(m, s, b)  PyType_FromModuleAndSpec(m, s, b)
  #define __Pyx_PyCMethod  PyCMethod
#endif
...
```
And there is the problem, I was defining `Py_LIMITED_API=0x03090000`, but the preprocessor guard checked for `0x030900B1`, and things did not work as I expected, that is, the `PyType_FromModuleAndSpec` API was not being used, despite being available in the Python 3.9 limited API.

Changing my definition to `Py_LIMITED_API=0x030900F0` adding the final (`F`) hex digid, and of course this fixes the problem. I could leave it there assuming it was my fault, and move on.
HOWEVER, to my defense, in the relevant Python documentation about [Py_LIMITED_API](https://docs.python.org/3/c-api/stable.html#c.Py_LIMITED_API) there are two examples where the final ("F") hex digit is not added in the hex version value.
MOREOVER, in CPython header files, all of the preprocessor guards checking for `Py_LIMITED_API` use hex versions like `0x03090000` *without* the final (`F`) hex digit.
THEREFORE, I would argue that other users may experience the same issue I encountered.

Should Cython double check and eventually fix all the guards involving `__PYX_LIMITED_VERSION_HEX` to remove the alpha (`A`) or beta (`B`) hex digits from the hex version strings, following closely what CPython's headers do and what users would expect?
Of course, I understand this rule may not be applied to new stuff from in-development releases like currently Python 3.14, but the fix should go to everything Python 3.13 and older.

dalcinl:
* Replace use of alfa/beta+level markers with 00
* Consistently use uppercase hexadecimal digits

Fixes #6629

### DIFF
--- a/Cython/Utility/Builtins.c
+++ b/Cython/Utility/Builtins.c
@@ -148,7 +148,7 @@ static CYTHON_INLINE PyObject *__Pyx_GetAttr3(PyObject *, PyObject *, PyObject *
 //@requires: Exceptions.c::PyErrFetchRestore
 //@requires: Exceptions.c::PyErrExceptionMatches
 
-#if __PYX_LIMITED_VERSION_HEX < 0x030d00A1
+#if __PYX_LIMITED_VERSION_HEX < 0x030d0000
 static PyObject *__Pyx_GetAttr3Default(PyObject *d) {
     __Pyx_PyThreadState_declare
     __Pyx_PyThreadState_assign
@@ -162,7 +162,7 @@ static PyObject *__Pyx_GetAttr3Default(PyObject *d) {
 
 static CYTHON_INLINE PyObject *__Pyx_GetAttr3(PyObject *o, PyObject *n, PyObject *d) {
     PyObject *r;
-#if __PYX_LIMITED_VERSION_HEX >= 0x030d00A1
+#if __PYX_LIMITED_VERSION_HEX >= 0x030d0000
     int res = PyObject_GetOptionalAttr(o, n, &r);
     // On failure (res == -1), r is set to NULL.
     return (res != 0) ? r : __Pyx_NewRef(d);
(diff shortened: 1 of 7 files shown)


Cython/Utility/Builtins.c
```diff
@@ -148,7 +148,7 @@ static CYTHON_INLINE PyObject *__Pyx_GetAttr3(PyObject *, PyObject *, PyObject *
 //@requires: Exceptions.c::PyErrFetchRestore
 //@requires: Exceptions.c::PyErrExceptionMatches
 
-#if __PYX_LIMITED_VERSION_HEX < 0x030d00A1
+#if __PYX_LIMITED_VERSION_HEX < 0x030D0000
```
scoder, line 151: I chose 'd' here because 'D' looks too close to '0'.

scoder: Similarly, `B` looks pretty close to `0`, too, in some fonts.

dalcinl: No, no my friend... the real problem is you eyes are aging... :rofl: 

OK, now back to being serious: You want me to revert all these uppercase changes? Or maybe make everything lowercase, at least when `__PYX_LIMITED_VERSION_HEX` is involved? CPython headers are not consistent about this admittedly silly matter.

dalcinl: Uppercase changes reverted.

da-woods (approved):

scoder:
Thanks

Issue #6704, opened 2025-03-06
[BUG][3.1] Infinite Loop/OOM in generate_wrapper_functions with a static method override


pilleye:
### Describe the bug

Overriding a static method in a child class causes an infinite loop in `generate_wrapper_functions`, where `entry.prev_entry == entry`.

This issue only exists with 3.1/master and the C++ backend. Not reproducible on 3.0 or the 3.1 with the C backend.

### Code to reproduce the behaviour:

Cherry-pick this commit: https://github.com/pilleye/cython/commit/dbeb85b25a8091ddeaa213d5d9805ef150f316ee

This issue only exists on 3.1/master with the C++ backend:
```sh
python runtests.py -3 --backends cpp -vv static_method_inheritance
```

Minimal repro:
```cython
cdef class Foo:
    @staticmethod
    cdef Foo meth(Foo this):
        return Foo()

cdef class OtherFoo(Foo):
    @staticmethod
    cdef OtherFoo meth(Foo this):
        return OtherFoo()
```

### Expected behaviour

Should either not infinite loop or have validation against unexpected inputs.

### OS

Reproducible on Linux and Mac, don't have access to a Windows machine.

### Python version

3.12 and 3.13

### Cython version

3.1/master

PR #6705, opened 2025-03-07
Fix incorrect code generation for method override

pilleye:
#3235 subtly changed the behavior of `declare` to return a reference to an already declared `Entry` in the case of overriding a parent method. 

This caused `cfunc_entries` to have two entries pointing to the same `Entry` object, whereas `cfunc_entries` is expected to have two different `Entry` objects, one for the parent method that is being inherited, and one of the new override from the child class.

This PR fixes this bug by copying `Entry` instead of returning a reference. 

Closes https://github.com/cython/cython/issues/6704

### DIFF
--- a/tests/run/static_method_inheritance.pyx
+++ b/tests/run/static_method_inheritance.pyx
@@ -1,0 +1,19 @@
+# mode: run
+
+cdef class A:
+    pass
+
+cdef class B(A):
+    pass
+
+cdef class Foo:
+    @staticmethod
+    cdef A meth():
+        return A()
+
+
+cdef class Bar(Foo):
+    @staticmethod
+    cdef B meth():
+        return B()
+
--- a/tests/run/static_method_inheritance.pyx
+++ b/tests/run/static_method_inheritance.pyx
@@ -1,19 +1,36 @@
 # mode: run
 
 cdef class A:
-    pass
+    @staticmethod
+    def moo():
+        return "moo"
 
 cdef class B(A):
-    pass
+    @staticmethod
+    def moo():
+        return "boo"
 
 cdef class Foo:
     @staticmethod
     cdef A meth():
         return A()
 
-
 cdef class Bar(Foo):
     @staticmethod
     cdef B meth():
         return B()
+
+def call_foo():
+    """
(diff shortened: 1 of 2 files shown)


tests/run/static_method_inheritance.pyx
```diff
@@ -0,0 +1,36 @@
...
+
+cdef class A:
+    @staticmethod
+    def moo():
+        return "moo"
+
+cdef class B(A):
+    @staticmethod
+    def moo():
+        return "boo"
```
scoder, line 11: Let's make the test easier to understand and more helpful for debugging in failure cases.
```suggestion
cdef class A:
    @staticmethod
    def moo():
        return "mooA"

cdef class B(A):
    @staticmethod
    def moo():
        return "mooB"
```

tests/run/static_method_inheritance.pyx
```diff
@@ -0,0 +1,36 @@
...
+
+cdef class Foo:
+    @staticmethod
+    cdef A meth():
+        return A()
+
+cdef class Bar(Foo):
+    @staticmethod
+    cdef B meth():
+        return B()
```
scoder, line 21:
```suggestion
cdef class GetBaseA:
    @staticmethod
    cdef A meth():
        return A()

cdef class GetSubB(GetBaseA):
    @staticmethod
    cdef B meth():
        return B()
```

tests/run/static_method_inheritance.pyx
```diff
@@ -0,0 +1,36 @@
...
+    'moo'
+    """
+    return Foo.meth().moo()
+
+def call_bar():
+    """
+    >>> call_bar()
+    'boo'
+    """
+    return Bar.meth().moo()
```
scoder, line 35:
```suggestion
def call_base():
    """
    >>> call_base()
    'mooA'
    """
    return GetBaseA.meth().moo()

def call_sub():
    """
    >>> call_sub()
    'mooB'
    """
    return GetSubB.meth().moo()
```

scoder (commented):
Good naming also helps in tests.

scoder:
Thanks. This looks right to me.

PR #6711, opened 2025-03-10
Fix warning in file extension/mode mismatch

P403n1x87:
We fix a coding error in the warning emission logic that had a wrong variable reference.


P403n1x87:
I have discovered this while trying to compile a `.pyx` source that has `# distutils: language = c++` at the very top into an output file with the `.cpp` extension. The problem does not occur with the latest stable version.

PR #6726, opened 2025-03-14
Fix C compiler warnings

dalcinl:
Includes cherry-pick of #6636.

### DIFF
--- a/Cython/Compiler/Nodes.py
+++ b/Cython/Compiler/Nodes.py
@@ -5036,7 +5036,7 @@ class OverrideCheckNode(StatNode):
             code.error_goto_if_null(func_node_temp, self.pos)))
         code.put_gotref(func_node_temp, py_object_type)
 
-        code.putln("if (!__Pyx_IsSameCFunction(%s, (void*) %s)) {" % (func_node_temp, method_entry.func_cname))
+        code.putln("if (!__Pyx_IsSameCFunction(%s, (void(*)(void)) %s)) {" % (func_node_temp, method_entry.func_cname))
         self.body.generate_execution_code(code)
         code.putln("}")
 
--- a/Cython/Compiler/Nodes.py
+++ b/Cython/Compiler/Nodes.py
@@ -4181,14 +4181,19 @@ class DefNodeWrapper(FuncDefNode):
         if self.starstar_arg or self.star_arg:
             self.generate_stararg_init_code(max_positional_args, code)
 
-        code.putln('{')
         all_args = tuple(positional_args) + tuple(kw_only_args)
         non_posonly_args = [arg for arg in all_args if not arg.pos_only]
-        non_pos_args_id = ','.join(
-            ['&%s' % code.intern_identifier(arg.entry.name) for arg in non_posonly_args] + ['0'])
-        code.putln("PyObject ** const %s[] = {%s};" % (
-            Naming.pykwdlist_cname,
-            non_pos_args_id))
+        accept_kwd_args = non_posonly_args or self.starstar_arg
+
+        code.putln('{')
+        if accept_kwd_args:
+            non_pos_args_id = ','.join([
+                f'&{code.intern_identifier(arg.entry.name)}'
+                for arg in non_posonly_args
+            ] + ['0'])
+            code.putln("PyObject ** const %s[] = {%s};" % (
+                Naming.pykwdlist_cname,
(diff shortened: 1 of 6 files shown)


Cython/Compiler/Nodes.py
```diff
@@ -4184,11 +4184,14 @@ def generate_tuple_and_keyword_parsing_code(self, args, code, decl_code):
...
         all_args = tuple(positional_args) + tuple(kw_only_args)
         non_posonly_args = [arg for arg in all_args if not arg.pos_only]
-        non_pos_args_id = ','.join(
-            ['&%s' % code.intern_identifier(arg.entry.name) for arg in non_posonly_args] + ['0'])
-        code.putln("PyObject ** const %s[] = {%s};" % (
-            Naming.pykwdlist_cname,
-            non_pos_args_id))
+        if non_posonly_args:
+            non_pos_args_id = ','.join([
+                '&%s' % code.intern_identifier(arg.entry.name)
```
scoder, line 4189:
```suggestion
                f'&{code.intern_identifier(arg.entry.name)}'
```

Cython/Compiler/Nodes.py
```diff
@@ -4184,11 +4184,14 @@ def generate_tuple_and_keyword_parsing_code(self, args, code, decl_code):
...
-            Naming.pykwdlist_cname,
-            non_pos_args_id))
+        if non_posonly_args:
+            non_pos_args_id = ','.join([
+                f'&{code.intern_identifier(arg.entry.name)}'
+                for arg in non_posonly_args
+            ] + ['0'])
+            code.putln("PyObject ** const %s[] = {%s};" % (
+                Naming.pykwdlist_cname,
+                non_pos_args_id))
```
h-vetinari, line 4194: This could also use an f-string, just needs to double the `{` to get a literal.
```suggestion
            code.putln(f"PyObject ** const {Naming.pykwdlist_cname}[] = {{{non_pos_args_id}}};")
```

scoder: Yeah, I had considered it but found the triple braces too easy to miss on a quick read and the `{%s}` to be clearer. I already hit a few cases where I found it sad that f-strings don't make the code look clearer, as in this case.

Not sure. Do you vote for it or just propose it as an alternative?

h-vetinari: Not important to me (and I don't pretend to have a vote :)), just thought I'd point it out.

scoder, line 4194: Here is a different proposal:
```suggestion
        if non_posonly_args:
            non_posonly_args_init = '{%s,0}' % ','.join([
                f'&{code.intern_identifier(arg.entry.name)}'
                for arg in non_posonly_args
            ])
            code.putln(
                f"PyObject ** const {Naming.pykwdlist_cname}[] = {non_posonly_args_init};")
```

scoder:
Either way, CI is unhappy because it's also used in other cases, even if it's empty. The condition is really `if accept_kwd_args:`, which is set further down but could just be moved up.

dalcinl:
> CI is unhappy because the code is also used in other cases

Sorry, I did not run the full test suite. I'll try later via `CYTHON_UNUSED` or  `CYTHON_UNUSED_VAR`.

scoder:
> The condition is really `if accept_kwd_args:`, which is set further down but could just be moved up.

I still think this is the right fix.

dalcinl:
> I still think this is the right fix.

Sorry, I missed your comment. I'll look into it.

dalcinl:
@scoder I consolidated all of my warning-related fixes (including the cherry-pick of #6636) in this PR.

Cython/Utility/TypeConversion.c
```diff
@@ -806,23 +806,23 @@ static PyObject* __Pyx_PyUnicode_FromOrdinal_Padded(int value, Py_ssize_t ulengt
 
         char *cpos = chars + sizeof(chars);
         if (value < 0x800) {
-            *--cpos = (char) (0b10000000 | (value & 0x3f));
```
dalcinl, line 809: GCC `-pedantic` warns about binary constants being C23 or GCC extension, so I decided to go old-fashion.

dalcinl:
@scoder Should I merge (via cmd line and push) or you do it?

scoder:
Thanks

Issue #6691, opened 2025-02-23
[ENH] Generate shared utility module in `cythonize()`


matusvalo:
### Is your feature request related to a problem? Please describe.

PR #6531 introduces CLI argument for generating shared module. The workflow using new argument does not integrate well into setuptools machinery because it needs one extra step of calling cython.

### Describe the solution you'd like.

We should integrate generating shared module in `cythonize()` to generate shared utility extention module together with compilation of cython files. Hence, for following code structure:

```
mypkg/
+-- __init__.py
+-- subpkg1/
    +-- __init__.py
    +-- module.pyx
```

we should be able to write following `setup.py` file:

```python
from Cython.Build import cythonize
from setuptools import setup, Extension

extensions = [
    Extension("*", ["**/*.pyx"]),
]

setup(
  ext_modules = cythonize(extensions, shared_utility_qualified_name='mypkg.shared._cyutility')
)
```

build system will produce following structure:

```
mypkg/
+-- __init__.py
+-- shared/
|   +-- __init__.py
|   +-- _cyutility.cpython-311-x86_64-linux-gnu.so  
+-- subpkg1/
    +-- __init__.py
    +-- module.cpython-311-x86_64-linux-gnu.so
```

where:
1. `_cyutility.cpython-311-x86_64-linux-gnu.so` is shared utility extension module automatically generated and compiled by the build system.
2. `module.cpython-311-x86_64-linux-gnu.so` is compiled with `shared_utility_qualified_name` directive set to `mypkg.shared._cyutility`
3. build system will create all necessary directories and `__init__.py` files if they do not exists to ensure that shared utility is importable under provided full qualified module name.
4. build system will automatically compile all extensions passed to `cythonize()` with proper directive to use utility code from shared utility extention module. 
### Describe alternatives you've considered.

N/A

### Additional context

See discussion in https://mail.python.org/pipermail/cython-devel/2025-February/005466.html

PR #6728, opened 2025-03-14
Generate and use shared library for MemoryView Utility code [Context attribute]

matusvalo:
This is variant of #6531 with `shared_utility_qualified_name` as `Context` attribute:

```python
######## setup.py ########

import numpy

extensions = [
    Extension("*", ["**/*.pyx"], include_dirs=[numpy.get_include()]),
    Extension("pkg2.CythonShared", sources=["pkg2/CythonShared.c"])
]

setup(
  ext_modules = cythonize(extensions, shared_utility_qualified_name = 'pkg2.CythonShared')
)
```

> [!NOTE] 
> ~There is missing peace of code for including shared library as dependency.~ Moreover, we should define how we will proceed with #6691 (we are introducing the parameter but the logic for creation of package is not present)


Cython/Compiler/ModuleNode.py
```diff
@@ -177,7 +177,7 @@ def extend_if_not_in(L1, L2):
             # Ensure that we don't generate import code for these entries!
             for entry in scope.c_class_entries:
                 entry.type.module_name = self.full_module_name
-                entry.type.scope.directives["internal"] = True
+                entry.type.scope.directives["internal"] = internalise_c_class_entries
 
             self.scope.merge_in(scope)
```
scoder, line 182: `merge_scope=True` seems to do something entirely separate. It should be a separate function. And `internalise_c_class_entries()` could also be a separate function that gets called in the appropriate place(s).

scoder (commented):
Looks good to me, overall. As I said, how exactly we enable the flag in `Context` is a matter of taste, not obligation.

Cython/Compiler/MemoryView.py
```diff
@@ -835,26 +845,68 @@ def use_cython_array_utility_code(env):
...
+                    overlapping_utility,
+                    copy_contents_new_utility,
+                    ],
+    )
+    memviewslice_declare_code.requires.append(memoryview_utility_code)
+    copy_contents_new_utility.requires.append(memoryview_utility_code)
+    return memoryview_utility_code, memviewslice_init_code
+
+@Utils.cached_function
+def _get_memoryview_shared_utility_code(shared_utility_qualified_name):
```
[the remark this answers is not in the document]
scoder: I just noticed that this doesn't work. Passing the `context` dict misfits the `@cached_function`. Actually, even using it seems already wrong. I'd remove the decorator.

scoder (commented):
On a second look, it seems that you're passing a whole context argument in several places that don't need the complete set of information. If it's enough to pass the shared module name, then pass the shared module name instead of the complete context. If there's more to pass, then it depends. But currently, both cases don't seem separated.

Additionally, the term `context` is used for different things all across the code, also in this PR. That's unhelpful for the understanding of the code.

Cython/Compiler/ModuleNode.py
scoder: As far as I can tell, we are only using this feature in utility code and not pxd injected code. Isn't this something that `UtilityCode.get_tree()` could do already?

Cython/Compiler/ExprNodes.py
```diff
@@ -5511,7 +5515,11 @@ def analyse_as_type(self, env):
             if not self.start and not self.stop:
                 # memory view
                 from . import MemoryView
-                env.use_utility_code(MemoryView.view_utility_code)
+                env.use_utility_code(
+                    MemoryView.get_view_utility_code(
+                        env.shared_utility_qualified_name
```
scoder, line 5520:
```suggestion
                        env.context.shared_utility_qualified_name
```

Cython/Compiler/MemoryView.py
```diff
@@ -756,7 +756,7 @@ def _resolve_NameNode(env, node):
     except AttributeError:
         raise CompileError(node.pos, INVALID_ERR)
 
-    viewscope = env.global_scope().context.cython_scope.viewscope
+    viewscope = env.context.cython_scope.viewscope
```
scoder, line 759: This is unrelated to this PR, but – do we really need to go through the context for this? Is there no direct way just through the scopes? :-o

Cython/Build/SharedModule.py
```diff
@@ -0,0 +1,74 @@
...
+from Cython.Compiler.StringEncoding import EncodedString
+from Cython.Compiler.Scanning import FileSourceDescriptor
+
+def create_shared_library_pipeline(context, scope, options, result):
+
+    def generate_tree_factory(context):
+        def generate_tree(compsrc):
+
+            # Force to generate __Pyx_ExportFunction
+            Options.cimport_from_pyx = True
```
scoder, line 17: As a simple mitigation for this global setting, we should at least reset the original value after our use.

Cython/Compiler/Main.py
```diff
@@ -92,7 +94,7 @@ def __init__(self, include_directories, compiler_directives, cpp=False,
     @classmethod
     def from_options(cls, options):
         return cls(options.include_path, options.compiler_directives,
-                   options.cplus, options.language_level, options=options)
+                   options.cplus, options.language_level, options.shared_utility_qualified_name, options=options)
```
scoder, line 97: Seeing this, it really makes no sense. The `Context` knows its `.options` and we're only duplicating the option into the `Context` here.  Just go through the options then. That will mean that several places that now pass the `Context` down into a function can instead pass only `context.options`. Again, less dependencies.

scoder:
Regarding the CLI, I think both alternatives would work, and both are probably fine. However, yours is already implemented and my proposal requires some more details to be clarified. So, I'm fine with keeping the CLI as it now is.

> But I can rename that variable to `template_context`

Sounds good to me.

> I am not sure whether there is other code which can replace new `merge_scope()` function...

Do you mean other code that could use it? Doesn't matter. It's a good as it is and as it is used.

Cython/Compiler/CmdLine.py
```diff
@@ -158,6 +157,11 @@ def create_cython_argparser():
                            'deduced from the import path if source file is in '
                            'a package, or equals the filename otherwise.')
     parser.add_argument('-M', '--depfile', action='store_true', help='produce depfiles for the sources')
+    parser.add_argument("--generate-shared", dest='shared_c_file_path', action='store', type=str,
+                        help='Generates shared module with specified name.')
+    parser.add_argument("--shared", dest='shared_utility_qualified_name', action='store', type=str,
+                        help='Imports utility code from shared module specified by fully qualified module name.')
```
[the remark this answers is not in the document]
scoder: > rewriting whole cython CLI into subcommands

That would be a major disruption for users. I don't think we can do that.

> I don't think we need users to specify a C file path. That just seems redundant.

I'm fine with keeping the two CLI options.

Cython/Build/SharedModule.py
```diff
@@ -0,0 +1,63 @@
...
+
+    # "cimport_from_pyx=True" to force generating __Pyx_ExportFunction
+    parse = Pipeline.parse_stage_factory(context, cimport_from_pyx=True)
+
+    def generate_tree_factory(context):
+        def generate_tree(compsrc):
+            tree = parse(compsrc)
+
+            scope.use_utility_code(
+                    MemoryView.get_view_utility_code(context.shared_utility_qualified_name))
```
scoder, line 21:
```suggestion
            scope.use_utility_code(
                MemoryView.get_view_utility_code(options.shared_utility_qualified_name))
```

Cython/Compiler/Pipeline.py
```diff
@@ -45,7 +45,7 @@ def parse(source_desc):
         return tree
     return parse
 
-def generate_pyx_code_stage_factory(options, result, cimport_from_pyx=False):
+def generate_pyx_code_stage_factory(options, result, cimport_from_pyx=Options.cimport_from_pyx):
```
scoder, line 48: This nails the option at import time, regardless of the value at runtime. If users change it after importing Cython, this will fail to produce the intended result.

It's apparently needed across the whole compiler run, from parse to code generation. Thus, it shouldn't be set just here and during parsing. We should change the setting for the whole run of the `SharedModule` pipeline. You can add a pipeline function as reset step after the code generation.

Cython/Build/SharedModule.py
```diff
@@ -27,15 +26,23 @@ def generate_tree(compsrc):
 
         return generate_tree
 
+    def set_cimport_from_pyx(cimport_from_pyx):
+        def inner(node):
+            Options.cimport_from_pyx = cimport_from_pyx
+            return node
+        return inner
```
scoder, line 33: Almost. We still need to make sure to reset the option to its original value in the end.

```suggestion

    orig_cimport_from_pyx = Options.cimport_from_pyx

    def set_cimport_from_pyx(cimport_from_pyx):
        def inner(node):
            Options.cimport_from_pyx = cimport_from_pyx
            return node
        return inner
```

Cython/Build/SharedModule.py
```diff
@@ -27,15 +26,23 @@ def generate_tree(compsrc):
...
         generate_tree_factory(context),
         *Pipeline.create_pipeline(context, 'pyx', exclude_classes=()),
         Pipeline.inject_pxd_code_stage_factory(context),
         Pipeline.inject_utility_code_stage_factory(context, internalise_c_class_entries=False),
         Pipeline.inject_utility_pxd_code_stage_factory(context),
         Pipeline.abort_on_errors,
-        # "cimport_from_pyx=True" to force generating __Pyx_ExportFunction
-        Pipeline.generate_pyx_code_stage_factory(options, result, cimport_from_pyx=True),
+        Pipeline.generate_pyx_code_stage_factory(options, result),
+        set_cimport_from_pyx(False),
```
scoder, line 45:
```suggestion
        set_cimport_from_pyx(orig_cimport_from_pyx),
```

scoder:
Thanks, @matusvalo. I know that this took a while to get accepted in the end, so, thanks for pushing all the way through to get a very long-standing feature request solved. Even the ticket, long after the first discussions, dates back to 2018.

PR #6738, opened 2025-03-18
Avoid a performance panelty in CPython 3.12/13 when creating a StopIteration

scoder:
Turns out that CPython<3.14 doesn't benefit from `__Pyx_PyObject_CallOneArg()` for exceptions because `BaseException` does not implement vectorcall yet.

See https://github.com/cython/cython/issues/6710


scoder:
Looking at the latest benchmark runs on 3.12 and 3.13, there does seem to be a positive effect in the relevant benchmarks.

PR #6748, opened 2025-03-21
Test coercing Python objects to/from `string_view`

jakirkham:
Add a test to try coercing Python objects (like `bytes` and typed-memoryviews) to and from `std::string_view`.


tests/run/cpp_stl_string_view.pyx
```diff
@@ -35,6 +35,30 @@ def test_return_to_py():
...
+    """
+    if not isinstance(py_str, bytes):
+        return
+
+    cdef string_view sv1 = py_str
+    assert len(py_str) == sv1.size()
+    assert <const char*>py_str == sv1.data()
+
+    cdef const char[:] mv1 = sv1
+    assert len(mv1) = sv1.size()
```
da-woods, line 51:
```suggestion
    assert len(mv1) == sv1.size()
```

da-woods (commented):
See also #6655 (where I try to apply the same rules as for `const char*`). I don't think that contradicts anything here though

tests/run/cpp_stl_string_view.pyx
```diff
@@ -35,6 +35,30 @@ def test_return_to_py():
...
+    assert <const char*>py_str == sv1.data()
+
+    cdef const char[:] mv1 = sv1
+    assert len(mv1) == sv1.size()
+    if py_str:
+        assert &mv1[0] == sv1.data()
+
+    cdef const char[:] mv2 = py_str
+    cdef string_view sv2 = mv2
+    assert len(mv2) = sv2.size()
```
jakirkham, line 57:
```suggestion
    assert len(mv2) == sv2.size()
```

jakirkham:
Thanks for the pointers and suggestions! 🙏

Having better error checking sounds like a great idea

Wanted to cover these in the tests as passing a `bytes` object straight into lower-level C++ without copying (thanks to `string_view`) would satisfy a recurring request over the years

Given how it easy it is to do, agree it is important to have clear guidance in the form of compiler errors when using temporaries

tests/run/cpp_stl_string_view.pyx
```diff
@@ -35,6 +35,30 @@ def test_return_to_py():
...
+    >>> test_coercion_py(b_asdf_5)
+    """
+    if not isinstance(py_str, bytes):
+        return
+
+    cdef string_view sv1 = py_str
+    assert len(py_str) == sv1.size()
+    assert <const char*>py_str == sv1.data()
+
+    cdef const char[:] mv1 = sv1
```
jakirkham, line 50: Looks like this case fails

Was hoping it would do something equivalent to

```cython
cdef const char[:] mv1 = <char[:sv1.length():1]><const char*>sv1.data()
```

However it looks like this doesn't work for `std::string` either. Plus this is a non-owning instance. So it makes sense why we might want to add more hoops to jump through for this case

tests/run/cpp_stl_string_view.pyx
```diff
@@ -35,6 +35,25 @@ def test_return_to_py():
...
 
+def test_coercion_py(py_str):
+    """
+    >>> b_asdf_5 = b_asdf * 5
+    >>> test_coercion_py(b_asdf_5)
+    """
+    if not isinstance(py_str, bytes):
+        return
+
+    cdef string_view sv1 = py_str
```
jakirkham, line 46: Looks like this doesn't work out of the box. FWIW `std::string` does

Maybe we can try typing it as `bytes`

```suggestion
    cdef string_view sv1 = <bytes>py_str
```

tests/run/cpp_stl_string_view.pyx
da-woods: > `cdef const char[:] mv1 = <char[:sv1.length():1]><const char*>sv1.data()`

Shouldn't that be

```
cdef const char[:] mv1 = <const char[:sv1.length():1]><const char*>sv1.data()
```

?

I think it's reasonable to jump through hoops for this and strings because the memoryview can't own them

jakirkham: Both seemed to work, but agree `const char` would be better

Excluding this case makes sense. Didn't quite follow the last bit

tests/run/cpp_stl_string_view.pyx
```diff
@@ -35,6 +35,25 @@ def test_return_to_py():
...
+    """
+    if not isinstance(py_str, bytes):
+        return
+
+    cdef string_view sv1 = <bytes>py_str
+    assert len(py_str) == sv1.size()
+    assert <const char*>py_str == sv1.data()
+
+    cdef const char[:] mv = py_str
+    cdef string_view sv2 = mv
```
jakirkham, line 51: Looks like the last assignment breaks down on the C++ side

Wondering if we need to extend some existing Cython logic

This in particular stick out, but maybe there are other code paths worth looking at

https://github.com/cython/cython/blob/bf59121ba5c9f79b4bbfeab05407c15e07e4cb15/Cython/Utility/CppConvert.pyx#L4-L35

jakirkham:
Huzzah! 🎉

This new test now passes 😄

David could you please take another look?

tests/run/cpp_stl_string_view.pyx
```diff
@@ -35,6 +36,19 @@ def test_return_to_py():
...
+    """
+    >>> b_asdf_5 = b_asdf * 5
+    >>> test_coercion_py(b_asdf_5)
+    """
+    if not isinstance(py_str, bytes):
+        return
+
+    cdef string_view sv1 = <bytes>py_str
+    assert len(py_str) == sv1.size()
+    assert <uintptr_t><const char*>py_str == <uintptr_t>(sv1.data())
```
da-woods, line 49:
```suggestion
    assert <const char*>py_str == sv1.data()
```

I don't think going through uintptr_t adds anything here

tests/run/cpp_stl_string_view.pyx
```diff
@@ -4,6 +4,7 @@
 
 cimport cython
 
+from libc.stdint cimport uintptr_t
```
da-woods, line 7:
```suggestion
```

da-woods (commented):
I've made a brief simplification. Other than that I'm happy to have a test.

tests/run/cpp_stl_string_view.pyx
jakirkham: So the intent was to make sure the pointer addresses are compared as opposed to the strings' contents (`strcmp`)

Cython tends to treat `uintptr_t` more like an integer where `const char*` tends to be treated like a string. So was trying to ensure the former through the additional cast

da-woods: Ah fair enough. I don't think that's necessary in this case, but being explicit about the intent seems reasonable so let's go with that.

jakirkham:
Thanks David! 🙏

Issue #6750, opened 2025-03-21
[ENH]


jakirkham:
### Is your feature request related to a problem? Please describe.

Cython currently handles converting a few Python `bytes`-like types (`bytes` & `bytearray`) to C++'s `std::string`. However some other objects like typed-memoryviews (`const char[:]`) can't be compiled. Even though both `bytes` and `bytearray` can be passed into functions that have an appropriate typed-memoryview argument. To make it easier to work with other `bytes`-like objects that have a byte-based type (`char`, `unsigned char`, etc.), it would be desirable to also add conversions for these types to `std::string`. IOW allowing the second case below to work as well

```cython
from libcpp.string cimport string as std_string


cpdef bytes_to_std_string(bytes b):
    cdef std_string s = b  # works


cpdef chars_to_std_string(const char[:] b):
    cdef std_string s = b  # cython error
```

jakirkham:
Related `object`s can be converted at runtime to `std::string` for certain types. However a similar issue can be seen as above

```cython
from libcpp.string cimport string as std_string


cpdef object_to_std_string(o):
    cdef std_string s = o
```

```python
object_to_std_string(b"abc")                # works
object_to_std_string(bytearray(b"abc"))     # works
object_to_std_string(memoryview(b"abc"))    # fails
```

It would be nice to have the last case also work as long as the is some kind of byte (`char`, `unsigned char`, `uint8_t`, etc.)

scoder:
I agree that this is desirable. This can be implemented in
https://github.com/cython/cython/blob/bf59121ba5c9f79b4bbfeab05407c15e07e4cb15/Cython/Utility/CppConvert.pyx#L4-L16

jakirkham:
Would this also make sense for a C-string? IOW the last case below

```cython
# cython: language_level=3


cpdef bytes_to_c_string(bytes b):
    cdef const char* s = b  # works


cpdef chars_to_c_string(const char[:] b):
    cdef const char* s = b  # cython error
```

scoder:
Hmm. It's easy for the bytes case because the lifetime of the pointer is tied to the bytes object owning the data. I guess it's similar for the memoryview where the pointer would be tied to the lifetime of the view (not the buffer owner).

OTOH, we already have `s = &b[0]` which is commonly used. I consider it less obvious than a simple assignment, although once you know it, it's not difficult to remember. If we allow simple assignments, we'd have two ways to do it. Not sure if that's necessary.

user202729:
[...]

PR #6752, opened 2025-03-21
[WIP] Handle `bytes`-like objects in C-string conversion

jakirkham:
Fixes https://github.com/cython/cython/issues/6750

In `__Pyx_PyObject_AsStringAndSize` handle `bytes`-like objects (such as `const char[:]`) by adding support for [the Python Buffer Protocol]( https://docs.python.org/3/c-api/buffer.html ).

### DIFF
--- a/Cython/Utility/TypeConversion.c
+++ b/Cython/Utility/TypeConversion.c
@@ -290,13 +290,24 @@ static CYTHON_INLINE const char* __Pyx_PyObject_AsStringAndSize(PyObject* o, Py_
         return PyByteArray_AsString(o);
 #endif
 
-    } else
-    {
+    } elif(PyBytes_Check(o)) {
         char* result;
         int r = PyBytes_AsStringAndSize(o, &result, length);
         if (unlikely(r < 0)) {
             return NULL;
         } else {
+            return result;
+        }
+    } else {
+        char* result;
+        Py_buffer view;
+        int r = PyObject_GetBuffer(o, &view, PyBUF_SIMPLE);
+        if (unlikely(r < 0)) {
+            return NULL;
+        } else {
+            result = (char*) view.buf;
+            *length = view.len;
+            PyBuffer_Release(view);
             return result;
         }
     }


Cython/Utility/TypeConversion.c
```diff
@@ -290,15 +290,26 @@ static CYTHON_INLINE const char* __Pyx_PyObject_AsStringAndSize(PyObject* o, Py_
...
         char* result;
         int r = PyBytes_AsStringAndSize(o, &result, length);
         if (unlikely(r < 0)) {
             return NULL;
         } else {
             return result;
         }
+    } else {
+        char* result;
+        Py_buffer view;
```
jakirkham, line 303: Seeing some compile error suggesting `Py_buffer` is unavailable in some builds. Guessing this is related to Python Limited API support. To fix that looks like we want something like this

https://github.com/cython/cython/blob/bf59121ba5c9f79b4bbfeab05407c15e07e4cb15/Cython/Utility/StringTools.c#L767-L799

jakirkham: Looking at that code, it doesn't adapt that well to here. The idea of `__Pyx_PyObject_AsStringAndSize` is to get a pointer to underlying memory in the object. However the other code creates a temporary `bytes` object, whose memory it uses within the lifespan of that temporary. Given `__Pyx_PyObject_AsStringAndSize` has no real way to keep temporary objects (like `bytes`) alive, passing pointers to said temporary objects back to users is UB. So we can't use that

Maybe there are some other options I'm overlooking. Though a simple solution would just be to error in the Python Limited API case

scoder: > keep temporary objects (like `bytes`) alive

This also applies to buffers, though. The buffer view is only guaranteed to be usable until we call `PyBuffer_Release()`. After that, the buffer is allowed to become unavailable and should not be used any more.

Buffers need to be handled in place, where and as long as they are used.

jakirkham: Yep agreed, which we do in this code currently

Am more meaning I don't see a good fallback for the Limited API on Python pre-3.11. Though could just be missing alternatives

If you have suggestions, please let me know 🙂

jakirkham: Or actually does your comment ( https://github.com/cython/cython/issues/6750#issuecomment-2744947269 ) mean we should consider moving this code from `__Pyx_PyObject_AsStringAndSize` to the `string.from_py` functions?

Because then we could use the same fallback mechanism

scoder: > consider moving this code from `__Pyx_PyObject_AsStringAndSize` to the `string.from_py` functions

Yes. It's safe to do there but not in `__Pyx_PyObject_AsStringAndSize`.

da-woods: The best general solution is probably `https://docs.python.org/3/c-api/arg.html#c.PyArg_ParseTuple` with `z#`,

If you look at the implementation of that it allows anything that implements the buffer protocol but doesn't specifically have `bf_releasebuffer` (which implies that the buffer is good so long as the object exists). That's "borrowed" but no worse than what we have for bytes.

jakirkham: That's a good suggestion!

Haven't used the `PyArg_*Parse*` functions much before

Tried rewriting this below

Cython/Utility/TypeConversion.c
```diff
@@ -290,15 +290,22 @@ static CYTHON_INLINE const char* __Pyx_PyObject_AsStringAndSize(PyObject* o, Py_
...
         char* result;
         int r = PyBytes_AsStringAndSize(o, &result, length);
         if (unlikely(r < 0)) {
             return NULL;
         } else {
             return result;
         }
+    } else {
+        char* result;
+        int r = PyArg_Parse(o, "y#", &result, &length);
```
jakirkham, line 303: Used [`PyArg_Parse`]( https://docs.python.org/3/c-api/arg.html#c.PyArg_Parse ) as it looks like that can take a single value

Also from [the "strings and buffers" section]( https://docs.python.org/3/c-api/arg.html#strings-and-buffers ) used `y#` as it looks like it is limited to `bytes`-like objects where a pointer and size can be returned

jakirkham: Should add was able to test this out locally and it worked ok

tests/run/cpp_stl_string_ascii_auto_encoding_str.pyx
```diff
@@ -20,7 +20,7 @@ def test_conversion(py_obj):
     True
     >>> test_conversion(123)  # doctest: +ELLIPSIS
     Traceback (most recent call last):
-    TypeError: expected ..., int found
+    TypeError: a bytes-like object is required, not 'int'
```
jakirkham, line 23: Looks like the syntax of the traceback message changes to look more like this

Are we ok with that? If so, can push similar changes to the test suite

Alternatively we can change the error messaging in the function

Cython/Utility/TypeConversion.c
scoder: I was generally unimpressed by the performance of the `PyArg_*()` functions and tend to avoid them if possible. Not sure if it makes a difference here, given that the most important cases are already handled up-front and we end up asking the owner to fill our buffer struct etc. These things usually tend to harm smaller data buffers more than larger ones.

If it turns out to be a visible factor, we could try to inline the relevant code for `y#` specifically. Cython has helper macros to look up slots in both the normal and Limited C-API.

tests/run/cpp_stl_string_ascii_auto_encoding_str.pyx
scoder: We try to avoid also error message changes if it's somewhat reasonable, but try to follow what CPython outputs (at the Python level).

Cython/Utility/TypeConversion.c
da-woods: The only thing there is hat you can't use `PyType_GetSlot` to "backdate" support in the Limited API so I think `PyArg_Parse` may be the only option there. Agree with backdating it elsewhere.

tests/run/cpp_stl_string_ascii_auto_encoding_str.pyx
jakirkham: The change in the error message is a result of the fact that the `else` case was handling `bytes` objects. Now it is handling any `bytes`-like object

This matches what one would get from `memoryview`. For example

```python
In [1]: memoryview(1)
Cell In[1], line 1
----> 1 memoryview(1)

TypeError: memoryview: a bytes-like object is required, not 'int'
```

Not sure what the equivalent of `PyBytes_AsStringAndSize` would have been in Python

<hr>

If we still want the same error message as before, probably the simplest option is to move the `bytes` handling case back into the `else`. So we would check for a non-`bytes` object for our new case in `else if`

Cython/Utility/TypeConversion.c
jakirkham: Tried the following approaches

```cython
from cpython.getargs cimport PyArg_Parse
from cpython.buffer cimport (
    PyBUF_SIMPLE, Py_buffer,
    PyObject_GetBuffer, PyBuffer_Release,
)


cpdef noop_to_c_string(b):
    pass


cpdef pybuf_to_c_string(b):
    cdef Py_buffer pb
    cdef const char* s
    cdef Py_ssize_t n
    try:
        PyObject_GetBuffer(b, &pb, PyBUF_SIMPLE)
        s = <const char*>pb.buf
        n = pb.len
    finally:
        PyBuffer_Release(&pb)


cpdef pyarg_to_c_string(b):
    cdef const char* s
    cdef Py_ssize_t n
    PyArg_Parse(b, "y#", &s, &n)
```

Then used the string `b"abc"` to get the following results

```python
%timeit noop_to_c_string(b)
15.3 ns ± 0.0764 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)

%timeit pybuf_to_c_string(b)
18.7 ns ± 0.259 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)

%timeit pyarg_to_c_string(b)
31.3 ns ± 0.31 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

So agree it is slower to use `PyArg_Parse` and is probably best saved as a fallback for the limited API on Python pre-3.11

As `PyObject_GetBuffer`, is already calling the slot functions itself and a Python could be compiled with build time optimizations, likely that is about as fast as gets

Can add a compile time branch in this `else` to capture these two cases

jakirkham: Actually another option would be to use [the old-style buffer protocol]( https://docs.python.org/3.12/c-api/objbuffer.html#old-buffer-protocol ). This is part of the stable API and was around until Python 3.12

If we did this we could call [`PyObject_AsCharBuffer`]( https://docs.python.org/3.12/c-api/objbuffer.html#c.PyObject_AsCharBuffer ) like so

```suggestion
        int r = PyObject_AsCharBuffer(o, &result, &length);
```

tests/run/cpp_stl_string_ascii_auto_encoding_str.pyx
scoder: I think the `expected x, y found` message is more suitable for one concrete expected type. It doesn't fit if multiple types are acceptable. The `bytes-like object` message seems better adapted to that case. Feel free to change it.

Cython/Utility/TypeConversion.c
scoder: > AFAICT it doesn't check whether releasing the buffer is required. It just always releases the buffer

But that doesn't make it any better than what the current master branch does, right?

jakirkham: For now have added `Py_buffer` approach for most cases (non-Limited API on any Python version or Python 3.11+ w/Limited API). So that ensures the bulk of cases get the more performant approach. Also added a check that the slot `bf_releasebuffer` is `NULL` and raised an error if not

For all other cases (so Limited API on 3.8, 3.9 or 3.10), we fallback to `PyArg_Parse`. It is slower, but it still provides the new behavior in those cases. Should add [the `PyArg_Parse` case does check that the buffer does not need to be released]( https://github.com/python/cpython/blob/8a00c9a4d2ce9d373b13f8f0a2265a65f4523293/Python/getargs.c#L1245-L1250 )

tests/run/cpp_stl_string_ascii_auto_encoding_str.pyx
jakirkham: Included a custom error at the end. Updated any failing case to `goto` it

All successful cases `return` avoiding it entirely

Cython/Utility/TypeConversion.c
```diff
@@ -286,20 +291,61 @@ static CYTHON_INLINE const char* __Pyx_PyObject_AsStringAndSize(PyObject* o, Py_
...
+
+    error:
+        // Clear existing errors and return our standardized one
+        PyErr_Clear();
+        __Pyx_TypeName result_type_name = __Pyx_PyType_GetFullyQualifiedName(Py_TYPE(result));
+        PyErr_Format(PyExc_TypeError,
+            "a bytes-like object is required, not '" __Pyx_FMT_TYPENAME "'",
+            result_type_name);
+        *length = -1;
+        return NULL;
```
jakirkham, line 348: Added a generic error for all cases to use here. Please let me know what you think about this and if there are things we want to improve about it

Cython/Utility/TypeConversion.c
jakirkham: Have changed the logic to use `Py_buffer` where possible

Now `PyArg_Parse` is used as fallback

Cython/Utility/TypeConversion.c
```diff
@@ -286,20 +291,61 @@ static CYTHON_INLINE const char* __Pyx_PyObject_AsStringAndSize(PyObject* o, Py_
...
+        char* result;
+        int r;
+#if !CYTHON_COMPILING_IN_LIMITED_API || __PYX_LIMITED_VERSION_HEX >= 0x030B0000
+        // Check that `o`:
+        // * Supports the Python Buffer Protocol
+        // * There is no need to keep the buffer around (as we cannot return it)
+        PyBufferProcs *pb = Py_TYPE(o)->tp_as_buffer;
+        if (pb == NULL || pb->bf_getbuffer == NULL || pb->bf_releasebuffer != NULL) {
+            goto error;
+        }
```
jakirkham, line 316: Have added this guard to only use the Python Buffer Protocol in either non-Limited API builds or Limited API builds on Python 3.11+. In those cases this path should be ok

Also added an explicit check that the Python Buffer Protocol does work on this object. This check can only be done under the conditions laid out previously. However this does help us pick out any cases where the object doesn't support the Python Buffer Protocol or it does, but depends on releasing the buffer after

Please let me know if there are other cases we should check for or if this needs tweaking

Cython/Utility/TypeConversion.c
```diff
@@ -286,20 +291,61 @@ static CYTHON_INLINE const char* __Pyx_PyObject_AsStringAndSize(PyObject* o, Py_
...
+
+        // Try to acquire buffer from `o`
+        Py_buffer view;
+        r = PyObject_GetBuffer(o, &view, PyBUF_SIMPLE);
+        if (unlikely(r < 0)) {
+            goto error;
+        } else {
+            result = (char*) view.buf;
+            *length = view.len;
+            PyBuffer_Release(&view);  // Release to clean up buffer (decref object `o`)
```
jakirkham, line 326: We still release the buffer as it does some cleanup of the buffer object itself (namely [decref-ing the associated object]( https://github.com/python/cpython/blob/52b5eb95b770fa00ebbd449ba40cab4a0e7c7df7/Objects/abstract.c#L817-L818 ))

Thought it better to call the standard function here instead of bypassing it by doing our own version of this cleanup

Cython/Utility/TypeConversion.c
```diff
@@ -276,30 +276,81 @@ static CYTHON_INLINE const char* __Pyx_PyUnicode_AsStringAndSize(PyObject* o, Py
...
+
+    error:
+        // Clear existing errors and return our standardized one
+        PyErr_Clear();
+        __Pyx_TypeName o_type_name = __Pyx_PyType_GetFullyQualifiedName(Py_TYPE(o));
+        PyErr_Format(PyExc_TypeError,
+            "a bytes-like object is required, not '" __Pyx_FMT_TYPENAME "'",
+            o_type_name);
+        *length = -1;
+        return NULL;
```
jakirkham, line 353: Added a generic error for all cases to use here. Please let me know what you think about this and if there are things we want to improve about it

Cython/Utility/TypeConversion.c
```diff
@@ -276,30 +276,87 @@ static CYTHON_INLINE const char* __Pyx_PyUnicode_AsStringAndSize(PyObject* o, Py
...
+            PyExc_RuntimeWarning,
+            "Cython: Pointer taken from resizable `bytearray` object. "
+            "Consider using `unsigned char[::1]` instead. "
+            "This will also ensure the underlying `bytearray` is not resized while held."
+        );
+        if (unlikely(r < 0)) {
+            return NULL;
+        } else {
+            return result;
+        }
```
jakirkham, line 299: Tried to add a warning in relation to issue: https://github.com/cython/cython/issues/6760

Since we are providing an alternative path, which would avoid that bug here, it makes sense to encourage its use

Please let me know what you think

Cython/Utility/TypeConversion.c
```diff
@@ -276,30 +276,86 @@ static CYTHON_INLINE const char* __Pyx_PyUnicode_AsStringAndSize(PyObject* o, Py
...
+        int r = PyErr_Warn(
+            PyExc_RuntimeWarning,
+            "Cython: Pointer taken from resizable `bytearray` object. "
+            "To restrict resizing instead use `unsigned char[::1]`."
+        );
+        if (unlikely(r < 0)) {
+            return NULL;
+        } else {
+            return result;
+        }
```
jakirkham, line 298: Tried to add a warning in relation to issue: https://github.com/cython/cython/issues/6760

Since we are providing an alternative path, which would avoid that bug here, it makes sense to encourage its use

Please let me know what you think

Cython/Utility/TypeConversion.c
```diff
@@ -276,30 +276,86 @@ static CYTHON_INLINE const char* __Pyx_PyUnicode_AsStringAndSize(PyObject* o, Py
...
     } else
 #endif
 
     if (PyByteArray_Check(o)) {
+        int r = PyErr_Warn(
+            PyExc_RuntimeWarning,
+            "Cython: Pointer taken from resizable `bytearray` object. "
+            "To restrict resizing instead use `unsigned char[::1]`."
+        );
```
jakirkham, line 293: Tried to add a warning in relation to issue: https://github.com/cython/cython/issues/6760

Since we are providing an alternative path, which would avoid that bug here, it makes sense to encourage its use

Please let me know what you think

Cython/Utility/TypeConversion.c
```diff
@@ -276,37 +276,88 @@ static CYTHON_INLINE const char* __Pyx_PyUnicode_AsStringAndSize(PyObject* o, Py
 static CYTHON_INLINE const char* __Pyx_PyObject_AsStringAndSize(PyObject* o, Py_ssize_t *length) {
 #if __PYX_DEFAULT_STRING_ENCODING_IS_ASCII || __PYX_DEFAULT_STRING_ENCODING_IS_UTF8
     if (PyUnicode_Check(o)) {
-        return __Pyx_PyUnicode_AsStringAndSize(o, length);
+        const char* result = __Pyx_PyUnicode_AsStringAndSize(o, length);
+        if (unlikely(result == NULL)) {
+            goto error;
+        } else {
+            return result;
+        }
```
scoder, line 284: We usually just put simple error gotos into one line. And I prefer keeping the normal case separate from the error case, not joined with `else`. The same applies to other cases further down.
```suggestion
        if (unlikely(!result)) goto error;
        return result;
```

Cython/Utility/TypeConversion.c
```diff
@@ -276,37 +276,88 @@ static CYTHON_INLINE const char* __Pyx_PyUnicode_AsStringAndSize(PyObject* o, Py
...
     } else
 #endif
 
     if (PyByteArray_Check(o)) {
+        int r = PyErr_WarnEx(
+            PyExc_RuntimeWarning,
+            "Cython: Pointer taken from resizable `bytearray` object. "
+            "To restrict resizing instead use `unsigned char[::1]`.",
```
scoder, line 292: At first read, this seems to suggest using `::1` specifically to solve this problem, but that has nothing to do with it.
```suggestion
        int r = PyErr_WarnEx(
            PyExc_RuntimeWarning,
            "Cython: Pointer taken from resizable `bytearray` object. "
            "To restrict resizing instead use a memory view, e.g. `unsigned char[::1]`.",
```

Cython/Utility/TypeConversion.c
```diff
@@ -276,37 +276,88 @@ static CYTHON_INLINE const char* __Pyx_PyUnicode_AsStringAndSize(PyObject* o, Py
...
             return result;
         }
     } else {
         char* result;
-        int r = PyArg_Parse(o, "y#", &result, &length);
+        int r;
+#if !CYTHON_COMPILING_IN_LIMITED_API || __PYX_LIMITED_VERSION_HEX >= 0x030B0000
+        // Check that `o`:
+        // * Supports the Python Buffer Protocol
+        // * There is no need to keep the buffer around (as we cannot return it)
```
scoder, line 321:
```suggestion
        // Check that `o`:
        // * supports the Python Buffer Protocol
        // * does not require keeping the buffer around (as we cannot return it)
```

Cython/Utility/TypeConversion.c
scoder: If you want this to work in the Limited API, you need to use `__Pyx_PyType_GetSlot()`.

Cython/Utility/TypeConversion.c
scoder: > > consider moving this code from `__Pyx_PyObject_AsStringAndSize` to the `string.from_py` functions
> 
> Yes.

I think this got lost. It's still better to use the buffer protocol directly in places where we need the data within a well-defined scope. That's still more general than letting `__Pyx_PyObject_AsStringAndSize` exclude types that require buffer releasing.

Cython/Utility/TypeConversion.c
```diff
@@ -276,30 +276,75 @@ static CYTHON_INLINE const char* __Pyx_PyUnicode_AsStringAndSize(PyObject* o, Py
...
+        return result;
+    } else {
+        char* result;
+        int r;
+#if !CYTHON_COMPILING_IN_LIMITED_API || __PYX_LIMITED_VERSION_HEX >= 0x030B0000
+        // Check that `o`:
+        // * supports the Python Buffer Protocol
+        // * does not require keeping the buffer around (as we cannot return it)
+        PyBufferProcs *pb = __Pyx_PyType_GetSlot(o, tp_as_buffer, PyBufferProcs);
+        if (pb == NULL || pb->bf_getbuffer == NULL || pb->bf_releasebuffer != NULL) goto error;
```
scoder, line 315: I think you probably need to request both slot functions separately, not the buffer struct slot.

jakirkham:
```suggestion
        getbufferproc* getbuffer = __Pyx_PyType_GetSlot(o, bf_getbuffer, getbufferproc);
        releasebufferproc* releasebuffer = __Pyx_PyType_GetSlot(o, bf_releasebuffer, releasebufferproc);
        if (pb == NULL || getbuffer == NULL || releasebuffer != NULL) goto error;
```

Cython/Utility/CppConvert.pyx
```diff
@@ -9,10 +9,31 @@ cdef extern from *:
...
+    cdef const char* data = NULL
+
+    if string_like_t is const_char_mv:
+        data = &(<const_char_mv>o)[0]
+        length = len(<const_char_mv>o)
+    elif string_like_t is const_uchar_mv:
+        data = &(<const_uchar_mv>o)[0]
+        length = len(<const_uchar_mv>o)
+    else:
+        data = __Pyx_PyObject_AsStringAndSize(o, &length)
```
jakirkham, line 35: Was thinking of how best to capture this logic. Took a first pass using `fused` types with typed-`memoryview`s. Imagine there are other changes we want here, but before getting too carried away want to get a sense of whether this is going in the general direction of what we want or if we want something totally different. May need a bit more guidance if the latter

Cython/Utility/TypeConversion.c
jakirkham: Ok tried to pull this out into `string.from_py` below: https://github.com/cython/cython/pull/6752#discussion_r2963755080

However would not be surprised if that needs revisions or we want a different approach. Would welcome more guidance on how we should go from there

Cython/Utility/CppConvert.pyx
jakirkham, line 58: Am also wondering about improving the experience going back from typed-`memoryview`s to Python. Though this is a separate question and may even need a different approach. Have raised it in issue: https://github.com/cython/cython/issues/7583

Cython/Utility/CppConvert.pyx
da-woods: I think the fused function approach is unlikely to work. `cdef` fused functions are dispatched at compile-time.  So what will actually happen is that Cython will generate 6 different versions of the code with a different cname. But there's in `PyrexTypes` that will do this dispatch.

PR #6755, opened 2025-03-23
Don't include absolute paths when using `--embed-positions`

tobiasdiez:
Currently, invoking cython with `embed-positions` from a folder not containing the source code, will result in the absolute paths being embedded. This is fixed by always embedding the relative paths.

Since there are a couple of places where a relative path of a `SourceDescriptor` is needed, a new method `get_relative_path` is introduced and reused.

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -961,12 +961,8 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
         code.putln("static const char* const %s[] = {" % Naming.filetable_cname)
         if code.globalstate.filename_list:
             for source_desc in code.globalstate.filename_list:
-                file_path = source_desc.get_filenametable_entry()
-                if isabs(file_path):
-                    # never include absolute paths
-                    file_path = source_desc.get_description()
                 # Always use / as separator
-                file_path = pathlib.Path(file_path).as_posix()
+                file_path = source_desc.get_relative_path().as_posix()
                 escaped_filename = as_encoded_filename(file_path)
                 code.putln('%s,' % escaped_filename.as_c_string_literal())
         else:
--- a/Cython/Compiler/Scanning.py
+++ b/Cython/Compiler/Scanning.py
@@ -4,22 +4,25 @@
 #
 
 
+from pathlib import Path
+
 import cython
+
 cython.declare(make_lexicon=object, lexicon=object,
                print_function=object, error=object, warning=object,
                os=object, platform=object)
 
 import os
 import platform
-from unicodedata import normalize
 from contextlib import contextmanager
+from unicodedata import normalize
 
 from .. import Utils
-from ..Plex.Scanners import Scanner
 from ..Plex.Errors import UnrecognizedInput
-from .Errors import error, warning, hold_errors, release_errors, CompileError
-from .Lexicon import any_string_prefix, make_lexicon, IDENT
+from ..Plex.Scanners import Scanner
+from .Errors import CompileError, error, hold_errors, release_errors, warning
 from .Future import print_function
+from .Lexicon import IDENT, any_string_prefix, make_lexicon
 
 debug_scanner = 0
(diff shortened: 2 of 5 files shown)


Cython/Compiler/Scanning.py
```diff
@@ -4,22 +4,25 @@
 #
 
 
+from pathlib import Path
+
```
scoder:
```suggestion
```

Cython/Compiler/Scanning.py
```diff
@@ -4,22 +4,25 @@
...
 import cython
+
 cython.declare(make_lexicon=object, lexicon=object,
                print_function=object, error=object, warning=object,
                os=object, platform=object)
```
scoder:
```suggestion
               os=object, platform=object, Path=object)
```

Cython/Compiler/Scanning.py
```diff
@@ -4,22 +4,25 @@
...
 
 import os
 import platform
-from unicodedata import normalize
 from contextlib import contextmanager
+from unicodedata import normalize
```
scoder, line 17:
```suggestion
from pathlib import Path
from unicodedata import normalize
```

scoder (commented):
It looks like you always call `.as_posix()` on the return value of `get_relative_path()`. So, why return a `pathlib.Path` instead of a plain string directly?

Cython/Compiler/Scanning.py
```diff
@@ -274,6 +284,9 @@ def get_lines(self, encoding=None, error_handling=None):
     def get_description(self):
         return self.name
 
+    def get_relative_path(self):
+        return Path(self.get_description())
+
     get_error_description = get_description
 
     def get_filenametable_entry(self):
```
scoder, line 281: We previously used `"stringsource>"` here and you changed it to `self.name`. Why?

tobiasdiez: I'm not sure how the previous code worked in this regard. If you would use `stringsource>` here, then above in the `Compilor/<x>Nodes` you would call `Path("stringsource>")` which doesn't look right. I suspect that it got filtered out by the `isabs` check, which then falls back to the description > name.

tobiasdiez:
@scoder Anything left to do here?

antonio-rojas:
Ping. Having this merged is important to fix issues with the meson build of sagemath.

dkwo:
@tobiasdiez Do you think you can rebase this patch? For me, it does not apply cleanly on top of 3.2.4.

tobiasdiez:
@dkwo it says there are no conflicts with the the base branch, so if you take the whole branch as a patch it should apply cleanly.

@scoder could we please get this in?

dkwo:
For me it does not apply cleanly on top of master branch: I had to edit by hand a few changes to `Scanning` file.
Hopefully it will get merged soon, and thanks!

Cython/Compiler/ModuleNode.py
da-woods: Probably want to remove these - I think they're both unused (although `isabs` was used before your change)

da-woods (commented):
It looks good to me.

My main worry is that I don't quite know where all these paths are actually used. I think we have:
* code objects (probably mostly cosmetic but I imagine some people are using them in debugger type tools)
* tracebacks (again, mostly cosmetic but not helpful if they break)
* docstrings (mostly cosmetic, probably some people using them. I think this is what sage is trying to use)

My view is that we should probably merge it into the master branch (on the basis that it's consistent with what we try to do generally) and see if it causes any issues. But it's definitely the sort of change that feels a bit like it's going to cause unexpected problems.

Cython/Compiler/ModuleNode.py
tobiasdiez: Thanks, removed them.

da-woods:
Thanks. I think it makes sense to merge this and see how it works.

tobiasdiez:
Thanks for the review and merge!

PR #6768, opened 2025-03-29
Updated TypeSlots.py

### LINKED ISSUE #3056: Specifying __bool__ and __nonzero__
Consider the following case
```
cdef class Foo(object):
    def __bool__(self):
        return True
    def __nonzero__(self):
        return self.__bool__()
```

This causes a `RecursionError` when calling `bool(Foo())`

I know the work around is specifying only `__bool__` but it's still a bug IMHO

kekwlboy12469:
fixes #3056
**what was the problem?**
that defining both __bool__ and __nonzero__ methods in a cdef class can lead to a RecursionError which will ig cause infinite loop
so what i did was i created a function where if  __nonzero__ is called it will directly call _bool_ which will solve the infinite loop .

### DIFF
--- a/Cython/Compiler/TypeSlots.py
+++ b/Cython/Compiler/TypeSlots.py
@@ -1129,13 +1129,11 @@ class SlotTable:
         #
         #-------------------------------------------------------------------------
 
+        def resolve_nonzero_as_bool(name):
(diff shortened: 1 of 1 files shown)


scoder:
That's not a correct implementation.

PR #6797, opened 2025-04-18
DOC: Update Sage Notebook instructions in quickstart build

### LINKED ISSUE #6501: [BUG] Outdated documentation point to sage notebook
### Describe the bug

Currently the documentation http://docs.cython.org/en/latest/src/quickstart/build.html#using-the-sage-notebook shows the old sage notebook, which is no longer in use.

Nowadays Sage uses jupyter as its back-end, so the syntax is actually `%%cython` (by default Sage has its own implementation of `%%cython` cell magic, though `%load_ext cython` converts to the `%%cython` one)

We can just delete the last part and mention
* Sage has a slightly-different `%%cython` : https://doc.sagemath.org/html/en/reference/repl/sage/repl/ipython_extension.html#sage.repl.ipython_extension.SageMagics.cython that is loaded by default;
* otherwise, the procedure for IPython/Jupyter notebook also works for sage command-line/sage notebook.

user202729:
closes #6501 

see the issue for more context.


docs/src/quickstart/build.rst
```diff
@@ -113,13 +113,11 @@ For more information about the arguments of the ``%%cython`` magic, see
...
-
-  For users of the Sage math distribution, the Sage notebook allows
-  transparently editing and compiling Cython code simply by typing
-  ``%cython`` at the top of a cell and evaluate it.  Variables and
-  functions defined in a Cython cell imported into the running session.
-
+   For users of the SageMath distribution, the Sage notebook uses Jupyter by default.
+   Sage provides its own implementation of the ``%%cython`` cell magic (see
+   `Sage Magics <https://doc.sagemath.org/html/en/reference/repl/sage/repl/ipython_extension.html#sage.repl.ipython_extension.SageMagics.cython>`_)
+   which is loaded by default. Alternatively, you can overwrite it with version provided by Cython,
```
scoder, line 119:
```suggestion
   which is loaded by default.
   Alternatively, you can overwrite it with the implementation provided by Cython,
```

scoder:
Thanks

PR #6852, opened 2025-05-09
Optimize `range(start, end, step)` when `step` is known to be an unsigned integer

bbb23exposed:
Optimize `range(start, end, step)` when `step` is known to be an unsigned integer. If `step` is unsigned, it is also positive, so we can skip the business of dealing with a potentially unsigned value.

### DIFF
--- a/Cython/Compiler/Optimize.py
+++ b/Cython/Compiler/Optimize.py
@@ -806,20 +806,32 @@ class IterationTransform(Visitor.EnvTransform):
         args = range_function.arg_tuple.args
         if len(args) < 3:
             step_pos = range_function.pos
-            step_value = 1
             step = ExprNodes.IntNode(step_pos, value='1', constant_result=1)
+            step_value = 1
+            abs_step_value = 1
+            neg_step_value = False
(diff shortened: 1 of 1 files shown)


da-woods (commented):
This looks reasonable to me.

It probably does need at least a little testing though to show it works. I think the relevant tests are in tests/run/for_in_range_T372.pyx and tests/run/reversed_iteration.pyx.

scoder:
Looks good to me too. As said, tests are needed, e.g. for different integer value sizes. You can also look at the HTML results of the pycoverage CI job to see if there's anything obviously missing tests in your code.

scoder:
> the relevant tests are in tests/run/for_in_range_T372.pyx and tests/run/reversed_iteration.pyx.

@bbb23exposed, would you mind adding some tests?

bbb23exposed:
> > the relevant tests are in tests/run/for_in_range_T372.pyx and tests/run/reversed_iteration.pyx.
> 
> @bbb23exposed, would you mind adding some tests?

I'll do it tomorrow or two days from now.

PR #6877, opened 2025-05-16
Use PyErr_WarnFormat to avoid %zd format warning

daniel-larraz:
`PyErr_WarnFormat` uses `PyUnicode_FromFormat()` to format the warning message, which is more portable than `PyOS_snprintf`, since the latter depends on compiler support for format specifiers. Specifically, MinGW-w64 emits the following warnings when using `%zd` on Windows:

```
 warning: unknown conversion type character 'z' in format [-Wformat=]
|             "Expected %zd from C header, got %zd from PyObject",
|                        ^
 warning: unknown conversion type character 'z' in format [-Wformat=]
|             "Expected %zd from C header, got %zd from PyObject",
|                                               ^
 warning: too many arguments for format [-Wformat-extra-args]
|             "%s.%s size changed, may indicate binary incompatibility. "
|             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
|             "Expected %zd from C header, got %zd from PyObject",
|             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/Cython/Utility/ImportExport.c
+++ b/Cython/Utility/ImportExport.c
@@ -477,7 +477,6 @@ static PyTypeObject *__Pyx_ImportType_$cyversion(PyObject *module, const char *m
     size_t size, size_t alignment, enum __Pyx_ImportType_CheckSize_$cyversion check_size)
 {
     PyObject *result = 0;
-    char warning[200];
     Py_ssize_t basicsize;
     Py_ssize_t itemsize;
 #if CYTHON_COMPILING_IN_LIMITED_API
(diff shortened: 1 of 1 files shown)


scoder:
Ah, nice, sure. That was a left-over from the time when Py2 was still alive.

daniel-larraz:
As I understand it, the CI failure is not related to my code. Please let me know if any action is required on my part.

PR #6968, opened 2025-06-15
Sets both .not_none and .or_none to False for annotated arguments.

jpe:
This makes the none flags for cpdef f(a: T) equivalent to those for cpdef(T a) which fixes possible cross module call segfaults.

Added run/cross_mod_nonecheck to check that something tests for None before accessing an attribue on a C pointer -- doesn't segfault because it's returning part of the memory for the None instance.

Note that not all tests succeed. If there is interest in merging this, I can fixup the other tests and / or limit the scope of this.

### DIFF
--- a/Cython/Compiler/Nodes.py
+++ b/Cython/Compiler/Nodes.py
@@ -1037,6 +1037,13 @@ class CArgDeclNode(Node):
             elif not self.or_none and arg_type.can_be_optional():
                 self.not_none = True
 
+            # Set both .not_none and .or_none to False so that cpdef f(a: T)
+            # and cpdef f(a: T | None) are equivalent to cpdef f(T a) --
+            # cpdef f(T not None a) and cpdef f(T or None a) are currently
+            # not allowed
+            self.not_none = False
+            self.or_none = False
+
         if arg_type:
             self.type_from_annotation = True
         return arg_type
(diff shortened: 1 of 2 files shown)


Cython/Compiler/Nodes.py
```diff
@@ -1037,6 +1037,13 @@ def inject_type_from_annotations(self, env):
             elif not self.or_none and arg_type.can_be_optional():
                 self.not_none = True
 
+            # Set both .not_none and .or_none to False so that cpdef f(a: T)
+            # and cpdef f(a: T | None) are equivalent to cpdef f(T a) --
+            # cpdef f(T not None a) and cpdef f(T or None a) are currently
+            # not allowed
+            self.not_none = False
+            self.or_none = False
```
scoder, line 1045: Doing this unconditionally seems wrong, give how much effort we invest into correctly setting these two flags just above.

jpe: It is necessary if the function is called from another module because the only form acceptable in a .pxd file is def f(T a): which means that both not_none and or_none will be false when generating code for another module. This can and does lead to segfaults with cross module calls. I had another pull request to allow type hints in .pxd files but that wasn't going anywhere so I created this (and am using it internally) so that I don't get segfaults.

I now think that allowing type hints in .pxd files is the preferred approach so that whatever call convention is used with a type hinted function within a module is also used for cross module calls.

PR #6986, opened 2025-06-24
Compress the function line tables as Python bytes objects

scoder:
It is now much more efficient to store them together with the other compressed Python bytes object contents, than to store them as plain const C strings and constructing the bytes objects at import time anyway.


scoder:
For the benchmark modules, the size reduction is 1 - 9%, depending on the number and length of the functions in the module.

PR #7099, opened 2025-08-20
HTML annotation UI: toggle +/- symbols for expand/collapse state

Zaczero:
This change makes it slightly more readable to see which lines have been expanded.

image

### DIFF
--- a/Cython/Compiler/Annotate.py
+++ b/Cython/Compiler/Annotate.py
@@ -108,9 +108,11 @@ class AnnotationCCodeWriter(CCodeWriter):
 
     # on-click toggle function to show/hide C source code
     _onclick_attr = ' onclick="{}"'.format((
-        "(function(s){"
-        "    s.display =  s.display === 'block' ? 'none' : 'block'"
-        "})(this.nextElementSibling.style)"
+        "(function(f, s, c) {"
+        "    c = f.nodeValue == '+';"
+        "    s.display = c ? 'block' : 'none';"
+        "    f.nodeValue = c ? '−' : '+'"
+        "})(this.firstChild, this.nextElementSibling.style)"
         ).replace(' ', '')  # poor dev's JS minification
     )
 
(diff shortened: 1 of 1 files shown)


Cython/Compiler/Annotate.py
```diff
@@ -108,9 +108,9 @@ def _css(self):
...
     # on-click toggle function to show/hide C source code
     _onclick_attr = ' onclick="{}"'.format((
-        "(function(s){"
-        "    s.display =  s.display === 'block' ? 'none' : 'block'"
-        "})(this.nextElementSibling.style)"
+        "f = this.firstChild, s = this.nextElementSibling.style, "
+        "c = f.nodeValue == '+', s.display = c ? 'block' : 'none', "
+        "f.nodeValue = c ? '-' : '+'"
         ).replace(' ', '')  # poor dev's JS minification
     )
```
scoder: The function argument was meant to avoid defining global names, which your change now does. Please keep the global namespace clean.

scoder:
I think that the feature is an improvement, good idea.

Zaczero:
I have taken the opportunity to replace the ASCII `-` (hyphen-minus) with UTF-8 `−` (minus), as the toggle between +/− feels more natural (they have the same width):

image

Cython/Compiler/Annotate.py
```diff
@@ -108,9 +108,11 @@ def _css(self):
 
     # on-click toggle function to show/hide C source code
     _onclick_attr = ' onclick="{}"'.format((
-        "(function(s){"
-        "    s.display =  s.display === 'block' ? 'none' : 'block'"
-        "})(this.nextElementSibling.style)"
+        "(function(f, s, c) {"
```
scoder, line 112:
```suggestion
        # Use local JS variables by declaring them as function arguments.
        "(function(f, s, c) {"
```

scoder:
Thanks

PR #7132, opened 2025-09-03
Feature/fix tutorial setup py

akettmann:
ref: https://docs.cython.org/en/latest/src/tutorial/cython_tutorial.html#id4

If you follow the tutorial verbatim, setup.py will fail with the following error:

```
setup.py build_ext --inplace 
error: Multiple top-level modules discovered in a flat-layout: ['primes_python', 'primes_python_compiled'].

To avoid accidental inclusion of unwanted files or directories,
setuptools will not proceed with this build.

If you are trying to create a single distribution with multiple modules
on purpose, you should not rely on automatic discovery.
Instead, consider the following options:

1. set up custom discovery (`find` directive with `include` or `exclude`)
2. use a `src-layout`
3. explicitly set `py_modules` or `packages` with a list of names

To find more information, look for "package discovery" on setuptools docs.
```

Accounting for all of the files resolves the issue.

### DIFF
--- a/docs/src/tutorial/cython_tutorial.rst
+++ b/docs/src/tutorial/cython_tutorial.rst
@@ -370,6 +370,7 @@ Now the ``setup.py`` looks like this:
                     ['primes.pyx',                  # Cython code file with primes() function
                      'primes_python_compiled.py'],  # Python code file with primes() function
                     annotate=True),                 # enables generation of the html annotation file
+                py_modules=["primes_python.py"],    # Tells setuptools to include this Python module as well
             )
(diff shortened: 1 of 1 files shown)


akettmann:
cython==3.1.3
setuptool==80.9.0

scoder:
Ok, thanks. Looks like that's needed now.

akettmann:
> Ok, thanks. Looks like that's needed now.

Happy to help. Just working through any rough edges while I get more acquainted before I try and actually help.

Issue #7134, opened 2025-09-04
Warning about integer truncation in generated C code


dalcinl:
### Describe your issue

Since commit d95049a10888bdf996fdf7f60aca19410d74d9ab, Cython generates the following code:

```c
#if CYTHON_USE_PYLONG_INTERNALS
static PyObject* __Pyx_Unpacked___Pyx_PyLong_MultiplyCObj( ...
    ...
    long b, x;
    ...
    if (likely(__Pyx_PyLong_IsCompact(op2))) {
        b = __Pyx_PyLong_CompactValue(op2);
    } else {
```

Note that `PyUnstable_Long_CompactValue` returns `Py_ssize_t`, but the result is assigned to a `long`. 
In fact, I'm getting a corresponding warning from the C compiler on Windows:
https://github.com/mpi4py/mpi4py/actions/runs/17461524191/job/49587112025#step:9:228
```console
src/mpi4py/MPI.c(298497): warning C4244: '=': conversion from 'Py_ssize_t' to 'long', possible loss of data
```

Looking at the CPython PyLong implementation, I not 100% sure integer truncation can ever occur in practice (max 30 bits in digit?). Therefore, maybe the trivial fix is to just add a `(long)` cast:

```diff
diff --git a/Cython/Utility/Optimize.c b/Cython/Utility/Optimize.c
index d714042f1..1bd9dddb5 100644
--- a/Cython/Utility/Optimize.c
+++ b/Cython/Utility/Optimize.c
@@ -1250,7 +1250,7 @@ static {{c_ret_type}} __Pyx_Unpacked_{{cfunc_name}}(PyObject *op1, PyObject *op2
 
     // handle most common case first to avoid indirect branch and optimise branch prediction
     if (likely(__Pyx_PyLong_IsCompact({{pyval}}))) {
-        {{ival}} = __Pyx_PyLong_CompactValue({{pyval}});
+        {{ival}} = (long) __Pyx_PyLong_CompactValue({{pyval}});
     } else {
         const digit* digits = __Pyx_PyLong_Digits({{pyval}});
         const Py_ssize_t size = __Pyx_PyLong_SignedDigitCount({{pyval}});
```

dalcinl:
cc @scoder As the original author of that refactoring, let me know what you think. If you indeed agree that my patch above is good enough to silence the (otherwise inconsecuential) warning, I'll put a PR together for your review.

scoder:
[...]

PR #7160, opened 2025-09-17
Use a fast and safe conversion from PyLong to C long.

scoder:
Previously, we assumed that a compact PyLong value always fits into a C `long`, which is usually true (<= 30 bits) but doesn't need to stay that way, given the `Py_ssize_t` return value when converting it.

Fixes https://github.com/cython/cython/issues/7134


dalcinl (approved):

scoder:
3.1.x commit is https://github.com/cython/cython/commit/74706d67dc88fa68d101b261e60e355497fe94c0

PR #7138, opened 2025-09-04
Bug/Documentation example doesn't work #3069

akettmann:
@scoder You had opinions regarding documentation on my other PR. Do you mind if I use your judgement here?

the literal ticket was: [this ](https://github.com/cython/cython/issues/3069) example doesn't work, which I fixed. 

the extra problem: I fixed it but this path (specifying `language` to the `cythonize` function) is deprecated (https://github.com/cython/cython/blob/93afd3c980907c9350c2bc1bcd9aa639a31a2ee9/Cython/Build/Dependencies.py#L727-L729).

I added a note in my second commit in this PR. Do you want to keep this in the documentation ? Do you want to strengthen the deprecation note? Is there a different style I should use? 

Thanks,
Andrew Kettmann

PR #7195, opened 2025-10-08
Docs: Fix C++ wrapping example in user guide

### LINKED ISSUE #3069: Documentation example doesn't work
The documentation example from  http://docs.cython.org/en/latest/src/userguide/wrapping_CPlusPlus.html#specify-c-language-in-setup-py

```

setup(ext_modules = cythonize(
           "rect.pyx",                 # our Cython source
           sources=["Rectangle.cpp"],  # additional source file(s)
           language="c++",             # generate C++ code
      ))
```

throws error:

> ValueError: got unknown compilation option, please remove: sources

If this to be believed (https://github.com/fginter/dep_search/issues/13), it doesn't work since 0.26.

kianacaster:
The setup.py example for wrapping C++ code was outdated and caused a `ValueError`.

This commit updates the example to use `setuptools.Extension` to correctly specify the C++ language and source files.

The `cython_test` directory is added to provide a complete, working example that can be used for testing and verification.

Closes #3069  'Documentation example doesn't work'

### DIFF
--- a/docs/src/userguide/wrapping_CPlusPlus.rst
+++ b/docs/src/userguide/wrapping_CPlusPlus.rst
@@ -667,9 +667,9 @@ in ``libcpp.typeindex``.
 Specify C++ language in setup.py
 ================================
 
-Instead of specifying the language in the source file (using a ``distutils``
-directive), it is possible to declare it in the :file:`setup.py` file.
-This is done by creating an :class:`~setuptools.Extension` object for your
+The target language (C or C++) can be specified either in each source file
+(as shown below) or in the :file:`setup.py` file.
+For the latter, create an :class:`~setuptools.Extension` object for your
 extension::
 
    from setuptools import Extension, setup
@@ -678,8 +678,11 @@ extension::
    setup(
        ext_modules=cythonize([
            Extension(
+               # Declare the extension module name (including package).
                "rect",
+               # List the main extension file together with additional source files.
                sources=["rect.pyx", "Rectangle.cpp"],
+               # Let Cython generate C++ code and setuptools use the C++ compiler.
                language="c++",
            )
        ])


da-woods:
I think https://github.com/cython/cython/pull/7138 probably does the same thing in a slightly simpler way (it shows a simple version without `Extension` which is what we were trying to do in the first example). So I think we should probably prefer that one instead. 

Thanks though

da-woods:
I think actually we should use the text you have here - it's clear and straightforward.

I don't think we should use your cython_test files though.  We do actually test simple examples from our documentation, but multi-step build-system examples like this are quite hard to test seamlessly. It would be possible to make this example into a test in our actual test-suite (as a srctree file) but it wouldn't be kept up-to-date with the docs so there isn't too much value in it.

Could your remove the new cython_test files and I'll merge this?

docs/src/userguide/wrapping_CPlusPlus.rst
```diff
@@ -667,17 +667,23 @@ in ``libcpp.typeindex``.
 Specify C++ language in setup.py
 ================================
 
-Instead of specifying the language and the sources in the source files, it is
-possible to declare them in the :file:`setup.py` file::
+Instead of specifying the language in the source file (using a ``distutils``
+directive), it is possible to declare it in the :file:`setup.py` file.
+This is done by creating an :class:`~setuptools.Extension` object for your
+extension::
```
scoder:
```suggestion
The target language (C or C++) can be specified either in each source file
(as shown below) or in the :file:`setup.py` file.
For the latter, create an :class:`~setuptools.Extension` object for your
extension::
```

docs/src/userguide/wrapping_CPlusPlus.rst
```diff
@@ -667,17 +667,23 @@ in ``libcpp.typeindex``.
...
-              sources=["Rectangle.cpp"],  # additional source file(s)
-              language="c++",             # generate C++ code
-         ))
+   setup(
+       ext_modules=cythonize([
+           Extension(
+               "rect",
+               sources=["rect.pyx", "Rectangle.cpp"],
+               language="c++",
+           )
```
scoder, line 687:
```suggestion
           Extension(
               # Declare the extension module name (including package).
               "rect",
               # List the main extension file together with additional source files.
               sources=["rect.pyx", "Rectangle.cpp"],
               # Let Cython generate C++ code and setuptools use the C++ compiler.
               language="c++",
           )
```

scoder (commented):
I've reworded the initial paragraph to lower the surprise level ("instead of x", where x is first mentioned further down) and added back comments in the `Extension()` call to give a complete and well explained example (even though only part of it is the topic here).

scoder:
Thanks

PR #7197, opened 2025-10-09
🧪 Fix CI for Python 3.14.0(t)

webknjaz:
With CPython turning its π series from rc into final, the `coverage_*` end-to-end tests started failing.

It was confirmed that the CI passes with Python 3.14.0rc3 and older. But the final release makes said test crash. This is because it never ran on release candidate of 3.14 — the test-runner skips coverage tests on pre-releases: https://github.com/cython/cython/blob/5b14ba8/runtests.py#L248. Perhaps, that check should allow release candidates too.

The reason for the coverage tests failure is that coveragepy uses `sysmon` engine under Python 3.14 which ends up not printing coverage for the Cython module and the test does not find the filename in the output. But the test also disables sysmon while compiling. The included fix explicitly makes coveragepy use the former default of `ctrace` to be consistent with the Cython-side setting.

It's also failing on `pip install` locally, because PEP 517 build backend is isolated and the context `setup.py` (having said import) is running in does not have access to the "outer" Cython copy. It is unclear why this works in older runtimes.

This patch forces the PEP 517 build to happen with isolation turned off.

### DIFF
--- a/tests/run/coverage_cmd_src_pkg_layout.srctree
+++ b/tests/run/coverage_cmd_src_pkg_layout.srctree
@@ -2,6 +2,7 @@
 # tag: coverage,trace
 
 """
+# Install without build isolation as this currently fails in Py3.14.0.
 PYTHON -m pip install --no-build-isolation .
 PYTHON setup.py build_ext --inplace
 PYTHON -m coverage run --source=pkg coverage_test.py


.github/workflows/ci.yml
```diff
@@ -49,8 +49,8 @@ jobs:
           - "3.12"
           - "3.13"
           - "3.13t"
-          - "3.14-dev"
-          - "3.14t-dev"
+          - "3.14.0-rc.3"
+          - "3.14.0t-rc.3"
```
webknjaz: This syntax doesn't work for FT. But that's fine. Sinc non-FT is failing the same way, it'll give us enough indication of whether this was the culprit.

tests/run/coverage_cmd_src_pkg_layout.srctree
```diff
@@ -2,28 +2,39 @@
...
 MODULES = [
-        Extension("pkg.module1", ["src/pkg/module1.pyx"]),
-        ]
+    Extension(
+        "pkg.module1",
+        ["src/pkg/module1.pyx"],
+        define_macros=[
+            ('CYTHON_TRACE', '1'),
+            ('CYTHON_USE_SYS_MONITORING', sys_monitoring_setting),
+        ],
```
webknjaz: @da-woods any idea why moving the macros here doesn't work? The docs imply this should be fine:
> pass the C macro definition `CYTHON_TRACE=1` to the C compiler (e.g. using the setuptools option `define_macros`).

*(https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives)*

webknjaz: Though https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#c-macro-defines suggests using `extra_compile_args` which don't work either.

webknjaz:
@da-woods please feel free to add commits on top of this PR if required for quicker merge. I believe this patch has priority over others since the CI will keep blocking PRs until it's fixed.

.github/workflows/ci.yml
```diff
@@ -382,11 +382,12 @@ jobs:
         with:
           # List special Pythons first to keep the 'normal' one as (last) "python3.xy".
           python-version: |
+            3.14t
```
scoder: We should probably switch to Py3.14t as the base line for FT Python benchmarks (because it's much more advanced), but that's not a part of this PR.
```suggestion
```

tests/run/coverage_cmd_src_pkg_layout.srctree
```diff
@@ -2,7 +2,7 @@
 # tag: coverage,trace
 
 """
-PYTHON -m pip install .
+PYTHON -m pip install --no-build-isolation .
```
scoder, line 6: Why is this needed? I don't think we have any dependencies in this test that would benefit from this option.

scoder: Ok, you mentioned it in your commit message for this change. As I wrote, I don't see a reason for this either, but whatever, it doesn't hurt. Deserves a comment, though.

scoder:
```suggestion
# Install without build isolation as this currently fails in Py3.14.0.
PYTHON -m pip install --no-build-isolation .
```

scoder:
Thanks!

tests/run/coverage_cmd_src_pkg_layout.srctree
webknjaz: @scoder when I was the CI failures, I was trying to repro locally and saw those import errors. So I originally thought this was the root cause. With multiple things failing, I focused on this.
Later, I realized that the traceback I was seeing locally was not happening in the CI but since I was hitting it locally, I figured that other people's local setups might get the same error. I'm not sure why exactly but I do have a user-global pip config with a few options set in it, though I didn't think to check that earlier.

Because this test was failing for me and in the CI, I assumed it was a 3.14 thing. But actually, I was getting this crash on 3.13 locally. So the code comment you added is a little inaccurate.

I later realized that but didn't clarify in the PR description as I was digging into the actual root cause and eventually forgot to document this bit.. Still, I believe this option is useful to keep since it makes the expected pip behavior explicit rather than influenced by something unidentified. The comment just needs to be improved a bit.

.github/workflows/ci.yml
webknjaz: Yeah, I originally thought this would mostly be a PR making it more explicit that the CI now chooses the stable release. That pre-release notion of `-dev` being a pre-release threw me off the right path for a while and I wanted to make sure it's clear in the CI definition.

webknjaz:
@scoder was that one CI job that failed flaky? https://github.com/cython/cython/actions/runs/18390769342/job/52400309314?pr=7197#step:5:20763

I didn't get to look into it before you hit merge.

tests/run/coverage_cmd_src_pkg_layout.srctree
webknjaz: FTR, this is the config I have:
```ini
$ cat ~/.config/pip/pip.conf
[global]
user = false
require-virtualenv = true

[install]
use-pep517 = true
```

But removing it didn't change anything — the test keeps failing if `--no-build-isolation` isn't passed. This is under pyenv-provided Pythons (3.13.7 and 3.14.0 at least). Within respective virtualenvs.

I'll submit a PR correcting the comment.

webknjaz: @scoder here's a better explanation: https://github.com/cython/cython/pull/7203

tests/run/coverage_cmd_src_pkg_layout.srctree
webknjaz: @scoder perhaps you could shed some light on why this didn't set the macros?

scoder: Not sure. I guess I'd also have expected them to work here.
But I think they're fine as `distutils` comments in the test files.

webknjaz: @scoder yep. But I was originally looking into making the value conditional based on the runtime, which would be nicer here. I think I also hit defs not working in `setup.py` in the past but never got to figuring out why.

PR #7205, opened 2025-10-10
Set explicit setuptools' PEP 517 build backend

webknjaz:
If it's unset, build front-ends will fall back to the legacy backend
which is discouraged and has a subtle difference in behavior.

---

📚 Change log preview: https://cython--7205.org.readthedocs.build/en/7183/src/changes.html#build-integration

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -140,6 +140,15 @@ Bugs fixed
 
 * Includes all fixes as of Cython 3.1.4.
 
+Build integration
+-----------------
+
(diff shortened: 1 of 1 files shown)


CHANGES.rst
webknjaz, line 1: 📚 Preview: https://cython--7205.org.readthedocs.build/en/7183/src/changes.html#build-integration

webknjaz:
Oh, so it's a bootstrapping problem: https://app.readthedocs.org/projects/cython/builds/29897382/.
I'll see if I can get around that.

PR #7213, opened 2025-10-12
Reduce visibility of the attributes in the Code.pxd

matusvalo:
Based on comment https://github.com/cython/cython/pull/6983#discussion_r2353652645

### DIFF
--- a/Cython/Compiler/Code.py
+++ b/Cython/Compiler/Code.py
@@ -712,7 +712,7 @@ class UtilityCode(UtilityCodeBase):
     def __hash__(self):
         return hash(self._parts_tuple)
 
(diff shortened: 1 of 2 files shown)


matusvalo:
Maybe it is good idea to group attributes based on visibility, but I wanted to have diff clean. But if there is a need for that I can do that...

Cython/Compiler/Code.py
scoder: This type declaration is not true. I can compare an object to whatever I want, not just its own kind.

(BTW, this last line should probably return `NotImplemented` rather than `False`, but I don't think it matters inside of Cython.)

matusvalo: Yes it was by mistake. i removed it already...

PR #7244, opened 2025-10-20
Simplify unpacked PyLong operation code and optimise it for Py3.12+

scoder:
Replace the negative switch-cases with a simple conditional negation. This moves the "signed digit count" sign condition in Py3.12 out of the digit count and puts it where it actually belongs.

This reduces and simplifies the code so that the C compiler can hopefully track the assignments. Previously, we got false positives warning about potentially unassigned variables.

The diff looks larger than it is because it also shifts the indentation a bit to make the C code align better.


scoder:
I'll merge this as the last bit for 3.2. It really resolves several issues with the previous implementation.

Issue #7233, opened 2025-10-18
[ENH] Compiling ./Demos on WIndows 10 fails to cheese.obj : error LNK2001: unresolved external symbol find_cheeses


MironAtHome:
### Is your feature request related to a problem? Please describe.

In my code, I would like to do [...]
```cython
# add use case related code here
```

MironAtHome:
I am seeing a compiler errors for unit tests, this pr picked up along the way. I will have something soon to fix once I am out of family errands. An extra space linter is on me, too ;)

MironAtHome:
Based on this log: https://github.com/BVLC/caffe/issues/6054
it could be that the compiler issue ( this is an educated guess ) is due to compatibility between python 3.x and 2.x somehow mixing between library reference, build of cythonized code and invoking test.
This error is related to the tests provided by tests/run/public_fused_types.srctree
The affected lines read:
```
File "b.pyx", line 2, in init b
cimport cython
cimport a as a_cmod
```
Error message:

> ImportError: dynamic module does not define module export function (PyInit_a)

I will attempt an repro shortly and see where it leads

MironAtHome:
Extra space was fixed. Linter checks passing.
Please share any concerns, in case additional changes.

PR #7247, opened 2025-10-22
Fix Demo modules

MironAtHome:
Changes:
subfolder callback, module cheese
  Fix __declspec(dllexport) function decorator for the function find_cheeses
  Modified cheesefinder.h and cheesefinder.c
subfolder libraries, module call_mymath
Fixes
1 same function decorator
  Modified mymath.h and mymath.c
2 add #include directive for the file "mymath.c"
  change #include directive for the file "math.h" to angle brackets to ensure proper C standard file included
3 modify file Setup.py
  add check not sys.platform.startswith("win") prior call to gcc compiler
  make call to cythonize conditional upon check for library file presence
  add second variant of cythonize for windows without libraries and library_dirs parameters
Modify Setup.py under Demos subfolder
1 add to  the list of exclusions callback and libraries subfolders and their respective modules
2 organize subset of custom extensions into array
3 ensure include_dirs for the extensions point to correct folders
4 ensure sources are correctly provided
Link issue https://github.com/cython/cython/issues/7233


Demos/setup.py
```diff
@@ -8,18 +8,31 @@
...
+    ext_def = [Extension("cheese",
+                            sources=["callback/*.pyx",
+                                     "callback/cheesefinder.c"],
+                            include_dirs=[get_include(),
+                                          "callback"]),
+               Extension("call_mymath",
+                            sources=["libraries/*.pyx",
+                                     "libraries/mymath.c"],
+                            include_dirs=[get_include(),
+                                          "libraries"]),
```
da-woods: This seems like it's covered by setup.py in "libraries" so shouldn't be handled here

Demos/setup.py
```diff
@@ -8,18 +8,31 @@
...
     from numpy import get_include
 except ImportError:
     pass
 else:
-    numpy_demo = [Extension("*",
+    ext_def = [Extension("cheese",
+                            sources=["callback/*.pyx",
+                                     "callback/cheesefinder.c"],
+                            include_dirs=[get_include(),
+                                          "callback"]),
```
da-woods: The "cheese" extension doesn't involve numpy so doesn't need to be in the `else` clause of `try: from numpy import get_include`

Demos/libraries/setup.py
```diff
@@ -10,23 +10,32 @@
...
+                  include_dirs=[os.getcwd()],  # path to .h file(s)
+                  library_dirs=[os.getcwd()],  # path to .a or .so file(s)
+                  libraries=['mymath'])
+    ])
+else:
+    ext_modules = cythonize([
+        Extension("call_mymath",
+                  sources=["call_mymath.pyx", "mymath.c"],
+                  include_dirs=[os.getcwd()])  # path to .h file(s)
+    ])
```
da-woods: The point of the demo is to show how to link to a library that already exists.

Manually building it into the Cython module is demonstrating something else.

Either we should make the demo work as intended on Windows (with a library) or we should just document that the demo doesn't work on Windows

Demos/setup.py
```diff
@@ -8,18 +8,31 @@
 from distutils.extension import Extension
 from Cython.Build import cythonize
 
-ext_modules = cythonize("**/*.pyx", exclude="numpy_*.pyx")
+ext_modules = cythonize("**/*.pyx",
+                            exclude=["numpy_*.pyx",
+                                     "callback/*.pyx",
+                                     "libraries/*.pyx"])
```
da-woods: Excluding callback and libraries looks right because they have their own setup files.

Demos/callback/cheesefinder.c
```diff
@@ -11,7 +11,7 @@ static char *cheeses[] = {
   0
 };
 
-void find_cheeses(cheesefunc user_func, void *user_data) {
+DLL_EXPORT void find_cheeses(cheesefunc user_func, void *user_data) {
```
da-woods: Running `python Setup.py build_ext --inplace` on the `Setup.py` in the callback folder works for me, so I don't think any changes are needed here.

MironAtHome: Did you run it on Windows?
On Windows building with cl.exe and completely removed DLL_EXPORT directive ( both .c and .h file ) the build fails with linker error message:
```
error: command 'C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.44.35207\\bin\\HostX86\\x64\\link.exe' failed with exit code 1120
```
and with DLL_EXPORT removed in .h file the linker, once again, fails with a different message:
```
cheesefinder.c
callback/cheesefinder.c(14): error C2375: 'find_cheeses': redefinition; different linkage
F:\HL\dev\git\ML\cython\Demos\callback\cheesefinder.h(8): note: see declaration of 'find_cheeses'
error: command 'C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.44.35207\\bin\\HostX86\\x64\\cl.exe' failed with exit code 2
```
Ok, the removal of DLL_EXPORT in .c file apparently does not break the code. Removing directive in .c file.
The whole excercise was to ensure portable code that works correctly on Windows platform with Microsoft C/C++ cl.exe compiler, and retains successful build on Linux. I was hoping portable build would be a welcome feature and help the Demos to be a little bit better. And left figuring out the precise makeup of DLL_EXPORT to a later day. Thank you for your help to push my lazy to refine this effort :)

Demos/libraries/setup.py
MironAtHome: Well, it didn't work on Windows to begin with, because Windows C compiler is not relaxing to old K&R style of C inferring header from .c file.
So, please understand my mindset, as I was not just fixing, but making a call, what feature works, and needs fix, and what breaks, and so, I had to chose how to support it, while enabling build to pass without failure, with an overriding goal of making code portable between platforms. So, if you think of it, as tuning and musical instrument, it's like tuning a string, and to make it resonate in a larger group of instruments, it had to be pulled to a higher pitch, where it sounds right in larger arrangement, with a reasonable effort.
The thing about enabling library build on windows is not trivial, because inferring through the rules, what compiler is used, is not a trivial task. Where GCC is considered a nearly default feature on Linux, CL.exe is not on Windows, and so, to put a line cl.exe mymath.c in else for platform 'nt' is not something I felt comfortable, and chasing Windows C compiler I just finished after learning Distutils CCompiler class with its cc= CCompiler.new_compiler('nt') followed by cc.cc call, basically, emulating how platform infers where to find C compiler on Windows OS. And given the possibility of portable MingW and CygWin makes for an overly complex for a demo case.
Plus, I didn't see a harm in introducing a cython feature, to showcase, how to correctly produce Python natively compiled module, by adding a few steps to cythonize call.
I will be happy to document regarding Windows omitting to provide a demo for Library, for the above reason, and provide example codes for all of the noted cases, including MingW and the rest, only ask it to be part of separate PR ( if this sounds reasonable ). And yeah, totally omitting creating module on windows for this demo case would require additional changes to the script actually executing the code, the run*.py script. Which does seem like an overkill in terms of conditionals.

Demos/setup.py
MironAtHome: Fixing

Demos/setup.py
MironAtHome: The include directive appeared superfluous, removing.
The additional .c file is required, else, build fails with error message.
```
   Creating library build\temp.win-amd64-cpython-312\Release\libraries\call_mymath.cp312-win_amd64.lib and object build\temp.win-amd64-cpython-312\Release\libraries\call_mymath.cp312-win_amd64.exp
call_mymath.obj : error LNK2001: unresolved external symbol sinc
build\lib.win-amd64-cpython-312\call_mymath.cp312-win_amd64.pyd : fatal error LNK1120: 1 unresolved externals
error: command 'C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.44.35207\\bin\\HostX86\\x64\\link.exe' failed with exit code 1120
```

Demos/setup.py
MironAtHome: Actually, excluding cheese and call_mymath modules is required to ensure setup.py completes free of failures, since both require building custom .c files. Those custom .c files fail the build, without additional "sources" parameter adding those explicitly to the build, else, the build fails with error message:
```
   Creating library build\temp.win-amd64-cpython-312\Release\libraries\call_mymath.cp312-win_amd64.lib and object build\temp.win-amd64-cpython-312\Release\libraries\call_mymath.cp312-win_amd64.exp
call_mymath.obj : error LNK2001: unresolved external symbol sinc
build\lib.win-amd64-cpython-312\call_mymath.cp312-win_amd64.pyd : fatal error LNK1120: 1 unresolved externals
error: command 'C:\\Program Files\\Microsoft Visual Studio\\2022\\Community\\VC\\Tools\\MSVC\\14.44.35207\\bin\\HostX86\\x64\\link.exe' failed with exit code 1120
```
My goal was very humble, ensure code builds without failure :)

Demos/callback/cheesefinder.c
MironAtHome: Please review changes, as per above.

MironAtHome:
Please review changes, as per above.
Most requests fixed.
Asking to document omission with regards to Windows for call_mymath into a separate PR ( to follow shortly ).
A through documentation would require covering 2 additional compilers, mingw and cygwin.
Potentially, I would like to add OneAPI Intel compiler into the fold, as it has some very nice features, so, it looks like a very - very nice opportunity to add this compiler to the fold of tools on the list of possible choices. In an ideal world to run a small code performance comparison. I do like racing :)

da-woods:
> A through documentation would require covering 2 additional compilers, mingw and cygwin.

It's very difficult to use anything except MSVC on Windows.  The most promising open-source option is clang-cl. This would probably work perfectly except that setuptools doesn't recognise it.

I really would not try to support or document mingw or cygwin.

Demos/setup.py
da-woods: I think I didn't make this clear.

This file should just exclude the libraries and callback directories not attempt to build them. If people want to build the libraries or callback examples then they should use the setup files in those directories.

Demos/setup.py
```diff
@@ -1,25 +1,61 @@
...
+import sysconfig
 sys.path.insert(0, "..")
 
 from distutils.core import setup
 from distutils.extension import Extension
 from Cython.Build import cythonize
 
-ext_modules = cythonize("**/*.pyx", exclude="numpy_*.pyx")
+ext_modules = cythonize("**/*.pyx",
+                            exclude=["numpy_*.pyx",
```
scoder: The indentation here looks rather random. Feel free to run black or blue over this file to clean it up (and maybe also the `setup.py` in `Demos/libraries/`). Note that we have a (soft) maximum line length of 120 characters.

Demos/libraries/setup.py
```diff
@@ -2,30 +2,47 @@
...
+    try:
+        print("building libmymath.a")
+        assert os.system(f"gcc -I{sysconfig.get_path("include")} -shared -fPIC -c mymath.c -o mymath.o") == 0
+        assert os.system("ar rcs libmymath.a mymath.o") == 0
+    except Exception as ex:
+        print(ex)
+        if not os.path.exists("libmymath.a"):
+            print("Error building external library, please create libmymath.a manually.")
+            sys.exit(1)
+elif sys.platform.startswith("win"):
```
scoder: Regarding the if-conditions here, either `sys.platform` starts with `win` or it does not. There is no need to repeat the condition. Since you seem to treat `win` as the special case, please put that case first (i.e. "if startswith('win') … else"), to make it clear what the "normal" case is.

Demos/libraries/setup.py
```diff
@@ -2,30 +2,47 @@
...
 
 # Here is how to use the library built above.
-ext_modules = cythonize([
-    Extension("call_mymath",
-              sources=["call_mymath.pyx"],
-              include_dirs=[os.getcwd()],  # path to .h file(s)
-              library_dirs=[os.getcwd()],  # path to .a or .so file(s)
-              libraries=['mymath'])
-])
+if os.path.exists("libmymath.a") or os.path.exists("mymath.lib"):
```
scoder: We call `sys.exit()` above if these files do not exist, so you don't need to make the following code conditional.

Demos/libraries/setup.py
```diff
@@ -2,30 +2,47 @@
...
+if not sys.platform.startswith("win"):
+    try:
+        print("building libmymath.a")
+        assert os.system(f"gcc -I{sysconfig.get_path("include")} -shared -fPIC -c mymath.c -o mymath.o") == 0
+        assert os.system("ar rcs libmymath.a mymath.o") == 0
+    except Exception as ex:
+        print(ex)
+        if not os.path.exists("libmymath.a"):
+            print("Error building external library, please create libmymath.a manually.")
+            sys.exit(1)
```
scoder: It looks like this part can be pulled out of the build cases and done generally, in the end. (I.e. do "try … if … elif" instead of "if … try … elif … try".)

Demos/callback/cheesefinder.h
```diff
@@ -1,2 +1,4 @@
-typedef void (*cheesefunc)(char *name, void *user_data);
-void find_cheeses(cheesefunc user_func, void *user_data);
+#include <exports.h>
```
scoder: Interesting. I didn't even know about `exports.h` in the CPython distribution. It seems badly named, way too generic to be a global import.

I also find it questionable whether it is helpful to use it here in the example. I understand that it's an easy way to get platform independent declarations, but external libraries will probably not make use of Python declarations, so this might be misleading for readers.

Maybe we should make it explicit that there is a difference for this between Windows and other systems? By spelling out the preprocessor conditions?

scoder (commented):
I wonder if it's really worth showing two ways to build the libraries. If setuptools can do it, that's probably enough. The intention here is not to show how to build an external library because users who want to integrate with an external library probably have it available or at least know how to build it.

Demos/setup.py
MironAtHome: It keeps inline with style I inherited and was there when I started.
I just tried to blend in.
Happy to match any style you could point to. Just want it to be somewhat consistent.
Looking at this specific line alignment, this was, again, in previous check - in.
In fact I specifically went back to check how style looked in the original file, and found that the square bracket was aligned with first double - quotation mark of the first line of the expression, and I did modify all similar to match original style. Just refresh the browser.
Once again, if this looks different from what is considered appropriate style now, happy to match. Please provide pointer to example. :)

Demos/libraries/setup.py
MironAtHome: Well, no, it's not binary. And this list is controlled by external party, which means, it can mutate for, say, mobile phone, VR glasses, or yet some other platform, like wearables. Though it does look a bit offish for general development coding, the point is, should someone decide to support some other platform, they add an condition and add a block for another os/compiler, however clumsy the logic appears.
In fact many - many teams maintain separate suites to build portable code for distinct platform. So, this, to some fine senses, overly explicit use of platform name, is in line with how in this specific case the condition is weighted.
As far as different build script styles, using gcc compiler command line for Linux and object model of Distutils for Windows, it goes back to my previous comment with dawood. GCC on LInux is primary compiler, Windows, potentially, offers multiple choices. So, the hope is that this somewhat easy to follow Distutils script could help to ease the curve for someone introducing Cython.
- quote

I wonder if it's really worth showing two ways to build the libraries. If setuptools can do it, that's probably enough. The intention here is not to show how to build an external library because users who want to integrate with an external library probably have it available or at least know how to build it.
- end of quote.

Well, look. I ran into a block of existing Cython code, that could very much benefit from splitting it in two parts, part implementing an C++ class, or a set of classes, that mostly encompass data structures, used throughout the rest of platform and mostly expressed in basic C++ data types. They won't mutated over the lifetime of application in ways, that warrant making them in Python. If those class(es) ever add a field or a few fields, it will still be base C/C++ data types. What can be gained from this. Every feature of C++ code. Exception handling, encapsulation, performance(!!!) and integration with algorithmic / compiled code layer, that delivers exceptional performance, while keeping all the modern features today developer considers a norm ( they weren't a very short time ago ). This is precisely the scenario I have coded for, an hybrid C/C++ & Cython. Whether to compile C/C++ block as a distributable library or compile to object code, is a different question. There is a value in using separate library build approach where Cython code is integrated to facade or interface.
And so, I see value in what I have presented. It may be different from what you had in mind, but it is still of value.

Demos/callback/cheesefinder.h
MironAtHome: Well, it is considered a good taste to use, as much as possible, platform features, and add new features only when those are really needed.
The only time it is advised ( by rules of coding excellence ) is to define own macros when platform either 
- lacking one
- or carries significant bug impacting functionality

As far as this codebase goes, I feel we are safe.
Plus, a gentle suggestion to explore existing platform features, rather then reinventing the wheel, is considered an advantage amongst tenured professors of CS cathedra.

Demos/libraries/setup.py
scoder: > should someone decide to support some other platform, they add an condition and add a block for another os/compiler

Thus my request to reverse the condition. Special cases first, to allow adding other special cases easily, then the default case at the end.

However, as I wrote, we may not need the condition at all and could probably get away with using only setuptools to build the external library in a platform-agnostic way.

Demos/setup.py
scoder: It's not a big file, nor an important one. Just run a code formatter over it.

Demos/libraries/setup.py
MironAtHome: Agreed, the variant with code matching proposed logic was checked in about 10 hours ago.
Could it be just a refresh of a browser was keeping you a step away from seeing logic as per description.

Demos/libraries/setup.py
MironAtHome: Yep, just a timing of the review vs. code, matching requested.

Demos/callback/cheesefinder.h
scoder: > it is considered a good taste to use, as much as possible, platform features, and add new features only when those are really needed.

I agree in general. However, this is example code that should resemble a real-world scenario. And that scenario is unlikely to involve a library that uses Python headers and definitions, because the goal is to write code that newly connects Python to an existing C library. Thus my line of thought above. Please read it and tell me your opinion about it.

Demos/libraries/setup.py
scoder: It seems that you changed one file but not the other.

Note that you can give a better error message by assigning the platform specific names of the library to a variable and formatting it into the message.

Demos/callback/cheesefinder.h
MironAtHome: Replaced is the exports.h with Python.h.

Demos/libraries/setup.py
MironAtHome: Is it possible to do this in next iteration of Demos?
On my mind this edition is an improvement over what it was before sufficient for iterative check - in.
It would be nice if a sense of perfection was balanced with adding support to platform.
On my part, I will keep adding features. However, my current ( like next 24 hours goal ) is to use what I learned on a version of library called spaCy to remove the lock on it on python 2.7.
Not suggesting to approve something unfit. It just seems like what is being discussed is in the real of perfect vs. step forward.

Demos/setup.py
MironAtHome: I did run formatter using VS Code "black" extension.
Could it be this formatter didn't format the way expected.
Please do comment.
Once again, happy to adjust. Specific example would be appreciated.

Demos/libraries/setup.py
MironAtHome: Variables applied to handle platform specific library name.
Added *.split(os.path.pathsep) to make directory list parameter portable
Hopefully this closes the gap

Demos/libraries/setup.py
```diff
@@ -1,33 +1,60 @@
...
-        print("Error building external library, please create libmymath.a manually.")
-        sys.exit(1)
+    if not sys.platform.startswith("win"):
+        print(f"building {libname}")
+        assert (
+            os.system(
+                f"gcc -I\"{"\" -I\"".join(py_include.split(os.path.pathsep))}\" -shared -fPIC -c {c_sources} -o mymath.o"
+            )
+            == 0
+        )
```
scoder, line 27: This is barely readable to me – way too many special characters and escapes. And it hides an important call (with side effects) inside of an `assert` statement, which may not get executed in some configurations. (I'm aware that it was like this before, but now is the best time to clean it up.) We should also use `subprocess` instead of `os.system`.

```suggestion
            c_includes = [f'-I{path}' for path in py_include.split(os.path.pathsep)]
            subprocess.check_call(["gcc", "-shared", "-fPIC", "-c"] + c_includes + [c_sources, "-o", "mymath.o"])
```

By its name, `c_sources` should rather be a list of strings than a string. That would also prevent (hypothetical) issues with spaces in file paths etc.

The same applies to the main `Demos/setup.py` script.

Demos/libraries/mymath.h
```diff
@@ -1 +1,3 @@
-double sinc(double);
+#include <Python.h>
+
+Py_EXPORTED_SYMBOL double sinc(double);
```
scoder, line 3: I still think that a C library example should not be using `Python.h` and `Py_EXPORTED_SYMBOL`. It's a C library, not a Python extension.

MironAtHome: So, you mean to say you have a very strong emphasis on how to craft external C libraries, without Python provisions?
I understand, it just looked to me that a gentle suggestion to study what is available in Python ( and seeing how it supports portable code ) would help.
And it does look aesthetically pleasing to an eye a bit better than lines of #define(s) I had earlier.

Demos/libraries/setup.py
MironAtHome: We can't assume that on Windows include directories will not have spaces, hence, need to wrap in quotes.
So, some hairiness may be a bit of a tax on the eyesight, but does help to ensure error free execution.
Thanks for the suggestion regarding subprocess. I will take a look if I can do a jump to setuptools at once, or use this suggestion, since it does look more like a change, over iterative fix.

MironAtHome:
I am a bit concerned that the current failure
image
flagged formatting of the following lines:
```
setup(
    name = 'Demos',
    ext_modules = ext_modules,
)
```
Here is the error message:
`/home/runner/work/cython/cython/Demos/setup.py:77: [E113] unexpected indentation`
and here is screenshot of the line affected:
image
same, as in present time main branch.
In fact all lines flagged in the failed test formatted similarly.
Perhaps the style checker is the trouble and the pull request can be forced over this spuriously raised code formatting related issue?

Demos/libraries/setup.py
scoder: > We can't assume that on Windows include directories will not have spaces, hence, need to wrap in quotes.

I don't think we need to do that if we pass the arguments separately (as in the `subprocess` call that I suggested), rather than pushing them through a command line string.

Demos/libraries/mymath.h
scoder: > you have a very strong emphasis on how to craft external C libraries

No, I just think we shouldn't give wrong impressions in our documentation.

PR #7251, opened 2025-10-23
Add support for Extention arguments for shared module

matusvalo:
This PR allows to pass parameters to extention of shared module:

```python
extensions = [
    Extension("*", ["deleteme.pyx"]),
    Extension("_cyutility", sources=["_cyutility.c"], extra_compile_args=['-O3']),
]

setup(
  ext_modules = cythonize(extensions, shared_utility_qualified_name = "_cyutility")
)
```


scoder (approved):
Looks good to me.
I just noticed that we're apparently not writing module metadata (the JSON header) into the shared module. Could be worth adding as well.

scoder:
Seems fine to include in 3.2.

PR #7274, opened 2025-10-30
Fix wrapping_CPlusPlus example: use Extension() for C++ sources

Hasan-8326:
🧩 Summary

This pull request fixes the outdated C++ wrapping example in
docs/examples/userguide/wrapping_CPlusPlus/setup.py.

🐛 Problem

The example previously used:

setup(ext_modules=cythonize("rect.pyx", sources=["Rectangle.cpp"], language="c++"))

With Cython ≥ 0.26, this raises:

ValueError: got unknown compilation option, please remove: sources

because cythonize() no longer supports the sources argument directly.

🔧 Fix

Rewrote the setup script to use a proper Extension() definition, which is the correct and modern syntax:

from setuptools import setup
from setuptools.extension import Extension

extensions = [
    Extension(
        name="rect",
        sources=["rect.pyx", "Rectangle.cpp"],
        language="c++",
    )
]

setup(
    ext_modules=cythonize(extensions)
)

✅ Result

Works correctly with modern Cython (tested on Cython 0.29.12, Python 3.10).

Prevents ValueError during build.

Keeps the “Wrapping C++” user guide example up-to-date and functional.

🧪 Testing

Built locally with:

python3 setup.py build_ext --inplace
python3 -c "import rect; print(rect.area(3,4))"

Output confirmed: 12.0

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,17 +158,17 @@ jobs:
           - os: ubuntu-22.04
             python-version: "3.11"
             backend: "c,cpp"
-            env: { LIMITED_API: "--limited-api" }
+            env: { LIMITED_API: "--limited-api --abi3audit" }
             extra_hash: "-limited_api"
           - os: ubuntu-22.04
             python-version: "3.12"
             backend: "c,cpp"
-            env: { LIMITED_API: "--limited-api" }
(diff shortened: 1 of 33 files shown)


scoder:
Thanks, but this was already resolved in https://github.com/cython/cython/pull/7195

PR #3629, opened 2020-05-22
FIX: cygdb for recent virtualenvs like pyenv and venv (GH-1961)

ashutoshvarma:
This fix the `activate_this.py` not found error in cygdb as describe in issue #1961 
```
gdb command file: Activating virtualenv: /home/varma/Projects/pyxpdf/venv; path_to_activate_this_py: /home/varma/Projects/pyxpdf/venv/bin/activate_this.py
Traceback (most recent call last):
  File "<string>", line 8, in <module>
FileNotFoundError: [Errno 2] No such file or directory: '/home/varma/Projects/pyxpdf/venv/bin/activate_this.py'
```

This remove the dependence on `activate_this.py` to activate virtualvenvs.

PR #7285, opened 2025-11-04
Fix bugs in cygdb script for Windows usage

merlinz01:
Windows uses backslashes in file paths, so the simple string insertion of `__file__` resulted in a `SyntaxError` complaining about truncated Unicode escapes.

Virtual environments on Windows use the `Scripts` directory instead of the `bin` directory, so the loading of `activate_this.py` raised a `FileNotFoundError`.

I have verified that these changes fix the issues on my machine.

Possibly related: #3629, #7016

### DIFF
--- a/Cython/Debugger/Cygdb.py
+++ b/Cython/Debugger/Cygdb.py
@@ -53,10 +53,8 @@ def make_command_file(path_to_debug_info, prefix_code='',
                 import sys
                 virtualenv = os.getenv('VIRTUAL_ENV')
                 if virtualenv:
-                    if sys.platform == "win32":
-                        path_to_activate_this_py = os.path.join(virtualenv, 'Scripts', 'activate_this.py')
-                    else:
-                        path_to_activate_this_py = os.path.join(virtualenv, 'bin', 'activate_this.py')
+                    scripts_dir = 'Scripts' if sys.platform == "win32" else 'bin'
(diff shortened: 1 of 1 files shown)


Cython/Debugger/Cygdb.py
```diff
@@ -50,17 +50,21 @@ def make_command_file(path_to_debug_info, prefix_code='',
...
                 # Activate virtualenv, if we were launched from one
                 import os
+                import sys
                 virtualenv = os.getenv('VIRTUAL_ENV')
                 if virtualenv:
-                    path_to_activate_this_py = os.path.join(virtualenv, 'bin', 'activate_this.py')
+                    if sys.platform == "win32":
+                        path_to_activate_this_py = os.path.join(virtualenv, 'Scripts', 'activate_this.py')
+                    else:
+                        path_to_activate_this_py = os.path.join(virtualenv, 'bin', 'activate_this.py')
```
scoder: In conditionals, I often prefer to make it clear that we are always setting the variable, just with different values. That's more difficult to see if you have entirely separate if-branches.
```suggestion
                    scripts_dir = 'Scripts' if sys.platform == "win32" else 'bin'
                    path_to_activate_this_py = os.path.join(virtualenv, scripts_dir, 'activate_this.py')
```

PR #7293, opened 2025-11-07
Clean up dependencies for memslice copy code

scoder:
When using a shared module for the memory view utility code, the slice-copy code is only needed in the shared module, not in the user code modules.

Clean up the dependencies along the way to simplify the correct inclusion.

Reported in https://github.com/scipy/scipy/issues/23932#issuecomment-3500037604


scoder:
> how I missed the warning

Might have been me messing this up when I moved stuff around in https://github.com/cython/cython/pull/7210

Issue #7298, opened 2025-11-07
[BUG] Incorrect Unicode character formatting in f-strings


scoder:
### Describe the bug

Related to https://github.com/cython/cython/issues/7240, when formatting a Unicode character value with the `:c` format and short padding (< 250 characters), Cython can generate code that builds an incorrect string. This is due to an off-by-one error in the `CIntToPyUnicode` utility code that inserts a spurious character (from potentially uninitialised memory) before the formatted character.
https://github.com/cython/cython/blob/e16ba11fc555c42afee71e5722828a6dd06eb12e/Cython/Utility/TypeConversion.c#L936-L938

The bug was introduced in commit https://github.com/cython/cython/commit/e1c98faaf7c6d1c92d02644046a2f5a63195a803 from PR https://github.com/cython/cython/pull/6342

### Code to reproduce the behaviour:

```cython
# The following generates e.g. the string "    \x00x" instead of the expected "    x"
assert f"{ord('x'):5c" == "    x"
```

### Cython version

Cython 3.1+

PR #7299, opened 2025-11-07
Fix an off-by-one error when padding Unicode characters with the ':c' formatting type

scoder:
Also fix an exception when trying to format lone surrogates this way.

And, along the way, I finally split the `:c` formatting from the other integer formatting functions because it really is a different thing. The decision is currently taken from a C macro because the (cached) type conversion function cannot currently depend on the format string but only on the Cython type of the formatted C value. The C compiler should be happy to discard the unused function respectively.

Closes https://github.com/cython/cython/issues/7298


scoder:
3.1.x commit: https://github.com/cython/cython/commit/fef8cf02de0987cb6b0d91a7653534758f94d5ca

Also merged into 3.2.x.

PR #7344, opened 2025-11-20
When coercing between list and tuple, remember the control flow state etc.

scoder:


scoder:
Merging this since it's probably uncontroversial.

PR #7359, opened 2025-11-23
Improve errors when assigning to const

scoder:
We cannot currently assign to `const` C arrays and probably shouldn't accept non-trivial values as C init for the `const` variable declaration.


Cython/Compiler/Nodes.py
```diff
@@ -6299,8 +6299,13 @@ def analyse_types(self, env, use_temp=0):
         self.rhs = self.rhs.analyse_types(env)
 
         if self.lhs.is_name and self.lhs.entry.type.is_const:
-            if env.is_module_scope and self.lhs.entry.init is None and self.rhs.has_constant_result():
-                self.lhs.entry.init = self.rhs.constant_result
+            if self.lhs.entry.type.is_array:
+                error(self.pos, f"Assignment to const array '{self.lhs.name}'. Assign to a pointer variable instead.")
+            elif env.is_module_scope and self.lhs.entry.init is None and self.rhs.has_constant_result():
+                if isinstance(self.rhs.constant_result, (list, tuple, set, dict, frozenset)):
```
matusvalo, line 6305: I think tests should contain for all possible instances test (currently there is only `list`)

matusvalo (commented):
I have only one small comment to tests of the PR..

Cython/Compiler/Nodes.py
scoder: Done.

scoder:
I'm postponing this to 3.2.3 since I'd like to have a safe "usable for all" release for 3.2.2 and adding a new error message without any real world testing contradicts that.

PR #7405, opened 2025-12-11
Apply code style to non-py docs; allow whitespace in docs

da-woods:
1. Apply code-style checks to Cython files in the docs.
2. Allow spurious whitespace (i.e. blank lines with a bit of space and newlines at the end of file) for both .py and .pyx files. We'd already used this in .pyx files to make the spacing match, so it seems worthwhile to allow this for .py files too.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -1864,7 +1864,7 @@ class TestCodeFormat(unittest.TestCase):
         result = style.check_files(paths)
         total_errors += result.total_errors
 
-        # chcks for non-Python docs source files
+        # checks for non-Python docs source files
         paths = []
         for codedir in ['docs']:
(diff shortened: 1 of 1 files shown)


runtests.py
```diff
@@ -1846,6 +1853,24 @@ def breakpoint_check(physical_line):
...
+        # checks for non-Python source files
+        paths = []
+        for codedir in ['Cython', 'Demos', 'pyximport']:  # source_dirs:
+            paths += glob.iglob(os.path.join(self.cython_dir, codedir + "/**/*.p[yx][xdi]"), recursive=True)
+        style = pycodestyle.StyleGuide(config_file=config_file, select=non_python_list)
+        print("")  # Fix the first line of the report.
+        result = style.check_files(paths)
+        total_errors += result.total_errors
+
+        # chcks for non-Python docs source files
```
```suggestion
        # checks for non-Python docs source files
```

Issue #7401, opened 2025-12-10
[BUG]: changes to static arrays breaks blosc2 compilation


tacaswell:
### Describe the bug

The development branch of cython fails to build blosc2 with this error  

```
    [287/289] Generating blosc2_ext.c
      FAILED: [code=1] blosc2_ext.c /tmp/tmp_3tpq57y/build/blosc2_ext.c
      cd /tmp/tmp_3tpq57y/build && /home/tcaswell/.virtualenvs/cp315t/bin/python3 -m cython /tmp/pip-install-9lmh_w66/blosc2_b4787ab9e7c44645be8ca9c6384382e6/src/blosc2/blosc2_ext.pyx --output-file blosc2_ext.c

      Error compiling Cython file:
      ------------------------------------------------------------
      ...
              if BLOSC2_USER_REGISTERED_CODECS_START <= cparams.compcode <= BLOSC2_USER_REGISTERED_CODECS_STOP\
                      and cparams.compcode in blosc2.ucodecs_registry.keys():
                  raise ValueError("Cannot use multi-threading with user defined Python codecs")

              ufilters = [BLOSC2_USER_REGISTERED_FILTERS_START <= filter <= BLOSC2_USER_REGISTERED_FILTERS_STOP
                        for filter in cparams.filters]
                        ^
      ------------------------------------------------------------
      /tmp/pip-install-9lmh_w66/blosc2_b4787ab9e7c44645be8ca9c6384382e6/src/blosc2/blosc2_ext.pyx:768:18: Compiler crash in IterationTransform

      ModuleNode.body = StatListNode(blosc2_ext.pyx:11:0)
      StatListNode.stats[56] = StatListNode(blosc2_ext.pyx:761:0)
      StatListNode.stats[0] = CFuncDefNode(blosc2_ext.pyx:761:0,
          args = [...]/1,
          c_compile_guard = '',
          modifiers = [...]/0,
          node_positions = [...]/68,
          outer_attrs = [...]/2,
          visibility = 'private')
      CFuncDefNode.body = StatListNode(blosc2_ext.pyx:762:4)
      StatListNode.stats[0] = IfStatNode(blosc2_ext.pyx:762:4)
      IfStatNode.if_clauses[0] = IfClauseNode(blosc2_ext.pyx:762:7)
      IfClauseNode.body = StatListNode(blosc2_ext.pyx:763:8)
      StatListNode.stats[1] = SingleAssignmentNode(blosc2_ext.pyx:767:19)
[...]

PR #7406, opened 2025-12-12
Allow enum values as C array dimensions

scoder:
Fixes https://github.com/cython/cython/issues/7401

### DIFF
--- a/tests/run/carray_iteration_pyx.pyx
+++ b/tests/run/carray_iteration_pyx.pyx
@@ -1,0 +1,259 @@
+# mode: run
+# tag: carray, forin, genexpr, generators, comprehension, bytes, unicode
+
+# Tests marked with "@cython.test_fail_if_path_exists()" are not currently optimised (but might be later.)
+
+cimport cython
+
+
+cdef extern from *:
+    """
+    typedef enum {
+        EXT_ENUM_VALUE = 5,
+    } ExtEnum;
+    """
+    enum ExtEnum:
+        EXT_ENUM_VALUE
+
+cdef enum InternalEnum:
+    INTERNAL_ENUM_VALUE = 5
+
+
+# Externally defined enum
+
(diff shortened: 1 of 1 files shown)


tacaswell:
I can confirm that this fixes #7401

matusvalo:
I think this PR is missing tests. I created one - feel free to adjsut it and include it to the PR

```cython
# mode: run

PYTHON main.py


######## blosc2.h ########

enum {
  BLOSC2_MAX_FILTERS = 6,
};
typedef struct {
  int filters[BLOSC2_MAX_FILTERS];
} blosc2_cparams;


######## blosc2.pyx ########

cdef extern from "blosc2.h":
    ctypedef enum:
        BLOSC2_MAX_FILTERS

    ctypedef struct blosc2_cparams:
        int filters[BLOSC2_MAX_FILTERS]

cdef main(blosc2_cparams *cparams):
        return [f for f in cparams.filters]

def test():
    cdef blosc2_cparams p = blosc2_cparams({1,2,3,4,5,6})
    return main(&p)

######## setup.py ########

from setuptools import setup

setup(
    ext_modules = cythonize(["blosc2.pyx"],
        compiler_directives={'language_level': 3})
)


######## main.py ########

import blosc2
result = blosc2.test()
assert [1,2,3,4,5,6] == result, result
```

scoder:
Sorry, forgot to commit the test file. It's based on the C array iteration tests, just with non-int constants as array size.

scoder:
We currently store the array size as semantics free value, either integer or name string. It would seem better to store it as a syntax node (`IntNode` or `NameNode`) to remember its type and entry. Here, I only create a `RawCNameExprNode` of type `Py_ssize_t`, pretty much blindly. We lose the information whether a named integer is signed or unsigned, for example, or whether it's referring to an enum or const variable.

The simple solution here should work for what we're doing with the array size, but it is less than ideal.

PR #7409, opened 2025-12-13
Make the array data property noexcept

### LINKED ISSUE #7408: A compiler warning
### Describe your issue

```
[sagelib-10.8.rc2] [spkg-install]   warning: /home/zhongcx/sage/local/var/lib/sage/venv-python3.14/lib/python3.14/site-packages/Cython/Includes/cpython/array.pxd:104:37: Implicit noexcept declaration is deprecated. Function declaration should contain 'noexcept' keyword.
```
It should add ``noexcept`` @da-woods

da-woods:
For me it's adding in exception checks (which are especially bad for use within nogil blocks). It certainly can't raise an exception in CPython.

Closes #7408

### DIFF
--- a/Cython/Includes/cpython/array.pxd
+++ b/Cython/Includes/cpython/array.pxd
@@ -101,7 +101,7 @@ cdef extern from *:  # Hard-coded utility code hack.
             arraydescr* ob_descr    # struct arraydescr *ob_descr;
 
         @property
-        cdef inline __data_union data(self) nogil:
+        cdef inline __data_union data(self) noexcept nogil:
             return __Pyx_PyArray_Data(self)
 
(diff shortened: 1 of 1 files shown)


da-woods:
I've labelled this as 3.2.4 although potentially it's a candidate for 3.2.3 if that isn't yet released. It isn't super important though

Edit: Looks like 3.2.3 is tagged, so too late for that.

scoder:
> Looks like 3.2.3 is tagged, so too late for that.

It's not the first time I add `…-1` etc. tags, usually due to build issues. I don't know when 3.2.4 will become a thing, so why not use `3.2.3-1` for a quick fix for once…

PR #7410, opened 2025-12-14
Opaque objects for coroutines/generators/async utility code

da-woods:
...but not the closures at this stage. This largely just follows the pattern from the CythonFunction utility code.

Part of the 3.15 Stable ABI+freethreading support.


Cython/Utility/ModuleSetupCode.c
```diff
@@ -890,6 +890,13 @@ static CYTHON_INLINE int __Pyx__IsSameCFunction(PyObject *func, void (*cfunc)(vo
     #define __PYX_SHARED_RELATIVE_OFFSET 0
     #define CYTHON_OPAQUE_SHARED_TYPES 0
 #endif
+#if CYTHON_OPAQUE_OBJECTS && CYTHON_USE_FREELISTS
+    // It's a bit difficult (but not impossible) to work out how much memory to zero out in this case
+    // so turn it off in the short-term.
```
da-woods, line 901: ... although actually they're already off in freethreading so maybe they just have to remain off here too.

PR #7415, opened 2025-12-15
Escape coverage page URL

matusvalo:
When git branch contains `/`, the coverage publish fails. This PR escape the name to replace `/` with escape character (`%2F`).

> [!NOTE] 
> CI/CD seems to pass but I don't know how to verify the uploaded coverage file.


da-woods:
> CI/CD seems to pass but I don't know how to verify the uploaded coverage file.

Go to the "actions" tab (once it's finished running). Click "coverage" on the left-hand side of the screen. Then click on an individual run.  Here's a link to one of mine just as an example https://github.com/cython/cython/actions/runs/20212046917

matusvalo:
> > CI/CD seems to pass but I don't know how to verify the uploaded coverage file.
> 
> Go to the "actions" tab (once it's finished running). Click "coverage" on the left-hand side of the screen. Then click on an individual run. Here's a link to one of mine just as an example https://github.com/cython/cython/actions/runs/20212046917

Hmm it seems to work it seems (I used branch name containing `/` on purpose of this PR)

scoder:
Looking at the original command, I can see that we allow shell injection here. We should do more than just `/` escaping.
According to [SO](https://stackoverflow.com/a/78420438), we can pass the name through shell escaping via an env variable and then let Python's `urllib.quote()` output a properly escaped URL.

matusvalo:
> According to [SO](https://stackoverflow.com/a/78420438), we can pass the name through shell escaping via an env variable and then let Python's urllib.quote() output a properly escaped URL.

Done

Tools/escape_url.py
matusvalo, line 1: This script is failry simple and not providing any CLI interface. Since main/the only usage is CI/CD I dont mind much but can be done something more inteligent...

.github/workflows/coverage.yml
```diff
@@ -50,8 +51,9 @@ jobs:
       - name: Copy coverage HTML
         run: |
           if [ -d './coverage-report-html' ]; then
+            ESCAPED_HEAD=$(printf "%s" "$HEAD" | python Tools/escape_url.py)
```
scoder: We might get away with a one-liner like this:
```suggestion
            ESCAPED_HEAD=$(python3 -c "import sys; from urllib.parse import quote; print(quote(sys.argv[1]))" "$HEAD")
```

matusvalo: Hmm I thought the idea is to go with `printf(1)`.  In any case I am 100% fine with this approach - it even simplify the change.

da-woods:
This looks OK to me. I think I can't merge it though because it doesn't trigger a bunch of the "required" tests

PR #7417, opened 2025-12-18
Reduce C code overhead in parser and tempita

scoder:


Cython/Compiler/Parsing.py
```diff
@@ -4593,25 +4593,21 @@ def p_group_pattern(s: PyrexScanner):
...
-                s.next()
-            else:
-                if opener == '(' and len(patterns) == 1:
-                    s.error("tuple-like pattern of length 1 must finish with ','")
-                break
-        s.expect(closer)
-        return MatchCaseNodes.MatchSequencePatternNode(pos, patterns=patterns)
-    else:
-        s.error("Expected '[' or '('")
+    assert s.sy in ('(', '[')
```
da-woods, line 4597: While you're changing this, there's a comment elsewhere in the file that says

```
* starts with "{" is a sequence_pattern
* starts with "[" is a mapping_pattern
```

which is the wrong way around (and almost certainly my fault). But might be worth correcting.

da-woods:
Other than that it looks right to me

PR #7418, opened 2025-12-18
Optimise object comparisons against single Unicode character literals

scoder:
This is used a lot in our parser.

This turned out a bit tricky because we cannot easily change the signature of optimised C comparison functions, and we also cannot reverse the order of the operands here due to the way cascaded comparisons work.

Thus, I chose to generate a simple macro for each character that we compare against, named after the character code, and then let the C preprocessor inject the integer of  the character code in its replacement. This might appear a bit wasteful, but usually, the number of different characters that user code compares against is rather limited.

### DIFF
--- a/Cython/Utility/StringTools.c
+++ b/Cython/Utility/StringTools.c
@@ -292,7 +292,7 @@ return_ne:
 #define __Pyx_PyObject_Equals_uchar(s1, s2, ch2, equals, s1_is_str) (\
     ((s1) == (s2)) ? ((equals) == Py_EQ) : \
     ((s1) == Py_None) ? ((equals) == Py_NE) : \
-    __Pyx__PyObject_EqualsUCS4(s1, s2, equals) \
+    PyObject_RichCompareBool(s1, s2, equals) \
     )
 
 #else
@@ -301,19 +301,13 @@ return_ne:
     ((s1) == Py_None) ? ((equals) == Py_NE) : \
     (likely((s1_is_str) || PyUnicode_CheckExact(s1)) ? \
         __Pyx__PyUnicode_EqualsUCS4(s1, ch2, equals) : \
-        __Pyx__PyObject_EqualsUCS4(s1, s2, equals) \
+        PyObject_RichCompareBool(s1, s2, equals) \
     ))
 
 static CYTHON_INLINE int __Pyx__PyUnicode_EqualsUCS4(PyObject* s1, Py_UCS4 ch2, int equals); /*proto*/
 #endif
 
-static CYTHON_INLINE int __Pyx__PyObject_EqualsUCS4(PyObject* s1, PyObject* s2, int equals); /*proto*/
-
 //////////////////// UnicodeEqualsUCS4 ////////////////////
-
-static CYTHON_INLINE int __Pyx__PyObject_EqualsUCS4(PyObject* s1, PyObject* s2, int equals) {
-    return (s1 == Py_None) ? (equals == Py_NE) : PyObject_RichCompareBool(s1, s2, equals);
-}
 
 #if !(CYTHON_COMPILING_IN_PYPY || CYTHON_COMPILING_IN_LIMITED_API || CYTHON_COMPILING_IN_GRAAL)
 static CYTHON_INLINE int __Pyx__PyUnicode_EqualsUCS4(PyObject* s1, Py_UCS4 ch2, int equals) {


Cython/Utility/StringTools.c
```diff
@@ -277,6 +276,92 @@ static CYTHON_INLINE int __Pyx_PyUnicode_Equals(PyObject* s1, PyObject* s2, int
...
+
+//////////////////// UnicodeEqualsUCS4.proto ////////////////////
+
+#if CYTHON_COMPILING_IN_PYPY || CYTHON_COMPILING_IN_LIMITED_API || CYTHON_COMPILING_IN_GRAAL
+#define __Pyx_PyObject_Equals_uchar(s1, s2, ch2, equals, s1_is_str) \
+    __Pyx__PyObject_EqualsUCS4(s1, s2, equals)
+
+#else
+#define __Pyx_PyObject_Equals_uchar(s1, s2, ch2, equals, s1_is_str) (\
+    likely((s1_is_str) ? ((s1) != Py_None) : PyUnicode_CheckExact(s1)) ? \
```
da-woods: I wonder if it's worth adding `s1 == s2`?

Mainly because on CPython all `chr(0)` to `chr(255)` strings will be interned. So it's reasonably likely to give you a quick answer.

It's slightly hard to know though because I expect these comparisons are `False` more often than they're `True`. So it gives you a quicker answer for `True` but a slower answer for `False`

Cython/Utility/StringTools.c
```diff
@@ -277,6 +276,92 @@ static CYTHON_INLINE int __Pyx_PyUnicode_Equals(PyObject* s1, PyObject* s2, int
...
+    int kind;
+
+    if (unlikely(__Pyx_PyUnicode_READY(s1) < 0)) goto bad;
+    length = __Pyx_PyUnicode_GET_LENGTH(s1);
+    #if !CYTHON_ASSUME_SAFE_SIZE
+    if (unlikely(length < 0)) goto bad;
+    #endif
+    if (length != 1) goto return_ne;
+
+    kind = __Pyx_PyUnicode_KIND(s1);
```
da-woods: Since we exclude the Limited API here, I wonder if using `__Pyx_PyUnicode_KIND` rather than just `PyUnicode_KIND` is making it a little hard to see what's actually going on.

Cython/Utility/StringTools.c
```diff
@@ -277,6 +276,92 @@ static CYTHON_INLINE int __Pyx_PyUnicode_Equals(PyObject* s1, PyObject* s2, int
...
+        } else if (kind == PyUnicode_4BYTE_KIND) {
+            void *data1 = __Pyx_PyUnicode_DATA(s1);
+            ch1 = __Pyx_PyUnicode_READ(kind, data1, 0);
+        } else {
+            goto return_ne;
+        }
+    } else {
+        if (kind == PyUnicode_4BYTE_KIND){
+            void *data1 = __Pyx_PyUnicode_DATA(s1);
+            ch1 = __Pyx_PyUnicode_READ(kind, data1, 0);
```
da-woods:
```suggestion
            ch1 = PyUnicode_4BYTE_DATA(kind)[0];
```

I'd expect the C compiler to be able to see through this and eliminate the dead branches of `PyUnicode_READ` itself, but arguably it makes it slightly easier for it.

(the same probably applies above too)

Cython/Utility/StringTools.c
scoder: We intern many module strings, which makes it even more likely that we end up with identical string objects in comparisons. And the pointer comparison is so cheap that it's almost certainly worth it. I'll add it to both variants.

Cython/Utility/StringTools.c
scoder: I thought about that, too. It's tempting but also means that we'd rely on the far away declaration to keep using the same guard in the future. It's easy to forget about this guard here if we modify the macro guards, e.g. to support a new Python version or implementation. I went with the local macros so that we don't have to think about this later. I'll simplify this, though, following your comment below.

Cython/Utility/StringTools.c
scoder: I was trusting in the C compiler here as well (and haven't verified it). My intuition was that `PyUnicode_READ()` would be more likely available across Python implementations than the specific `PyUnicode_?BYTE_DATA()` macros, but then, let's keep the code simple until we see that it can't be.

PR #7422, opened 2025-12-19
docs: update autopxd FAQ section with current project info

elijahr:
## Summary

Updates the FAQ section "How do I automatically generate Cython definition files from C (.h) or C++ (.hpp) header files?" with current, accurate information about available tools.

## Changes

- Updated autopxd2 link to point to the actively maintained repository (https://github.com/elijahr/python-autopxd2)
- Added information about libclang backend support for C++ parsing
- Reorganized content into a cleaner bullet-point format
- Added note about how autopxd2 and autowrap can be used together
- Fixed RST formatting (the previous version used markdown link syntax)

## Motivation

The previous FAQ entry pointed to outdated repositories and didn't reflect the current state of the autopxd2 project, which now includes:
- Full C++ support via libclang
- Template, namespace, and class handling
- Automatic system header detection

This aligns with the wiki page at https://github.com/cython/cython/wiki/AutoPxd which was recently updated with similar information.

## Checklist

- [x] Documentation-only change (no code changes)
- [x] Follows existing FAQ formatting conventions
- [x] No tests required

### DIFF
--- a/docs/src/userguide/faq.rst
+++ b/docs/src/userguide/faq.rst
@@ -618,6 +618,15 @@ How do I automatically generate Cython definition files from C (.h) or C++ (.hpp
 **Answer**: Two actively maintained tools can help automate the creation of
 Cython bindings from C/C++ headers. They serve complementary purposes:
 
+**autowrap** generates complete ``.pyx`` wrapper modules from annotated ``.pxd`` files.
+
+- Takes hand-written or generated ``.pxd`` files with special annotations
+- Produces Python extension modules with automatic type conversions
+- Supports STL containers, exception handling, and operator overloads
+- Useful when fine-grained control over the Python API is needed
+
+https://github.com/OpenMS/autowrap
+
 **autopxd2** generates ``.pxd`` declaration files directly from C/C++ headers.
 
 - Parses unmodified header files using libclang (recommended) or pycparser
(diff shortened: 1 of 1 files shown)


da-woods:
Sorry for a being a bit slow to review this.

My view based on a quick reading:
* Yes - I agree with you that autopxd is dead and your project is active so that we should update to autopxd2.
* The link to autowrap is now https://github.com/OpenMS/autowrap - if we're changing it then we should update it at the same time.
* My impression is that autowrap is more widely used (just based on the crude metric of https://pypistats.org/packages/autowrap vs https://pypistats.org/packages/autopxd2). On that basis it should probably remain listed first.
* @timosachsenberg and @jpfeuffer - you look like the current most active autowrap contributors. Could one of you to have a quick read of the autowrap section here and check that you agree it's a fair summary? Thanks :)

docs/src/userguide/faq.rst
```diff
@@ -612,33 +612,33 @@ name) to the list of files to be compiled for the extension.
...
-[pycparser](https://github.com/eliben/pycparser) to parse the definitions, so
-the only requirement beyond python dependencies is a C preprocessor on PATH.
+- Takes hand-written or generated ``.pxd`` files with special annotations
+- Produces Python extension modules with automatic type conversions
+- Supports STL containers, exception handling, and operator overloads
+- Useful when fine-grained control over the Python API is needed
 
-https://github.com/gabrieldemarmiesse/python-autopxd2 (A friendly fork of
-python-autopxd, supporting recent Python versions)
+https://github.com/uweschmitt/autowrap
```
jpfeuffer: We transferred the repo to our OpenMS organisation (but of course github redirects). Otherwise, that is a good short description!

elijahr:
@da-woods thanks for the review! My thinking with the ordering was to highlight the workflow of using autopxd to convert C headers to pxd, and then using autowrap to convert those pxd to pyx - but I don't know how common that pattern is, and I agree that putting the more widely used package foremost makes sense. PR has been updated. Thanks!

PR #7423, opened 2025-12-20
Escape spaces in depfile paths

lesteve:
The issue was found in meson with Cython depfiles whose paths contain spaces see https://github.com/mesonbuild/meson/issues/15227.

I ended up only escaping spaces not backslashes to follow what `gcc -M` does. rust-bindgen decided to escape both spaces and back-slashes, see https://github.com/rust-lang/rust-bindgen/pull/2503.
 
This PR does seems to fix the meson issue seen in https://github.com/mesonbuild/meson/issues/15227.

### DIFF
--- a/tests/build/depfile.srctree
+++ b/tests/build/depfile.srctree
@@ -1,7 +1,31 @@
 """
+CYTHONIZE -M foo.pyx
 CYTHONIZE -M 'folder with spaces/foo.pyx'
 PYTHON check.py
 """
+
+######## foo.pyx ########
+
+from bar cimport empty
+
+include "baz.pxi"
+
+empty()
+print(foo())
+
+######## baz.pxi ########
+
+from cython.cimports.libc.math import sin
+
+def foo():
+    return "foo"
+
+print(sin(0))
+
+######## bar.pxd ########
+
+cdef inline void empty():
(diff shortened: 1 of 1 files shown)


lesteve:
I am confident the CI will be green on Windows now, so I am marking this as ready for review.

da-woods:
Would it be possible to make these tests additional to the existing ones rather than replacing the existing ones? Otherwise looks fine to me

scoder:
Sorry, what I think David meant was to add the files next to the existing ones in the same `.srctree` test file, just not replacing the existing ones. I don't think this merits a completely separate test file (`.srctree` tests are fairly costly).

lesteve:
> Would it be possible to make these tests additional to the existing ones rather than replacing the existing ones?

Sure, I just did this in my last commit.

lesteve:
> Sorry, what I think David meant was to add the files next to the existing ones in the same .srctree test file

Ah OK, sorry I misunderstood, will fix accordingly, thanks!

lesteve:
Hopefully now this is closer to what you had in mind

da-woods:
This looks good to me.

da-woods:
Thanks.

3.2.x commit is df5af3619

PR #7288, opened 2025-11-05
Allow Cython to understand subscripted types

matusvalo:
This PR adds coercion and inference for `list[TYPE]`. It allows cython optimize following examples:

```python
a: list[str] = 'Foo'
b: list[str] = 'Bar'
print(a[0] + b[0])
```
The PR allows cython to optimise `a[0] + b[0]` to use `__Pyx_PyUnicode_Concat__Pyx_ReferenceSharing_OwnStrongReferenceInPlaceSafe()` instead of `PyNumber_Add()`

> [!NOTE]  
> The same effect can be done when code is rewritten as follows:
> ```python
> a: list[str] = 'Foo'
> b: list[str] = 'Bar'
> c: str = a[0]
> d: str = d[0]
> print(c + d)
> ``` 

PR also implements inference allowing to write efficiently loops:

```python
b: cython.int = 1
a: list[cython.int] = [2]
for c in a:
    print(b + c)
```

The example above infer type of `c` variable and declare `__pyx_v_c` as `int`. Hence the `b + c` line is generated as `__pyx_v_b + __pyx_v_c` instead of `PyNumber_Add()`

> [!NOTE]  
> The same effect can be done when code is rewritten as follows:
> ```python
> b: cython.int = 1
> a: list[cython.int] = [2]
> c: cython.int
> for c in a:
>     print(b + c)
> ``` 
> Also note that `infer_types` directive must be set to `True`

> [!CAUTION]
> ~~This PR is just for early feedback. The changes are just exploratory containing a lot of dirty workarounds.~~ Caveats:
> * ~~Only `list` is supported for now. It can be easilly extended to `set`. `dict` and `tuple` are more complex and maybe can be done in second round~~ Types `list`, `set`, `frozenset` and `dict` are supported. `tuple` is not in scope of this PR.
> ~~* PR supports coercion and inference for 1 level (`list[tuple]`) all higher nested levels (`list[list[int]]`) looses type information (they work correctly but the generated code is not optimised)~~

## TODO:
* [x] Make `get_container_type()` available for all Cython types (see https://github.com/cython/cython/pull/7288#pullrequestreview-3467280514)
* [x] implement casts like `list_long = <list[cython.long]> x` (or `cython.cast(…)`) to change the item type inference from C int to C long. Even `<list[cython.float]> x` seems a reasonable thing to do in that regard. (see https://github.com/cython/cython/pull/7288#issuecomment-3535132454)
[...]

PR #7424, opened 2025-12-20
Infer types of some subscripted container methods

matusvalo:
This PR is a followup of #7288. ~~Currently marked as draft, since #7288 needs to go first.~~

~~Note for review: The change was implemented only by adjusting `find_return_type_of_builtin_method()` and `inferred_method_return_types`~~


Cython/Compiler/Builtin.py
```diff
@@ -556,7 +559,7 @@ def _generate_divmod_function(scope, argument_types):
         ljust='T',
         lower='T',
         lstrip='T',
-        maketrans='dict[int,object]',  # staticmethod
+        maketrans='dict[int,int]',  # staticmethod
```
matusvalo: Based on typeshed here should be `int` not `object`: https://github.com/python/typeshed/blob/cc1c684fe5c8a614808c5d019ade3a876a3a0c95/stdlib/builtins.pyi#L623

matusvalo:
I think nested types needs to be managed. Following code:

```cython
cdef list[list[int]] a = [[1]]
print(cython.typeof(a.pop().pop()))
print(cython.typeof(a[0][0]))
```
Is printing, what is not optimal:
```
Python object
int
```

Update:

The issue is not related to nesting types (the reproducer works even for `list[int]` type). The issue is that when `.pop()` is called, Cython handles return value via

https://github.com/cython/cython/blob/87f82b7d7c84886396e15f0c09910011188047af/Cython/Compiler/ExprNodes.py#L6465

This function forces the return line to be pyobject:

https://github.com/cython/cython/blob/87f82b7d7c84886396e15f0c09910011188047af/Cython/Compiler/ExprNodes.py#L6277-L6282

I am inclined to leave this improvement to other PR...

Cython/Compiler/Tests/TestBuiltin.py
```diff
@@ -8,6 +8,7 @@
     unsafe_compile_time_methods, is_safe_compile_time_method,
     builtin_scope,
 )
+from Cython.Compiler.Symtab import ModuleScope
```
scoder:
```suggestion
from ..Symtab import ModuleScope
```

Cython/Compiler/Tests/TestBuiltin.py
```diff
@@ -20,6 +21,7 @@
 class TestBuiltinReturnTypes(TimedTest):
     def test_find_return_type_of_builtin_method(self):
         # It's enough to test the method existence in a recent Python that likely has them.
+        scope = ModuleScope('test', None, None)
         look_up_methods = sys.version_info >= (3,10)
```
scoder, line 25: The comment belongs to the version line.

Cython/Compiler/Builtin.py
scoder: It seems that `object` is correct here:
```python
>>> str.maketrans({'a': None})
{97: None}
```

Cython/Compiler/Tests/TestBuiltin.py
```diff
@@ -20,6 +21,7 @@
 class TestBuiltinReturnTypes(TimedTest):
     def test_find_return_type_of_builtin_method(self):
         # It's enough to test the method existence in a recent Python that likely has them.
+        scope = ModuleScope('test', None, None)
```
scoder: `scope` is a very generic name that says nothing about the reason why we need this.
```suggestion
        test_module_scope = ModuleScope('test', None, None)
```

Cython/Compiler/Tests/TestBuiltin.py
```diff
@@ -37,17 +39,28 @@ def test_find_return_type_of_builtin_method(self):
 
             for method_name, return_type_name in methods.items():
                 builtin_type = builtin_scope.lookup(type_name).type
-                return_type = find_return_type_of_builtin_method(builtin_type, method_name)
+                return_type = find_return_type_of_builtin_method(1, scope, builtin_type, method_name)
```
scoder: `1` is a magic constant here, it doesn't say what it is. I had to look up the function signature in order to understand it and the first argument is actually `pos`, the usual three-item position tuple. If it really doesn't matter what we pass here, I'd rather create a variable `pos = None` and pass that in, to make it clear what we are passing it for and that that we're not actually passing a position.

The same applies to the `specialize_here()` call below.

Cython/Compiler/Builtin.py
```diff
@@ -496,6 +496,8 @@ def _generate_divmod_function(scope, argument_types):
 
 # When updating this mapping, also update "unsafe_compile_time_methods" below
 # if methods are added that are not safe to evaluate at compile time.
+# 'T' - type identical to type in dictionary key
+# 'I' - type from index - e.g. for list[int], I is `int`
```
```suggestion
# 'I' - type from subscript - e.g. for list[int], I is `int`
```

Cython/Compiler/Builtin.py
```diff
@@ -656,24 +662,31 @@ def _generate_divmod_function(scope, argument_types):
 inferred_method_return_types['frozenset'].update(inferred_method_return_types['set'])
 
 
-def find_return_type_of_builtin_method(builtin_type, method_name):
-    type_name = builtin_type.name
+def find_return_type_of_builtin_method(pos, env, builtin_type, method_name):
+    type_name = tn.name if (tn := builtin_type.get_container_type()) else builtin_type.name
```
scoder: I find this line barely readable.
```suggestion
    container_type = builtin_type.get_container_type()
    type_name = container_type.name if container_type else builtin_type.name
```

Cython/Compiler/Builtin.py
```diff
@@ -656,24 +662,31 @@ def _generate_divmod_function(scope, argument_types):
...
+    subscripted_types = ()
     if type_name in inferred_method_return_types:
         methods = inferred_method_return_types[type_name]
         if method_name in methods:
             return_type_name = methods[method_name]
             if '[' in return_type_name:
-                # TODO: Keep the "[...]" part when we add support for generics.
+                subscripted_types = return_type_name.partition('[')[2].partition(']')[0]
                 return_type_name = return_type_name.partition('[')[0]
```
```suggestion
                return_type_name, _, subscripted_types = return_type_name[:-1].partition('[')
```

Cython/Compiler/Tests/TestBuiltin.py
```diff
@@ -37,17 +39,28 @@ def test_find_return_type_of_builtin_method(self):
...
 
                 if return_type.is_builtin_type:
                     if '[' in return_type_name:
-                        return_type_name = return_type_name.partition('[')[0]
+                        subscripted_type_names = return_type_name.partition('[')[2].partition(']')[0].split(',')
+                        subscripted_type_names = [type_name if t == 'T' else t for t in subscripted_type_names]
+                        subscripted_types = [builtin_scope.lookup(t).type for t in subscripted_type_names]
+
+                        origin_type_name = return_type_name.partition('[')[0]
```
scoder:
```suggestion
                        origin_type_name, _, subscripted_type_names = return_type_name[:-1].partition('[')
                        subscripted_type_names = [type_name if t == 'T' else t for t in subscripted_type_names.split(',')]
                        subscripted_types = [builtin_scope.lookup(t).type for t in subscripted_type_names]
```

scoder: BTW, once we have `origin_type_name`, we can see if it's `"tuple"` and avoid all the rest (as long as we don't support `tuple`).

Cython/Compiler/Builtin.py
```diff
@@ -656,24 +662,31 @@ def _generate_divmod_function(scope, argument_types):
...
             if return_type_name == 'T':
                 return builtin_type
-            if 'T' in return_type_name:
-                return_type_name = return_type_name.replace('T', builtin_type.name)
+            if return_type_name == 'I':
+                return indexed_type if (indexed_type := builtin_type.infer_indexed_type()) else PyrexTypes.py_object_type
```
```suggestion
                return builtin_type.infer_indexed_type() or PyrexTypes.py_object_type
```

Cython/Compiler/ExprNodes.py
```diff
@@ -6462,7 +6462,7 @@ def analyse_types(self, env):
             self.arg_tuple = TupleNode(self.pos, args = self.args)
             self.arg_tuple = self.arg_tuple.analyse_types(env).coerce_to_pyobject(env)
             self.args = None
-            self.set_py_result_type(function, func_type)
+            self.set_py_result_type(function, env, func_type)
```
scoder: It looks funny to pass the arguments in this order. Could you move `env` first?

Cython/Compiler/Tests/TestBuiltin.py
matusvalo: Fixed.

Cython/Compiler/ExprNodes.py

Cython/Compiler/Tests/TestBuiltin.py
matusvalo: Fixed.

Cython/Compiler/Builtin.py
matusvalo: Hmm strange. Why then typeshed is promoting it as `[int, int]`? I am not arguing your conclusion, it just confuses me. The issue is that having it as `object` also breaks tests :-/. It was convinient to have it as `[int,int]`. I will have a closer look on.

da-woods: Isn't that typeshed line saying "if you pass in two `int` then you get `dict[int, int]`?

But there's also two other overloads which return a `dict[int, something else]`

scoder: > having it as `object` also breaks tests

I guess that's due to https://github.com/cython/cython/issues/7653 ?

matusvalo: No I think it is another issue. #7653 is very specific issue not replicable only with simple object subscription.

matusvalo: > > having it as `object` also breaks tests
> 
> I guess that's due to [#7653](https://github.com/cython/cython/issues/7653) ?

This is the error:
```
ERROR: test_find_return_type_of_builtin_method (Cython.Compiler.Tests.TestBuiltin.TestBuiltinReturnTypes.test_find_return_type_of_builtin_method)
  File "/home/dev/cython/Cython/Compiler/Tests/TestBuiltin.py", line 42, in test_find_return_type_of_builtin_method
    return_type = find_return_type_of_builtin_method(1, test_module_scope, builtin_type, method_name)
  File "/home/dev/cython/Cython/Compiler/Builtin.py", line 687, in find_return_type_of_builtin_method
    subscripted_types = [builtin_scope.lookup(t).type for t in subscripted_types.split(',')]
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'type'
```

Cython/Compiler/Tests/TestBuiltin.py
matusvalo: Fixed.

Cython/Compiler/Builtin.py
matusvalo: fixed.

Cython/Compiler/Builtin.py
matusvalo: Fixed.

Cython/Compiler/Builtin.py
matusvalo: Fixed.

Cython/Compiler/Builtin.py
matusvalo: Fixed.

Cython/Compiler/Tests/TestBuiltin.py
matusvalo: Fixed.

matusvalo:
OK I resolved all comments except `maketrans()`

matusvalo:
OK 01653185effb79c0f4eb29b903566604b884df03 is fixing the issue with `maketrans()`

Cython/Compiler/Tests/TestBuiltin.py
```diff
@@ -37,17 +41,31 @@ def test_find_return_type_of_builtin_method(self):
...
                     if '[' in return_type_name:
-                        return_type_name = return_type_name.partition('[')[0]
+                        origin_type_name, _, subscripted_type_names = return_type_name[:-1].partition('[')
+                        if origin_type_name == 'tuple':
+                            # Currently, cython does not support tuple with subscript
+                            return_type_name = origin_type_name
+                        else:
+                            subscripted_type_names = [type_name if t == 'T' else t for t in subscripted_type_names.split(',')]
+                            subscripted_type_entries = [builtin_scope.lookup(t) for t in subscripted_type_names]
+                            subscripted_types = [e.type if e else py_object_type for e in subscripted_type_entries]
```
scoder: Just a suggestion, but we can bring this together into one pass.
```suggestion
                            subscripted_types = [
                                entry.type if entry else py_object_type
                                for entry in map(builtin_scope.lookup, subscripted_type_names)
                            ]
```

Cython/Compiler/Builtin.py
```diff
@@ -656,24 +662,32 @@ def _generate_divmod_function(scope, argument_types):
...
+                subscripted_type_names = subscripted_type_names.replace('T', builtin_type.name)
             if return_type_name == 'bint':
                 return PyrexTypes.c_bint_type
             elif return_type_name == 'Py_ssize_t':
                 return PyrexTypes.c_py_ssize_t_type
-            return builtin_scope.lookup(return_type_name).type
+            container_type = builtin_scope.lookup(return_type_name).type
+            if subscripted_type_names:
+                subscripted_type_entries = [builtin_scope.lookup(t) for t in subscripted_type_names.split(',')]
+                subscripted_types = [e.type if e else PyrexTypes.py_object_type for e in subscripted_type_entries]
```
scoder: Same as above, just a suggestion.
```suggestion
                subscripted_types = [
                    entry.type if entry else PyrexTypes.py_object_type
                    for entry in map(builtin_scope.lookup, subscripted_type_names.split(','))
                ]
```

scoder (approved):

matusvalo:
It seems other PRs are failing with the same sanitizers. Hence, marking auto-merge

PR #7425, opened 2025-12-21
Match targets

da-woods:
Part of #4029

This enable patterns that involve assignment to targets (i.e. `case 1 as x` or `case x`).

### DIFF
--- a/Cython/Compiler/MatchCaseNodes.py
+++ b/Cython/Compiler/MatchCaseNodes.py
@@ -506,7 +506,6 @@ class OrPatternNode(PatternNode):
 
     def analyse_pattern_expressions(self, subject_node, env):
         self.alternatives = [
-        self.alternatives = [
             a.analyse_pattern_expressions(subject_node, env)
             for a in self.alternatives
         ]
@@ -518,7 +517,7 @@ class OrPatternNode(PatternNode):
         for a in self.alternatives:
             a_assignment = a.generate_target_assignments(subject_node)
             if a_assignment:
-                # In an "or" pattern we need we will need to chose which targets to assign to
+                # In an "or" pattern we will need to chose which targets to assign to
                 # based on which alternative matches.
                 error(self.pos, "Assignment in 'or' patterns is not yet handled")
                 assignments.append(a_assignment)


Cython/Compiler/FlowControl.py
```diff
@@ -946,6 +946,82 @@ def visit_IfStatNode(self, node):
             self.flow.block = None
         return node
 
+    def visit_MatchNode(self, node):
+        # https://peps.python.org/pep-0634/#side-effects-and-undefined-behavior
+        # The specification of structural pattern matching gives
+        # a reasonable amount of freedom about when a binding takes
+        # effect (especially for matches that fail after a few completed
+        # steps).
+        # To make it easier to reason about which variables around bound,
```
scoder: Did you mean this?
```suggestion
        # To make it easier to reason about which variables are unbound,
```

Cython/Compiler/FlowControl.py
```diff
@@ -946,6 +946,82 @@ def visit_IfStatNode(self, node):
...
+                    self.flow.block.add_child(next_block)
+            elif isinstance(case, SubstitutedMatchCaseNode):
+                self.flow.nextblock()
+                if guard_block:
+                    guard_block.add_child(self.flow.block)
+                    guard_block = None
+                self._visit(case.body)
+                orig_parent = self.flow.block
+            else:
+                assert False, case
```
scoder, line 1009: It's sad that this diverges from the visitor pattern. IIUC, it's mostly because you are tracking the `guard_block` horizontally along the way, which doesn't fit well with depth first traversal, right?

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -165,10 +165,14 @@ def analyse_case_expressions(self, subject_node, env):
         # coerce_to_boolean.coerce_to_simple is taken from analyse_temp_boolean_expression
         # and ensures that self.comp_node.generate_disposal_code is trivial and so
         # it doesn't matter if it's skipped in one branch. IfClauseNode relies on the same mechanism.
-        self.pattern.comp_node = self.pattern.comp_node.coerce_to_boolean(env).coerce_to_simple(env)
+        self.pattern.comp_node = self.pattern.comp_node.coerce_to_boolean(
+            env
+        ).coerce_to_simple(env)
```
```suggestion
        self.pattern.comp_node = self.pattern.comp_node.coerce_to_boolean(env).coerce_to_simple(env)
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -165,10 +165,14 @@ def analyse_case_expressions(self, subject_node, env):
...
+        if self.target_assignments:
+            self.target_assignments = self.target_assignments.analyse_expressions(env)
         if self.guard:
-            # analyse_temp_boolean_expression ensures that self.guard.generate_disposal_code is trivial
-            # and so it doesn't matter if it's skipped in one branch.
+            # analyse_temp_boolean_expression ensures that means that self.guard.generate_disposal_code
+            # is trivial and so it doesn't matter if it's skipped in one branch.
```
```suggestion
            # analyse_temp_boolean_expression ensures that self.guard.generate_disposal_code is trivial
            # and so it doesn't matter if it's skipped in one branch.
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -306,12 +314,42 @@ def generate_comparison_disposal_code(self, code):
     def free_comparison_temps(self, code):
         self.comp_node.free_temps(code)
 
+    def generate_target_assignments(self, subject_node):
+        # Generates the assignment code needed to initialize all the targets.
+        # Returns either a StatListNode or None
```
```suggestion
        # Generates the assignment code needed to initialize all the targets.
        # Returns either a StatListNode or None.
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -306,12 +314,42 @@ def generate_comparison_disposal_code(self, code):
...
+        assignments.extend(self.generate_main_pattern_assignment_list(subject_node))
+        if assignments:
+            return Nodes.StatListNode(self.pos, stats=assignments)
+        else:
+            return None
+
+    def generate_main_pattern_assignment_list(self, subject_node):
+        # generates assignments for everything except the "as_target".
+        # Override in subclasses.
+        # Returns a list of Nodes
```
scoder: I think `as_target` is now called `as_targets`. Is this comment still true then?
```suggestion
        # Generates assignments for everything except the "as_targets".
        # Override in subclasses.
        # Returns a list of Nodes.
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -378,6 +416,30 @@ def get_simple_comparison_node(self, subject_node):
...
+
+    def generate_main_pattern_assignment_list(self, subject_node):
+        if self.target:
+            return [
+                Nodes.SingleAssignmentNode(
+                    self.pos, lhs=self.target.clone_node(), rhs=subject_node
+                )
+            ]
+        else:
+            return []
```
scoder, line 422:
```suggestion
    def generate_main_pattern_assignment_list(self, subject_node):
        if not self.target:
            return []
        return [Nodes.SingleAssignmentNode(self.pos, lhs=self.target.clone_node(), rhs=subject_node)]
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -378,6 +416,30 @@ def get_simple_comparison_node(self, subject_node):
...
+    def analyse_pattern_expressions(self, subject_node, env):
+        if self.is_star:
+            return super(MatchAndAssignPatternNode, self).analyse_pattern_expressions(
+                subject_node, env
+            )
+        else:
+            self.comp_node = self.get_comparison_node(subject_node).analyse_expressions(
+                env
+            )
+            return self
```
scoder, line 431:
```suggestion
        if self.is_star:
            return super().analyse_pattern_expressions(subject_node, env)
        else:
            comp_node = self.get_comparison_node(subject_node)
            self.comp_node = comp_node.analyse_expressions(env)
            return self
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -453,11 +515,22 @@ def analyse_declarations(self, env):
             a.analyse_declarations(env)
 
     def analyse_pattern_expressions(self, subject_node, env):
-        for a in self.alternatives:
-            a = a.analyse_pattern_expressions(subject_node, env)
+        self.alternatives = [
+            a.analyse_pattern_expressions(subject_node, env) for a in self.alternatives
+        ]
```
scoder, line 511:
```suggestion
        self.alternatives = [
            a.analyse_pattern_expressions(subject_node, env)
            for a in self.alternatives
        ]
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -453,11 +515,22 @@ def analyse_declarations(self, env):
...
         self.comp_node = self.get_comparison_node(subject_node).analyse_temp_boolean_expression(env)
         return self
 
+    def generate_main_pattern_assignment_list(self, subject_node):
+        assignments = []
+        for a in self.alternatives:
+            a_assignment = a.generate_target_assignments(subject_node)
+            if a_assignment:
+                # Switch code paths depending on which node gets assigned
+                error(self.pos, "Need to handle assignments in or nodes correctly")
```
scoder: I'm not sure I understand this combination of comment and error.

Cython/Compiler/MatchCaseNodes.py
da-woods: Yes - still true.

Cython/Compiler/MatchCaseNodes.py
da-woods: I've changed it to:

```
# In an "or" pattern we need we will need to chose which targets to assign to
                # based on which alternative matches.
                error(self.pos, "Assignment in 'or' patterns is not yet handled")
```

It's mostly just a placeholder TODO that'll get removed when "or" patterns are fully implemented.

Cython/Compiler/FlowControl.py
da-woods: Yes I think it's mostly the guard block that complicates it.

We need to do the assignments before the guard block (because the guard block needs to be able to use the assigned names). So that means we have to do:

```
if case:
   assignments
   if guard:
       body
   else:
      goto next_cast
```

Which isn't a bit of logic that's reallly possible to write in sequential Python if-statements.

But I think it is necessary to keep the case, the assignments and the guard as separate steps. And that means that either the guard or the body needs an odd jump.

da-woods: This bit doesn't get changed after this PR, so if you have a better approach then it shouldn't cause problems though

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -526,8 +518,9 @@ def generate_main_pattern_assignment_list(self, subject_node):
         for a in self.alternatives:
             a_assignment = a.generate_target_assignments(subject_node)
             if a_assignment:
-                # Switch code paths depending on which node gets assigned
-                error(self.pos, "Need to handle assignments in or nodes correctly")
+                # In an "or" pattern we need we will need to chose which targets to assign to
+                # based on which alternative matches.
```
```suggestion
                # In an "or" pattern we will need to chose which targets to assign to
                # based on which alternative matches.
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -516,7 +506,9 @@ def analyse_declarations(self, env):
 
     def analyse_pattern_expressions(self, subject_node, env):
         self.alternatives = [
```
```suggestion
```

Cython/Compiler/FlowControl.py
scoder: I don't think it's too bad. It's quite contained. We can afford to clean it up when we find a better way to do it.

PR #7426, opened 2025-12-22
Avoid opaque locals() in string formatting

scoder:
Especially when passing `locals()` into templates, it becomes really unclear which variables are used where or not at all.


da-woods:
Just updating to trigger a rerun. There were a few failed tests that looked unrelated.

Looks good to me though.

PR #7429, opened 2025-12-23
Fix --embed-modules for package submodules

taimoorzulfiqar:
Sanitise module names passed to --embed-modules so generated C symbols are valid
(replace dots with underscores, handle non-ASCII via punycode and leading-digit prefix).
Update Embed.c template to use sanitized names and add a srctree test exercising
embedding a packaged submodule.

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -3572,14 +3572,43 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
             wmain = "wmain"
         else:
             wmain = Options.embed
+        def _sanitise_embed_name(name):
+            # name: EncodedString or str with dotted module name
+            try:
+                is_ascii = name.isascii()
+            except Exception:
+                # fallback if not EncodedString
+                is_ascii = all(ord(c) < 128 for c in name)
+            if is_ascii:
+                cname = name.replace('.', '_')
+            else:
+                puny = name.encode('punycode') if isinstance(name, str) else name.encode('punycode')
+                if isinstance(puny, bytes):
+                    puny = puny.replace(b'-', b'_').decode('ascii')
+                else:
+                    puny = str(puny).replace('-', '_')
+                cname = 'U_' + puny
+            if cname and cname[0].isdigit():
+                cname = '_' + cname
+            return cname
+
+        # Build a structure with import name and C identifier name for each embed module
+        embed_modules_ctx = []
+        for m in Options.embed_modules:
+            embed_modules_ctx.append((m, _sanitise_embed_name(m)))
+
+        # main module also needs a C-name (sanitised) for the PyInit symbol
+        main_cname = _sanitise_embed_name(env.module_name)
+
         main_method = TempitaUtilityCode.load_cached(
-                "MainFunction", "Embed.c",
-                context={
-                    'module_name': env.module_name,
-                    'module_is_main': module_is_main,
-                    'main_method': Options.embed,
-                    'wmain_method': wmain,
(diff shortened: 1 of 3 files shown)


taimoorzulfiqar79:
Fix for issue #7365

Cython/Compiler/ModuleNode.py
```diff
@@ -3572,14 +3572,43 @@ def generate_main_method(self, env, code):
             wmain = "wmain"
         else:
             wmain = Options.embed
+        def _sanitise_embed_name(name):
+            # name: EncodedString or str with dotted module name
```
scoder, line 3576: We should normalise this to `EncodedString` and then make use of the existing functionality around that.

Cython/Compiler/ModuleNode.py
```diff
@@ -3572,14 +3572,43 @@ def generate_main_method(self, env, code):
...
+                is_ascii = all(ord(c) < 128 for c in name)
+            if is_ascii:
+                cname = name.replace('.', '_')
+            else:
+                puny = name.encode('punycode') if isinstance(name, str) else name.encode('punycode')
+                if isinstance(puny, bytes):
+                    puny = puny.replace(b'-', b'_').decode('ascii')
+                else:
+                    puny = str(puny).replace('-', '_')
+                cname = 'U_' + puny
```
scoder, line 3590: There's code for this in `Symtab.punicodify_name`.

Cython/Compiler/ModuleNode.py
```diff
@@ -3572,14 +3572,43 @@ def generate_main_method(self, env, code):
...
+            if cname and cname[0].isdigit():
+                cname = '_' + cname
```
scoder, line 3592: Why do you think this is necessary?

Cython/Compiler/ModuleNode.py
```diff
@@ -3572,14 +3572,43 @@ def generate_main_method(self, env, code):
...
+            return cname
+
+        # Build a structure with import name and C identifier name for each embed module
+        embed_modules_ctx = []
+        for m in Options.embed_modules:
+            embed_modules_ctx.append((m, _sanitise_embed_name(m)))
```
scoder, line 3598: This is a prime candidate for a list comprehension.

Cython/Compiler/ModuleNode.py
```diff
@@ -3572,14 +3572,43 @@ def generate_main_method(self, env, code):
...
-                    'wmain_method': wmain,
-                    'embed_modules': tuple(Options.embed_modules)})
+            "MainFunction", "Embed.c",
+            context={
+                'module_name': env.module_name,
+                'module_cname': main_cname,
+                'module_is_main': module_is_main,
+                'main_method': Options.embed,
+                'wmain_method': wmain,
+                'embed_modules': tuple(embed_modules_ctx)})
```
scoder, line 3611: It really makes no sense to use a tuple here, so let's fix that along the way.
```suggestion
                'embed_modules': embed_modules_ctx})
```

tests/build/embed_modules_package.srctree
scoder, line 1: There is already a test in `tests/build/embed_modules.srctree`. Please extend that instead of duplicating the entire complicated build setup into a new file. You probably just need to add a second module to it, with a Unicode name.

scoder (commented):
Thanks. Have you looked at https://github.com/cython/cython/pull/7368 before submitting this, the PR linked from the ticket?

da-woods:
> handle non-ASCII via punycode

I don't believe this will work (because I don't think the inittab mechanism can ever handle non-ascii).

I could be wrong, but if you're claiming it works then it should be tested

PR #7430, opened 2025-12-23
Pattern matching for sequences

da-woods:
Part of #4029

There's a few Cython-specific decisions to flag: 
* typed memoryviews are treated as sequences
* C tuples are treated as sequences
* Some constant length sequences are resolved at compile-time and other branches dropped (mainly for ctuples)


Cython/Compiler/MatchCaseNodes.py
```diff
@@ -80,27 +84,54 @@ def refactor_cases(self):
...
+        self.cases = [c for c in self.cases if c is not None]
         return self
 
     def generate_execution_code(self, code):
+        if self.sequence_mapping_temp:
+            self.sequence_mapping_temp.allocate(code)
+            code.putln(
+                "%s = 0; /* sequence/mapping test temp */"
+                % self.sequence_mapping_temp.result()
+            )
```
```suggestion
            code.putln(f"{self.sequence_mapping_temp.result()} = 0; /* sequence/mapping test temp */")
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -80,27 +84,54 @@ def refactor_cases(self):
...
+            # For things that are a sequence at compile-time it's difficult
+            # to avoid generating the sequence mapping temp. Therefore, silence
+            # an "unused error"
+            code.putln("(void)%s;" % self.sequence_mapping_temp.result())
```
```suggestion
            code.putln("(void){self.sequence_mapping_temp.result()};")
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -150,22 +182,39 @@ def validate_irrefutable(self):
             return
         self.pattern.validate_irrefutable()
 
+    def is_sequence_or_mapping(self):
+        return isinstance(
+            self.pattern, (MatchSequencePatternNode, MatchMappingPatternNode)
+        )
```
```suggestion
        return isinstance(self.pattern, (MatchSequencePatternNode, MatchMappingPatternNode))
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -150,22 +182,39 @@ def validate_irrefutable(self):
...
+    def analyse_case_expressions(self, subject_node, env, sequence_mapping_temp):
+        with local_errors(True) as errors:
+            self.pattern = self.pattern.analyse_pattern_expressions(env, sequence_mapping_temp)
+            self.comp_node = self.pattern.get_comparison_node(subject_node, sequence_mapping_temp)
+            self.comp_node = self.comp_node.analyse_types(env)
+
+        if self.comp_node and self.comp_node.is_literal:
+            self.comp_node.calculate_constant_result()
+            if not self.comp_node.constant_result:
+                # we know this pattern can't succeed. Ignore any errors and return None
```
```suggestion
                # We know this pattern can't succeed. Ignore any errors and return None.
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -80,27 +84,54 @@ def refactor_cases(self):
...
     def generate_execution_code(self, code):
+        if self.sequence_mapping_temp:
+            self.sequence_mapping_temp.allocate(code)
+            code.putln(
+                "%s = 0; /* sequence/mapping test temp */"
+                % self.sequence_mapping_temp.result()
+            )
+            # For things that are a sequence at compile-time it's difficult
+            # to avoid generating the sequence mapping temp. Therefore, silence
+            # an "unused error"
```
```suggestion
            # an "unused error".
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -176,12 +225,22 @@ def analyse_case_expressions(self, subject_node, env):
...
-        self.pattern.generate_comparison_evaluation_code(code)
-        code.putln("if (%s) { /* pattern */" % self.pattern.comparison_result())
-        self.pattern.generate_comparison_disposal_code(code)
-        self.pattern.free_comparison_temps(code)
+        self.pattern.allocate_subject_temps(code)
+        self.comp_node.generate_evaluation_code(code)
+
+        end_of_case_label = code.new_label()
+
+        code.putln("if (!%s) { /* !pattern */" % self.comp_node.result())
```
```suggestion
        code.putln(f"if (!{self.comp_node.result()}) ""{ /* !pattern */")
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -224,29 +287,42 @@ class PatternNode(Node):
...
+       returns an ExprNode which can be evaluated to determine if the
+       pattern has matched.
+       While generating the comparison we try quite hard not to
+       analyse it until right at the end, because otherwise it'll lead
+       to a lot of repeated work for deeply nested patterns.
+    4. In the code generation stage, PatternNodes hardly generate any
+       code themselves. However, they do set up whatever temps they
+       need (mainly for sub-pattern subjects), with "allocate_subject_temps",
+       "release_subject_temps", and "dispose_of_subject_temps" (which
+       they also call recursively on their sub-patterns)
```
scoder: Minor formatting, quotes, etc.
```suggestion
    How these nodes are processed:
    1. During "analyse_declarations", "PatternNode.generate_target_assignments()"
       is called on the main "PatternNode" of the case. This calls its
       sub-patterns ".generate_target_assignments()" recursively.
       This creates a "StatListNode" that is held by the "MatchCaseNode".
    2. In the "analyse_expressions" phases, the "MatchCaseNode" calls
       "PatternNode.analyse_pattern_expressions", which calls its sub-pattern recursively.
    3. At the end of the "analyse_expressions" stage, the "MatchCaseNode"
       class "PatternNode.get_comparison_node()" (which calls
       "PatternNode.get_comparison_node()" for its sub-patterns).
       This returns an ExprNode which can be evaluated to determine if the
       pattern has matched.
       While generating the comparison we try quite hard not to
       analyse it until right at the end, because otherwise it'll lead
       to a lot of repeated work for deeply nested patterns.
    4. In the code generation stage, "PatternNodes" hardly generate any
       code themselves. However, they do set up whatever temps they need
       (mainly for sub-pattern subjects), with "allocate_subject_temps",
       "release_subject_temps", and "dispose_of_subject_temps" (which
       they also call recursively on their sub-patterns).
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -284,43 +360,35 @@ def get_simple_comparison_node(self):
         """
         raise NotImplementedError
 
+    def get_comparison_node(self, subject_node, sequence_mapping_temp=None):
+        error(self.pos, "This type of pattern is not currently supported %s" % self)
```
```suggestion
        error(self.pos, f"This type of pattern is not currently supported: {self}")
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -284,43 +360,35 @@ def get_simple_comparison_node(self):
...
                 child.validate_irrefutable()
 
-    def analyse_pattern_expressions(self, subject_node, env):
-        error(self.pos, "This type of pattern is not currently supported")
-        return self
-
-    def calculate_result_code(self):
-        return self.comp_node.result()
+    def analyse_pattern_expressions(self, env, sequence_mapping_temp):
+        error(self.pos, "This type of pattern is not currently supported %s" % self)
```
```suggestion
        error(self.pos, f"This type of pattern is not currently supported {self}")
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -499,23 +574,26 @@ def get_simple_comparison_node(self, subject_node):
             )
         return binop
 
+    def get_comparison_node(self, subject_node, sequence_mapping_temp):
```
```suggestion
    def get_comparison_node(self, subject_node, sequence_mapping_temp=None):
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -412,31 +492,26 @@ def is_simple_value_comparison(self):
 
     def get_simple_comparison_node(self, subject_node):
         assert self.is_simple_value_comparison()
-        return ExprNodes.BoolNode(self.pos, value=True)
+        return self.get_comparison_node(subject_node, None)
```
```suggestion
        return self.get_comparison_node(subject_node)
```

.github/workflows/ci.yml
```diff
@@ -144,7 +144,7 @@ jobs:
             extra_hash: "-all"
           # Linting
           - os: ubuntu-22.04
-            python-version: "3.9"
+            python-version: "3.13"
```
da-woods: I was getting code style failures. I think because it was trying to use syntax which Python 3.9 doesn't understand (and code-style goes through the AST)

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -51,15 +54,16 @@ def refactor_cases(self):
...
+                    match_node=self
                 )
                 if_clause = Nodes.IfClauseNode(
                     c.pos,
                     condition=c.pattern.get_simple_comparison_node(subject),
                     body=body,
                 )
-                assignments = c.pattern.generate_target_assignments(subject)
+                # Passing None for env is safe only because we know it's a simple value comparison.
+                assignments = c.pattern.generate_target_assignments(subject, None)
```
da-woods: This `None` seems a bit dubious to me. It works and is designed to work, but I wonder if there's a better way.

.github/workflows/ci.yml
da-woods: Unfortunately that causes it to pick up another (deliberate) code-style error that `# noqa` fails to silence, and adding `tests/run/fstring.pyx` to be excluded in the pycodestyle config also fails to silence.

da-woods:
There's a potential complication that I realised while thinking about what `TPFLAGS_SEQUENCE` actually means on other PRs:

In this PR I've implemented it as `len(p)` followed by `p[0]`, `p[1]`, ... and potentially `p[-2]`, `p[-1]` if there are nodes to the right of `*`.

In CPython they've implemented it as `len(p)` followed by `iter(p)` (plus a few special-cases for known types).

So my version is only correct if we take the definition of "sequence" to mean that `list(iter(s))`, `[ s[i] for i in range(len(s)) ]`, and `[ s[i] for i in range(-len(s)-1, 0) ]` should be equivalent. I think that's a reasonable definition but it doesn't seem to explicitly be written down anywhere.

The pattern matching PEP is a bit vague on the subject. However it generally takes the view that a sensible implementation is a valid implementation and is intentionally not too prescriptive about the exact order of operations. I think that makes my implementation justifiable. It wouldn't be too hard to revise if any good, justifiable examples of people doing something different appear in future

scoder (commented):
Looks like I had a couple of older review comments lying around that I hadn't posted yet. I hadn't finished the review at the time, but here they are.

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+
+////////////////////// OtherSequenceSliceToList.proto //////////////////////
+
+static PyObject *__Pyx_MatchCase_OtherSequenceSliceToList(PyObject *x, Py_ssize_t start, Py_ssize_t end); /* proto */
+
+////////////////////// OtherSequenceSliceToList //////////////////////////
+
+// This is substantially based off ceval unpack_iterable.
+// It's also pretty similar to itertools.islice
+// Indices must be positive - there's no wraparound or boundschecking
```
scoder:
```suggestion
// This is substantially based off ceval unpack_iterable.
// It's also pretty similar to itertools.islice.
// Indices must be positive - there's no wraparound or boundschecking.
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+    list = PyList_New(total);
+    if (!list) {
+        return NULL;
+    }
+
+    slot = __Pyx_PyObject_TryGetSubSlot(x, tp_as_sequence, sq_item, ssizeargfunc);
+    if (!slot) {
+        #if !defined(Py_LIMITED_API) && !defined(PySequence_ITEM)
+        // PyPy (and maybe others?) implements PySequence_ITEM as a function. In this case
+        // it's slightly more efficient than using PySequence_GetItem since it skips negative indices
```
scoder:
```suggestion
        // PyPy (and maybe others?) implements PySequence_ITEM as a function. In this case.
        // it's slightly more efficient than using PySequence_GetItem since it skips negative indices.
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+////////////////////// TupleSliceToList.proto //////////////////////
+
+static PyObject *__Pyx_MatchCase_TupleSliceToList(PyObject *x, Py_ssize_t start, Py_ssize_t end); /* proto */
+
+////////////////////// TupleSliceToList //////////////////////////
+//@requires: OtherSequenceSliceToList
+//@requires: ObjectHandling.c::TupleAndListFromArray
+
+// Note that this should also work fine on lists (if needed)
+// Indices must be positive - there's no wraparound or boundschecking
```
scoder:
```suggestion
// Note that this should also work fine on lists (if needed).
// Indices must be positive - there's no wraparound or boundschecking.
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+
+static PyObject *__Pyx_MatchCase_TupleSliceToList(PyObject *x, Py_ssize_t start, Py_ssize_t end) {
+#if !CYTHON_COMPILING_IN_CPYTHON
+    return __Pyx_MatchCase_OtherSequenceSliceToList(x, start, end);
+#else
+    PyObject **array;
+
+    (void)__Pyx_MatchCase_OtherSequenceSliceToList; // clear unused warning
+
+    array = PySequence_Fast_ITEMS(x);
```
scoder: Does it _need_ to work on a list? Otherwise, `&PyTuple_GET_ITEM(x, 0)` would probably be faster and use less code, if we already know that we have a tuple.

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+static int __Pyx_MatchCase_IsSequence(PyObject *o, unsigned int *sequence_mapping_temp); /* proto */
+
+//////////////////////////// IsSequence /////////////////////////
+//@requires: ABCCheck
+
+static int __Pyx_MatchCase_IsSequence(PyObject *o, unsigned int *sequence_mapping_temp) {
+#if PY_VERSION_HEX >= 0x030A0000 && !(CYTHON_COMPILING_IN_LIMITED_API || CYTHON_COMPILING_IN_PYPY)
+    return __Pyx_PyType_HasFeature(Py_TYPE(o), Py_TPFLAGS_SEQUENCE);
+#else
+    // Py_TPFLAGS_SEQUENCE doesn't exit.
```
scoder, line 226: Since the feature check is supposedly quite quick, we could still check for it in older Python versions and if it's set, return immediately. We allow setting it from Cython now, so it's at least not impossible that it's used.

Also, shouldn't we always be doing the check in the Limited API? We won't know the Python version upfront there.

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+    *sequence_mapping_temp = abc_result;
+    if (*sequence_mapping_temp & __PYX_DEFINITELY_SEQUENCE_FLAG) {
+        return 1;
+    }
+
+    // array.array is a more complicated check (and unfortunately isn't covered by
+    // collections.abc.Sequence on Python <3.10).
+    // Do the test by checking the module name, and then importing/testing the class
+    // It also doesn't give perfect results for classes that inherit from both array.array
+    // and a mapping
```
scoder:
```suggestion
    // Do the test by checking the module name, and then importing/testing the class.
    // It also doesn't give perfect results for classes that inherit from both array.array
    // and a mapping.
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+    if (sequence_mapping_temp) {
+        // maybe we already know the answer
+        if (*sequence_mapping_temp & __PYX_DEFINITELY_SEQUENCE_FLAG) {
+            return 1;
+        }
+        if (*sequence_mapping_temp & __PYX_DEFINITELY_NOT_SEQUENCE_FLAG) {
+            return 0;
+        }
+    } else {
+        // Probably quicker to just assign it and not check from here
```
scoder:
```suggestion
        // Probably quicker to just assign it and not check from here.
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+        sequence_mapping_temp = &dummy;
+    }
+
+    // Start by check a known list of types
```
scoder:
```suggestion
    // Start by checking a known list of types.
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+            result = (__PYX_DEFINITELY_SEQUENCE_FLAG | __PYX_DEFINITELY_NOT_MAPPING_FLAG);
+            break;
+        } else if (is_subclass_mapping && !is_subclass_sequence) {
+            result = (__PYX_DEFINITELY_NOT_SEQUENCE_FLAG | __PYX_DEFINITELY_MAPPING_FLAG);
+            break;
+        }
+    }
+    // If we get to the end of the loop without breaking then neither type is in
+    // the MRO, so they've both been registered manually. We don't know which was
+    // registered first so accept the object as either as a compromise
```
scoder:
```suggestion
    // If we get to the end of the loop without breaking then neither type is in
    // the MRO, so they've both been registered manually. We don't know which was
    // registered first so accept the object as either as a compromise.
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+    if (0) {
+        loop_error:
+        PyErr_Clear();
+    }
```
scoder: Rather than a slightly obscure `if (0)` that is not unused code, you could move the label down below the `return` and jump back to `end` afterwards.

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+            goto end;
+        } else /* mapping_result == 1 */ {
+            if (sequence_first && !sequence_result) {
+                result |= __PYX_DEFINITELY_MAPPING_FLAG;
+                goto end;
+            }
+        }
+    }
+    if (!sequence_first) {
+        // here we know mapping_result is true because we'd have returned otherwise
```
scoder:
```suggestion
        // Here we know mapping_result is true because we'd have returned otherwise.
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+}
+
+// the result is defined using the specification for sequence_mapping_temp
+// (detailed in "is_sequence")
+static unsigned int __Pyx_MatchCase_ABCCheck(PyObject *o, int sequence_first, int definitely_not_sequence, int definitely_not_mapping) {
+    // in Python 3.10 objects can have their sequence bit set or their mapping bit set
+    // but not both. Practically this translates to "which type is registered first".
+    // In Python < 3.10 we can only determine this if they're direct bases (by looking
+    // at the MRO order). If they're registered manually then we can't tell
+
```
scoder, line 75:
```suggestion
    // In Python 3.10+, objects can have their sequence bit set or their mapping bit set
    // but not both. Practically, this translates to "which type is registered first".
    // In Python < 3.10 we can only determine this if they're direct bases (by looking
    // at the MRO order). If they're registered manually then we can't tell.
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+#define __PYX_SEQUENCE_MAPPING_ERROR (1U<<4)  // only used by the ABCCheck function
+
+static int __Pyx_MatchCase_InitAbcType(PyObject *abc_module, PyObject **abc_type, PyObject *name) {
+    if (*abc_type) return 0;
+    *abc_type = PyObject_GetAttr(abc_module, name);
+    return *abc_type ? 0 : -1;
+}
+
+// the result is defined using the specification for sequence_mapping_temp
+// (detailed in "is_sequence")
```
scoder:
```suggestion
// The result is defined using the specification for sequence_mapping_temp
// (detailed in "is_sequence").
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+    return PyDict_CheckExact(o);
+}
+
+static int __Pyx_MatchCase_IsExactNeitherSequenceNorMapping(PyObject *o) {
+    if (PyByteArray_Check(o) || PyBytes_Check(o) || PyUnicode_Check(o)) {
+        return 1;  // these types are deliberately excluded from the sequence test
+            // even though they look like sequences for most other purposes.
+            // Leave them as inexact checks since they do pass
+            // "isinstance(o, collections.abc.Sequence)" so it's very hard to
+            // reason about their subclasses
```
scoder:
```suggestion
        // These types are deliberately excluded from the sequence test
        // even though they look like sequences for most other purposes.
        // Leaving them as inexact checks since they do pass
        // "isinstance(o, collections.abc.Sequence)" so it's very hard to
        // reason about their subclasses.
        return 1;
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+    }
+    if (o == Py_None || PyLong_CheckExact(o) || PyFloat_CheckExact(o)) {
```
scoder: `bool`, `complex`, `slice`, sets, others…

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
...
+        // that it isn't a mapping too
+        return 1;
+    }
+    return 0;
+}
+
+static CYTHON_INLINE int __Pyx_MatchCase_IsExactMapping(PyObject *o) {
+    // Py_Dict is the only regularly used mapping type
+    // "types.MappingProxyType" also exists but is correctly covered by
+    // the isinstance(o, Mapping) check
```
scoder:
```suggestion
    // Py_Dict is the only regularly used mapping type.
    // "types.MappingProxyType" also exists but is correctly covered by
    // the isinstance(o, Mapping) check.
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,386 @@
+///////////////////////////// ABCCheck //////////////////////////////
+
+#if PY_VERSION_HEX < 0x030A0000 || CYTHON_COMPILING_IN_LIMITED_API || CYTHON_COMPILING_IN_PYPY
+static CYTHON_INLINE int __Pyx_MatchCase_IsExactSequence(PyObject *o) {
+    // is one of the small list of builtin types known to be a sequence
```
scoder:
```suggestion
    // Is one of the small list of builtin types known to be a sequence.
```

scoder (commented):
Review of the C parts, will look through the Python node changes next.

Cython/Utility/MatchCase.c
da-woods: This is mainly intended as an optimization I think so I didn't intend to be too comprehensive. (It's also in "fallback" code so maybe not important.) The `bytearray`, `bytes`, `str` check above is necessary but these ones aren't

I'll take the view that `bool` is almost certainly worthwhile. I'll add the others you list because they're considerably cheaper than what follow.  I don't think it's worth looking spending too much time being comprehensive here

Cython/Utility/MatchCase.c
da-woods: `Py_TPFLAGS_SEQUENCE` was deliberately not added to the Limited API so the check isn't supposed to be possible there.

However, there is a runtime source of the flag available. I think that might be worthwhile so I'll add that in.

Cython/Utility/ModuleSetupCode.c
```diff
@@ -760,6 +760,19 @@ static int __Pyx_init_co_variables(void); /* proto */
   #define __Pyx_PyObject_GC_IsFinalized(o) PyObject_GC_IsFinalized(o)
 #endif
 
+#if CYTHON_COMPILING_IN_LIMITED_API
```
da-woods, line 734: This (and the other changes in ModuleSetupCode.c) are new and should let us do fairly quick checks at runtime on recent versions of Python.

I decided it wasn't ultra-simple to apply them to the indexing code so I haven't done that here. It might be worth a look at some point though.

Or it might be worth moving this to a separate block of utility code and including this only on demand instead of universally.

Cython/Utility/MatchCase.c
scoder: I was going to write that we know the macro value and can probably still read the feature flag through the Limited API function, but that might not apply to other implementations of the Limited API. Still a hypothetical issue at this time, though, since there's only CPython that has an implementation.

Did you try asking the `HasFeature` function in the Limited API for the feature? I wouldn't mind taking that route, maybe restricted to CPython's Limited API implementation (if we can detect that).

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,408 @@
...
+// (detailed in "is_sequence").
+static unsigned int __Pyx_MatchCase_ABCCheck(PyObject *o, int sequence_first, int definitely_not_sequence, int definitely_not_mapping) {
+    // in Python 3.10 objects can have their sequence bit set or their mapping bit set
+    // but not both. Practically this translates to "which type is registered first".
+    // In Python < 3.10 we can only determine this if they're direct bases (by looking
+    // at the MRO order). If they're registered manually then we can't tell
+    // In Python 3.10+, objects can have their sequence bit set or their mapping bit set
+    // but not both. Practically, this translates to "which type is registered first".
+    // In Python < 3.10 we can only determine this if they're direct bases (by looking
+    // at the MRO order). If they're registered manually then we can't tell.
```
scoder: This got duplicated somehow.
```suggestion
    // In Python 3.10+, objects can have their sequence bit set or their mapping bit set
    // but not both. Practically, this translates to "which type is registered first".
    // In Python < 3.10 we can only determine this if they're direct bases (by looking
    // at the MRO order). If they're registered manually then we can't tell.

```

da-woods: The "apply changes" interface was really struggling for some reason. It wouldn't let me commit them as a batch earlier

Cython/Utility/MatchCase.c
da-woods: > Still a hypothetical issue at this time, though, since there's only CPython that has an implementation.

I think GraalPy has an ambition to implement it. But hasn't done so yet.

> I was going to write that we know the macro value and can probably still read the feature flag through the Limited API function, but that might not apply to other implementations of the Limited API.

I think in principle CPython could argue that they could change the flag to be a different number in a future Python version. We all know they probably won't. But I'm trying to be nice and respect the flexibility that they're trying to give themselves.

da-woods: But I have added the check to the Limited API - just with a value that I've extracted at module import (so it should be robust to any future changes)

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -624,7 +998,347 @@ class SubstitutedIfStatListNode(Nodes.StatListNode):
...
+        elif self.base.type.is_pyobject:
+            result = self.generate_for_pyobject()
+        else:
+            # Some other type (probably a ctuple).
+            # Just slice it, copy it to a list and hope it works
+            result = self.generate_via_slicing(env)
+        return result.analyse_types(env)
+
+
+class CompilerDirectivesExprNode(ExprNodes.ProxyNode):
```
scoder: This could probably go to `ExprNodes.py` and share code with `CompilerDirectivesNode` in `Nodes.py`. I'm not sure if a mixin class could help to handle both uniformly in the transforms? It seems risky to require taking both into account for essentially doing the same thing.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -624,7 +998,347 @@ class SubstitutedIfStatListNode(Nodes.StatListNode):
...
+    subexprs = ["arg"]
+    type = PyrexTypes.py_object_type
+
+    def analyse_types(self, env):
+        return self.arg.analyse_types(env).coerce_to_pyobject(env)
+
+
+class LazyCoerceToBool(ExprNodes.ExprNode):
+    """
+    Just calls "self.arg.coerce_to_bool" when it's analysed,
```
scoder: Wrong method name.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -284,43 +353,35 @@ def get_simple_comparison_node(self):
...
-        return self.comp_node.result()
-
-    def generate_comparison_disposal_code(self, code):
-        self.comp_node.generate_disposal_code(code)
-
-    def free_comparison_temps(self, code):
-        self.comp_node.free_temps(code)
-
-    def generate_target_assignments(self, subject_node):
+    def generate_target_assignments(self, subject_node, env):
```
scoder: I only noticed this now, but `node.generate_*()` is the prefix that is commonly used for code generation methods. Maybe use `.create_*` instead?

Cython/Compiler/MatchCaseNodes.py
scoder: I'd at least make it a keyword argument to avoid a bare `None` argument.
```suggestion
                assignments = c.pattern.generate_target_assignments(subject, env=None)
```
That said, the `refactor_cases` method seems to be called only from the `ConstantFolding` transform, which doesn't currently track scopes but potentially could do so. If it's important enough. Not sure if this case makes a case for that, though…

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -80,27 +84,51 @@ def refactor_cases(self):
             new_cases.append(SubstitutedMatchCaseNode(
                 current_if_statement.pos, body=current_if_statement
             ))
-        self.cases = new_cases
+        # eliminate optimized cases
+        self.cases = [c for c in new_cases if c is not None]
```
scoder: I think we can avoid this comprehension if we move the `is not None` condition to the `new_cases.append(c)` shortly above. That seems to be the only case where `None` values can be introduced.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -80,27 +84,51 @@ def refactor_cases(self):
...
         if not self.subject.arg.is_literal:
             self.subject.arg = self.subject.arg.coerce_to_temp(env)
-        subject = self.subject_clonenode
-        self.cases = [c.analyse_case_expressions(subject, env) for c in self.cases]
+        subject = self.subject_clonenode.analyse_expressions(env)
+        self.cases = [
+            c.analyse_case_expressions(subject, env, self.sequence_mapping_temp)
+            for c in self.cases
+        ]
+        self.cases = [c for c in self.cases if c is not None]
```
scoder: We can avoid overwriting the attribute twice here.
```suggestion
        analysed_cases = []
        for c in self.cases:
            c = c.analyse_case_expressions(subject, env, self.sequence_mapping_temp)
            if c is not None:
                analysed_cases.append(c)
        self.cases = analysed_cases
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -80,27 +84,51 @@ def refactor_cases(self):
...
     def analyse_expressions(self, env):
+        sequence_mapping_count = 0
+        for c in self.cases:
+            if c.is_sequence_or_mapping():
+                sequence_mapping_count += 1
+        if sequence_mapping_count >= 2:
+            self.sequence_mapping_temp = AssignableTempNode(
+                self.pos, PyrexTypes.c_uint_type
+            )
+            self.sequence_mapping_temp.is_addressable = lambda: True
```
scoder: If this functionality is needed, it could probably be configured in the constructor of this new (sub-)class rather than injecting a method-like thing after the fact.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -150,22 +179,37 @@ def validate_irrefutable(self):
...
-        self.pattern = self.pattern.analyse_pattern_expressions(subject_node, env)
-        self.original_pattern = self.pattern
+    def analyse_case_expressions(self, subject_node, env, sequence_mapping_temp):
+        with local_errors(True) as errors:
+            self.pattern = self.pattern.analyse_pattern_expressions(env, sequence_mapping_temp)
+            self.comp_node = self.pattern.get_comparison_node(subject_node, sequence_mapping_temp)
+            self.comp_node = self.comp_node.analyse_types(env)
+
+        if self.comp_node and self.comp_node.is_literal:
+            self.comp_node.calculate_constant_result()
```
scoder: This method is designed to end in arbitrary failures that the `ConstantFolding` transform catches and ignores. I'm not sure it's safe to call it outside of that transform. (EDIT: And either way, I wouldn't _expect_ it to get called outside that transform.)

scoder (commented):
I still haven't gone through all of this – it's quite a bite. Here are some minor comments that I had along the way.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -624,7 +998,347 @@ class SubstitutedIfStatListNode(Nodes.StatListNode):
...
+        return self.arg.analyse_boolean_expression(env)
+
+def generate_binop_tree_from_list(pos, operator, list_of_tests):
+    """
+    Given a list of operands generates a roughly balanced tree:
+    (test1 op test2) op (test3 op test4)
+    This is better than (((test1 op test2) op test3) op test4)
+    because it generates a shallower tree of nodes so is
+    less likely to overflow the compiler
+    """
```
scoder, line 1279: It also generates fewer operations. In the left-to-right nested case, all operators must be evaluated. In the balanced case, the RHS operators will be skipped entirely (for `and`) if the LHS already failed.

Given Python's short-circuiting semantics, I wonder if this is something we can apply to chains of `and` operators in general. I think `or` isn't as simple because it might pick up the RHS evaluation at any point.

da-woods: Looking at the code for chains of `and` I think `goto __pyx_L3_bool_binop_done;` keeps the number of comparisons to a minimum so there may not be great advantages to applying it generally.

If I remember right, it was definitely needed here to avoid recursion errors with large sequences.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -624,7 +997,301 @@ class SubstitutedIfStatListNode(Nodes.StatListNode):
...
+
+
+class AssignableTempNode(ExprNodes.TempNode):
+    lhs_of_first_assignment = True  # assume it can be assigned to once
+    _assigned_twice = False
+    _is_addressable = False
+
+    def __init__(self, pos, *args, **kwds):
+        self._is_addressable = kwds.pop('is_addressable', False)
+        super().__init__(pos, *args, **kwds)
```
scoder, line 1053:
```suggestion
    def __init__(self, pos, *args, is_addressable=False, **kwds):
        self._is_addressable = is_addressable
```

Cython/Compiler/ParseTreeTransforms.py
scoder, line 2096: Might be a reason to make the method public so that it's just
```python
    def visit_CompilerDirectivesMixin(self, node):
        with node.apply_directives(self.module_scope):
            self.visitchildren(node)
        return node
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,404 @@
...
+    }
+    return 0;
+}
+
+static CYTHON_INLINE int __Pyx_MatchCase_IsExactMapping(PyObject *o) {
+    // Py_Dict is the only regularly used mapping type.
+    // "types.MappingProxyType" also exists but is correctly covered by
+    // the isinstance(o, Mapping) check.
+    return PyDict_CheckExact(o);
+}
```
scoder, line 21: frozendict ?

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,404 @@
...
+        // it's slightly more efficient than using PySequence_GetItem since it skips negative indices.
+        slot = PySequence_ITEM;
+        #else
+        slot = PySequence_GetItem;
+        #endif
+    }
+
+    for (i=start; i<end; ++i) {
+        PyObject *obj = slot(x, i);
+        if (unlikely(!obj || __Pyx_PyList_SET_ITEM(list, i-start, obj) == -1)) {
```
scoder: We can make it clearer that both conditions are unlikely independently of each other.
```suggestion
        if (unlikely(!obj) || unlikely(__Pyx_PyList_SET_ITEM(list, i-start, obj) == -1)) {
```

Cython/Utility/ModuleSetupCode.c
```diff
@@ -1315,10 +1328,84 @@ static int __Pyx_init_co_variables(void) {
...
+        count += (flag & 1);
+        flag >>= 1;
+    }
+    return count;
+}
+
+// It's quite possible that the flags aren't reliably available at runtime.
+// Therefore most errors are ignored and we just leave the values as 0.
+static int __Pyx_init_tpflags_variables(void) {
+    if (__Pyx_Runtime_TPFLAGS_MAPPING != 0 && __Pyx_Runtime_TPFLAGS_MAPPING != 0) {
```
scoder: Should one of these have been `…_SEQUENCE` ?

Cython/Utility/ModuleSetupCode.c
```diff
@@ -731,6 +731,19 @@ static int __Pyx_init_co_variables(void); /* proto */
   #define __Pyx_PyObject_GC_IsFinalized(o) PyObject_GC_IsFinalized(o)
 #endif
 
+#if CYTHON_COMPILING_IN_LIMITED_API
+// These can be deduced at runtime and are enough of an optimization that
+// it's worth doing (while still respecting the decision not to add them to
+// the Limited API).
+static long __Pyx_Runtime_TPFLAGS_SEQUENCE;
+static long __Pyx_Runtime_TPFLAGS_MAPPING;
```
scoder: Flags should generally be `unsigned long` (as is the return type of `PyType_GetFlags()`).

scoder (commented):
I think I looked through most of this now. Some more comments below, but it's mostly ok so that we can merge it and continue with the next steps.

There is a lot of legacy code for "Py<3.10", which nowadays really means only Py3.9. Since pattern matching was added in Py3.10, it's not certain that users really expect us to backport it to Py3.9. The code is there, so that's that, but there's certainly the option of requiring Py3.10+ (also in the Limited API) for modules that use pattern matching and just fail hard to C-compile otherwise.

Since you've written the code and would probably be the first to maintain it, I'll leave the decision to you whether you want to include the fallback code or remove it.

scoder:
The build currently fails with
```
@cython.cfunc
def p_or_test(s: PyrexScanner):
    return p_rassoc_binop_expr(s, 'or', p_and_test)
                                        ^
Cython/Compiler/Parsing.py:193:40: Cannot assign type 'object (PyrexScanner, int __pyx_skip_dispatch)' to 'p_sub_expr_func' (alias of 'object (*)(PyrexScanner)')
```
Meaning "cannot assign cpdef function to function pointer". Not sure right now what changed here.

scoder:
Could you try moving the `CompilerDirectivesMixin` base class first in the class declarations, before the actual node class?

Cython/Utility/MatchCase.c
da-woods: Updating to make sure that frozendict works and is optimized was on my list of things to do to the `mapping` PR. But I'll update this now since you spotted it.

da-woods:
> There is a lot of legacy code for "Py<3.10", which nowadays really means only Py3.9. Since pattern matching was added in Py3.10, it's not certain that users really expect us to backport it to Py3.9. The code is there, so that's that, but there's certainly the option of requiring Py3.10+ (also in the Limited API) for modules that use pattern matching and just fail hard to C-compile otherwise.
> 
> Since you've written the code and would probably be the first to maintain it, I'll leave the decision to you whether you want to include the fallback code or remove it.

I think I'd prefer to keep the general principle of "backport features where we can" so I'll leave it in. I'm fairly confident that it shouldn't need too much maintenance (probably because no-one will ever actually end up running it...).

da-woods:
> Could you try moving the `CompilerDirectivesMixin` base class first in the class declarations, before the actual node class?

Yes that fixed it

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,405 @@
...
+    return list;
+}
+
+////////////////////// TupleSliceToList.proto //////////////////////
+
+static PyObject *__Pyx_MatchCase_TupleSliceToList(PyObject *x, Py_ssize_t start, Py_ssize_t end); /* proto */
+
+////////////////////// TupleSliceToList //////////////////////////
+//@requires: OtherSequenceSliceToList
+//@requires: ObjectHandling.c::TupleFromArray
```
scoder:
```suggestion
//@requires: ObjectHandling.c::ListFromArray
```

Cython/Utility/ModuleSetupCode.c
```diff
@@ -1315,10 +1328,84 @@ static int __Pyx_init_co_variables(void) {
     Py_DECREF(inspect);
     return result ? 0 : -1;
 }
+
+static int __Pyx_init_tpflags_bitcount(long flag) {
```
```suggestion
static int __Pyx_init_tpflags_bitcount(unsigned long flag) {
```

Cython/Utility/MatchCase.c
```diff
@@ -0,0 +1,405 @@
...
+    ssizeargfunc slot;
+
+    list = PyList_New(total);
+    if (!list) {
+        return NULL;
+    }
+
+    slot = __Pyx_PyObject_TryGetSubSlot(x, tp_as_sequence, sq_item, ssizeargfunc);
+    if (!slot) {
+        #if !defined(Py_LIMITED_API) && !defined(PySequence_ITEM)
```
scoder: Do you really mean to test for `Py_LIMITED_API` here?
```suggestion
        #if !CYTHON_COMPILING_IN_LIMITED_API && !defined(PySequence_ITEM)
```

scoder:
CI is green, high time to get this merged. :)

PR #7431, opened 2025-12-26
Improve benchmarks scaling for sub-benchmarks

scoder:
… by scaling to the fastest benchmark instead of the slowest.

Scale back the timings of sub-benchmarks to the outer scale count to report comparable timings (which will be divided by the outer scale for reporting the per-loop runtime).


scoder:
Merged as part of https://github.com/cython/cython/pull/7454

PR #7432, opened 2025-12-26
Speed up sequence subscripting using Py_TPFLAGS_SEQUENCE

scoder:
Python semantics dictate that we first try the mapping protocol and then the sequence protocol for subscripting. When the index is a C integer, we can optimise perfectly for list/tuple, but all other sequences suffer from having to build a Python `int` object for the index to pass it through the mapping lookup if they implement that (e.g. to support extended slicing, like NumPy arrays).

Python 3.10 added type markers (for pattern matching) for explicitly declaring a type as sequence or mapping, called `Py_TPFLAGS_SEQUENCE` and `Py_TPFLAGS_MAPPING`, which can now be checked for quite quickly. If a type is marked as sequence but still implements mapping lookups for slicing, and it supports sequence subscripting, we can avoid the Python `int` creation of the mapping protocol and go straight through the sequence index lookup.

With this change, indexing into Python's `array.array` and `memoryview` types is ~60% faster in a micro-benchmark.
Using a C integer as dict key got slightly slower but is resolved by adding a separate up-front special case.

Benchmark is based on https://github.com/cython/cython/pull/7431

See
https://docs.python.org/3/c-api/typeobj.html#c.Py_TPFLAGS_SEQUENCE
https://github.com/cython/cython/issues/1807
https://github.com/pandas-dev/pandas/pull/55915
https://github.com/pandas-dev/pandas/pull/55179#issuecomment-1802484187

### DIFF
--- a/Tools/ci-run.sh
+++ b/Tools/ci-run.sh
@@ -81,7 +81,7 @@ if [[ $PYTHON_VERSION == "3.1"[2-9]* || $PYTHON_VERSION == *"-dev" || $PYTHON_VE
   python -m pip install -U pip wheel setuptools || exit 1
 else
   # Drop dependencies cryptography and nh3 (purely from twine) when removing support for PyPy3.10.
-  python -m pip install -U pip "setuptools<60" wheel<0.46 twine "cryptography<42" "nh3<0.2.19" || exit 1
+  python -m pip install -U pip "setuptools<60" "wheel<0.46" "twine" "cryptography<42" "nh3<0.2.19" || exit 1
 fi
 if [[ $PYTHON_VERSION != *"t" && $PYTHON_VERSION != *"t-dev" && $PYTHON_VERSION != "graalpy"* ]]; then
   # twine is not installable on freethreaded Python due to cryptography requirement


da-woods:
I think this is reasonable although it's arguably a slight abuse of the flag (which is really only documented as being for pattern matching).

I think a type type that gives a different answer if you go through `tp_sequence` and `tp_mapping` is pretty broken. So from that point of view it seems fine to treat this flag as a strong hint and if anyone actually notices a difference in behaviour it's most likely on them.

Cython/Utility/ModuleSetupCode.c
```diff
@@ -747,10 +747,10 @@ static int __Pyx_init_co_variables(void); /* proto */
   #define Py_TPFLAGS_HAVE_FINALIZE 0
 #endif
 #ifndef Py_TPFLAGS_SEQUENCE
-  #define Py_TPFLAGS_SEQUENCE 0
+  #define Py_TPFLAGS_SEQUENCE (1 << 5)
```
da-woods:
```suggestion
  #define Py_TPFLAGS_SEQUENCE (CYTHON_COMPILING_IN_LIMITED_API ? 0 : 1 << 5)
```

It looks like these are purposefully excluded from the Limited API (i.e. they're in a big `#ifndef Py_LIMITED_API` block (https://github.com/python/cpython/blob/9712dc1d9eb03ffa96ed746d20bd43239d251ba7/Include/object.h#L535) unlike some of the other flags here which just become available in later versions.

I think we should probably follow the spirit of the Limited API and not define them there.

I don't think that should affect of the behaviour in this PR (which I think is all `!CYTHON_COMPILING_IN_LIMITED_API`) so it's more about making it harder to use them by accident in future.

Cython/Utility/ModuleSetupCode.c
```diff
@@ -747,10 +747,10 @@ static int __Pyx_init_co_variables(void); /* proto */
...
 #endif
 #ifndef Py_TPFLAGS_MAPPING
-  #define Py_TPFLAGS_MAPPING 0
+  #define Py_TPFLAGS_MAPPING (1 << 6)
```
da-woods:
```suggestion
  #define Py_TPFLAGS_MAPPING (CYTHON_COMPILING_IN_LIMITED_API ? 0 : 1 << 6)
```

scoder:
NumPy will set the sequence flag in the future: https://github.com/numpy/numpy/issues/30519
Given the expected speedups, I'll merge this as soon as CI is happy.

Issue #5729, opened 2023-09-27
[ENH] special-case `__getitem__(self, Py_ssize_t index)`


scoder:
### Is your feature request related to a problem? Please describe.

We currently generate a wrapper for the `sq_item` C-API slot that takes the C integer index, creates a Python int object from it, and forwards it to `mp_subscript`. If users define their `__getitem__` method with a C integer type as argument, then this becomes 1) pass C integer, 2) convert to Python int, 3) pass into `mp_subscript`, 4) unpack C integer object from Python int object, 5) run the method with it.

### Describe the solution you'd like.

For code like
```cython
def __getitem__(self, Py_ssize_t index):
    ...
```
Cython could put a wrapper into `mp_subscript` instead of `sq_item`, which would unpack the Python index value into the expected C integer and then call `sq_item`.

### Additional context

One drawback is that Python (and Cython 3.0) check for the `mp_subscript` slot first and, if available, prefer it over `sq_item`. That is already a problem now.

Maybe we could avoid setting `mp_subscript` all together if the `__getitem__` argument has a C integer type?

scoder:
> Maybe we could avoid setting `mp_subscript` all together if the `__getitem__` argument has a C integer type?

Exploring this some more, if it hits the case that calling code can pass a C integer into `sq_item()` directly, it's much faster by ~60% according to tests in https://github.com/cython/cython/pull/7434.

If it needs to test the (then not implemented) mapping slot first and only then the sequence slot, requiring it to unpack an existing Python integer, we lose a tiny bit of performance compared to directly passing the Python integer into the mapping slot and unconditionally unpacking it on the other side. However, that is dwarfed by the large gain of allowing foreign code to pass C integers through the sequence interface.

In the end, it's a trade-off. Code that is mostly used by Python code that indexes with a Python integer will suffer a bit. If the subscripting is mostly done by C code that can avoid passing through a Python integer, we'll gain a lot. Not implementing the mapping slots allows this code to bypass the "mapping first" semantics in Python that would otherwise induce inefficient index passing.

Users have a way to avoid the downside by simply not typing their index argument but converting it inside of the method. Thus, I think the decision is clear. If users decide to type the index argument as a C integer, we should interpret it as a sign to implement only the sequence slot and not the mapping slot.

scoder:
Quoting from https://github.com/cython/cython/pull/7435#issuecomment-3722589798 :

[...]

PR #7434, opened 2025-12-27
Implement sequence and mapping protocols depending on user signature

scoder:
Implement both as before, but prefer sq_item() over mp_subscript() if the index argument of the user implemented "`__getitem__`" method is a C integer. Do the same split for setitem/delitem.

Closes https://github.com/cython/cython/issues/5729

As noted in the ticket, an alternative would be to not implement `mp_subscript()` etc. at all (only `sq_item()`) if the corresponding special method uses a C integer index. This would be faster for existing usages in Cython modules and Python would then do the unpacking itself.

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -2237,7 +2237,7 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
                 "r = Py_TYPE(o)->tp_as_sequence->sq_item(o, x);")
             code.putln("#else")
             code.putln(
-                "r = ((binaryfunc)PyType_GetSlot(Py_TYPE(o), Py_sq_item))(o, x);")
+                "r = ((ssizeargfunc)PyType_GetSlot(Py_TYPE(o), Py_sq_item))(o, x);")
             code.putln("#endif")
             code.putln(
                 "return r;")


Cython/Compiler/ModuleNode.py
```diff
@@ -2190,27 +2188,62 @@ def generate_clear_function(self, scope, code, cclass_entry):
...
+            code.putln(
+                "Py_ssize_t x = __Pyx_PyIndex_AsSsize_t(i); if (unlikely(x == -1 && PyErr_Occurred())) return 0;")
+            # Note that PyType_GetSlot only works on heap-types before 3.10, so not using type slots
+            # and defining cdef classes as non-heap types is probably impossible
+            code.putln("#if CYTHON_USE_TYPE_SLOTS || (!CYTHON_USE_TYPE_SPECS && __PYX_LIMITED_VERSION_HEX < 0x030A0000)")
+            code.putln(
+                "r = Py_TYPE(o)->tp_as_sequence->sq_item(o, x);")
+            code.putln("#else")
+            code.putln(
+                "r = ((binaryfunc)PyType_GetSlot(Py_TYPE(o), Py_sq_item))(o, x);")
```
da-woods: `(binaryfunc)` doesn't look right to me here. I think it should be `(ssizeargfunc)`

scoder: Thanks. There was obviously a lot of back and forth copying and editing going on on my side…

scoder:
> not implement `mp_subscript()` etc. at all (only `sq_item()`) if the corresponding special method uses a C integer index

I tried it by simply setting `tp_as_mapping` to 0 for a suitable extension type. It doesn't seem to make a clear difference.

Thus, intuitively, it seems better to avoid lying about the capabilities of the type and only expose the straight efficient path through `sq_item()`. That allows calling (i.e. subscripting) code to take appropriate decisions itself.

See my comment in https://github.com/cython/cython/issues/5729#issuecomment-3694570997

da-woods:
> Thus, intuitively, it seems better to avoid lying about the capabilities of the type and only expose the straight efficient path through `sq_item()`. That allows calling (i.e. subscripting) code to take appropriate decisions itself.

Yes - that seems sensible. Exposing only the efficient path is also likely to work well with any future improvements in the Python interpreter (e.g. if they implement something like tagged integers and can avoid a Python object conversion)

scoder:
I think this is clearly superseded by https://github.com/cython/cython/pull/7435. Even if we decide to keep the slot implemented (or come back to that later), the other PR is a better basis for future changes.

PR #7435, opened 2025-12-28
Do not implement mapping slots if sequence type method has sequence index argument types

scoder:
If users explicitly declare a type as collection type `sequence`, and implement the subscript special methods with an integer index argument, it seems better to avoid lying about the capabilities of the type and only expose the straight efficient path through `sq_item()`. That allows calling (i.e. subscripting) code to take appropriate decisions itself.

Closes https://github.com/cython/cython/issues/5729
Based on https://github.com/cython/cython/pull/7434
See comment in https://github.com/cython/cython/pull/7434#issuecomment-3694571349

### DIFF
--- a/Demos/benchmarks/bm_getitem.py
+++ b/Demos/benchmarks/bm_getitem.py
@@ -20,7 +20,7 @@ class GetItemExt:
 
 
 # Decorator @collection_type() is not supported by older Cython versions.
-#[3.3+getitem_split] #@cython.collection_type("sequence")
+#[3.3+] #@cython.collection_type("sequence")
 @cython.cclass
 class GetItemExtSequence:
     _len: cython.int
@@ -33,7 +33,7 @@ class GetItemExtSequence:
         return 5  # do not waste time on PyLong_FromSsize_t()
 
 # Decorator value 'mapping' is not supported by older Cython versions.
-#[3.3+getitem_split] #@cython.collection_type("mapping")
+#[3.3+] #@cython.collection_type("mapping")
 @cython.cclass
 class GetItemExtMapping:
     _len: cython.int


Cython/Compiler/ModuleNode.py
```diff
@@ -2190,81 +2188,153 @@ def generate_clear_function(self, scope, code, cclass_entry):
         code.putln("}")
         code.exit_cfunc_scope()
 
-    def generate_getitem_int_function(self, scope, code):
-        # This function is put into the sq_item slot when
-        # a __getitem__ method is present. It converts its
-        # argument to a Python integer and calls mp_subscript.
+    def generate_getitem_function(self, scope, code):
+        # Implement both 'sq_item()' and 'mp_subscript()'.
+        # Use a delegating wrapper for the less suitable slot.
```
da-woods: I don't think this comment is (quite) true any more (at least it doesn't do that for `mp_subscript`)

Cython/Compiler/ModuleNode.py
```diff
@@ -2190,81 +2188,153 @@ def generate_clear_function(self, scope, code, cclass_entry):
...
+        if not is_sequence_get:
+            code.start_slotfunc(scope, PyrexTypes.py_objptr_type, "mp_subscript", "PyObject *o, PyObject *i", needs_funcstate=False)
+            code.putln(f"return {get_entry.func_cname}(o, i);")
+            code.putln(
+                "}")
+            code.exit_cfunc_scope()
+
     def generate_ass_subscript_function(self, scope, code):
         # Setting and deleting an item are both done through
         # the ass_subscript method, so we dispatch to user's __setitem__
```
da-woods: Again, this maybe needs updating?

da-woods (commented):
Try

```
cdef class Base:
    def __setitem__(self, Py_ssize_t i, o):
        print(f"__setitem__{i, o}")

cdef class Derived(Base):
    def __delitem__(self, i):
        print(f"__delitem__{i}")

b = Base()
b[1] = "a"

d = Derived()
d[1] = "a"
```

This gives me:

```
NotImplementedError: Subscript assignment not supported by sqmap.Derived
```

while I think it should probably succeed.

(I think there's also an alternative case the other way around, but that probably needs a non-Cython non-builtin extern base class to reach).

scoder:
> Try: … This gives me: …

Right, inheritance doesn't look right yet when mixing the two protocols. That might have been one of the reason why previously both were always implemented.

I'll generate tests for all the cases, also for Python inheritance from extension types, and see what that gives.

scoder:
What's actually tricky is the other way round, overriding the Mapping protocol with the Sequence protocol, e.g. for getitem:
```cython
cdef class Base:
    def __getitem__(self, i):
        print(f"__getitem__{i}")

cdef class Derived(Base):
    def __getitem__(self, Py_ssize_t i):
        print(f"__getitem__{i}")
```
Here, the derived type only implements `sq_item`, but the `mp_subscript` slot of the base type is inherited and takes precedence, so subscripting still calls that instead and ignores the subtype's method (unless explicitly going through the Sequence protocol).

da-woods:
I guess the simplest thing would be to implement both slots if there's any inheritance. Obviously that's less efficient but it's fairly straightforward. And typically container types probably wouldn't have multi-layer inheritance anyway.

I'm not sure if that quite fixes my inheritance issue (but I suspect that's relatively easily handled by trying both options)

scoder:
> implement both slots if there's any inheritance

Done.

> I'm not sure if that quite fixes my inheritance issue

I chose to look up the basetype methods to get the full picture, which resolves this particular issue.

It does not help with the case of an extern base class that we don't see into.

> I suspect that's relatively easily handled by trying both options

Trying both options can easily lead to infinite recursion. If we only ever go one way, that's easy to prevent, but to try both, we probably have to look at the function pointers in the call hierarchy to see if we're calling ourselves. That's brittle.

da-woods:
Is there a potential problem with negative numbers?

Let's suppose I'm implementing a wrapping around the C++ class `std::unordered_map<int, double>` (essentially a `dict` with integer keys).

In that case I'm likely to implement `__getitem__(self, int key)` which I think would trigger your sequence code path. However, the sequence lookup implies that negative numbers get wrapped. That almost certainly isn't what is intended.

Also: if I'm implementing an actual sequence I probably to accept `slice` too (so in an ideal world I'd have 2 functions - an optimized path for `int` indices and but also a function that handles slices to make it behave like the Python sequence types.

To mean that suggests that most fully-featured sequence types will get missed by this optimization, but typed mapping types will probably be included.

da-woods:
A few more thoughts on negative numbers:

* We could tie this feature to `@cython.collection_type('sequence')`. That'd work well with #7432 but wouldn't provide any benefit by default.
* if we don't define `sq_length` then the auto-wraparound doesn't happen. So that would mean the number gets passed directly to `__getitem__` as is.
  * That'd would alter any current code that's manually doing `PySequence_GetItem` on `cdef classes`. The biggest user of `PySequence_GetItem` I can find with a quick github search is Pandas. They don't appear to do it with negative numbers, but it might be worth a manual check.
  * One option there is to define `sq_item` for any `__getitem__` with an integer type but _only_ define `sq_length` for things tagged `@cython.collection_type('sequence')`. That would avoid the conversions but might be a bit subtle.

scoder:
I think your observations regarding negative indexing are correct. That is one of the reasons why the mapping protocol is generally preferred over the sequence protocol if both are implemented.

> The biggest user of `PySequence_GetItem` I can find with a quick github search is Pandas

Apparently, you suggested that they use it in https://github.com/cython/cython/issues/1807#issuecomment-1802656280, leading to https://github.com/pandas-dev/pandas/pull/55915.

> We could tie this feature to `@cython.collection_type('sequence')`.

I think that's the safest bet. The way I've now written it in this PR is centrally taking the decision in `TypeSlots.py`, so it's easy to change and adapt. I'm fine with restricting it to the safe cases for now (especially given that `collection_type` is essentially a new feature that existing code doesn't use).

scoder:
Now I think the only remaining difference to before is that we don't generate `mp_subscript` and friends for `collection_type("sequence")` if the slot methods uses a C integer index type. Not a big difference in the end.

scoder:
@da-woods I think your concerns are resolved by restricting this to `collection_type("sequence")`, which a mapping type would not have. So users must explicitly opt into the sequence protocol and will only get it exclusively (without mapping protocol) if their subscript methods use integer indices.

Can we merge this?

docs/src/userguide/special_methods.rst
```diff
@@ -454,13 +454,21 @@ https://docs.python.org/3/reference/datamodel.html#emulating-container-types
 +-----------------------+---------------------------------------+-------------+-----------------------------------------------------+
 | __delitem__ 	        | self, x 	  	                |             | del self[x]                                         |
 +-----------------------+---------------------------------------+-------------+-----------------------------------------------------+
-| __getslice__ 	        | self, Py_ssize_t i, Py_ssize_t j 	| object      | self[i:j]                                           |
+| __contains__ 	        | self, x 	                        | int 	      | x in self                                           |
++-----------------------+---------------------------------------+-------------+-----------------------------------------------------+
+
+To target specifically the sequence protocol (rather than the more general mapping protocol), implement the signatures as follows.
```
da-woods: This should probably be updated to mention `collection_type`

PR #7438, opened 2025-12-29
Improve fstring error reporting

### LINKED ISSUE #7436: possible incorrect error message during compilation of invalid f-string
Please consider `t.pyx`
```
DEF value = 1

print(f'value {value)}')
```

Compile with Cython 3.2.3
```
$ cython t.pyx

Error compiling Cython file:
...
DEF value = 1

print(f'value {value)}')
                        ^
t.pyx:3:24: Unclosed string literal
```

I would expect better error, for example Python gives
```
SyntaxError: f-string: unmatched ')'
```

da-woods:
Error on miscounted brackets inside an fstring expression rather rather than waiting until the string has finished to complain. Closes #7436


scoder:
Seems helpful, but adding the reported issue as error test would be nice to make sure this actually prints the right error and position.

da-woods:
> Seems helpful, but adding the reported issue as error test would be nice to make sure this actually prints the right error and position.

Added - unfortunately the errors end up stopping the compilation so it isn't possible to bundle multiple errors into a single file

PR #7439, opened 2025-12-31
Avoid immortalizing anything shared on freethreading

da-woods:
Py_SET_REFCNT isn't thread-safe so there's the
risk of races if another thread is accessing it at the same time. This will mostly apply to things that are already immortal so I don't think we'll lose much from the check.

On Python 3.13t we have to use a slightly weaker check than the full "IsUniquelyReferenced" - I think this is fine because we know that our "1" reference is not yet accessible anywhere external, however I've still chosen to use the full check where available.

### DIFF
--- a/Cython/Compiler/Code.py
+++ b/Cython/Compiler/Code.py
@@ -2434,7 +2434,7 @@ class GlobalState:
         # because setting the refcount isn't thread-safe. The chances are that most of the constants
         # that this applies to are already immortal though so that isn't a great loss.
         writer.putln("#if PY_VERSION_HEX < 0x030E0000")
-        writer.putln("if (Py_REFCNT(table[i]) == 1)")
+        writer.putln("if (_Py_IsOwnedByCurrentThread(table[i]) && Py_REFCNT(table[i]) == 1)")
         writer.putln("#else")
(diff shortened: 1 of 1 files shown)


da-woods:
I think the main source of "shared but not immortal" objects is interned unicode strings, which are already excluded. Therefore I'm not too worried about any *real* bugs here.

ngoldbaum:
I happened to notice this doing end-of-month reporting for the free-threading project here at Quansight - hope you don't mind me randomly jumping in.

> On Python 3.13t we have to use a slightly weaker check than the full "IsUniquelyReferenced"

You can do a little better than that if you follow what happens in the pythoncapi-compat header: https://github.com/python/pythoncapi-compat/blob/11cb80f2652cb2fe5231bf60b9dd98c83a4e25f4/pythoncapi_compat.h#L2220-L2227

da-woods:
> You can do a little better than that if you follow what happens in the pythoncapi-compat header

Thanks. That's useful. I've added a `_Py_IsOwnedByCurrentThread` check. I'm not worried about the case where the refcount is 0 local, 1 non-local because we *know* we have 1 currently inaccessible local reference ourselves so I've kept it simple and not copied the full atomic reads.

PR #7442, opened 2026-01-02
Remove exclusion of collection_type in test

da-woods:
See 9f518b7c3b6792a3f969617a11da83e600ac8764.

This should have been caught by the test, but I'd specifically excluded it when I first added `collection_type` for internal use only..


scoder:
3.2.x commit in https://github.com/cython/cython/commit/d347c82898f4e90139b63510d984860762c01f1c

Issue #7443, opened 2026-01-03
[BUG] Auto-pickle is inefficient in 3.2


scoder:
### Describe the bug

The optimisation in https://github.com/cython/cython/commit/065b8eaa131637549fa3250b14d5f45bd9ad475d left a spurious trailing comma in the comparison code (line 2326) that turned a plain value into a non-empty tuple, which is always true. As effect, the pickling code always stored the object state, even when empty.

### Cython version

3.2.0b1 - 3.2.3

PR #7444, opened 2026-01-03
Fix accidental inefficiency in auto-pickling that always stored the object state

scoder:
The condition code contained a trailing comma which changed it into a non-emtpy tuple that is always true. This was introduced in 3.2.0b1 by https://github.com/cython/cython/commit/065b8eaa131637549fa3250b14d5f45bd9ad475d

Fixes https://github.com/cython/cython/issues/7443

### DIFF
--- a/tests/run/reduce_pickle.pyx
+++ b/tests/run/reduce_pickle.pyx
@@ -319,3 +319,35 @@ else:
         >>> b == NoPyMembers(i=11, x=1.5) or b
         True
         """
+
+OLD_310_PICKLE = b'''\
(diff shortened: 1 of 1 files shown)


da-woods:
If I understand his correctly, it only picks a more efficient version of saving the data but Cython 3.2.3 pickles can still be on 3.2.4 (and the other way around)?

I've usually thought we don't want to guarantee any long-term stability for our pickled data but we probably want to avoid breakages in minor versions if possible. But I don't think there's any breakage here?

scoder:
Right, I think it's all safe. The variant that avoids setstate just stores a smaller tuple, but both variants can be unpickled.

scoder:
I actually tested it and the 3.2.4 pickle output will be the same as pre-3.2.0 again.

PR #7445, opened 2026-01-05
Fix graalpy CI

Ariouz:
Hi, I'm still working on this PR, what would be an acceptable run time for the graalpy ci run ?

### DIFF
--- a/tests/graal_bugs.txt
+++ b/tests/graal_bugs.txt
@@ -96,6 +96,7 @@ memoryview
 cythonarray
 dict_iter_unpack
 asyncio_leaks
+test_benchmark_timer_ns
 
 # probably just GC?
 double_dealloc_T796


scoder:
> what would be an acceptable run time for the graalpy ci run ?

Depends. You definitely want to make use of ccache since a lot of time is spent in the C compiler. And that needs a warm cache to be useful. So you'll likely have longer run times in the beginning that will then decrease once there's a successful run. So just to give a number, anything below one hour might still be acceptable for an initial complete run (and it looks like you're not there yet).

da-woods:
The GraalPy run was originally disabled because the list of failing tests had changed dramatically when GraalPy was updated. So it wasn't the run time that was the main problem - just that it was failing and it looked like a lot of work to fix (and work that would probably need repeating at the next version update too). 

I think it used to be about an hour, and that'd probably be OK still.

I kind of disagree with @scoder about ccache - GraalPy ran Cython really slowly so the C compile time is comparatively less important. Ccache is still worth having but may not get you dramatic improvements.

da-woods:
It looks like you have one failing test. And it's really just testing a benchmark tool which probably isn't hugely important if it works on GraalPy

Ariouz:
Looks like the list is now up to date, its runs in [~1hr](https://github.com/Ariouz/cython/actions/runs/20986665513/job/60322102236)
We'll look into some fails to fix them on our side, if this looks good to you I'll set the PR as ready

tests/graal_bugs.txt
```diff
@@ -75,6 +132,70 @@ reimport_from_subinterpreter
...
+int_literals
+py_mix_by_neg1
+cpython_capi
+pyarray
+pyint
 
 # appears to run forever
 common_include_dir
+
+# concurrent.futures.process.BrokenProcessPool
```
da-woods, line 154: For reference that just means that the test has crashed - it's just multiprocessing responding to one of the processes stopping suddenly.

(Not proposing a change here - just explaining the message).

da-woods:
I'll merge this - it's probably slightly slower than we'd like but we can live with it

PR #7446, opened 2026-01-06
Performance tests: Skip measuring python command duration when GNU time is not installed

matusvalo:


matusvalo:
When GNU time is not installed it is confusing because terminal understands time command (shell builtin) but yet running perf test fails. Maybe the message can be further improved for better explaining reason of the fail.

scoder:
Ah, yes, that's an issue. However, instead of telling users that their system doesn't work, I'd rather disable the `time` calls if the program isn't available. It's just additional benchmark numbers, if they are missing, so what.

matusvalo:
>  I'd rather disable the time calls 

Done

PR #7452, opened 2026-01-09
Special case Python comparisons of int/float objects

scoder:
Use dedicated utility code to compare int/float objects based on injected type checks.

### DIFF
--- a/Cython/Utility/Optimize.c
+++ b/Cython/Utility/Optimize.c
@@ -1073,6 +1073,7 @@ static CYTHON_INLINE {{c_ret_type}} __Pyx_PyObject_Compare{{'' if ret_type.is_py
 /////////////// PyObjectCompare ///////////////
 
 {{py: c_ret_type = 'PyObject*' if ret_type.is_pyobject else 'int'}}
+{{py: func_suffix = f"{'' if ret_type.is_pyobject else 'Bool'}{op}"}}
 {{py: return_true = 'goto __pyx_return_true'}}
 {{py: return_false = 'goto __pyx_return_false'}}
 {{py: return_error = "return NULL" if ret_type.is_pyobject else "return -1"}}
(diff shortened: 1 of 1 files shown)


da-woods:
I'm a bit sceptical of the `int`-`float` comparisons in the `has_object` case. They generate quite a bit of code and I suspect they aren't used all that much. (Obviously if you know you have an `int` and `float` up-front then it makes sense).

scoder:
> I'm a bit sceptical of the `int`-`float` comparisons in the `has_object` case.

My intuition is that we won't know the Python type in many if not most cases, so the object case is important. And at least comparing floats to `0` or other integers is not so uncommon – but we have better dedicated code for the case of a known constant already.

I have to agree with your hunch that most comparisons to non-literals will be homogeneous. I looked at the assembly and the generated mixed comparison functions are about 250-600 bytes of code on my side. The simple float-float comparison is not even 100 bytes small, int-int is about 150 bytes. So it adds less than 400 bytes per comparison to include the mixed comparisons. It's certainly visible but I'm not sure I'd consider it all too bad.

I thought about extracting them into a separate function but then reusing that function is a bit tricky because it's really used twice, just with swapped arguments (and, thus, operator). We don't really have a good way to define utility code dependencies on templated code.

scoder:
I split it into separate functions and that might actually have made it even faster, probably due to better branch prediction. We'll see what the benchmarks say.

Cython/Utility/Optimize.c
```diff
@@ -1065,6 +1065,309 @@ static PyObject* __Pyx__PyNumber_PowerOf2(PyObject *two, PyObject *exp, PyObject
...
+        other_type = type2 if operand == 'op1' else type1
+        if other_type == expected:
+            check = f"likely({check})"
+    return check
+}}
+
+{{if type1 in ('object', 'float') and type2 in ('object', 'int')}}
+// Less likely, non-inlined comparison of float and int.
+static {{if (type1, type2) == ('float', 'int')}}CYTHON_INLINE{{endif}} {{c_ret_type}}
+__Pyx_PyObject_CompareFloatInt{{'' if ret_type.is_pyobject else 'Bool'}}{{op}}_{{type1}}_{{type2}}(PyObject *op1, PyObject *op2) {
```
da-woods: The other option for deduplicating these (rather than adding `{{type1}}_`{{type2}}` to the name, would be to enclose them in an

```
#ifndef __PYX_COMPARE_FLOAT_INT_{{OP}}_DEFINED
#define __PYX_COMPARE_FLOAT_INT_{{OP}}_DEFINED
// ...
#endif
```

That doesn't let you have different versions of `CYTHON_INLINE` but it does get rid of duplicate code.

scoder: … although the intention for this is to get it inlined in most cases, so there won't be much of an advantage for the binaries in the end. It's still a good idea, there's no need to let the C compiler analyse the function more than once. Thanks.

scoder:
The comparisons are now faster throughout, even for the Limited API.

The module sizes went up by a couple of KB, which is to be expected due to inlining.

Given Cython's intention to generate fast code, even at the expense of generating more of it, I think this is a reasonable tradeoff.

da-woods:
Yeah I agree - it seems worthwhile.

I'm not too worried about module sizes in general - I'm mostly just in favour of trying to fix excessive duplication where it exists

PR #7454, opened 2026-01-09
Benchmarks: Improve autoscaling and add a benchmark for subscripting

scoder:
While working on the subscript benchmark, I noticed that the autoscaling (originally copied from Python's timeit command) often misbehaved for short-running benchmarks. Here's a better and faster implementation.

I also improved the overall reporting by scaling the separately adjusted sub-benchmark results back to the overall scale for the benchmark module.

To make use of new features in newer Cython versions, I added a way to mark code lines
as "version x.y+ only" and comment them out otherwise.

And I improved the output of errors and warnings and started showing the timings as early as possible to give an idea of the behaviour without waiting for the whole run to finish.


scoder:
Merging this to make use of it in my other PRs.

PR #7456, opened 2026-01-11
Test limited api on higher versions

da-woods:
This does a few things:
* Adds a `--no-compile` argument to `runtests.py` to let you go through the test suite without recompiling anything but just rerunning the tests.
* Rerun the tests on higher Python versions in the CI for the Limited API jobs.


da-woods:
* The 3.9 tests are mostly just failing because a few tests use the buffer protocol but aren't correctly labelled as such. I think I'll take them out and do them separately.

It mostly works I think. Problems:
* One problem is that it relies on the tests being assigned to the same numbered runners for the file paths to match. That's a little fragile, but in the worst case some things will be skipped.
* Files aren't correctly tagged for the limited API (with abi3.so). I can't persuade setuptools to cooperate here. That doesn't matter for the "run" tests because we can use the exact .so filename. It does break all the end to end tests though. I think those just need to be disabled for now.

da-woods:
I'm merging this as soon as it passes because I've had to fix this up for conflicts enough times and I'm running out of enthusiasm for doing it again.

runtests.py
```diff
@@ -2499,6 +2522,10 @@ def main():
         "--no-cython", dest="with_cython",
         action="store_false", default=True,
         help="do not run the Cython compiler, only the C compiler")
+    parser.add_argument(
+        "--no-compile", dest="with_compile",
+        action="store_false", default=True,
+        help="do not run either either Cython or the C compiler")
```
```suggestion
        help="do not run either Cython or the C compiler")
```

scoder (approved):

PR #7458, opened 2026-01-11
Export PYTHON_GIL=0 for all free-threading tests

da-woods:
Not just 3.13t


da-woods:
3.2.x commit ed4abcf07

PR #7464, opened 2026-01-15
Report also size of the shared utility module

matusvalo:


scoder:
Looks ok, although the `.c` file suffix is unusual.

However, this is untested since we don't run the benchmarks with the shared module. Most of them also won't benefit from it, so it would waste a lot of processing time.

But, since we are generally interested in reducing the overall module sizes independent of benchmarks, maybe we can start measuring the total (stripped) size of all _test_ modules in our CI run (or at least one of the runs) and output that in the end? Maybe also the average size, for comparison.

We could also have a complete CI test job with a shared module stored in one place, to see if that works properly across all test modules. And then report also the total module size for that run, to allow a comparison.

matusvalo:
> Looks ok, although the .c file suffix is unusual.

Yes it is stupid workaround to keep thinks simple...

> We could also have a complete CI test job with a shared module stored in one place, to see if that works properly across all test modules. And then report also the total module size for that run, to allow a comparison.

I created #7479 for that. The size reporting seems to be done in #7478. Is this PR not needed then? I still find it kind of useful but I am not insisting on it.

PR #7466, opened 2026-01-15
Rewrite import-star name matching to avoid lengthy chains of if-elif comparisons

scoder:
Instead, use a bit of binary search on the name's first character to reduce the number of comparisons.

Reduces both the overall runtime overhead as well as the storage size.
E.g. `tests/memoryview/memslice.so` shrinks from 1134512 to 1129296 bytes, -5KB or 0.5%.

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -6,9 +6,10 @@
 import cython
 cython.declare(Naming=object, Options=object, PyrexTypes=object, TypeSlots=object,
                error=object, warning=object, py_object_type=object, UtilityCode=object,
-               EncodedString=object, re=object)
+               EncodedString=object, itertools=object, operator=object, re=object)
 
 from collections import defaultdict
+import itertools
 import json
 import operator
 import os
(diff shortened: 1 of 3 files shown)


Cython/Compiler/ModuleNode.py
```diff
@@ -2889,64 +2890,148 @@ def create_import_star_conversion_utility_code(self, env):
...
+        done = code.new_label("done")
+
+        entries = env.entries
+
+        # Generate assignment code.
+
+        entry_names = [
+            name for name, entry in entries.items()
+            if entry.is_cglobal and entry.used and not entry.type.is_const
+        ]
```
scoder, line 2912: The list of names that we allow for assignments would probably benefit from excluding entries defined in internal scopes (e.g. utility code).
```suggestion
         entry_names = [
             name for name, entry in entries.items()
             if entry.is_cglobal and entry.used and not entry.type.is_const
+            and not entry.scope.is_internal
         ]
```
We should then also declare the `NonManglingScope` of Cython utility code as `.is_internal = True`, i.e.
```patch
diff --git a/Cython/Compiler/UtilityCode.py b/Cython/Compiler/UtilityCode.py
index 9ad24e4f5..b0a6822a4 100644
--- a/Cython/Compiler/UtilityCode.py
+++ b/Cython/Compiler/UtilityCode.py
@@ -13,6 +13,7 @@ import io
 
 
 class NonManglingModuleScope(Symtab.ModuleScope):
+    is_internal = True
 
     def __init__(self, prefix, *args, **kw):
         self.prefix = prefix
```

Cython/Compiler/ModuleNode.py
```diff
@@ -2889,64 +2890,148 @@ def create_import_star_conversion_utility_code(self, env):
...
+            code.putln("const char* failed_type;")
+            code.putln("CYTHON_UNUSED_VAR(failed_type);")
+            compare_prefixes(0, len(groups))
+
+        # Reject assignments to internal types.
+
+        internal_names = [
+            name for name, entry in entries.items()
+            if entry.is_type
+        ]
```
scoder, line 2981: `.is_type` refers to all types defined in the module (or its Cython utility code), which seems excessive. What might have been meant at the time could be `.is_cclass`, or any Python types that the module defines.
```suggestion
         internal_names = [
             name for name, entry in entries.items()
-            if entry.is_type
+            if entry.is_cclass or (entry.scope.is_internal and entry.is_cglobal and not entry.type.is_const)
         ]
```

scoder (commented):
I was actually surprised by the list of names that this compared against. Some memoryview names seem to have slipped in there that I doubt were intended (or anticipated).

scoder:
I created a second PR for the functional changes based on this one.
https://github.com/cython/cython/pull/7467

scoder:
Assuming that the name list changes are ok, this is superseded by https://github.com/cython/cython/pull/7467.

PR #7467, opened 2026-01-16
Reduce the list of names checked in import-star

scoder:
Some memoryview (and other internal) names seem to have slipped in the list of names that star-imports compared against, that I doubt were intended (or anticipated). This PR changes the lists to something more reasonable and helpful.

This PR also takes the opportunity to unify the mapping code of the assignments/rejection name lists by using Python strings for everything, which we probably have or need anyway, thus avoiding additional C string storage for them.

The `memslice.pyx` test currently allows assignments to these names (in the master branch):

<details>

```python
['PyExc_IndexError',
 'PyExc_ValueError',
 'Py_None',
 '__pyx_collections_abc_Sequence',
 '__pyx_memoryview_thread_locks',
 '__pyx_memoryview_thread_locks_used',
 'contiguous',
 'generic',
 'global_A',
 'global_B',
 'indirect',
 'indirect_contiguous',
 'strided']
```
</details>

and explicitly prevents assignments to these names:

<details>

```python
['MockBuffer',
 'CharMockBuffer',
 'IntMockBuffer',
 'UnsignedIntMockBuffer',
 'ShortMockBuffer',
 'UnsignedShortMockBuffer',
 'FloatMockBuffer',
 'DoubleMockBuffer',
 'ObjectMockBuffer',
 'IntStridedMockBuffer',
 'ErrorBuffer',
 'MyStruct',
 'SmallStruct',
 'NestedStruct',
 'PackedStruct',
 'NestedPackedStruct',
 'MyStructMockBuffer',
 'NestedStructMockBuffer',
 'PackedStructMockBuffer',
 'NestedPackedStructMockBuffer',
 'LongComplex',
 'LongComplexMockBuffer',
 'TestIndexSlicingDirectIndirectDims',
 'ArrayStruct',
 'PackedArrayStruct',
 'TestAttrs',
 'PaddedAtEnd',
 'AlignedNested',
 'PaddedAtEndNormal',
 'AlignedNestedNormal',
 'A',
 'B',
 'C',
 'D',
 'NoneSliceAttr',
 'PyObject',
 'array',
 'td_cy_int',
 'td_h_short',
 'td_h_double',
 'td_h_ushort',
 'td_h_cy_short',
 '__pyx_opt_args_8memslice_cdef_function',
 '__pyx_opt_args_8memslice_cdef_function2',
 'FusedStruct',
 'FusedPadded',
 'slice_1d',
 'slice_2d',
 '__pyx_opt_args_8memslice__get_empty_object_slice',
 '__pyx_ctuple_long_struct',
 '__pyx_ctuple_long',
 '__pyx_defaults',
 '__pyx_defaults1',
 '__pyx_ctuple_Py_ssize_t_struct',
 '__pyx_ctuple_Py_ssize_t',
 '__pyx_ctuple_int__and_int_struct',
 '__pyx_ctuple_int__and_int',
 '__pyx_ctuple_int_struct',
 '__pyx_ctuple_int',
 '__pyx_ctuple_signed__space_char__and_signed__space_char_struct',
 '__pyx_ctuple_signed__space_char__and_signed__space_char',
 '__pyx_ctuple_unsigned__space_char__and_unsigned__space_char_struct',
 '__pyx_ctuple_unsigned__space_char__and_unsigned__space_char',
 '__pyx_ctuple_char__and_char_struct',
 '__pyx_ctuple_char__and_char',
 '__pyx_ctuple_int__const__struct',
 '__pyx_ctuple_int__const_',
 '__pyx_ctuple_float__const__struct',
 '__pyx_ctuple_float__const_',
 '__pyx_ctuple_8memslice__dunder_pyx_t_8memslice_td_cy_int_struct',
 '__pyx_ctuple_8memslice__dunder_pyx_t_8memslice_td_cy_int',
 '__pyx_ctuple_8memslice_td_h_short_struct',
 '__pyx_ctuple_8memslice_td_h_short',
 '__pyx_ctuple_8memslice__dunder_pyx_t_8memslice_td_h_cy_short__const__struct',
 '__pyx_ctuple_8memslice__dunder_pyx_t_8memslice_td_h_cy_short__const_',
 '__pyx_ctuple_8memslice_td_h_ushort__const__struct',
 '__pyx_ctuple_8memslice_td_h_ushort__const_',
 '__pyx_ctuple_8memslice_td_h_double__const__struct',
 '__pyx_ctuple_8memslice_td_h_double__const_',
 '__pyx_ctuple_6279fd__signed__space_char__and_signed__space_char__an__etc_struct',
 '__pyx_ctuple_6279fd__signed__space_char__and_signed__space_char__an__etc',
 '__pyx_ctuple_26112e__signed__space_char__const___and_signed__space___etc_struct',
 '__pyx_ctuple_26112e__signed__space_char__const___and_signed__space___etc',
 '__pyx_ctuple_int__and_int__and_int__and_int__and_int_struct',
 '__pyx_ctuple_int__and_int__and_int__and_int__and_int',
 '__pyx_ctuple_0bdb1c__int__const___and_int__const___and_int__const____etc_struct',
 '__pyx_ctuple_0bdb1c__int__const___and_int__const___and_int__const____etc',
 '__pyx_ctuple_signed__space_char__and_int_struct',
 '__pyx_ctuple_signed__space_char__and_int',
 '__pyx_ctuple_signed__space_char__const___and_int__const__struct',
 '__pyx_ctuple_signed__space_char__const___and_int__const_',
 '__pyx_ctuple_bca720__signed__space_char__and_int__and_signed__space__etc_struct',
 '__pyx_ctuple_bca720__signed__space_char__and_int__and_signed__space__etc',
 '__pyx_ctuple_21c681__signed__space_char__const___and_int__const___a__etc_struct',
 '__pyx_ctuple_21c681__signed__space_char__const___and_int__const___a__etc',
 '__pyx_ctuple___dunderpyx_t_long_double_complex_struct',
 '__pyx_ctuple___dunderpyx_t_long_double_complex',
 '__pyx_ctuple_long__space_double__and_long__space_double_struct',
 '__pyx_ctuple_long__space_double__and_long__space_double',
 '__pyx_ctuple_int__and_int__and_int_struct',
 '__pyx_ctuple_int__and_int__and_int',
 '__pyx_ctuple_long__and_long__and_long_struct',
 '__pyx_ctuple_long__and_long__and_long',
 '__pyx_ctuple_Py_ssize_t__and_Py_ssize_t__and_Py_ssize_t_struct',
 '__pyx_ctuple_Py_ssize_t__and_Py_ssize_t__and_Py_ssize_t',
 '__pyx_ctuple_size_t_struct',
 '__pyx_ctuple_size_t',
 '__pyx_ctuple_long__and_long_struct',
 '__pyx_ctuple_long__and_long',
 '__pyx_ctuple_Py_ssize_t__and_Py_ssize_t_struct',
 '__pyx_ctuple_Py_ssize_t__and_Py_ssize_t',
 '__pyx_ctuple_int__and_int__and_int__and_int_struct',
 '__pyx_ctuple_int__and_int__and_int__and_int',
 '__pyx_ctuple_long__and_int_struct',
 '__pyx_ctuple_long__and_int',
 '__pyx_ctuple_int__and_int__and_long_struct',
 '__pyx_ctuple_int__and_int__and_long',
 '__pyx_scope_struct__testcase',
 '__pyx_scope_struct_1_test_arg_in_closure',
 '__pyx_scope_struct_2___pyx_f_8memslice_arg_in_closure_cdef',
 '__pyx_scope_struct_3_test_local_in_closure',
 '__pyx_scope_struct_4_genexpr',
 '__pyx_scope_struct_5_genexpr',
 'base_type',
 'PyTypeObject',
 'struct_type',
 '__pyx_memoryview',
 '__Pyx_memviewslice',
 '__pyx_buffer',
 '__Pyx_TypeInfo',
 'Enum',
 'memoryview',
 '_memoryviewslice',
 'PyThread_type_lock',
 'uintptr_t',
 '__pyx_atomic_int_type',
 '__pyx_ctuple_char__ptr_struct',
 '__pyx_ctuple_char__ptr',
 '__Pyx_StructField',
 '__Pyx_BufFmt_StackElem',
 '__pyx_typeinfo_string']
```
</details>

This PR turn that into these allowed assignments:
<details>

```python
['global_A', 'global_B']
```
</details>

and the rejection list is just the list of cdef class names defined in the module:

<details>

```python
['MockBuffer',
 'CharMockBuffer',
 'IntMockBuffer',
 'UnsignedIntMockBuffer',
 'ShortMockBuffer',
 'UnsignedShortMockBuffer',
 'FloatMockBuffer',
 'DoubleMockBuffer',
 'ObjectMockBuffer',
 'IntStridedMockBuffer',
 'ErrorBuffer',
 'MyStructMockBuffer',
 'NestedStructMockBuffer',
 'PackedStructMockBuffer',
 'NestedPackedStructMockBuffer',
 'LongComplexMockBuffer',
 'TestIndexSlicingDirectIndirectDims',
 'NoneSliceAttr']
```
</details>

which seems much more reasonable to me. All internally known names that get imported are simply assigned to new Python attributes.

Closes https://github.com/cython/cython/issues/4931

### DIFF
--- a/tests/run/starimport_cimport.srctree
+++ b/tests/run/starimport_cimport.srctree
@@ -64,6 +64,9 @@ globals()[''] = 6
 cdef int contiguous
 cdef object öbject
 
+cdef enum strided:
+    indirect_contiguous
+
 from all_things import *
 
 assert 'cython' in globals()
@@ -80,6 +83,9 @@ assert '__all__' not in globals()  # not in "__all__"
 
 cdef int indirect
 cdef object öbject
+
+cdef enum contiguous:
+    strided
 
 from noall_things import *
 


da-woods:
We've definitely had problems with this before - https://github.com/cython/cython/issues/4927, https://github.com/cython/cython/issues/4931 (specifically where the names are actually macros). So it'd be good to fix properly.

(I haven't yet looked at any of the detail of this PR at all)

scoder:
Thanks for the issue pointers. Yes, I think this should resolve https://github.com/cython/cython/issues/4931. Maybe with a bit more tweaking, can't say yet.

scoder:
I'm wondering if rejecting the assignments is actually a good way of dealing with this. Doesn't it just needlessly get in the way if we reject an assignment to something internal, instead of creating a Python attribute for it? We could simply ignore internally defined C variables when special casing them for assignments and leave it to the `SetAttr` fallback to do the normal Python thing.

The only thing that comes to my mind for which rejection is reasonable is cdef classes. We really don't want those to be replaced. But even cpdef functions are fine to be reassigned by a star-import, given that the user pretty much said so. And anything internal will stay untouched and internal, even if a Python attribute gets created under the same name.

scoder:
I've taken the opportunity of changing the names as a good time to remove the C string name processing and replace it with Python strings. Most of the names are already available as Python strings in one way or another (we are importing them as Python names from a foreign module). Thus, we can intern them and compare the string pointers quite quickly (assuming that interning does not fail, which would defeat the pointer comparison).

I also changed to a simple C switch statement based on the first name letter, so that we can easily integrate both the C name assignments and the name rejection into the same scheme.

Cython/Compiler/ModuleNode.py
```diff
@@ -2889,70 +2890,131 @@ def create_import_star_conversion_utility_code(self, env):
...
+        code.putln("PyUnicode_InternInPlace(&__pyx_interned_name);")
+        # Replace borrowed argument reference with borrowed local reference.
+        code.putln("py_name = __pyx_interned_name;")
+
+        # Specific C assignments.
+
+        entries = env.entries
+        cglobal_names = [
+            name for name, entry in entries.items()
+            if not entry.scope.is_internal and not name.startswith('__pyx_') and (
```
da-woods: is it worth making `startswith('__pyx_')` case insensitive?

tests/run/starimport_cimport.srctree
```diff
@@ -37,3 +38,119 @@ from star_cimport import *
...
+indirect_contiguous = 8
+strided = 9  # not in "__all__"
+
+######## noall_things.pyx ########
+
+cython = 5
+contiguous = 6
+indirect = 7
+indirect_contiguous = 8
+strided = 9
```
da-woods: newline

tests/run/starimport_cimport.srctree
```diff
@@ -1,18 +1,19 @@
...
+PYTHON -c "import star_import_all_things_cclass"
+PYTHON -c "import star_import_noall_things_cclass"
+PYTHON -c "import star_import_all_null"
 
 ######## setup.py ########
 
-from distutils.core import setup
-from Cython.Distutils import build_ext
-from Cython.Distutils.extension import Extension
+from typing import Type
```
da-woods: `Type` doesn't look to be used

Cython/Compiler/ModuleNode.py
```diff
@@ -2889,70 +2890,131 @@ def create_import_star_conversion_utility_code(self, env):
...
+                entry.is_cglobal and entry.used and not entry.type.is_const
```
da-woods, line 2923: The other thing that might be worth considering is C enum values (if they aren't already excluded by something else).

Cython/Compiler/ModuleNode.py
```diff
@@ -2889,70 +2890,131 @@ def create_import_star_conversion_utility_code(self, env):
...
+                    type_test = entry.type.type_test_code(
+                        env, "o")
+                    code.putln("if (!(%s)) %s;" % (
+                        type_test,
+                        code.error_goto(entry.pos)))
+                code.putln("Py_INCREF(o);")
+                code.put_decref(entry.cname, entry.type, nanny=False)
+                code.putln("%s = %s;" % (
+                    entry.cname,
+                    PyrexTypes.typecast(entry.type, py_object_type, "o")))
```
da-woods, line 2948: I think this needs to be `Py_SETREF` (or probably a manual version of it that's in the Limited API).  Because the decref can trigger arbitrary code in destructors.

Cython/Compiler/ModuleNode.py
```diff
@@ -2889,70 +2890,131 @@ def create_import_star_conversion_utility_code(self, env):
...
+                code.put_goto(done)
+            elif entry.type.create_from_py_utility_code(env):
+                # if available, utility code was already created in self.prepare_utility_code()
+                code.putln(entry.type.from_py_call_code(
```
da-woods, line 2953: At least for memoryviewslices I think this is missing some reference counting on the existing value

Cython/Compiler/ModuleNode.py
```diff
@@ -2889,70 +2890,131 @@ def create_import_star_conversion_utility_code(self, env):
...
+            code.putln("const char* failed_type = NULL;")
+
+            code.putln("Py_UCS4 __pyx_start_letter = __Pyx_PyUnicode_READ_CHAR(py_name, 0);")
+            code.putln("#if !CYTHON_ASSUME_SAFE_MACROS")
+            code.putln("if (unlikely(__pyx_start_letter) == (Py_UCS4) -1) goto bad;")
+            code.putln("#endif")
+
+            code.putln("switch (__pyx_start_letter) {")
+            for start_letter, names in prefix_groups:
+                code.putln(f"case {ord(start_letter)}:")
```
da-woods, line 2977: I guess this probably works fine, but maybe we should have at least one test with a non-ascii `start_letter`

Cython/Compiler/ModuleNode.py
scoder: Why not. I didn't find names in there that have a `__Pyx_` prefix, and this is more of a stop-gap solution. A clean fix would probably involve marking these entries somehow as internal, rather than relying on specific names. But at this stage, any Cython-internal names are worth hiding.

Cython/Compiler/ModuleNode.py
scoder: I changed it to use a temp to decref after setting to global variable.

Cython/Compiler/ModuleNode.py
scoder: Enums seem to be covered. I added a test.

scoder: I ran into https://github.com/cython/cython/issues/7473 while investigating this.

Cython/Compiler/ModuleNode.py
scoder: Probably. But since this code didn't change, I'll leave it to a separate PR and consider this one done.

da-woods: Yeah - it's broken both before and after. Just to save time, here's some code that shows the problem:

```
# import_mview.pyx
ba = bytearray(b'abcd')

cdef char[:] view = ba

try:
    ba.append(ord(b'e'))  # should fail - locked
except BufferError:
    pass
else:
    print("FAIL")
    exit(1)

from something_else import *

print(chr(view[0]))  # has been re-assigned (so prints '1')

ba.append(ord(b'4'))  # should work
```

```
# something_else.py
view = bytearray(b'123')
```

PR #7468, opened 2026-01-17
Traverse/clear cached cfunction module state

da-woods:
Obsoletes #3703

(I've attached a "coauthored by" tag because I've taken the useful/still relevant bits from that PR)


scoder:
Not sure if this is important enough to backport it to 3.2.x. Might be ok to only go into 3.3.

da-woods:
> Not sure if this is important enough to backport it to 3.2.x. Might be ok to only go into 3.3.

Yes agree. It'd be nice if "cleanup on module shutdown" was correct eventually, but this is a small incremental improvement and no-one will be relying on it.

PR #7471, opened 2026-01-19
Refactor int conversion

scoder:
In C, switch() is only officially defined on integer types up to `long` and we were using it for `Py_ssize_t` values. This is probably fine for most compilers and platforms and even an implicit downcast seems unlikely to pose problems in real code, but it still made me try a refactoring of the conversion code that was apparently worth it.

The integer conversion code is now split into separate C functions that make it easier for the C compiler to reason about inlining. The micro-benchmarks that I added show no degradations but suggest tiny improvements for integer unpacking.

### DIFF
--- a/Demos/benchmarks/bm_fib.py
+++ b/Demos/benchmarks/bm_fib.py
@@ -3,19 +3,27 @@ import time
 
 
 @cython.ccall
-def fib(x: float) -> float:
-    return 1 if x < 2 else fib(x-2) + fib(x-1)
+def fib_float(x: float) -> float:
+    return 1 if x < 2 else fib_float(x-2) + fib_float(x-1)
+
+@cython.ccall
+def fib_int(x: int) -> int:
+    return 1 if x < 2 else fib_int(x-2) + fib_int(x-1)
+
+@cython.ccall
+def fib_cull(x: cython.ulonglong) -> cython.ulonglong:
+    return 1 if x < 2 else fib_cull(x-2) + fib_cull(x-1)
 
 
-def run_fib(N, scale: cython.int, timer=time.perf_counter):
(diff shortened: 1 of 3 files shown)


scoder:
I think this is done. The benchmarks don't really show anything interesting, it seems the same as before given the usual variations.

da-woods:
> In C, switch() is only officially defined on integer types up to `long` and we were using it for `Py_ssize_t` values.

I couldn't immediately find anything telling me that. The [C standard* appears to say](https://c0x.shape-of-code.com/6.8.4.2.html) "The controlling expression of a switch statement shall have integer type.", and [looks to define "integer type"](https://c0x.shape-of-code.com/6.2.5.html) pretty generally.

With that said, if the refactoring seems positive then I don't have any issue with it..

But I'm not sure switch on `Py_ssize_t` is something we have to worry about generally.

(* or at least what looks to be an unofficial copy of the C standard....)

scoder:
> The [C standard* appears to say](https://c0x.shape-of-code.com/6.8.4.2.html) "The controlling expression of a switch statement shall have integer type.", and [looks to define "integer type"](https://c0x.shape-of-code.com/6.2.5.html) pretty generally.

I might have misread my sources then. Thanks for checking. There are a couple of good C references around, but they are not always conclusive regarding specific details.

I think the refactoring is worth it, especially moving the compact-int case to the top and splitting the negative/non-negative cases completely. For the rest, I don't think it matters much whether we use `switch()` or if-statements. The C compiler will take its own decisions anyway and that only happens after handling the most likely cases of reasonably small integers.

da-woods:
Yeah - I have no problem with the refactoring - I just don't think we need to make a complete survey of switch statements elsewhere.

PR #7476, opened 2026-01-22
Don't check alignment on the opaque object ABI

ngoldbaum:
This fixes the compiler errors related to alignment I was seeing over at https://github.com/numpy/numpy/issues/30704#issuecomment-3786016359. 

@da-woods said that he had disabled these checks but I think this extra bit is needed to actually disable them for all types.

Note that the PR is targeting https://github.com/cython/cython/tree/freethreading-limited-api-preview

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -4082,7 +4082,11 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
             # working when a builtin type is imported but not actually used.
             code.putln('0, 0,')
         else:
+            code.putln('#if !CYTHON_OPAQUE_OBJECTS')
             code.putln(f'sizeof({objstruct}), {alignment_func}({objstruct}),')
+            code.putln('#else')
+            code.putln('0, 0,')
+            code.putln("#endif")
         code.putln('#else')
         code.putln(f'sizeof({sizeof_objstruct}), {alignment_func}({sizeof_objstruct}),')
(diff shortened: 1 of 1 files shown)


da-woods:
I think this is too general. If you do:

```
from somemod cimport C
```

then the current version _does_ check the size and alignment successfully.

Looking in more detail at the Numpy issue, it's reporting

```
sizeof(PyObject), __PYX_GET_STRUCT_ALIGNMENT_3_3_0a0(PyObject),
```

which is for `PyObject` specifically (which is opaque).

(Edited out request for more details... I don't really need it I think)

da-woods:
It's almost certainly from https://github.com/numpy/numpy/blob/a0479caeeae7699ebec03b381778da45efd4305f/numpy/__init__.cython-30.pxd#L823-845.

I haven't attempted to do `extern` `cdef classes` yet. They're probably going to be tricky because you really have to know exactly how they're defined. I'll have a more detailed look soon but I don't think this is quite the solution.

ngoldbaum:
> I haven't attempted to do extern cdef classes yet. They're probably going to be tricky because you really have to know exactly how they're defined. I'll have a more detailed look soon but I don't think this is quite the solution.

Makes sense! I'll close this since this is definitely wrong.

da-woods:
Extern types are going to be a hard problem, but I think I can safely skip the checks for empty extern types (like what Numpy's using) because there are no attributes to mis-access, and inheritance will be opaque anyway. I'll try to do that later today.

Issue #7474, opened 2026-01-21
[ENH] Add `str`/`bytes`/`bytearray` to fast path comparisons


dg-pb:
### Is your feature request related to a problem? Please describe.

As noted in https://github.com/cython/cython/issues/7447#issuecomment-3766371063, I think there might be some value including strings into fast path of comparisons.

More and more often I find myself working with type-agnostic algorithms that involve element comparisons.
Although they need to be type-agnostic, naturally `str` remains a common type.

Example that pin-points performance benefit:
```cython
def search(element, iterable):
    i: cython.Py_ssize_t = 0
    el1: object | str = element
    el2: object | str
    for el2 in iterable:
        if el2 == el1:
            return i
        i += 1
```
Timings:
```python
%timeit search('a', '_' * 100_000 + 'a')
# 5.3 ms (pure python)
# 1.6 ms (cython with `object`)
# 0.4 ms (cython with `str`)
```

Of course, this shows maximum benefit.
For real applications, it is rare that comparisons comprise such large portion of computation.
Timings for KMP with the same setup as above:
```python
s1 = '_' * 10 + 'a'
s2 = '_' * 100_000 + 'a'
%timeit KMPattern(s1).search(s2)
# 17 ms (pure python)
# 4.3 ms (cython with `object`)
# 2.2 ms (cython with `str`)
```

So even for more involved algorithms, the performance benefit can be non-trivial.

---

So I think `str` application might be useful.

I don't know about `bytes` / `bytearray` - these so far have been rarer in my applications.
Mainly due to the fact that a large portion of cases are single character, thus elements most often become `int`.

However, I suspect that:
1. These could be useful - after all most of string algorithms in `sdlib` can handle all: `str`, `bytes` and `bytearray`
2. The logic to do this is likely very similar as for `str`

---

So in short, if overhead for non-applicable cases is negligible (I think you have a good idea on this from https://github.com/cython/cython/issues/7447) then maybe this is worth consideration.

scoder:
See https://github.com/cython/cython/pull/7477

PR #7477, opened 2026-01-23
Include str/bytes/bytearray in optimised comparison code

scoder:
Followup to https://github.com/cython/cython/pull/7452

Closes https://github.com/cython/cython/issues/7474

### DIFF
--- a/Cython/Utility/Optimize.c
+++ b/Cython/Utility/Optimize.c
@@ -1200,7 +1200,13 @@ static CYTHON_INLINE {{c_ret_type}} __Pyx_PyObject_Compare{{s1_prefix}}{{s2_pref
     if (length != length2) {{return_false if op == 'Eq' else return_true}};
 
     #endif
-    // len(s1) == len(s2) >= 1  (empty string is interned, and "s1 is not s2")
+
+    // len(s1) == len(s2)
+    {{if s1_type == 'bytearray'}}
+    if (length == 0) {{return_true if op == 'Eq' else return_false}};
+    {{else}}
+    // bytes: length >= 1  (empty bytes is singleton, and "s1 is not s2")
+    {{endif}}
 
     if (ps1[0] != ps2[0]) {{return_false if op == 'Eq' else return_true}};
     if (length == 1) {{return_true if op == 'Eq' else return_false}};


scoder:
I included mixed comparisons between bytes and bytearray because comparing a variable bytearray object to a fixed bytes object seems a rather natural operation – if comparisons happen at all.

Cython/Utility/ModuleSetupCode.c
```diff
@@ -582,8 +582,22 @@ typedef uintptr_t  __pyx_uintptr_t;
...
+#ifdef Py_UNREACHABLE
+  #define __Pyx_UNREACHABLE() Py_UNREACHABLE()
+#elif __Pyx_has_cbuiltin(__builtin_unreachable)
+  #define __Pyx_UNREACHABLE() __builtin_unreachable()
+#elif defined(__clang__) || defined(__INTEL_COMPILER) || (defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 5)))
+  #define __Pyx_UNREACHABLE() __builtin_unreachable()
+#elif defined(_MSC_VER)
+  #define __Pyx_UNREACHABLE() __assume(0)
+#elif !CYTHON_COMPILING_IN_LIMITED_API
+  #define __Pyx_UNREACHABLE() Py_FatalError("Unreachable C code path reached")
```
da-woods: `Py_FatalError` looks like it's in the Stable ABI:

https://github.com/python/cpython/blob/9181d776daf87f0e4e2ce02c08f162150fdf7d79/Misc/stable_abi.toml#L1630-L1631

da-woods:
In principle I think you can check `long`, `unicode` and `bytes` in one go with the various subclass flags: https://docs.python.org/3/c-api/typeobj.html#c.Py_TPFLAGS_LONG_SUBCLASS. You'd still need to do `CheckExact` afterwards but it'd reduce the number of branches before going to the generic code.

scoder:
> you can check `long`, `unicode` and `bytes` in one go with the various subclass flags

Good idea – but there is no `float` flag, nor `bytearray`. Meaning, we'd still need a float type check if that is allowed and cannot just jump over the type checks based on a flag check. Not sure if that's an oversight but it makes this less simple and less quick also.

scoder:
Actually, I'd expect a series of type pointer checks to be extremely fast, potentially even faster than an indirect branch based on a pointer dereference for reading the flags. Especially when branch prediction kicks in for a homogeneous collection. But that's just a guess. I doubt that it'd make a big difference in practice.

EDIT: The resulting assembly is literally a well predictable chain of cmp/je on a register:

<details>

```asm
static CYTHON_INLINE PyObject* __Pyx_PyObject_CompareEq_object_object(PyObject *op1, PyObject *op2, int pyop) {
    e86e:       48 83 ec 08             sub    $0x8,%rsp
    e872:       48 8b 47 08             mov    0x8(%rdi),%rax
    if (PyFloat_CheckExact(op1)) {
    e876:       48 3b 05 0b b6 43 00    cmp    0x43b60b(%rip),%rax        # 449e88 <PyFloat_Type@Base>
    e87d:       74 5a                   je     e8d9 <__Pyx_PyObject_CompareEq_object_object+0x6b>
    if (PyLong_CheckExact(op1)) {
    e87f:       48 3b 05 da b6 43 00    cmp    0x43b6da(%rip),%rax        # 449f60 <PyLong_Type@Base>
    e886:       0f 84 9c 00 00 00       je     e928 <__Pyx_PyObject_CompareEq_object_object+0xba>
    if (PyUnicode_CheckExact(op1)) {
    e88c:       48 3b 05 f5 b6 43 00    cmp    0x43b6f5(%rip),%rax        # 449f88 <PyUnicode_Type@Base>
    e893:       0f 84 bc 00 00 00       je     e955 <__Pyx_PyObject_CompareEq_object_object+0xe7>
    if (PyBytes_CheckExact(op1)) {
    e899:       48 3b 05 18 b6 43 00    cmp    0x43b618(%rip),%rax        # 449eb8 <PyBytes_Type@Base>
    e8a0:       0f 84 c8 00 00 00       je     e96e <__Pyx_PyObject_CompareEq_object_object+0x100>
    if (PyByteArray_CheckExact(op1)) {
    e8a6:       48 3b 05 03 b6 43 00    cmp    0x43b603(%rip),%rax        # 449eb0 <PyByteArray_Type@Base>
    e8ad:       75 40                   jne    e8ef <__Pyx_PyObject_CompareEq_object_object+0x81>
        if (op1 == op2) goto __pyx_return_true;
    e8af:       48 39 f7                cmp    %rsi,%rdi
    e8b2:       0f 84 23 01 00 00       je     e9db <__Pyx_PyObject_CompareEq_object_object+0x16d>
    e8b8:       48 8b 46 08             mov    0x8(%rsi),%rax
        if (PyByteArray_CheckExact(op2)) {
    e8bc:       48 3b 05 ed b5 43 00    cmp    0x43b5ed(%rip),%rax        # 449eb0 <PyByteArray_Type@Base>
    e8c3:       0f 84 d8 00 00 00       je     e9a1 <__Pyx_PyObject_CompareEq_object_object+0x133>
        if (PyBytes_CheckExact(op2)) {
    e8c9:       48 3b 05 e8 b5 43 00    cmp    0x43b5e8(%rip),%rax        # 449eb8 <PyBytes_Type@Base>
    e8d0:       75 1d                   jne    e8ef <__Pyx_PyObject_CompareEq_object_object+0x81>
            return __Pyx_PyObject_ComparePyByteArrayPyBytesEq(op1, op2);
    e8d2:       e8 87 ba ff ff          call   a35e <__Pyx_PyObject_ComparePyByteArrayPyBytesEq>
    e8d7:       eb 20                   jmp    e8f9 <__Pyx_PyObject_CompareEq_object_object+0x8b>
    e8d9:       48 8b 46 08             mov    0x8(%rsi),%rax
        if (PyFloat_CheckExact(op2)) {
    e8dd:       48 3b 05 a4 b5 43 00    cmp    0x43b5a4(%rip),%rax        # 449e88 <PyFloat_Type@Base>
    e8e4:       74 18                   je     e8fe <__Pyx_PyObject_CompareEq_object_object+0x90>
        if (PyLong_CheckExact(op2)) {
    e8e6:       48 3b 05 73 b6 43 00    cmp    0x43b673(%rip),%rax        # 449f60 <PyLong_Type@Base>
    e8ed:       74 32                   je     e921 <__Pyx_PyObject_CompareEq_object_object+0xb3>
    return PyObject_RichCompare(op1, op2, Py_EQ);
    e8ef:       ba 02 00 00 00          mov    $0x2,%edx
    e8f4:       e8 f7 83 ff ff          call   6cf0 <PyObject_RichCompare@plt>
```

</details>

scoder:
In my local and GHA-run micro benchmarks, `PyUnicode_Compare()` is consistently faster than anything I could do through the C-API. That might be due to PGO builds of CPython competing with non-PGO Cython modules, but few people build their own CPython (it will usually come as PGO binaries) and few people run PGO on their Cython modules, so that's not an unfair comparison. It also saves a lot of code to just use what CPython provides here.

scoder:
I think this is done. Microbenchmarks show considerable improvements for all comparison types, although the other benchmarks don't benefit much.

Cython/Utility/Optimize.c
```diff
@@ -1076,35 +1076,218 @@ static PyObject* __Pyx__PyNumber_PowerOf2(PyObject *two, PyObject *exp, PyObject
...
+// str comparisons
+
+{{if type1 in ('object', 'str') and type2 in ('object', 'str')}}
+#ifndef __Pyx_DEFINED_PyObject_CompareStrStr{{func_suffix}}
+#define __Pyx_DEFINED_PyObject_CompareStrStr{{func_suffix}}
+static CYTHON_INLINE {{c_ret_type}} __Pyx_PyObject_CompareStrStr{{func_suffix}}(PyObject* s1, PyObject* s2) {
+    {{if op in 'EqNe'}}
+    #if __PYX_LIMITED_VERSION_HEX >= 0x030e0000
+    int result = PyUnicode_Equal(s1, s2);
+    if (unlikely(result == -1)) {{return_error}};
```
da-woods, line 1135: According to the [documentation](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_Equal):

> The function always succeeds if a and b are str objects.

which I think should be true in the usage here? I don't know if that's enough justification to skip the check.

Cython/Utility/Optimize.c
```diff
@@ -1076,35 +1076,218 @@ static PyObject* __Pyx__PyNumber_PowerOf2(PyObject *two, PyObject *exp, PyObject
...
+    if (unlikely(PyBytes_AsStringAndSize(s2, &ps2, &length2) == -1)) {{return_error}};
+    {{else}}
+    ps2 = __Pyx_PyByteArray_AsString(s2); if (unlikely(!ps2)) {{return_error}};
+    length2 = __Pyx_PyByteArray_GET_SIZE(s2); if (unlikely(length2 == -1)) {{return_error}};
+    {{endif}}
+
+    if (length != length2) {{return_false if op == 'Eq' else return_true}};
+
+    #endif
+    // len(s1) == len(s2) >= 1  (empty string is interned, and "s1 is not s2")
```
da-woods: The comment won't be true for bytearrays though.

I think it works though because they're guaranteed to be `/0` terminated but the reason isn't right.

da-woods:
This looks to have broken something in PyArrow: https://github.com/apache/arrow/issues/49138

To me it looks like the code before the change was already broken (but potentially might not be for `bytes` rather than `str`)

PR #7479, opened 2026-01-24
Tests: Add a test run using shared utility module

matusvalo:
This PR:
* adds new option compiling tests with shared utility module
* adds separated CI run for testing shared utility module

> [!NOTE]  
> Running test with shared utility module enabled:
> * ~~is done in simple way. The shared utility module is compiled for **every** test what is suboptimal. This can be improved compiling once and just copying .so file. To be decided whether it is enough or PR needs more love...~~
> * is skipping all tests which are not compiled (pure python) or compilation is driven by test itself (.srctree tests)
>
> I was not sure which cases in the matrix should be enabled in CI/CD (which python version etc.), this is to be discussed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -250,18 +250,29 @@ jobs:
       - name: Run CI
         continue-on-error: ${{ matrix.allowed_failure || false }}
         env: ${{ matrix.env }}
-        run: |
-          bash -c "${{ !startsWith(runner.os, 'macOS') && 'stdbuf -e4M' || ''}}  bash ./Tools/ci-run.sh \
-            2> >(tee  ${{!startsWith(runner.os, 'macOS') && '-p' || ''}} \
-              >(${{ !startsWith(runner.os, 'macOS') && 'stdbuf -i4M' || ''}}  tail -20 > ci-summary.txt) \
-              >(${{ !startsWith(runner.os, 'macOS') && 'stdbuf -i4M' || ''}}  grep -F ' SLOWTEST ' >ci-slowtests.txt) \
-              >&2)"
-
-      - name: Report results
+        # Windows does not support job summary reports and has problems with pipes, so keep things simple there.
         run: |
           bash -c "\
-            if [ -s ci-slowtests.txt ]; then echo '| Slow tests | duration |'; echo '|:---|---:|'; sed -e 's/^.*\[/|    / ; s/: SLOWTEST / | / ; s/\].*/ |/' ci-slowtests.txt | sort -nr -k4; echo; fi; \
-            if [ -s ci-summary.txt ]; then grep -A20 'Most expensive pipeline stages' ci-summary.txt | sed -e '2,21s/^/    / ; 1,1s/:/:\n\n    /'; fi \
+            if [[ '${{runner.os}}' == 'Windows' ]]; then \
+              bash ./Tools/ci-run.sh; \
+            else \
+              ${{runner.os == 'Linux' && 'stdbuf -e4M' || ''}}  bash ./Tools/ci-run.sh \
+              2> >(tee  ${{runner.os == 'Linux' && '-p' || ''}} \
+                >(tail -20 > ci-summary.txt) \
+                >(grep -F ' SLOWTEST ' >ci-slowtests.txt) \
+                >&2); \
+            fi"
+
+      - name: Report results
+        if: runner.os != 'Windows'
+        run: |
+          bash -c "\
+            if [[ -s ci-slowtests.txt ]]; then \
+              echo '| Slow tests | duration |'; echo '|:---|---:|'; sed -e 's/^.*\[/|    / ; s/: SLOWTEST / | / ; s/\].*/ |/' ci-slowtests.txt | sort -nr -k4; echo; \
+            fi; \
+            if [[ -s ci-summary.txt ]]; then \
+              grep -A20 'Most expensive pipeline stages' ci-summary.txt | sed -e '2,21s/^/    / ; 1,1s/:/:\n\n    /'; \
+            fi; \
             " >> $GITHUB_STEP_SUMMARY
 
       - name: Upload HTML docs
--- a/runtests.py
+++ b/runtests.py
@@ -2100,7 +2100,7 @@ def collect_doctests(path, module_prefix, suite, selectors, exclude_selectors):
                         pass
 
 
-class EndToEndTest(unittest.TestCase):
+class EndToEndTest(TimedTest):
     """
     This is a test of build/*.srctree files, where srctree defines a full
     directory structure and its header gives a list of commands to run.
@@ -2129,11 +2129,16 @@ class EndToEndTest(unittest.TestCase):
         return "[%d] End-to-end %s" % (
             self.shard_num, self.name)
 
+    def id(self):
+        return f"etoe.{self.name}"
+
     def setUp(self):
         from Cython.TestUtils import unpack_source_tree
         _, self.commands = unpack_source_tree(self.treefile, self.workdir, self.cython_root)
+        super().setUp()
 
     def tearDown(self):
+        super().tearDown()
         if self.cleanup_workdir:
             for trial in range(5):
                 try:
--- a/Demos/benchmarks/util.py
+++ b/Demos/benchmarks/util.py
@@ -5,11 +5,15 @@
 import cython
 
 from functools import partial
+import sys
 import time
+
+# PyPy's JIT can induce high variance so we basically ignore this measure.
+IS_PYPY = sys.implementation.name == 'pypy'
 
 
 def repeat_to_accuracy(func, *args,
-                       variance_threshold: float = 5e-5,
+                       variance_threshold: float = 1. if IS_PYPY else 5e-5,
                        outlier_threshold: float = .12,
                        scale=1,
                        repeat=True,
--- a/tests/run/test_fstring.pyx
+++ b/tests/run/test_fstring.pyx
@@ -793,7 +793,7 @@ y = (
 
         # Test concatenating 2 largs fstrings.
         # self.assertEqual(eval(build_fstr(255)*256), (x+' ')*(255*256))
-        self.assertEqual(cy_eval(build_fstr(255)*8, x=x, width=width), (x+' ')*(255*8))  # CPython uses 255*256
+        self.assertEqual(cy_eval(build_fstr(255)*3, x=x, width=width), (x+' ')*(255*3))  # CPython uses 255*256
 
         s = build_fstr(253, '{x:{width}} ')
         # self.assertEqual(eval(s), (x+' ')*254)


.github/workflows/ci.yml
```diff
@@ -171,6 +169,12 @@ jobs:
             backend: "c,cpp"
             env: { LIMITED_API: "--limited-api" }
             extra_hash: "-limited_api"
+          # Cython Shared Module
+          - os: ubuntu-22.04
+            python-version: "3.12"
+            backend: "c,cpp"
+            env: { LIMITED_API: "--shared-module" }
+            extra_hash: "--shared-module"
```
matusvalo: Now I am thinking maybe the PR deserves better name for argument (and variables too)... `--shared-utility` or `--shared-utility-module`?

.github/workflows/ci.yml
```diff
@@ -171,6 +171,12 @@ jobs:
             backend: "c,cpp"
             env: { LIMITED_API: "--limited-api" }
             extra_hash: "-limited_api"
+          # Cython Shared Module
+          - os: ubuntu-22.04
+            python-version: "3.12"
+            backend: "c,cpp"
+            env: { LIMITED_API: "--shared-module" }
```
scoder:
```suggestion
            env: { SHARED_MODULE: "--shared-module" }
```

.github/workflows/ci.yml
scoder: If you ask me, `--shared-utility` is ok, `--shared-utility-module` is really clear but also a bit long if you need to type it more often (which you probably don't). `--shared-module` does seem rather generic and less clear. But we only have one shared module as it stands, so…

runtests.py
```diff
@@ -854,6 +855,9 @@ def handle_directory(self, path, context):
                     continue
                 if skip_limited(tags):
                     continue
+                if self.shared_module is not None and 'shared_module' not in tags['tag']:
+                    # It is redundant to run EndToEnd tests that are not related to using shared utility module
+                    continue
```
```suggestion
                if self.shared_module is not None and 'shared_module' not in tags['tag']:
                    # EndToEnd tests have their own build setup which doesn't use the shared module unless the tag says so.
                    continue
```

runtests.py
```diff
@@ -881,6 +885,9 @@ def handle_directory(self, path, context):
 
             if mode == 'run' and ext == '.py' and not self.cython_only and not filename.startswith('test_'):
                 # additionally test file in real Python
+                if self.shared_module:
+                    # Without compilation it does not make sense run it with shared utility module enabled
+                    continue
```
```suggestion
            if mode == 'run' and ext == '.py' and not filename.startswith('test_') and not (
                    self.cython_only or self.shared_module):
                # additionally test file in real Python
```

runtests.py
```diff
@@ -1207,8 +1216,19 @@ def runAbi3AuditTest(self):
...
 
+    def generateSharedModule(self):
+        from Cython.Build.SharedModule import generate_shared_module
+        from Cython.Compiler.Options import CompilationOptions
+        if self.language == 'cpp':
+            options = CompilationOptions(shared_c_file_path=f'{self.workdir}/{self.shared_module}.cpp')
+        else:
+            options = CompilationOptions(shared_c_file_path=f'{self.workdir}/{self.shared_module}.c')
+        if self.shared_module:
+            generate_shared_module(options)
```
scoder: `generate_shared_module()` has side-effects, it calls `Errors.reset()`, for example. I think we should avoid giving it side-effects.

Also, this doesn't seem to check if the shared module is already there. I don't think we should build it multiple times. Given its purpose of being reused (and thus saving space and time), we should also build it once in the whole run (like the refnanny) and then reuse it across all tests.

matusvalo: Calling directly `generate_shared_module()` was causing issues due mentioned side-effects. I fix it by running in separate subprocess as command...

matusvalo: > Also, this doesn't seem to check if the shared module is already there. I don't think we should build it multiple times. 

This is fixed now too. shared utility module is generated only once in the beginning of the test.

runtests.py
```diff
@@ -2812,6 +2827,51 @@ def time_stamper():
...
+        return
+
+    workdir = TemporaryDirectory()
+    shared_utility_c_file = os.path.join(workdir.name, f'{SHARED_UTILITY_MODULE_NAME}.c')
+    utility_gen_result = subprocess.run(
+        [
+            "python",
+            '-c' 'from Cython.Compiler.Main import main; main(command_line=1)',
+            '--generate-shared', shared_utility_c_file
+        ], stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding='utf8')
```
scoder: Can we use the "normal" command line interface for this? It would be nice if we could go through the same tool that regular users would use, i.e. probably the `cython` or `cythonize` commands.

runtests.py
```diff
@@ -16,6 +16,7 @@
 import subprocess
 import sys
 import sysconfig
+from tempfile import TemporaryDirectory
```
scoder: Could you move this down to the other from-imports?

runtests.py
matusvalo: Using cython command directly failed. I will investigate more about that...

matusvalo: Hmm it seems working though. Maybe I just messed something back then.

scoder:
Would "cythonize -b" work here? Then we could avoid duplicating the build setup.

scoder:
We should rebase this on https://github.com/cython/cython/pull/7489 once that is merged.

matusvalo:
> Would "cythonize -b" work here? Then we could avoid duplicating the build setup.

Honestly, I had **no** idea that `cythonize` could compile `.c` files! I thought that it requres .pyx/.py files...

da-woods:
> We should rebase this on https://github.com/cython/cython/pull/7489 once that is merged.

I've merged it. Superficially it doesn't conflict.

matusvalo:
> Would "cythonize -b" work here? Then we could avoid duplicating the build setup.

Implemented and master marged after https://github.com/cython/cython/pull/7489

matusvalo:
It seems another test is failing with shared utility:

Running `cython -2 --shared=_cython_shared memslice.pyx` will fail with

```
  File "/home/dev/cython/Cython/Compiler/Visitor.py", line 253, in _process_children
    result = self._visitchildren(parent, attrs, exclude)
  File "/home/dev/cython/Cython/Compiler/Visitor.py", line 220, in _visitchildren
    childretval = self._visitchild(child, parent, attr, None)
  File "/home/dev/cython/Cython/Compiler/Visitor.py", line 188, in _visitchild
    result = self._visit(child)
  File "/home/dev/cython/Cython/Compiler/Visitor.py", line 174, in _visit
    return handler_method(obj)
  File "/home/dev/cython/Cython/Compiler/Visitor.py", line 538, in visit_BinopNode
    return self._visit_binop_node(node)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/home/dev/cython/Cython/Compiler/Visitor.py", line 556, in _visit_binop_node
    node = self._dispatch_to_method_handler(
        special_method_name, None, False, type_name,
        node, None, [operand1, operand2], None)
  File "/home/dev/cython/Cython/Compiler/Visitor.py", line 695, in _dispatch_to_method_handler
    result = method_handler(
        node, function, arg_list, is_unbound_method)
  File "/home/dev/cython/Cython/Compiler/Optimize.py", line 3498, in _handle_simple_method_object___mul__
    return self._optimise_num_binop('Multiply', node, function, args, is_unbound_method)
           ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dev/cython/Cython/Compiler/Optimize.py", line 3601, in _optimise_num_binop
    elif node.type.is_pyobject:
         ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'is_pyobject'
```

Edit: Created an issue for that: #7497

scoder:
Now we are left with this error (basically throughout the memoryview tests):
```
wrapper_class.c: In function ‘__pyx_memoryview_check’:
wrapper_class.c:2532:61: error: ‘__pyx_mstatetype’ has no member named ‘__pyx_memoryview_type’
 2532 |   __pyx_t_1 = __Pyx_TypeCheck(__pyx_v_o, __pyx_mstate_global->__pyx_memoryview_type);
```

matusvalo:
> Now we are left with this error (basically throughout the memoryview tests):

There is one pending bug causing this issue: #7487

runtests.py
```diff
@@ -2824,6 +2837,34 @@ def time_stamper():
...
+            'python', 'cythonize.py', '-bi', shared_utility_c_file
+        ], stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding='utf8')
+    if compilation_result.returncode != 0:
+        raise RuntimeError(f"Shared utility compilation failed:\n{compilation_result.stdout}")
+
+    sys.path.append(workdir.name)
+    try:
+        yield workdir
+    finally:
+        workdir.cleanup()
```
```suggestion
    with TemporaryDirectory() as workdir:
        shared_utility_c_file = os.path.join(workdir.name, f'{SHARED_UTILITY_MODULE_NAME}.c')
        utility_gen_result = subprocess.run(
            [
                'python', 'cython.py', '--generate-shared', shared_utility_c_file
            ], stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding='utf8')
        if utility_gen_result.returncode != 0:
            raise RuntimeError(f"Shared utility generation failed:\n{utility_gen_result.stdout}")

        compilation_result = subprocess.run(
            [
                'python', 'cythonize.py', '-bi', shared_utility_c_file
            ], stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding='utf8')
        if compilation_result.returncode != 0:
            raise RuntimeError(f"Shared utility compilation failed:\n{compilation_result.stdout}")

        sys.path.append(workdir.name)
        yield workdir
```

runtests.py
```diff
@@ -1480,14 +1490,14 @@ def compile(self, test_directory, module, module_path, workdir, incdir,
                 with self.stats.time(self.name, self.language, 'cython'):
                     self.run_cython(
                         test_directory, module, module_path, workdir, incdir, annotate,
-                        evaluate_tree_assertions=evaluate_tree_assertions)
+                        evaluate_tree_assertions=evaluate_tree_assertions, extra_compile_options={'shared_utility_qualified_name': self.shared_utility})
```
```suggestion
                        evaluate_tree_assertions=evaluate_tree_assertions,
                        extra_compile_options={'shared_utility_qualified_name': self.shared_utility},
                    )
```

runtests.py
```diff
@@ -1480,14 +1490,14 @@ def compile(self, test_directory, module, module_path, workdir, incdir,
...
                 errors, warnings, perf_hints = sys.stderr.getall()
             finally:
                 sys.stderr = old_stderr
             if self.test_determinism and not expect_errors:
                 workdir2 = workdir + '-again'
                 os.mkdir(workdir2)
-                self.run_cython(test_directory, module, module_path, workdir2, incdir, annotate)
+                self.run_cython(test_directory, module, module_path, workdir2, incdir, annotate, extra_compile_options={'shared_utility_qualified_name': self.shared_utility})
```
```suggestion
                self.run_cython(
                    test_directory, module, module_path, workdir2, incdir, annotate,
                    extra_compile_options={'shared_utility_qualified_name': self.shared_utility},
                )
```

runtests.py
```diff
@@ -985,7 +993,8 @@ def build_test(self, test_class, path, workdir, module, module_path, tags, langu
                           stats=self.stats,
                           add_cython_import=add_cython_import,
                           extra_directives=extra_directives,
-                          abi3audit=self.abi3audit
+                          abi3audit=self.abi3audit,
+                          shared_utility=self.shared_utility
```
```suggestion
                          shared_utility=self.shared_utility,
```

runtests.py
scoder: Actually, I had thought we had implemented `cythonize --with-shared` (or similar) by now, but it seems that that's still missing and `cython` is the only command that can generate the shared utility module file. Only the `cythonize()` function has it. It should be added to the `cythonize` command as well, there's no reason it should be missing there.

scoder: I think you should be able to use something like `cythonize -s shared_utility_qualified_name=SHARED_UTILITY_MODULE_NAME -bi` with the file name of the shared module (and the other source files to compile) to generate and compile the shared module as part of the normal build. But there should really be a nicer command line option for this.

scoder:
This still fails in the graalpy job run:
```
   File "/home/runner/work/cython/cython/runtests.py", line 2857, in generate_shared_utility
    shared_utility_c_file = os.path.join(workdir.name, f'{SHARED_UTILITY_MODULE_NAME}.c')
                                         ^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'name'
```

matusvalo:
> Now we are left with this error (basically throughout the memoryview tests):
> 
> ```
> 
> wrapper_class.c: In function ‘__pyx_memoryview_check’:
> 
> wrapper_class.c:2532:61: error: ‘__pyx_mstatetype’ has no member named ‘__pyx_memoryview_type’
> 
>  2532 |   __pyx_t_1 = __Pyx_TypeCheck(__pyx_v_o,

Hmm strange. I clicked on auto merge after CI pass...

matusvalo:
Aah The test is not run in graalpy in CI pipeline...

scoder:
> I clicked on auto merge after CI pass

I didn't include all CI jobs in the required checks. Especially GraalPy used to be a bit flaky in the past, so requiring a clean run of all jobs can get in the way quite easily. I've added it now since David recently merged a fix for it.

Also, GHA requires listing each job in the required jobs explicitly. I cannot just say "all of them". That's annoying as well.

da-woods:
As a general rule, I wouldn't feel bad about just disabling individual tests on GraalPy if they aren't easy to make work.

matusvalo:
> This still fails in the graalpy job run:
> 
> ```
>    File "/home/runner/work/cython/cython/runtests.py", line 2857, in generate_shared_utility
>     shared_utility_c_file = os.path.join(workdir.name, f'{SHARED_UTILITY_MODULE_NAME}.c')
>                                          ^^^^^^^^^^^^
> AttributeError: 'str' object has no attribute 'name'
> ```

Should be fixed in #7538

PR #7480, opened 2026-01-24
Remove unused function in binops code

da-woods:
I think this is unused.

(Untested, but we'll find out pretty quickly if I'm wrong)


da-woods:
It fairly clearly is used...

PR #7483, opened 2026-01-25
Fix missed refcounting of heaptype types

da-woods:
Specifically for heaptypes that inherit from non-heaptypes we weren't reference-counting. This is most likely from builtin types, but could also be from extension types defined in other modules (although it'd be odd to be mixing heaptypes and non-heaptypes, it isn't forbidden).


tests/run/refcount_types.pyx
```diff
@@ -0,0 +1,48 @@
...
+    pass
+
+cdef class D(C):
+    pass
+
+class E(C):
+    pass
+
+cdef class L(list):
+    pass
```
da-woods, line 22: This was the only class here that I believe was broken. The others are tested just for a bit more completeness

Cython/Compiler/ModuleNode.py
```diff
@@ -1931,6 +1931,10 @@ def generate_dealloc_function(self, scope, code):
             if not base_type.is_builtin_type:
                 base_cname = code.name_in_slot_module_state(base_cname)
             tp_dealloc = TypeSlots.get_base_slot_function(scope, tp_slot)
+            if tp_dealloc is None:
+                code.putln("#if CYTHON_USE_TYPE_SPECS")
```
scoder, line 1930: Keeping only a borrowed reference to the type is probably ok since we know that it's owned by the instance that we read it from. But it seems at least worth a comment that we thought of it. It's not obvious given the intermediate `tp_dealloc()` upcalls.

PR #7485, opened 2026-01-27
Speed up add/sub/mul operators on int and float

scoder:
Introduces specialised replacements for `PyNumber_Add`, `PyNumber_Subtract` and `PyNumber_Multiply` that take care of available type information and provide fast paths for Python int and float objects found at runtime.


Cython/Utility/Optimize.c
```diff
@@ -1075,6 +1075,279 @@ static PyObject* __Pyx__PyNumber_PowerOf2(PyObject *two, PyObject *exp, PyObject
...
+        type_name_op1,
+        type_name_op2);
+    __Pyx_DECREF_TypeName(type_name_op1);
+    __Pyx_DECREF_TypeName(type_name_op2);
+}
+#endif
+#endif
+
+static PyObject* __Pyx_ReverseSlot_{{op_name}}_{{type1}}(PyObject *op1, PyObject *op2) {
+    #if CYTHON_USE_TYPE_SLOTS
```
da-woods: I don't think this is exactly the right guard.

`PyType_TryGetSubSlot` will work with `CYTHON_USE_TYPE_SLOTS` both true and false. (Although it'll return NULL for non-heaptypes in runtime versions < 3.10 when `CYTHON_USE_TYPE_SLOTS` is false).

`CYTHON_USE_TYPE_SLOTS` is really guarding direct access into the type object.

I think you want something like:

```
#if !CYTHON_USE_TYPE_SLOTS && __PYX_LIMITED_API_VERSION_HEX < 0x030A0000
if (__pyx_get_runtime_version() < 0x030A0000 && PyType_GetFlags(type_op2, Py_TPFLAGS_HEAPTYPE)) {
    return return {{op_name}}(op1, op2);
}
#else
// ... code as below, but you can use `__Pyx_PyType_GetSubSlot` because you don't have to guard against the heaptype case.
```

scoder: My intention was rather to restrict this code to the cases that we expect to be fast when going through type slots. `CYTHON_USE_TYPE_SLOTS` seemed a good indication for that. I think the `…Try…()` was mostly a relic of other things that I tried before. As you write, it can be removed if we keep the guard.

I'm not sure it's worth doing the whole dance also for the Limited API. It might be, but then it depends on the underlying implementation (which, admittedly, is CPython-only right now). Probably worth a try.

da-woods: Fair enough - that's a reasonable - I imagine `PyType_GetSlot` is somewhat slower. It's probably just the `TryGetSlot` that's a little misleading then.

scoder: I changed the guard to include the Limited API in Py3.10+. Let's see.

scoder:
This is ready to merge. I think I got it to a point where it's never slower than the master branch and visibly quicker in many cases, up to 15% when hitting the fast paths.

I moved the `float/float` fast path first because it's the shortest possible code and thus benefits the most from quick dispatch. `int` goes second, but the type pointer compare dispatch is really fast. All mixed cases now live in a separate function to reduce the inlining cost.

Cython/Utility/Optimize.c
```diff
@@ -1075,6 +1075,301 @@ static PyObject* __Pyx__PyNumber_PowerOf2(PyObject *two, PyObject *exp, PyObject
...
+#ifndef __Pyx_DEFINED_BinopTypeError
+#define __Pyx_DEFINED_BinopTypeError
+
+static void __Pyx_BinopTypeError(PyObject *op1, PyObject *op2, const char* op, int inplace) {
+    // op1 is either 'int' or 'float', op2 is unknown.
+    __Pyx_TypeName type_name_op1 = __Pyx_PyType_GetFullyQualifiedName(Py_TYPE(op1));
+    __Pyx_TypeName type_name_op2 = __Pyx_PyType_GetFullyQualifiedName(Py_TYPE(op2));
+    PyErr_Format(PyExc_TypeError,
+        "unsupported operand type(s) for %.2s%.1s: '%.5s' and '%.200s'",
+        op,
```
da-woods, line 1127: `op` is length 1 isn't it and and could be passed as a `char` rather than a C string?

Cython/Utility/Optimize.c
```diff
@@ -1075,6 +1075,301 @@ static PyObject* __Pyx__PyNumber_PowerOf2(PyObject *two, PyObject *exp, PyObject
...
+static PyObject* __Pyx_{{op_name}}_xfloat_{{type2}}(PyObject *op1, PyObject *op2, int inplace) {
+    {{if type2 in ('object', 'int')}}
+    if ({{is_type('op2', 'int')}}) {
+        double float_op1 = __Pyx_PyFloat_AS_DOUBLE(op1);
+        #if !CYTHON_ASSUME_SAFE_MACROS
+        if (unlikely((float_op1 == -1.) && PyErr_Occurred())) return NULL;
+        #endif
+
+        {{if c_op == '*'}}
+        if (float_op1 == 0.) return __Pyx_NewRef(op1);
```
da-woods: I can come up with a case where this contradicts Python here:

```
>>> 0.0*(2**1000000)
    0.0*(2**1000000)
    ~~~^^~~~~~~~~~~~
OverflowError: int too large to convert to float
```

with that I don't see any reason not to just return 0.

Cython/Utility/Optimize.c
```diff
@@ -1075,6 +1075,301 @@ static PyObject* __Pyx__PyNumber_PowerOf2(PyObject *two, PyObject *exp, PyObject
...
+        if (likely(slot_func)) {
+            PyObject *result = slot_func(op1, op2);
+            if (likely(result != Py_NotImplemented)) {
+                return result;
+            }
+            Py_DECREF(result);
+        }
+        {{if c_op == '+'}}
+        // Adding a number to a sequence is rather unusual.
+        slot_func = __Pyx_PyType_GetSubSlot(type_op2, tp_as_sequence, sq_concat, binaryfunc);
```
da-woods: As far as I can tell `PyNumber_Add(a, b)` only tries `a` to see if it's a sequence. I don't think `sq_concat` gets called with the reverse operands.

(This isn't based on testing though, because I don't think any of the built-in classes are would work for this)

Cython/Utility/Optimize.c
```diff
@@ -1075,6 +1075,301 @@ static PyObject* __Pyx__PyNumber_PowerOf2(PyObject *two, PyObject *exp, PyObject
...
```
da-woods: Same here for concat - I don't think this gets reversed.

Cython/Utility/Optimize.c
```diff
@@ -1075,6 +1075,301 @@ static PyObject* __Pyx__PyNumber_PowerOf2(PyObject *two, PyObject *exp, PyObject
...
+            else if (__Pyx_PyLong_IsZero(op2)) return __Pyx_NewRef(op1);
+            {{endif}}
+
+            {{endif}}
+            #endif
+
+            binaryfunc slot_func = __Pyx_PyType_GetSubSlot(&PyLong_Type, tp_as_number, nb_{{slot_name}}, binaryfunc);
+            if (likely(slot_func)) {
+                return slot_func(op1, op2);
+            }
```
da-woods, line 1368: Does this need to be skipped if `!CYTHON_USE_TYPE_SLOTS and Python <3.10`? (I know no-one will actually be using that combination outside the Limited API though, but you've accounted for it earlier)

Cython/Utility/Optimize.c
scoder: I'm slowly looking into including division, and `//` has two characters. If it stays as it is now, I agree that a `char` would be simpler and (thus) better, but it's not a big difference and not so likely that we'll stop right here.

Cython/Utility/Optimize.c
scoder: Good call, but I'm rather inclined to consider this exception an implementation detail. It's an edge case, though. We could argue that, if `0 * x` is special enough to break the rules, then why not `1 * x` ? And if `0 * x` succeeds, then why not `1e-300 * x` ?

We could emulate it by moving the special case after the unpacking of `op2`. I doubt that the slow-down would be relevant in real code – multiplication by 0 is not going to be performance critical anywhere. It might not even be worth the branch, but it seemed so cheap to reuse the zero-object in `op1`.

da-woods: I was really trying to see if large floats got rounded to infinity (which they don't).

I'm completely in favour of leaving this as an implementation detail.

> if `0 * x` is special enough to break the rules, then why not `1 * x`

The difference is that `1.0 * x` also requires a conversion to `float` (which would legitimately be an overflow error). So I don't think there's a working rule-break for that.

Cython/Utility/Optimize.c
scoder: Right. Sequence repetition gets reversed, but concat doesn't.

Cython/Utility/Optimize.c
scoder: It shouldn't be a heap type issue, `PyLong` is definitely a static type in Py3.9. And without that constraint, we'd always try the slot functions in the other places, so this doesn't need a guard.

Cython/Utility/Optimize.c
scoder: >> if `0 * x` is special enough to break the rules, then why not `1 * x`
> 
> The difference is that `1.0 * x` also requires a conversion to `float` (which would legitimately be an overflow error). So I don't think there's a working rule-break for that.

Right, although I was mostly trying to avoid constructing semantics based on implementation details – under which I'd also comprise the boundary that starts raising an `OverflowError`, however reasonable it is. If we'd make it an implementation detail that `0 * x` always succeeds, then users could accidentally write code that relies on that, although from our perspective it's just an optional optimisation.

I think we can reverse the unpacking here, unsafe first, safe last. That would make it more natural to keep the special cases.

PR #7489, opened 2026-01-31
Don't store test files by shard number

da-woods:
They *shouldn't* overlap and conflict because they don't when run as a single test. I want to do this (if it works) because it makes https://github.com/cython/cython/pull/7456 better - we don't have to worry about shard number when working out where files are stored because they're all stored in the same place.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -2611,6 +2611,7 @@ def main():
         keep_alive_interval = 10
     else:
         keep_alive_interval = None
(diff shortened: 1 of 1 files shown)


scoder:
One problem is that all shards try to clear the work directory before creating it, which creates chaos if some have already started filling it. That can probably be avoided by clearing the directory from the main thread before starting the shards.

scoder (approved):

PR #7490, opened 2026-01-31
Do not initialize CythonScope when non-type attribute of cython module is used

### LINKED ISSUE #7487: [BUG] Using `cython.pointer[...]` and `cython.const[...]` causes C compilation error when shared utility is used
### Describe your issue

Running  `carray_iteration` fails with C compilation error when shared utility is used:
```
carray_iteration.c: In function ‘__pyx_memoryview_check’:
carray_iteration.c:12642:61: error: ‘__pyx_mstatetype’ has no member named ‘__pyx_memoryview_type’
12642 |   __pyx_t_1 = __Pyx_TypeCheck(__pyx_v_o, __pyx_mstate_global->__pyx_memoryview_type);
```

Steps to reproduce:

1. Use branch from #7479 PR
2. Run test with shared utility enabled: ` python runtests.py --no-fork --shared-utility --no-capture --no-cpp carray_iteration`

Test fails with C compilation error mentioned above.

Note: regular test without `--shared-utility` is passes without issue.

> UPDATE:
> Much simpler reproducer is:
> ```
> import cython
> 
> @cython.cfunc
> def main(a: cython.pointer[int]):
 >    pass
> ```
> 
> Compiling using shared utility enabled casues C compilation to fail with mentioned error.

matusvalo:
When Cython analyses type annotation containing `cython.XYZ[...]` (`cython.const[]`, `cython.pointer[]`) it tries to look up Cython scope to find type matching `XYZ` name. 

https://github.com/cython/cython/blob/62a4106657b31485aed27ecdc42c68012b02cd8c/Cython/Compiler/ExprNodes.py#L4188
https://github.com/cython/cython/blob/62a4106657b31485aed27ecdc42c68012b02cd8c/Cython/Compiler/ExprNodes.py#L7962

There is no such type (`const`, `pointer`) hence Cython initializes CythonScope which ends up loading memoryview utility code.

https://github.com/cython/cython/blob/62a4106657b31485aed27ecdc42c68012b02cd8c/Cython/Compiler/CythonScope.py#L55
https://github.com/cython/cython/blob/62a4106657b31485aed27ecdc42c68012b02cd8c/Cython/Compiler/CythonScope.py#L136

This PR simply detects Cython scope valid attributes which occurs in type definition but are not type itself and just returns expected `None` without loading Cython scope.

Closes #7487


Cython/Compiler/CythonScope.py
```diff
@@ -6,6 +6,7 @@
 from . import MemoryView
 from .StringEncoding import EncodedString
 
+NON_TYPE_NAMES = {'pointer', 'const'}
```
da-woods: This is definitely missing `volatile`.

I'm a bit suspicious of this solution in general because I think the list is incomplete (and I don't think `volatile` makes it complete).

I think it'd be better to be more selective about loading the `view` submodule only when `view` is specifically looked. I don't know how hard that is though.

matusvalo: > I think it'd be better to be more selective about loading the view submodule only when view is specifically looked. I don't know how hard that is though.

The memoryview in CythonScope is needed only when `cython.view` is imported/used?

da-woods: I think so (but not 100% sure)

matusvalo: > I think it'd be better to be more selective about loading the view submodule only when view is specifically looked. I don't know how hard that is though.

I think it makes sense. For now, I expanded the list but I will try to investigate whether we can avoid loading MemoryView in case `cython.view` is not used/imported.

scoder: I agree with @da-woods that this is really a stopgap solution. We should keep looking for a better way to initialise and integrate only what is truly needed.

In the meantime, the change here (hopefully) allows us to run tests with the shared module which is dearly needed.

PR #7491, opened 2026-01-31
Docs: Mention type qualifiers in pure python tutorial

matusvalo:


docs/src/tutorial/pure.rst
```diff
@@ -232,6 +232,10 @@ Further pointer types can be constructed with ``cython.pointer[cython.int]``
 and arrays as ``cython.int[10]``. A limited attempt is made to emulate these
 more complex types, but only so much can be done from the Python language.
 
+Pure python mode also supports type qualifiers ``const`` and ``volatile``,
+which can be applied using ``cython.const[cython.int]`` and
+``cython.volatile[cython.int]``. For more details see :ref:`type_qualifiers` section.
+
 The Python types int, long and bool are interpreted as C ``int``, ``long``
```
matusvalo: Is this correct? `int`, `long` and `bool` are not interpreted as C types but rather as Python types. For C types one must use `cython.int`, etc.

da-woods: `bool` looks to be interpreted as a C type (although `Optional[bool]` isn't).

`int` is interpreted as the Python type though.

`long` gives gives me `Unknown type declaration 'long' in annotation, ignoring` with language level 3. It's converted to a Python int with language level 2. I suspect `long` is obscure and outdated enough to not be worth documenting.

scoder: I've corrected the documentation in
https://github.com/cython/cython/commit/3817e3da093c3bba6332b45e611c1e2748e26aba

PR #7495, opened 2026-02-02
Simplify __Pyx_PyUnicode_Join()

scoder:
Use more direct 'kind' instead of indirect 'max character' calculation to determine Unicode result kind in __Pyx_PyUnicode_Join(). This avoids an indirection, seems to make the code a bit faster and modules that use a lot of f-strings are slightly smaller.

For simple f-string patterns like pure value concatenation or when texts and placeholders alternate, we can easily replace the length and kind aggregation using C loops with appropriate steps, so I've implemented that, too.

I've removed the previous node result code deduplication here since I doubt that it ever had any effect. Values will usually reside in temps with distinct names, even when the values are identical.

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -3834,8 +3834,7 @@ class JoinedStrNode(ExprNode):
             if not indices:
                 return
 
-            # See guard in __Pyx_PyUnicode_Join() for using length and kind.
-            code.putln("#if CYTHON_USE_UNICODE_INTERNALS && CYTHON_ASSUME_SAFE_MACROS && !CYTHON_AVOID_BORROWED_REFS")
+            code.putln("#if __Pyx_PyUnicode_Join_CAN_USE_KIND_AND_LENGTH")
 
             if len(indices) == 1:
                 code.putln(f"{result_temp} {operator}= {cfunc_name}({values_array}[{indices[0]}]);")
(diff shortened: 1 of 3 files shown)


da-woods:
Could we completely `ifdef` out the length and kind calculations for the cases where they're unused? (Limited API, PyPy, GraalPy etc)

I suspect the kind is trivial to optimise out (because it's always 0 in the Limited API) but length probably isn't because it's an opaque function call that could at least have the side effect of setting an exception.

da-woods:
I couldn't see any obvious issues apart from that.

scoder:
> Could we completely `ifdef` out the length and kind calculations for the cases where they're unused?

I think it's problematic that `__Pyx_PyUnicode_Join()` is part of the shared module, so the two different implementations may need to work together in some situations. We can probably exclude the different runtimes that _cannot_ support the API, but we might run into issues if we base the distinction on specific feature flags. Users might want to set them at a per-module level. I'll see if I can make this safer.

da-woods:
> I think it's problematic that `__Pyx_PyUnicode_Join()` is part of the shared module

My first thought would be just to move it out of the shared module. I think the shared module should be memoryviews + "anything else that's convenient to add" and so we shouldn't feel bad about moving things in and out of it fairly freely.

----------------------

My second thought was:

the version based on `PyUnicode_Join` version will always work even if we've already calculated the length/kind. So in principle we could always include that and just have a special value of `length` and `kind` to indicate unset. So
* "optimized" call + "optimized" shared module - works
* "optimized" call + "unoptimized" shared module - works, but does some unnecessary length+kind calculations
* "unoptimized" call + "optimized" shared module  - works - falls back to `PyUnicode_Join`
* "unoptimized" call + "unoptimized" shared module - works

It's potentially a little fragile though because two separate places need to be kept in sync.

PR #7502, opened 2026-02-06
Assign int/float/bool Python types when coercing C values to Python

scoder:
We dropped the known type and instead inferred `object`.

While changing this, I noticed a bug where we inferred C `double` for `x = int(a) if condition else float(a)`, looking at the spanning type when we shouldn't. Testing this here but probably worth fixing independently.

### DIFF
--- a/Cython/Compiler/Optimize.py
+++ b/Cython/Compiler/Optimize.py
@@ -2368,7 +2368,7 @@ class OptimizeBuiltinCalls(Visitor.NodeRefCleanupMixin,
                     node.type.is_int and isinstance(arg, ExprNodes.BoolNode)):
                 return arg.coerce_to(node.type, self.current_env())
         elif isinstance(arg, ExprNodes.CoerceToPyTypeNode):
-            if arg.type is PyrexTypes.py_object_type:
+            if arg.type is PyrexTypes.py_object_type or arg.type.is_builtin_type:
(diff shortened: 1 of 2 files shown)


Cython/Compiler/Optimize.py
```diff
@@ -2368,7 +2368,7 @@ def visit_CoerceFromPyTypeNode(self, node):
                     node.type.is_int and isinstance(arg, ExprNodes.BoolNode)):
                 return arg.coerce_to(node.type, self.current_env())
         elif isinstance(arg, ExprNodes.CoerceToPyTypeNode):
-            if arg.type is PyrexTypes.py_object_type:
+            if arg.type is PyrexTypes.py_object_type or arg.type.is_builtin_type:
```
scoder, line 2371: I'm not entirely sure yet about this bit. It's worth growing our test cases in that corner. It has the potential to break cases where we jump though more than one type to achieve a conversion, like (pseudo-code) `x = <int> <pybool> int_val` or `y = <char> <pyint> large_int`, where we'd lose the value range safety of the Python object unpacking.

scoder: It's not a regression, though, so I'll leave the testing for later. We previously inferred `object`, so the cases that now change already fulfilled this condition before.

PR #7503, opened 2026-02-07
Fix limited API enum version check

da-woods:
Because it's runtime behaviour so on the Limited API it needs to be decided with the runtime version.

The change of declaration from `int` to `enum` is to avoid the problem with trying to overwrite names when using `from something import *`. I think that may be fixed anyway, but `enum` definitely avoids it.

### DIFF
--- a/Cython/Utility/CpdefEnums.pyx
+++ b/Cython/Utility/CpdefEnums.pyx
@@ -30,7 +30,7 @@ cdef extern from *:
     # Try to look up the module name dynamically if possible
 ], module=globals().get("__module__", '{{static_modname}}'))
 
-if (__PYX_LIMITED_VERSION_HEX >= 0x030B0000 or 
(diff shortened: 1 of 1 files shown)


Cython/Utility/CpdefEnums.pyx
```diff
@@ -27,7 +30,8 @@ cdef extern from *:
     # Try to look up the module name dynamically if possible
 ], module=globals().get("__module__", '{{static_modname}}'))
 
-if PY_VERSION_HEX >= 0x030B0000:
+if (__PYX_LIMITED_VERSION_HEX >= 0x030B0000 or
```
da-woods:
```suggestion
if (__PYX_LIMITED_VERSION_HEX >= 0x030B0000 or
```

da-woods:
3.2.x commit 40ffcdd57

PR #7505, opened 2026-02-07
Implement cdef property setters on ext-types

da-woods:
Previously we'd only supported getters.

With opaque objects, people exposing external types may want to hide being accessor functions more rather than exposing the class directly. This change should allow them to do so.

### DIFF
--- a/Cython/Compiler/Symtab.py
+++ b/Cython/Compiler/Symtab.py
@@ -2765,7 +2765,7 @@ class CClassScope(ClassScope):
         """
         property_entry = self.declare_property(name, doc=doc, ctype=type, pos=pos)
         cfunc_entry = property_entry.scope.declare_cfunction(
-            name=name,
+            name="__get__",
             type=PyrexTypes.CFuncType(
                 type,
                 [PyrexTypes.CFuncTypeArg("self", self.parent_type, pos=None)],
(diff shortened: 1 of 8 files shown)


Cython/Compiler/UtilNodes.py
```diff
@@ -387,3 +387,27 @@ def generate_result_code(self, code):
 
     def calculate_result_code(self):
         return str(int(self.in_nogil_context))
+
+
+class CPropertySetNode(ExprNodes.ExprNode):
+    subexprs = ['func_node']
+    arg1 : ExprNodes.RawCNameExprNode
+    func_node : ExprNodes.ExprNode  # most likely a call node
```
scoder: Probably, yes. So why not name the attribute `call_node` ? At least from the point of view of this node, it holds the node that implements the call.

Cython/Compiler/Symtab.py
```diff
@@ -3030,16 +3030,20 @@ def __init__(self, name, class_scope):
...
+            error(pos, "C property getter cannot return 'void'")
+        elif name == "__set__" and not type.return_type.is_void:
+            error(pos, "C property setter must return 'void'")
+
+        if name == "__get__" and len(type.args) != 1:
+            error(pos, "C property getter must have a single (self) argument")
+        elif name == "__set__" and len(type.args) != 2:
+            error(pos, "C property setter must have two arguments (self and value)")
+        elif not (type.args[0].type.is_pyobject or type.args[0].type is self.parent_scope.parent_type):
+            error(pos, "C property method must have a single (object) argument")
```
scoder: Wouldn't it be cleaner to separate the error checks by get/set?

scoder (approved):
Nice. Just a few style comments.

Cython/Compiler/Symtab.py
```diff
@@ -3030,16 +3030,23 @@ def __init__(self, name, class_scope):
...
-
+        if name=="__get__":
+            if type.return_type.is_void:
+                error(pos, "C property getter cannot return 'void'")
+            if len(type.args) != 1:
+                error(pos, "C property getter must have a single (self) argument")
+
+        if name=="__set__":
+            if not type.return_type.is_void:
+                error(pos, "C property setter must return 'void'")
```
scoder, line 3041: Why must setters return `void` ? What if they have to raise exceptions? Can't they return an `int` error code? Even your test uses `WriteUnraisable()` instead of propagating the exception (as I would have expected).

EDIT: PR to make them return an `int` error code is up at https://github.com/cython/cython/pull/7791

PR #7507, opened 2026-02-07
Refactor of refcounting functions in PyrexTypes

da-woods:
The main aim here is that the functions no longer take a `code` argument. The code argument was always a bit awkward I think (although the disadvantage of the updated approach is that working out whether to make the funcstate is "needs_refnanny" is a little more awkward).

The longer-term motivation here is to replace some uses of `INCREF` with `NEWREF`. This is mainly for freethreading-related reasons (because we typically do `INCREF` after assignment which is a little more awkward to make thread-safe). However, `NEWREF` is typically a code fragment rather than a complete line, so it fits better into this scheme than the current one.

### DIFF
--- a/Cython/Compiler/Code.py
+++ b/Cython/Compiler/Code.py
@@ -3106,8 +3106,8 @@ class CCodeWriter:
         from .PyrexTypes import py_object_type, typecast
         return typecast(py_object_type, type, cname)
 
-    def handle_refnanny(self, type, nanny=True):
-        if type.supports_refnanny and nanny:
+    def handle_refnanny(self, tp, nanny=True):
+        if tp.supports_refnanny and nanny:
             self.funcstate.needs_refnanny = True
 
     def put_gotref(self, cname, type):
--- a/Cython/Compiler/PyrexTypes.py
+++ b/Cython/Compiler/PyrexTypes.py
@@ -198,7 +198,7 @@ class PyrexType(BaseType):
     #  needs_refcounting     boolean     Needs code to be generated similar to incref/gotref/decref.
     #                                    Largely used internally.
     #  refcounting_needs_gil boolean     Reference counting needs GIL to be acquired.
-    #  supports_refnanny     boolean     Reference counting for can use refnanny
+    #  supports_refnanny     boolean     This type can use refnanny to check its reference counting.
     #  equivalent_type       type        A C or Python type that is equivalent to this Python or C type.
     #  default_value         string      Initial value that can be assigned before first user assignment.
     #  declaration_value     string      The value statically assigned on declaration (if any).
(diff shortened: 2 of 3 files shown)


Cython/Compiler/Code.py
```diff
@@ -3106,50 +3106,74 @@ def as_pyobject(self, cname, type):
         from .PyrexTypes import py_object_type, typecast
         return typecast(py_object_type, type, cname)
 
+    def handle_refnanny(self, type, nanny=True):
```
scoder: This should definitely become a final cdef function in `Code.pxd`.

Cython/Compiler/PyrexTypes.py
```diff
@@ -198,6 +198,7 @@ class PyrexType(BaseType):
     #  needs_refcounting     boolean     Needs code to be generated similar to incref/gotref/decref.
     #                                    Largely used internally.
     #  refcounting_needs_gil boolean     Reference counting needs GIL to be acquired.
+    #  supports_refnanny     boolean     Reference counting for can use refnanny
```
scoder: My English parser fails to work its way through this comment.

Cython/Compiler/Code.py
```diff
@@ -3106,50 +3106,74 @@ def as_pyobject(self, cname, type):
...
+        self.putln(type.get_xdecref_code(cname, nanny=nanny, have_gil=have_gil))
 
     def put_decref_clear(self, cname, type, clear_before_decref=False, nanny=True, have_gil=True):
-        type.generate_decref_clear(self, cname, clear_before_decref=clear_before_decref,
-                              nanny=nanny, have_gil=have_gil)
+        if nanny:
+            self.handle_refnanny(type)
+        self.putln(
+            type.get_decref_clear_code(cname, clear_before_decref=clear_before_decref,
+                              nanny=nanny, have_gil=have_gil))
```
scoder: Time to make the indentation less funny.
```suggestion
        self.putln(type.get_decref_clear_code(
            cname, clear_before_decref=clear_before_decref, nanny=nanny, have_gil=have_gil))
```

Cython/Compiler/Code.py
```diff
@@ -3106,50 +3106,74 @@ def as_pyobject(self, cname, type):
...
 
     def put_xdecref_clear(self, cname, type, clear_before_decref=False, nanny=True, have_gil=True):
-        type.generate_xdecref_clear(self, cname, clear_before_decref=clear_before_decref,
-                              nanny=nanny, have_gil=have_gil)
+        if nanny:
+            self.handle_refnanny(type)
+        self.putln(
+            type.get_xdecref_clear_code(cname, clear_before_decref=clear_before_decref,
+                              nanny=nanny, have_gil=have_gil))
```
scoder:
```suggestion
        self.putln(type.get_xdecref_clear_code(
            cname, clear_before_decref=clear_before_decref, nanny=nanny, have_gil=have_gil))
```

PR #7509, opened 2026-02-08
Add some missing `noexcept` to refnanny

da-woods:
Some of them are unnecessary and only really for documentation. However the internal call to `GOTREF` was generating exception handling (plus it had no return value to check), and the calls to `GIVEREF_and_report` were also generating exception handling.

### DIFF
--- a/Cython/Runtime/refnanny.pyx
+++ b/Cython/Runtime/refnanny.pyx
@@ -139,7 +139,7 @@ cdef PyObject* SetupContext(char* funcname, Py_ssize_t lineno, char* filename) e
     PyErr_Restore(type, value, tb)
     return result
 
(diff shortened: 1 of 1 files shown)


da-woods:
I'll put this for 3.3. The defect does technically exist in older versions (probably back to 3.0.x) but it's only a small amount of unnecessary reference checking in a mostly internal tool

scoder:
Is this a problem in the user module code? That uses external function pointers, I think, which shouldn't assume exception propagation.

da-woods:
> Is this a problem in the user module code? That uses external function pointers, I think, which shouldn't assume exception propagation.

Yes, the code that uses refnanny manually uses these functions via function pointers and it assumes that they don't propagate exceptions. So the change doesn't affect that.

PR #7514, opened 2026-02-09
Fix flow control issue with walrus+exceptions

### LINKED ISSUE #7462: [BUG] Cython flow analysis bug: walrus operator + raise inside try/except
### Describe the bug

An edge case with the walrus operator + raise inside try/except doesn't compile. I am porting existing tested production code to cython and found a subset of walrus operators could not compile due to this issue

```cython
async def fails_in_try():
    """
    Error on line 34: local variable 'x' referenced before assignment
    """
    try:
        if not (x := await get_value()):
            raise ValueError("Not found")

        return x["key"]  # <-- Cython thinks x might be unassigned
    except Exception:
        pass
```

### Expected behaviour

No error should be raised.

### OS

Ubuntu 25.10

### Python version

Python 3.13

### Cython version

Cython 3.2.4

### Additional context

```python
# =============================================================================
# WORKS: Same pattern outside try/except
# =============================================================================
async def works_outside_try():
    """This compiles fine - Cython correctly tracks the raise terminates flow."""
    if not (x := await get_value()):
        raise ValueError("Not found")

    return x["key"]  # <-- Cython knows x is assigned here


# =============================================================================
# WORKS: Separate assignment from condition workaround
# =============================================================================
async def workaround():
    """Separating assignment from condition works around the bug."""
    try:
        x = await get_value()
        if not x:
            raise ValueError("Not found")

        return x["key"]
    except Exception:
        pass
```

da-woods:
Fixes #7462

If an assignment node is in an exception block, then `mark_assignment` can change `self.flow.block`. This means that `self.flow.block` at the end of evaluation an `if` condition might be different to `self.flow.block` at the start. Therefore, we continue from the state at the end.

Also fix the equivalent thing in match-case guard blocks.

Also add a small print function just because it was useful for debugging this.

### DIFF
--- a/tests/run/extra_walrus.py
+++ b/tests/run/extra_walrus.py
@@ -454,7 +454,7 @@ def case_in_try_block_2(x):
         return f'Normal return {a}'
     except:
         return f'Except return {a}'
-    
+
 def match_in_try_block_2(x):
     """
     >>> match_in_try_block_2(True)
@@ -463,7 +463,10 @@ def match_in_try_block_2(x):
     'Normal return False'
     """
     try:
-        match (a := return_x(x)):
+        # Note that pycodestyle on older versions of Python
+        # enforces a weird spacing here and can't be turned
+        # off with noqa.  Undo this in future...
+        match(a := return_x(x)):
             case True:
                 raise RuntimeError
         return f'Normal return {a}'
@@ -501,11 +504,11 @@ def except_in_try_block(x):
             raise RuntimeError()
         except:
             pass
-            
+
         return f'Normal return {a.__name__}'
     except Exception:
         return f'Except return {a.__name__}'
-    
+
 def for_in_try_block_1(x):
     """
     >>> for_in_try_block_1(10)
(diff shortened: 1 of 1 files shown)


da-woods:
https://github.com/cython/cython/blob/7c47fd8857d8b5461bc8cb9a8eab9d4943844abd/Cython/Compiler/FlowControl.py#L793-L796

and

https://github.com/cython/cython/blob/7c47fd8857d8b5461bc8cb9a8eab9d4943844abd/Cython/Compiler/FlowControl.py#L821-L824

are where the re-assignments take place.

-------------------------------------------------------

However... putting this in draft for now, because it also affects `while` blocks too (and possibly some other things) so needs further work.

da-woods:
My view is it should be safe for 3.2.5.

Arguably the risk of breakage is more that generic loops might be broken rather than assignment expressions specifically. So the tests added here possibly wouldn't show it. But overall I think it should be tested pretty well.

da-woods:
3.2.x commit https://github.com/cython/cython/pull/7524/commits/71d6b77d277fc6d8802224d409cc4c8bd3fa602e

PR #7515, opened 2026-02-10
Allow casting C values to Python 'bool'

### LINKED ISSUE #7513: [BUG] Python bool casts are inconsistent
### Describe the bug

When casting different types to `<bool>`, the behaviour seems inconsistent.

```cython
# [1]
cdef int int_x = 5
cdef int int_bool_int_x = <int> <bool> int_x
# TypeError: Expected bool, got int

# [2]
cdef object x = 5
cdef char c = <char> <bint> x
# c = 1
cdef object x = 5
cdef char c = <char> bool(x)
# c = 1
cdef object x = 5
cdef char c = <char> <bool> x
# c = 5

# [3]
cdef object x = 555
cdef char c = <char> <bool> x
# OverflowError: value too large to convert to char
```

### Expected behaviour

1) We currently convert the C `int` to PyLong, then cast to PyBool (which fails in the type check) and then convert the result to C `int`. We should instead obey the cast and convert the C `int` to PyBool, then to C `int`. If we want to optimise, the way is through `bint` instead of  `bool`.
I would consider a successful conversion more helpful here than a (compile time) failure.

2) I think the `<bool>` cast ~should actually fail at runtime because~ we are trying to cast a Python object that is not `bool` to `bool`. If users want to do this, they can use `<bint>` or `bool()` to get an explicit conversion instead of a mere Python type cast. EDIT: We don't do type checks for explicit casts, so this is not a runtime error.
We might also want to issue warnings when users cast between incompatible known Python types (although that doesn't apply to `object` in this example code).

3) This is more or less the same as [2] and should fail with a `TypeError` on the first cast, not get through to the second.

**Alternatively:**
We could consider `<bool>` the Python equivalent of `<bint>` and always implement it as a truthiness test.
This wouldn't be quite in line with other Python types casts (which enforce a type tests), but would match the current behaviour of Python typed function arguments:

```cython
def bool_arg(x: bool):
    return x

# bool_arg(0) -> False
# bool_arg(1) -> True
# bool_arg(12345) -> True
```

One issue with this is that we'd introduce several ways to do it (`bool()`, `<bint>`, `<bool>` would all do the same thing), but without a way to get an actual `bool` type cast with type check if there is need for that. Users would have to resort to `isinstance()` and implement the error themselves. Not difficult, but it might be surprising to users that this is required, compared to other Python builtin type casts.

Given "explicit is better than implicit",
**my proposal is**:

- ~Make `<bool>` an explicit type cast and reject non-bool Python types.~
- Let `<bool>` work on C types but make the result always True/False or 0/1 (when aliasing to `bint`). Use the C type specific truthiness test for the conversion. (NOTE: this is really optional – it seems perfectly ok to require `bool()` or `<bint>` instead of allowing a cast to `bool`.)
- Require `bool()` or `<bint>` for boolean type conversions and (continue to) translate them to a truthiness test internally.

### OS

_No response_

### Python version

_No response_

### Cython version

master

### Additional context

This also applies to older Cython versions but should not change in a point release.

scoder:
We'd previously used the default type coercion and then tried to cast the resulting object to `bool`. That almost never works and leads to surprising errors. Instead, we can just allow `<bool> cvalue` and use the truthiness of the C value.

We now also warn about invalid casts between Python builtins that make no sense at runtime. This is especially interesting for `<bool>` since it might seem reasonable to assume a truthiness check also for Python objects, whereas the cast really only sets the type information. There's `bool()` for converting the type.

Closes https://github.com/cython/cython/issues/7513

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -8,7 +8,7 @@ cython.declare(error=object, warning=object, warn_once=object, InternalError=obj
                CompileError=object, UtilityCode=object, TempitaUtilityCode=object,
                StringEncoding=object, operator=object, local_errors=object, report_error=object,
                Naming=object, Nodes=object, PyrexTypes=object, py_object_type=object,
-               list_type=object, tuple_type=object, set_type=object, dict_type=object,
+               list_type=object, tuple_type=object, set_type=object, dict_type=object, bool_type=object,
                unicode_type=object, bytes_type=object, type_type=object,
                Builtin=object, Symtab=object, Utils=object, find_coercion_error=object,
                debug_disposal_code=object, debug_temp_alloc=object, debug_coercion=object,
@@ -44,7 +44,7 @@ from .PyrexTypes import c_char_ptr_type, py_object_type, typecast, error_type, \
 from . import TypeSlots
 from .Builtin import (
     list_type, tuple_type, set_type, dict_type, type_type,
-    unicode_type, bytes_type, bytearray_type,
+    unicode_type, bytes_type, bytearray_type, bool_type,
     slice_type
 )
 from . import Builtin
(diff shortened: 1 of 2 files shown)


Cython/Compiler/ExprNodes.py
```diff
@@ -11833,9 +11836,17 @@ def analyse_types(self, env):
         elif from_py and to_py:
             if self.typecheck:
                 self.operand = PyTypeTestNode(self.operand, self.type, env, notnone=True)
-            elif isinstance(self.operand, SliceIndexNode):
-                # This cast can influence the created type of string slices.
-                self.operand = self.operand.coerce_to(self.type, env)
+            else:
+                if self.type.is_builtin_type and self.operand.type.is_builtin_type and self.type != self.operand.type:
+                    if self.type is int_type and self.operand.type is bool_type:
```
```suggestion
                    if self.type.is_pyint_type and self.operand.type.is_pybool_type:
```

scoder:
I am not entirely sure that we should allow `<bool>` casts.
Also, this PR is partially broken by the change to allow subscripted builtin types. I'll change its status to draft.

PR #7517, opened 2026-02-10
BLD: avoid deprecated pypi trove classifier

neutrinoceros:
This avoids the following deprecation warning, which I hit when I built Cython in a CI env with `-Werror`

```
********************************************************************************
              Please consider removing the following classifiers in favor of a
      SPDX license expression:

              License :: OSI Approved :: Apache Software License

              See
      https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license
      for details.
      
      ********************************************************************************
```

setuptools 77.0.0 (yanked) was the first release to support the replacement, so I'm setting 77.0.1 as the lower bound.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools"]
+requires = ["setuptools>=77.0.1"]
 
 [tool.cibuildwheel]
 build-verbosity = 2
--- a/setup.py
+++ b/setup.py
@@ -454,10 +454,10 @@ def run_build():
 
         """) + collect_changelog(version),
         license='Apache-2.0',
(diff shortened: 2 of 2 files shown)


scoder:
Merci. setuptools 77.x is not even one year old, so I'll keep this change to Cython 3.3.

neutrinoceros:
suite yourself !

da-woods:
I think the important bits of this (i.e. the bit that causes the warning) were in https://github.com/cython/cython/commit/1a0f67b2894f571a35d7d196f1206c164a55cba0 and we don't _really_ need license_files given we pass `license='Apache-2.0'`.

So I'll close this. Thanks for pointing it out.

PR #7518, opened 2026-02-10
MNT: simplify cibuildwheel configuration

neutrinoceros:
With cibuildwheel 3.x, which you're already using, you don't need to explicitly skip pypy or graalpy: they're already unselected by default. There's also a simpler way to express the minimum Python version supported, that does *not* require setting up the full `[project]` table in `pyproject.toml`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,10 +2,9 @@
 requires = ["setuptools"]
 
 [tool.cibuildwheel]
+project-requires-python = ">=3.9"
 build-verbosity = 2
 skip = [
-    # Unsupported versions:
-    "cp36*", "cp37*", "cp38*",
     # These platforms are served with the generic Python wheel:
     "pp*",
     "gp*",


da-woods:
`project-requires-python` definitely looks like an improvement.

For me I'd probably prefer to be explicit about PyPy/GraalPy - just because it's a deliberate decision on our part but cibuildwheel could change in future.

neutrinoceros:
updated

scoder:
Merci.

scoder:
Oh, looks like I somehow missed the fact that the build had failed. I think it's just due to a misnaming of the option.
Probably fixed in https://github.com/cython/cython/commit/08559892bced0d0fe8f737f223ca582f76c1b41e

neutrinoceros:
Whoops, I think I misread [the docs](https://cibuildwheel.pypa.io/en/stable/options/#requires-python). On the bright side, it looks like I *also* missed that `cibuildwheel` understands `python_requires` as a keyword argument to `setuptools.setup`, which is what should be done here.

scoder:
> `cibuildwheel` understands `python_requires` as a keyword argument to `setuptools.setup`, which is what should be done here.

We already pass that option. I'll try if that suffices.

scoder:
> We already pass that option. I'll try if that suffices.

It doesn't. Time to revert.

neutrinoceros:
Sorry about this mishap. It looks like a bug in cibuildwheel and I might try fixing it there.

neutrinoceros:
Ah, I see what happened: `python_requires` is set as a dict value, while cibuildwheel only looks for it in arguments passed by keywords.
It would be easy to fix here, if only the dict wasn't populated only *conditionally*:
https://github.com/cython/cython/blob/f158aa964d4aced895f1a69fd9902e03ee996a5c/setup.py#L72
I don't actually understand why it's the case now. Any chance this was done for historical reasons that don't apply anymore ?

PR #7519, opened 2026-02-10
Don't store a PyCFunction in Limited API CyFunction

### LINKED ISSUE #7516: [BUG] Compiling with limited API changes lifetime of nested closures
### Describe the bug

If I compile the example pieces below _without_ the limited API:
```
$ python runner.py 
Calling capture
<__main__.Obj object at 0x77c76099ba10>
Tore down obj
Called capture
Exited scope of main()
```

In contrast _with_ the limited API:
```
$ python runner.py 
Calling capture
<__main__.Obj object at 0x78c88e9fba10>
Called capture
Exited scope of main()
Tore down obj
```

This means that defining nested closures in cython with the limited API is very unsafe because if they close over objects that otherwise would be collected via refcount-based destruction (and I've arranged that this is always possible) those same objects are now only destructed via GC-based refcycle breaking. If they rely on various bits of C state, it can be very difficult to arrange that the destruction order is safe.

```cython
# capture_lifetime.pyx
def capture(obj):
    def closure() -> None:
        print(obj)
    return closure()
```
```python
import capture_lifetime
import weakref

class Obj:
    def __init__(self):
        weakref.finalize(self, lambda: print("Tore down obj"))

def main():
    print("Calling capture")
    capture_lifetime.capture(Obj())
    print("Called capture")

if __name__ == "__main__":
    main()
    print("Exited scope of main()")
```
```python
# setup.py

setup(
    ext_modules=cythonize([
        Extension(
            name="capture_lifetime",
            sources=["capture_lifetime.pyx"],
            define_macros=[
                ("Py_LIMITED_API", 0x030B0000),
            ],
            py_limited_api=True,
        ),
    ]))
```

### Expected behaviour

Nested closures should have the same lifetime independent of whether the limited API is used.

### OS

_No response_

### Python version

3.13.11

### Cython version

3.2.4

### Additional context

_No response_

da-woods:
Fixes #7516. (I'm reluctant to fix it just for guarranteed destruction order, because I think that's a bit dangerous to promise, but I think the change is an overall improvement anyway).

Instead store the relevant data directly, which should hopefully improve things by giving slightly more direct access. I also spotted that the `self` attribute of CyFunctions was never changed and so we can reliably assume its value in most cases.

### DIFF
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -56,8 +56,8 @@ typedef struct {
 #if !CYTHON_OPAQUE_OBJECTS
     PyObject_HEAD
 #endif
-    // We can't "inherit" from func, but we can use it as a data store
-    PyObject *func;
+    PyMethodDef *func_methoddef;
+    PyObject *func_module;
 #else
     // PEP-573: PyCFunctionObject + mm_class
     PyCMethodObject func;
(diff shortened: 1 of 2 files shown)


da-woods:
> Nice cleanup, that's much better than before.

I think I was overly literal when I first translated this to the Limited API and just tried to reproduce the existing code.

> Changes the ABI, though, so definitely not for 3.2.x.

The shared module is disambiguated with the full version number so I don't actually think this is a problem. But I also don't think this needs to be in 3.2.x.

PR #7526, opened 2026-02-14
Fix f-string deduplication of simple "{s}" formatting for known str values.

scoder:
And use multiplication instead of repeated GET_LENGTH() calls for duplicate substrings in f-string joining.

This cuts some 90% runtime off the excessive f-string "many expressions" test.


scoder:
I'll merge this since it really reduces the test runtime a lot.

PR #7528, opened 2026-02-15
Fix using sizeof in exern arrays

### LINKED ISSUE #7451: [BUG] Compiler crash in 3.2.x when using sizeof() in C array dimension inside struct/union definition
### Describe the bug

Using sizeof(type) as an array dimension inside a struct or union definition causes a compiler crash with 
"TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'." This is a regression from Cython 3.1.x where the same code compiled successfully for extern block declarations.

test_sizeof.pxd:

```cython
from libc.stdint cimport uint8_t

cdef extern from "test.h":
    struct foo:
        uint8_t a
        uint8_t b

    union bar:
        foo foo
        uint8_t buf[sizeof(foo)]  # Crashes in 3.2.x
```

test.h:

```c
#include <stdint.h>

struct foo { uint8_t a, b; };

union bar {
    struct foo foo;
    uint8_t buf[sizeof(struct foo)];
};
```

test_sizeof.pyx:
```cython
cimport test_sizeof

def test():
    return sizeof(test_sizeof.foo)
```
  
Results in

```text

Error compiling Cython file:
...
        uint8_t a
        uint8_t b

    union bar:
        foo foo
        uint8_t buf[sizeof(foo)]  # Crashes in 3.2.x
                   ^
/tmp/cython_sizeof_bug/test_sizeof.pxd:10:19: Compiler crash in AnalyseDeclarationsTransform

File 'ModuleNode.py', line 286, in analyse_declarations: ModuleNode(test_sizeof.pxd:1:0,
    full_module_name = 'test_sizeof',
    is_pxd = True)
File 'Nodes.py', line 426, in analyse_declarations: StatListNode(test_sizeof.pxd:1:0)
File 'Nodes.py', line 481, in analyse_declarations: CDefExternNode(test_sizeof.pxd:3:0,
    include_file = 'test.h')
File 'Nodes.py', line 426, in analyse_declarations: StatListNode(test_sizeof.pxd:4:4)
File 'Nodes.py', line 1642, in analyse_declarations: CStructOrUnionDefNode(test_sizeof.pxd:8:4,
    in_pxd = True,
    kind = 'union',
    name = 'bar',
    visibility = 'extern')
File 'Nodes.py', line 1572, in analyse_declarations: CVarDefNode(test_sizeof.pxd:10:8,
    modifiers = [...]/0,
    visibility = 'extern')
File 'Nodes.py', line 651, in analyse: CArrayDeclaratorNode(test_sizeof.pxd:10:19,
    calling_convention = '')

  File "/home/thormick/gapit/gapit-modbus/.venv/lib/python3.14/site-packages/Cython/Compiler/Nodes.py", line 650, in analyse
    size = int(size)
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

### Expected behaviour

- Cython 3.1.x: Compiles successfully (the sizeof is passed through to C)
- Cython 3.2.x: Should also compile successfully

### OS

Linux

### Python version

3.14

### Cython version

3.2.4

### Additional context

The immediate cause seems to be how [a2a713e290](https://github.com/cython/cython/commit/a2a713e290#diff-8789a479ea870f7646251224385afa3ebffe9b6acab136ff55011b31386ddd84L625) removed `if size is not None`? No idea if that was done because changes for 3.2 made it so that it cannot be `None` anymore or why that was done.

da-woods:
Fixes #7451

It'd be nice if this worked in non-extern arrays too but that's a harder problem and not a regression so I haven't attempted to fix it.

An alternative fix would be to catch TypeError as well as ValueError

### DIFF
--- a/.github/cibuildwheel-requirement.txt
+++ b/.github/cibuildwheel-requirement.txt
@@ -1,2 +1,2 @@
 # This single dependency is in a pip dependency file so depandabot can handle it
-cibuildwheel==3.3.0
+cibuildwheel==3.3.1
--- a/.github/workflows/benchmarks-weekly.yml
+++ b/.github/workflows/benchmarks-weekly.yml
@@ -59,7 +59,7 @@ jobs:
           max-size: 250M
 
       - name: Setup python
-        uses: actions/setup-python@v6.1.0
+        uses: actions/setup-python@v6.2.0
         with:
           # List special Pythons first to keep the 'normal' one as (last) "python3.xy".
           python-version: |
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -61,7 +61,7 @@ jobs:
           max-size: 250M
 
       - name: Setup python
-        uses: actions/setup-python@v6.1.0
+        uses: actions/setup-python@v6.2.0
         with:
           # List special Pythons first to keep the 'normal' one as (last) "python3.xy".
           python-version: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -171,6 +171,12 @@ jobs:
             backend: "c,cpp"
             env: { LIMITED_API: "--limited-api" }
             extra_hash: "-limited_api"
+          # Cython Shared Module
+          - os: ubuntu-22.04
+            python-version: "3.12"
+            backend: "c,cpp"
+            env: { SHARED_UTILITY: "--shared-utility" }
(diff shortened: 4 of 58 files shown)


scoder:
3.2.x commit: https://github.com/cython/cython/commit/f9ae4d2554af5146a3b4b7096816e871e135b9a9

PR #7529, opened 2026-02-15
Separate out and simplify the common memoryview indexing cases in "_unellipsify()"

scoder:
Subscripting from Python went through a rather generic index analysis that spent a lot of time doing things that are rarely needed, especially not for the simple indexing cases. Specialising the different cases makes it 25-40% faster.

### DIFF
--- a/Cython/Utility/MemoryView.pyx
+++ b/Cython/Utility/MemoryView.pyx
@@ -400,7 +400,7 @@ cdef class memoryview:
             else:
                 PyThread_free_lock(self.lock)
 
-    cdef char *get_item_pointer(memoryview self, object index) except NULL:
+    cdef char *get_item_pointer(memoryview self, index: tuple) except NULL:
         cdef Py_ssize_t dim
         cdef char *itemp = <char *> self.view.buf
 
@@ -426,7 +426,7 @@ cdef class memoryview:
         if have_slices:
             return memview_slice(self, indices)
         else:
-            itemp = self.get_item_pointer(indices)
+            itemp = self.get_item_pointer(<tuple> indices)
             return self.convert_item_to_object(itemp)
 
     def __setitem__(memoryview self, object index, object value):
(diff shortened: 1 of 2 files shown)


Cython/Utility/MemoryView.pyx
```diff
@@ -669,34 +675,91 @@ cdef memoryview_cwrapper(object o, int flags, bint dtype_is_object, const __Pyx_
...
+        elif isinstance(item, slice):
+            have_slices = True
+        elif not PyIndex_Check(item):
+            _err_invalid_index(item)
         idx += 1
 
-    nslices = ndim - idx
-    return have_slices or nslices, tuple(result)
+    if first_ellipsis_index >= 0:
+        result = [slice(None)] * ndim
```
matusvalo, line 724: If I understand it correctly, this is pretty frequently called function (callgrind gives me around around 13M calls). Using something more efficient than list should not improve visibly performance?

matusvalo: Of course changing the `result` to some kind of array will make #7530 obsolete but I will live with it ;-)

scoder: > something more efficient than list

It's a fixed size list storing object references, that'd be hard to beat.

scoder: https://github.com/cython/cython/pull/7530 is a good idea, though. I hope you don't mind that I integrated it directly in https://github.com/cython/cython/pull/7529/commits/a8db84a6b91cbd8554f48aa27db68da8414e0a79 (rather than dealing with the merge conflicts).

matusvalo: > I hope you don't mind that I integrated it directly in a8db84a

Not at all.

scoder:
Merging this since it's such an enormous speedup. Several benchmarks show >= 40%.

PR #7530, opened 2026-02-15
Memoryview.pyx: Annotate argument as tuple

matusvalo:
PR annotates arguments as tuple to optimise slightly `enumerate()`:

https://github.com/cython/cython/blob/a3367d58137541111b83cc915d36a9d9d7d351cf/Cython/Utility/MemoryView.pyx#L401


scoder:
Thanks. Integrated into https://github.com/cython/cython/pull/7529

PR #7539, opened 2026-02-22
Fix cache ignoring `compiler_directives` when fingerprinting compiler options

### LINKED ISSUE #7532: [BUG] Cache fingerprinting ignores `compiler_directives`
### Describe the bug

The context is using `Cython+coverage.py`. To reproduce the bug:
 
* Build the package without tracing compiler directives/macros
*  Build with `cythonize.compiler_directives` with `linetrace=True` +  `Extensions.define_macros` with tracing macros
* The cache outputs a code not taking into account `linetrace`, but the macros are defined and passed on to the C/C++ compiler.
* No trace is generated at runtime and the Cython code is not part of the coverage report

I think the reason is because of the wrong assumption (`compiler_directives` does not influence code generation) that is written in a comment in `Compiler/Options.py:get_fingerprint`: 

The fingerprinting function returns a string that will be fed to `hash.update` in `Build/Cache.py:transitive_fingerprint`

``` python
def get_fingerprint(self):
    data = {} 
    for key, value in self.__dict__.items():
       if key in SOMEVALUES:
           ... # many elif
       elif key in ['compiler_directives']:
           # directives passed on to the C compiler do not influence the generated code
           continue
       ... # many other elif

    str_data = convert_to_string(data)
    return data
```

But that is wrong, as `Cython` with `linetrace=True` generates instrumentation code (such as in `Compiler/Code.py:write_trace_line()`).

_No response_

### Expected behaviour

I would expect Cython cache fingerprinting to take into account `compiler_directives` (or at least partially `compiler_directive.linetrace`).

### OS

Linux (official docker image python:3.13-slim)

### Python version

3.13.12

### Cython version

3.2.4

### Additional context

_No response_

v01dXYZ:
This PR suggests a conservative approach to fix the problem.

We could later refine it using an ignore list (we ignore a specific set of directives) or using a list of directives we know modify the code while considering the others do not.

Meant to fix #7532.

### DIFF
--- a/Cython/Compiler/Options.py
+++ b/Cython/Compiler/Options.py
@@ -738,8 +738,9 @@ class CompilationOptions:
                 # hopefully caching has no influence on the compilation result
                 continue
             elif key in ['compiler_directives']:
-                # directives passed on to the C compiler do not influence the generated C code
-                continue
+                # Some compiler directives do not influence the generated C code but others do
+                # (such as linetrace, profiling).
+                data[key] = value
             elif key in ['include_path']:
                 # this path changes which headers are tracked as dependencies,
                 # it has no influence on the generated C code


scoder:
We should add a test for the kind of setup that breaks without this change. You can look at the coverage tests in "tests/run/". The ".srctree" files are just multiple files concatenated into one (with test commands at the top) that the test runner unpacks into a directory. You can copy the `coverage_cmd` test and add two build setups to compare the option usage.

PR #7540, opened 2026-02-23
Implement parsing for "type parameter syntax"

da-woods:
The idea here is to provide enough to let people use it in pure-Python mode (mostly) without actually implementing it in Cython.

```
type X = something
```

is transformed into

```
X = None
```

just so that the name `X` exists. However, this wasn't simple to do for function and class type parameters. It would work for function type parameters, but the class ones have very unusual scoping rules that would need specific implementation.

I also skipped some checks for use of named expressions in them. This was mainly laziness - it only seems worth the effort if we actually tried to implement the runtime behaviour.

I added a disable-able warning that they're ignored - I think we want to warn people, but once people understand it they should be able to turn it off.

### DIFF
--- a/tests/run/basic_type_params.py
+++ b/tests/run/basic_type_params.py
@@ -1,5 +1,5 @@
 # mode: run
-# tag: pure3.12, warnings
+# tag: pure3.13, warnings
 
 # This is just to check that our minimal implementation of typeparams
 # produces usable functions and classes.  Beyond that, they don't do


tests/run/test_type_params.py
```diff
@@ -0,0 +1,1485 @@
+# tag: allow_unknown_names
```
da-woods, line 1: There's a fair number of failures if I don't do this. I'm hoping that most people won't actually access the names in real code.

tests/run/test_type_params.py
```diff
@@ -0,0 +1,1485 @@
...
+        outer()()
+        self.assertEqual(ns["X"], 2)
+
+    def _test_disallowed_expressions(self):
+        # The issue with yield/await in these is to do the scope which is
+        # evaluated too late to be seen by check_syntax_error in Cython
+        # (and actually, we've already dropped the type alias entirely
+        # by the time it is evaluated).
+        # Named expressions should also be disallowed by post-parse
+        # but this doesn't currently seem worth the effort.
```
da-woods: I've skipped this test.

In general I'd tried to do the "check_syntax_error" tests where possible, but all the examples were hard to make work.

da-woods: I've re-enabled these tests

tests/run/basic_type_params.py
```diff
@@ -0,0 +1,140 @@
+# mode: run
```
da-woods, line 1: And added some really simple tests for basic functionality.

So whatever we believe about whether this is worthwhile, it should be a little better tested at least

PR #7543, opened 2026-02-25
Refactor of tests of the builtin types

matusvalo:
Based on https://github.com/cython/cython/pull/7288#pullrequestreview-3701047472


Cython/Compiler/Optimize.py
```diff
@@ -225,16 +225,18 @@ def _optimise_for_loop(self, node, iterable, reversed=False):
             if annotation.is_subscript:
                 annotation = annotation.base  # container base type
 
-        if Builtin.dict_type in (iterable.type, annotation_type):
+        if (iterable.type and iterable.type.is_dict_type) or (annotation_type and annotation_type.is_dict_type):
```
matusvalo: This PR makes condition a little bit messy. I was thinking that if we want use attributes/methods for testing, we should not allow to have type as `None` because in that case we don't know where we forget to test `None` before accessing attribute. We should create `UnknownType` class with all attributes/methods. This should allow us to test safely the types in conditions...

Cython/Compiler/PyrexTypes.py
```diff
@@ -4852,6 +4857,27 @@ class BuiltinTypeConstructorObjectType(BuiltinObjectType, PythonTypeConstructorM
...
+    def is_set_type(self) -> bool:
+        return self.name == 'set'
+    
+    @property
+    def is_tuple_type(self) -> bool:
+        return self.name == 'tuple'
+
+    @property
+    def is_frozenset_type(self) -> bool:
+        return self.name == 'frozenset'
```
scoder: I think we tend to call these a lot more than we create the types, so bare attributes would be more efficient than properties.

Cython/Compiler/PyrexTypes.py
```diff
@@ -266,6 +266,11 @@ class PyrexType(BaseType):
     is_error = 0
     is_buffer = 0
     is_ctuple = 0
+    is_tuple_type = False
+    is_list_type = False
+    is_dict_type = False
+    is_set_type = False
+    is_frozenset_type = False
```
scoder: Looking through the usages, I wonder if additional flags like `is_anyset_type` and `is_tuple_or_list_type` would be helpful. Maybe a `@property` could work for these cases, although it shouldn't be much effort to set them at type creation time.

And while we're at it, why not an `is_builtin_sequence_type` as well …

Cython/Compiler/Optimize.py
scoder: I think the new version is fine. It makes it explicit that we are not expecting the type to be available at this point and took `None` into account.

Cython/Compiler/ExprNodes.py
```diff
@@ -12901,16 +12897,16 @@ def analyse_types(self, env):
...
-            elif operand2.type in builtin_sequence_types:
+            elif Builtin.is_sequence_type(operand2.type):
                 self.operand1 = operand1.coerce_to_index(env)
 
         return self.analyse_operation(env)
 
     @staticmethod
     def is_builtin_seqmul_type(type):
-        return type.is_builtin_type and type in builtin_sequence_types and type is not memoryview_type
+        return type.is_builtin_type and Builtin.is_sequence_type(type) and type is not memoryview_type
```
scoder: Seems worth reversing this condition for efficiency.
```suggestion
        return type.is_builtin_type and type is not memoryview_type and Builtin.is_sequence_type(type)
```

Cython/Compiler/PyrexTypes.py
```diff
@@ -279,6 +304,18 @@ class PyrexType(BaseType):
     default_value = ""
     declaration_value = ""
 
+    @property
+    def is_c_or_cpp_string(self) -> bool:
+        return self.is_string or self.is_cpp_string
+
+    @property
+    def is_bytes_or_str_or_bytearray(self) -> bool:
```
matusvalo, line 316: The names are rubbish. If you have good names please give me an idea.

Cython/Compiler/PyrexTypes.py
```diff
@@ -266,6 +274,23 @@ class PyrexType(BaseType):
     is_error = 0
     is_buffer = 0
     is_ctuple = 0
+
+    is_int_type = False
```
matusvalo: I would rename the attributes to ensure clear division between C/C++ attributes and python ones. E.g. `is_c_...` will be C attribute and `is_..._type` will be python ones... What do you think?

scoder:
>I would rename the attributes to ensure clear division between C/C++ attributes and python ones. E.g. `is_c_...` will be C attribute and `is_..._type` will be python ones... What do you think?

I agree that we should make the distinction clear but would suggest to mark it the other way round, i.e. "is_pyint_type" etc. for the Python types. The available C types are broader and seem more widely used in our code base, and thus, "is_int" seems a good way to test for any C integer type, whereas "is_pyint_type" is clearly testing for the specific Python builtin type.

We could argue whether it should be "is_pyint_type" or "is_pylong_type" to refer to the C-API names, but I feel that the Python names might be clearer while reading through our Python code.

scoder:
>+    @property
>+    def is_bytes_or_str_or_bytearray(self) -> bool:
>
>The names are rubbish. If you have good names please give me an idea.

I don't think we need properties for all combinations. "is_pybyteslike or is_pystr_type" should cover this case.

Although, "pybyteslike" might be unclear as it could allow the Python memoryview type. So, I guess, being explicit about the exact types, like "is_pybytes_or_bytearray_type", seems as clear as it gets.

Or maybe "is_pystring", without the "_type", as we do for "is_int" and "is_string" ? I think that would also be clear, at least for me.

Cython/Compiler/PyrexTypes.py
matusvalo: Fixed.

Cython/Compiler/PyrexTypes.py
matusvalo: I added some helper properties. We can tune it further if something is identified as useful

matusvalo:
OK I think I managed to implement all comments just few notes:
1. I renamed is_* attributes, I prefixed `py` only type names which collided with C counterparts.
2. I added missed types in docstring. There is one missing - `is_fake_reference` it would be great if someone who understand the meaning of it put there description
3. I **hope** that all builtin types are provided and I did not miss anything.

Let me know if there is anything else pending.

scoder:
>OK I think I managed to implement all comments

Cool, thanks!


>1. I renamed is_* attributes, I prefixed `py` only type names which collided with C counterparts.

Hmm. I find it hard to defend that we allow "is_int" and "is_bytes" next to each other, meaning entirely different things. I would like to make the distinction clearer, thus my proposal to prefix the types (as in, all builtin types) with "py".

matusvalo:
>  I would like to make the distinction clearer, thus my proposal to prefix the types (as in, all builtin types) with "py".

I will but later durin weekend.

Cython/Compiler/ExprNodes.py
```diff
@@ -1491,7 +1489,7 @@ def coerce_to_boolean(self, env):
     def coerce_to(self, dst_type, env):
         if dst_type == self.type:
             return self
-        if dst_type is py_object_type and self.type is Builtin.bool_type:
+        if dst_type is py_object_type and self.type.is_pybool:
```
matusvalo: I am 99% sure but I would like to ask to be sure. Can I also replace occurrences of `is py_object_type` with attribute test `.is_pyobject` ? The `is` test check whether the type *is* `PyObjectType` instance, the latter tests whether it is `PyObjectType` instance or instance of it subclass. Can I replace it automatically or I need to check case-by-case? Not 100% sure.

matusvalo:
> > I would like to make the distinction clearer, thus my proposal to prefix the types (as in, all builtin types) with "py".
> 
> I will but later durin weekend.

I fixed the comment. I found out one possible improvement with `py_object_type` tests... I will try to implement it and see whether tests breaks or not...

Cython/Compiler/ExprNodes.py
scoder: These are really different things (unfortunately). Any Python object type has `.is_pyobject == True`, so this is used to distinguish Python types from C types, whereas `type is py_object_type` is commonly used to detect if we know a specific type or have plain `object` as (unknown Python) type here.

matusvalo: I was afraid that it is not so simple. Based on discussion elsewhere should not we have more explicit tests like `.is_python_type` and `.is_plain_pyobject` - or something like that... I would not put it in this PR because I would like to move on with subscription PR which depends on this...

matusvalo: Or if you could lead me - does it make sense to replace all `is py_object_type` check with some attribute access I can do it in this PR if it is just simple (dump) replacement. I would like to ask also what attribute name to avoid back-and-forth discussion about name...

scoder: Since the "is plain/unknown object" test is checking for an exact type, I think `is py_object_type` is as explicit as it gets, and also unlikely to pose a problem in the future.

It's unfortunate that `is_pyobject` and `is_pyint` / `is_pylist` do different things now. I suggest to change it one more time (sorry) to `is_pyint_type` etc. to make it clearer that we're interested in a specific type, whereas `is_pyobject` is _not_ testing for a specific type but a generic type feature (like `is_int` etc.).

(I can see that that was my original proposal in https://github.com/cython/cython/pull/7543#issuecomment-3970739394)

matusvalo: > I suggest to change it one more time (sorry) to is_pyint_type etc. to make it clearer that we're interested in a specific type,

Done. Marking as resolved.

matusvalo:
OK finally done. I hope now it is in mergeable state.

Cython/Compiler/ExprNodes.py
```diff
@@ -4267,7 +4264,7 @@ def infer_type(self, env):
             elif base_type.is_pyunicode_ptr:
                 # sliced Py_UNICODE* strings must coerce to Python
                 return unicode_type
-            elif base_type in (unicode_type, bytes_type, bytearray_type, list_type, tuple_type):
+            elif base_type.is_bytes_or_str_or_bytearray or base_type.is_pylist_type or base_type.is_pytuple_type:
```
scoder: This could be a candidate for an `is_builtin_sequence` (which can be added later, if we feel like it).

Cython/Compiler/ExprNodes.py
```diff
@@ -5580,8 +5578,7 @@ def infer_type(self, env):
             return bytes_type
         elif base_type.is_pyunicode_ptr:
             return unicode_type
-        elif base_type in (bytes_type, bytearray_type, unicode_type,
-                           list_type, tuple_type):
+        elif base_type.is_bytes_or_str_or_bytearray or base_type.is_pylist_type or base_type.is_pytuple_type:
```
scoder: Another candidate for `is_builtin_sequence`.

Cython/Compiler/ExprNodes.py
```diff
@@ -5601,8 +5598,7 @@ def may_be_none(self):
         if base_type:
             if base_type.is_string:
                 return False
-            if base_type in (bytes_type, bytearray_type, unicode_type,
-                             list_type, tuple_type):
+            if base_type.is_bytes_or_str_or_bytearray or base_type.is_pylist_type or base_type.is_pytuple_type:
```
scoder: … and another.

Cython/Compiler/ExprNodes.py
```diff
@@ -9122,7 +9117,7 @@ def coerce_to(self, dst_type, env):
         if self.type.is_ctuple:
             if dst_type.is_ctuple and self.type.size == dst_type.size:
                 return self.coerce_to_ctuple(dst_type, env)
-            elif dst_type is tuple_type or dst_type is py_object_type:
+            elif dst_type.is_pytuple_type or dst_type is py_object_type:
```
scoder: I see this pattern a lot:  `is_pyxxx_type or is py_object_type`. What is usually meant here is "may be pyxxx type", in the sense that if it's really known to be `xxx`, or if we don't know anything about the (Python) type (i.e. it's `object`), then it might be of type `xxx`. If it's another known type, then definitely not.

Could this be worth secondary properties, `may_be_pyxxx_type` ?

Cython/Compiler/ExprNodes.py
```diff
@@ -9656,12 +9651,12 @@ def calculate_constant_result(self):
...
             result.extend(items)
-        if self.type is set_type:
+        if self.type.is_pyset_type:
             result = set(result)
-        elif self.type is tuple_type:
+        elif self.type.is_pytuple_type:
             result = tuple(result)
         else:
-            assert self.type is list_type
+            assert self.type.is_pylist_type
```
scoder: Let's give the assertion a helpful error message if it ever fails.
```suggestion
            assert self.type.is_pylist_type, self.type
```

Cython/Compiler/ExprNodes.py
```diff
@@ -9676,15 +9671,15 @@ def compile_time_value(self, denv):
...
             try:
                 result = set(result)
             except Exception as e:
                 self.compile_time_value_error(e)
-        elif self.type is tuple_type:
+        elif self.type.is_pytuple_type:
             result = tuple(result)
         else:
-            assert self.type is list_type
+            assert self.type.is_pylist_type
```
```suggestion
            assert self.type.is_pylist_type, self.type
```

Cython/Compiler/ExprNodes.py
```diff
@@ -9705,7 +9700,7 @@ def analyse_types(self, env):
             # strip this intermediate node and use the bare collection
             return args[0]
 
-        assert self.type in (set_type, list_type, tuple_type)
+        assert self.type.is_pyset_type or self.type.is_pylist_type or self.type.is_pytuple_type
```
```suggestion
        assert self.type.is_pyset_type or self.type.is_pylist_type or self.type.is_pytuple_type, self.type
```

Cython/Compiler/ExprNodes.py
```diff
@@ -12745,7 +12740,9 @@ def py_operation_function(self, code):
 
         return inplace_function_name if self.inplace else function_name
 
-    specialised_binop_types = (py_object_type, int_type, float_type)
+    @staticmethod
+    def is_specialised_binop_type(typ):
+        return typ is py_object_type or typ.is_int or typ.is_float
```
```suggestion
        return typ is py_object_type or typ.is_int_type or typ.is_float_type
```

Cython/Compiler/ExprNodes.py
```diff
@@ -12824,8 +12821,8 @@ def is_py_operation_types(self, type1, type2):
     def infer_builtin_types_operation(self, type1, type2):
         # b'abc' + 'abc' raises an exception in Py3,
         # so we can safely infer a mix here.
-        string_types = (bytes_type, bytearray_type, unicode_type)
-        if type1 in string_types and type2 in string_types:
+        if type1.is_bytes_or_str_or_bytearray and type2.is_bytes_or_str_or_bytearray:
+            string_types = (bytes_type, bytearray_type, unicode_type)
             return string_types[max(string_types.index(type1),
                                     string_types.index(type2))]
```
scoder: I think the previous version was better because it made it clearer why we were using `.index()` here. With this change, there is no guarantee any more that the types will actually be found in the tuple.

If you want to rewrite this, then we need a different algorithm, basically splitting the possible combinations into separate if-cases:

```suggestion
        if type1.is_pybytes_type:
            if type2.is_pybytes_type or type2.is_pybytearray_type:
                return type2  # bytes or bytearray
        elif type1.is_pybytearray_type:
            if type2.is_pybytes_type or type2.is_pybytearray_type:
                return type1  # bytearray
        elif type1.is_pystr_type and type2.is_pystr_type:
            return type1
```

Cython/Compiler/ExprNodes.py
```diff
@@ -12901,24 +12898,24 @@ def analyse_types(self, env):
                 return self.analyse_sequence_mul(env, operand1, operand2)
             elif operand2.is_sequence_constructor and operand2.mult_factor is None:
                 return self.analyse_sequence_mul(env, operand2, operand1)
-            elif operand1.type in builtin_sequence_types:
+            elif operand1.type and operand1.type.is_sequence:
```
scoder: Looks like we have our `is_builtin_sequence` already. :)

Cython/Compiler/ExprNodes.py
```diff
@@ -14908,7 +14907,7 @@ def __init__(self, arg, env, type=py_object_type):
                 self.type = Builtin.float_type
             self.target_type = self.type
         elif arg.type.is_string or arg.type.is_cpp_string:
-            if (type not in (bytes_type, bytearray_type)
+            if (not (type.is_pybytes_type or type.is_pybytearray_type)
```
scoder: I think an `and` chains better with the `and` that follows.
```suggestion
            if (not type.is_pybytes_type and not type.is_pybytearray_type
```

Cython/Compiler/Optimize.py
```diff
@@ -225,16 +225,18 @@ def _optimise_for_loop(self, node, iterable, reversed=False):
             if annotation.is_subscript:
                 annotation = annotation.base  # container base type
 
-        if Builtin.dict_type in (iterable.type, annotation_type):
+        if (iterable.type and iterable.type.is_pydict_type) or (annotation_type and annotation_type.is_pydict_type):
```
scoder: I think we should always have an `iterable.type` here but may not have an `annotation_type`.

```suggestion
        if iterable.type.is_pydict_type or (annotation_type and annotation_type.is_pydict_type):
```

Cython/Compiler/Optimize.py
```diff
@@ -225,16 +225,18 @@ def _optimise_for_loop(self, node, iterable, reversed=False):
...
                 return node
             return self._transform_dict_iteration(
                 node, dict_obj=iterable, method=None, keys=True, values=False)
 
-        if (Builtin.set_type in (iterable.type, annotation_type) or
-                Builtin.frozenset_type in (iterable.type, annotation_type)):
+        if (
+            (iterable.type and (iterable.type.is_pyset_type or iterable.type.is_pyfrozenset_type)) or
+            (annotation_type and (annotation_type.is_pyset_type or annotation_type.is_pyfrozenset_type))
+        ):
```
scoder: EDIT: Turns out, half of this condition is actually dead code since we never set `annotation_type` to anything but `None`. And since we now support parsing annotations properly into types, it's completely useless code and we can remove the whole annotation related code section above. So this is really just
```python
if iterable.type.is_pyset_type or iterable.type.is_pyfrozenset_type:
```

Cython/Compiler/PyrexTypes.py
```diff
@@ -190,6 +191,17 @@ class PyrexType(BaseType):
...
+    #  is_ctuple             boolean     Is a Cython ctuple type
+    #  is_pytuple_type       boolean     Is a Python tuple type
+    #  is_pylist_type        boolean     Is a Python list type
+    #  is_pydict_type        boolean     Is a Python dict type
+    #  is_pyset_type         boolean     Is a Python set type
+    #  is_pyfrozenset_type   boolean     Is a Python frozenset type
+    #  is_pybytes_type       boolean     Is a Python bytes type
+    #  is_pystr_type         boolean     Is a Python str type
+    #  is_pybytearray_type   boolean     Is a Python bytearray type
+    #  is_memoryview_type    boolean     Is a Python memoryview type
```
scoder: This should have a prefix, too. For consistency, but also to prevent confusion with Cython's memoryviews.
```suggestion
    #  is_pymemoryview_type  boolean     Is a Python memoryview type
```

Cython/Compiler/ExprNodes.py
matusvalo: Hmm here tests should fail... I suppose this is not covered by test suite somehow...

Cython/Compiler/PyrexTypes.py
matusvalo: fixed.

Cython/Compiler/Optimize.py
matusvalo: Hence, even the line (few lines above) 
```python
        if iterable.type.is_pydict_type or (annotation_type and annotation_type.is_pydict_type):
```
can be cleaned similarly too together with setting `annotation_type = None` I suppose

matusvalo: CI seems to be passing hence should be OK.

Cython/Compiler/ExprNodes.py
matusvalo: Hmm there is up to ~10 occurrences I was able to grep from the whole Cython codebase. So not sure if it is worth it. If you still want to have it I can implement it easily.

Cython/Compiler/ExprNodes.py
matusvalo: `is_sequence` include also `is_memoryview_type`:

```python
    @property
    def is_sequence(self) -> bool:
        return (
            self.is_pybytes_type or self.is_pystr_type or self.is_pybytearray_type or
            self.is_pymemoryview_type or self.is_pylist_type or self.is_pytuple_type
        )
```

The tests above are not having `self.is_pymemoryview_type` included. Is it OK?

matusvalo:
OK thank you for another round of review. I left only two open points:
* `may_be_pyxxx_type` - whether it is worth introducing new properties for ~6-7 occurrences
* `is_sequence` - not sure it is OK to include existing property for expressions not including `is_pymemoryview_type`

When I will have answers I can wrap up this PR.

Cython/Compiler/ExprNodes.py
da-woods: Both of the tests above would remain correct with `is_pymemoryview_type` added - they're saying "slicing returns the same type" which is also true for `memoryview`.

Edit: "both" is wrong - there's a third test above that I missed. But it looks like it'd also be correct if it included memoryview

Cython/Compiler/ExprNodes.py
scoder: I think `may_be_pyxxx_type` expresses the intent more clearly than `is_pyxxx_type or is py_object_type`. For me, that alone makes it worse adding the properties.

Cython/Compiler/ExprNodes.py
matusvalo: OK thank you @da-woods for your help. I implemented the changes then. I also renamed `is_sequence` to `is_builtin_sequence`.

Cython/Compiler/ExprNodes.py

matusvalo:
> OK thank you for another round of review. I left only two open points:
> 
>     * `may_be_pyxxx_type` - whether it is worth introducing new properties for ~6-7 occurrences
> 
>     * `is_sequence` - not sure it is OK to include existing property for expressions not including `is_pymemoryview_type`
> 
> 
> When I will have answers I can wrap up this PR.

OK both points should be fixed now. Please have a look. Big thanks for diligent review.

Cython/Compiler/PyrexTypes.py
```diff
@@ -1445,6 +1513,22 @@ class BuiltinObjectType(PyObjectType):
...
+        'list': 'is_pylist_type',
+        'dict': 'is_pydict_type',
+        'set': 'is_pyset_type',
+        'tuple': 'is_pytuple_type',
+        'frozenset': 'is_pyfrozenset_type',
+        'bytes': 'is_pybytes_type',
+        'str': 'is_pystr_type',
+        'bytearray': 'is_pybytearray_type',
+        'memoryview': 'is_pymemoryview_type',
+    }
```
scoder, line 1530: If you make the value side of this dict a list, you can include the other attributes like `is_bytes_or_str` and `is_builtin_sequence` directly, instead of implementing them as runtime properties. The attributes are certainly evaluated way more often than the types are created, so this would save time.

matusvalo: I will manage it in separate PR.

matusvalo: PR created: #7624

PR #7544, opened 2026-02-26
fix: replace 48 bare except clauses with except Exception

haosenwang1018:
Replace bare except clauses with except Exception.

### DIFF
--- a/Cython/Build/Dependencies.py
+++ b/Cython/Build/Dependencies.py
@@ -12,7 +12,7 @@ from collections.abc import Iterable
 
 try:
     import pythran
-except:
+except Exception:
     pythran = None
 
 from .. import Utils
@@ -1187,7 +1187,7 @@ if os.environ.get('XML_RESULTS'):
             try:
                 try:
                     func(*args)
-                except:
+                except Exception:
                     success = False
             finally:
                 t = time.time() - t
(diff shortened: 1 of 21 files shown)


scoder:
Thanks. I gave it a quick look.

Some of these seems too broad and would benefit from an even narrower exception type (like `ImportError` or `AttributeError`).

Some are in test files that we copied from CPython and shouldn't modify beyond the required adaptation. Their file names usually start with `test_…`.

I think there might also be a few cases (probably only in tests) where the bare except clause aims to suppress exceptions to prevent propagating them at the C level.

Either way, most of these need more thought than a quick search&replace. It might help to exclude the test files in a first round because they bare the highest risk of breakage on this kind of change.

da-woods:
This user has opened >400 similar PRs across many repositories in the last few days. They also look to be resubmitting the same PR multiple times in some places. 

I think they're effectively doing a denial of service attack on open source. We should close this PR and block them.

They're unlikely to ever be back to make a useful contribution.

scoder:
I was guessing as much, thus my limited effort in my reply.

It's still good to see the list of changes, though. A few of them are really worth making.

haosenwang1018:
Thanks for review context. Closing to rework into smaller scoped changes.

PR #7545, opened 2026-02-26
Implement support for PEP-814 frozendict

scoder:
Substitutes it with plain `dict` in Py<3.15.


Cython/Compiler/Builtin.py
```diff
@@ -432,6 +435,24 @@ def _generate_divmod_function(scope, argument_types):
...
+    ("frozendict", "&__Pyx_PyFrozenDict_Type", [
+                                    BuiltinMethod("__contains__",  "TO",   "b", "PyDict_Contains"),
+                                    BuiltinMethod("has_key",       "TO",   "b", "PyDict_Contains"),
+                                    BuiltinMethod("items",  "T",   "O", "__Pyx_PyDict_Items",
+                                                  utility_code=UtilityCode.load("py_dict_items", "Builtins.c")),
+                                    BuiltinMethod("keys",   "T",   "O", "__Pyx_PyDict_Keys",
+                                                  utility_code=UtilityCode.load("py_dict_keys", "Builtins.c")),
+                                    BuiltinMethod("values", "T",   "O", "__Pyx_PyDict_Values",
+                                                  utility_code=UtilityCode.load("py_dict_values", "Builtins.c")),
+                                    BuiltinMethod("viewitems",  "T",   "O", "__Pyx_PyDict_ViewItems",
```
da-woods: I'm not sure we need to provide the legacy Python2-era interface for a newly added type.

Cython/Utility/Builtins.c
```diff
@@ -808,3 +808,41 @@ static {{out_type}} __Pyx_PyMemoryView_Get_{{name}}(PyObject *obj) {
...
+// PyFrozenDict is not currently in the Limited API.
+#define __Pyx_PyFrozenDict_Type  PyFrozenDict_Type
+#define __Pyx_PyFrozenDict_New(it)  PyFrozenDict_New(it)
+#define __Pyx_PyFrozenDict_NewEmpty()  PyFrozenDict_New(NULL)
+#define __Pyx_PyFrozenDict_Check(obj)  PyFrozenDict_Check(obj)
+#define __Pyx_PyFrozenDict_CheckExact(obj)  PyFrozenDict_CheckExact(obj)
+#define __Pyx_PyAnyDict_Check(obj)  PyAnyDict_Check(obj)
+#define __Pyx_PyAnyDict_CheckExact(obj)  PyAnyDict_CheckExact(obj)
+#else
+// It is questionable to replace PyFrozenDict with PyDict in the Limited API of Py3.15,
```
da-woods: I think for the Limited API the right thing to do is to try to get it from the `types` module or `PyEval_GetBuiltins` (as in `CachedMethodType` or `Coro_CheckExact`) and use that if available.

Since even for Limited API<3.15 there's some chance that the type is actually available at runtime.

Cython/Compiler/Builtin.py
```diff
@@ -432,6 +435,18 @@ def _generate_divmod_function(scope, argument_types):
...
+    ("frozendict", "&__Pyx_PyFrozenDict_Type", [
+                                    BuiltinMethod("__contains__",  "TO",   "b", "PyDict_Contains"),
+                                    BuiltinMethod("has_key",       "TO",   "b", "PyDict_Contains"),
+                                    BuiltinMethod("items",  "T",   "O", "__Pyx_PyFrozenDict_Items",
+                                                  utility_code=UtilityCode.load("py_frozendict_items", "Builtins.c")),
+                                    BuiltinMethod("keys",   "T",   "O", "__Pyx_PyFrozenDict_Keys",
+                                                  utility_code=UtilityCode.load("py_frozendict_keys", "Builtins.c")),
+                                    BuiltinMethod("values", "T",   "O", "__Pyx_PyFrozenDict_Values",
+                                                  utility_code=UtilityCode.load("py_frozendict_values", "Builtins.c")),
+                                    BuiltinMethod("copy",   "T",   "T", "PyDict_Copy"),  # Returns frozendict in Py3.15.
```
da-woods: See https://github.com/python/cpython/pull/145542

It's probably worth holding off on this PR until the Python interface has settled a bit

da-woods:
I think the decisions on what Python will do here have probably settled now.

I still think the Limited API code needs to depend on the runtime version.

Cython/Utility/Builtins.c
```diff
@@ -809,3 +839,116 @@ static {{out_type}} __Pyx_PyMemoryView_Get_{{name}}(PyObject *obj) {
...
+        if (likely(type_name)) {
+            PyObject *frozendict_type = PyObject_GetItem(builtins, type_name);
+            Py_DECREF(type_name);
+            if (!frozendict_type && PyErr_ExceptionMatches(PyExc_KeyError)) {
+                PyErr_Clear();
+                type_name = PyUnicode_FromStringAndSize("dict", sizeof("dict")-1);
+                if (likely(type_name)) {
+                    frozendict_type = PyObject_GetItem(builtins, type_name);
+                    Py_DECREF(type_name);
+                }
```
da-woods:
```suggestion
                 Py_INCREF((PyObject*)&PyDict_Type);
                frozendict_type = &PyDict_Type;
```

This bit could be simplified a bit.

Cython/Utility/Builtins.c
```diff
@@ -809,3 +839,116 @@ static {{out_type}} __Pyx_PyMemoryView_Get_{{name}}(PyObject *obj) {
...
+    } else if (PyDict_Check(it)) {
+        return PyDict_Copy(it);
+    } else {
+        PyObject *dict = PyDict_New();
+        if (!dict) return NULL;
+        // PyDict_Merge() and friends do not handle arbitrary iterables. '|' does.
+        PyObject *result = PyNumber_InPlaceOr(dict, it);
+        Py_DECREF(dict);
+        return result;
+    }
```
da-woods:
```suggestion
    return PyObject_CallFunctionObjArgs(CGLOBAL(__Pyx_PyFrozenDictType), it, NULL);
```

Arguably you could do the more optimized fallback for the case where `dict` is substituting. I'm not convinced it's worth it here, but it's easy to do.

da-woods: Admittedly this is only currently used in ImportExport where it doesn't really matter, and the code generation in the `py315_really_has_frozendict` test just goes through `__Pyx_PyObject_FastCall((PyObject*)(__Pyx_PyFrozenDict_TypePtr), ...`

da-woods: Thinking about it more, the one usage of this function so far is in a context where it's probably better to return whichever type is quickest to construct rather than preferring `frozendict`.

If that's the case I wonder if the naming should be something like `__Pyx_PyFrozenDictOrDict_New` to make it clearer what it does (and then keep your current version which prefers `dict` in the limited API).

scoder: > the one usage of this function so far is in a context where it's probably better to return whichever type is quickest to construct

Good point. I'll change the usage to get a frozendict only when it's easily available.

scoder: > Admittedly this is only currently used in ImportExport

I guess I was anticipating a backport for the (yet to be added) C-API declarations for `Include/cpython/frozendict.pxd`. But we can probably start simple (as you suggest above) and then see if that's enough.

scoder:
I added the C-API declarations using the backport, so that users can more easily write C-API code that also compiles on older Python versions (and even does something reasonable there).

da-woods:
The utility code doesn't get auto-imported if you just use the C API functions (although I guess most real code would probably do something that uses the right utility code). e.g.

```
cimport cpython.frozendict

def f(o):
    return cpython.frozendict.PyAnyDict_Check(o)
```

fails for me with

```
error: implicit declaration of function ‘__Pyx_PyAnyDict_Check’; did you mean ‘__Pyx_PyDict_Items’? [-Wimplicit-function-declaration]
 2511 |   __pyx_t_1 = __Pyx_PyBool_FromLong(__Pyx_PyAnyDict_Check(__pyx_v_o)); if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 4, __pyx_L1_error)
      |                                     ^~~~~~~~~~~~~~~~~~~~~
      |                                     __Pyx_PyDict_Items
```.

My usual approach for similar things is to add a dummy function to the pxd file that triggers the right utility code. e.g.

```
cdef inline int __dummy_function(o):
    return isinstance(o, frozendict)
```

File: Cython/Utility/Builtins.c [hunk 6]
```diff
@@ -809,3 +839,103 @@ static {{out_type}} __Pyx_PyMemoryView_Get_{{name}}(PyObject *obj) {
...
+                frozendict_type = (PyObject*) &PyDict_Type;
+                Py_INCREF(frozendict_type);
+            }
+            CGLOBAL(__Pyx_PyFrozenDictType) = frozendict_type;
+        }
+    }
+} // error handling follows
+#endif
+
+/////////////// PyFrozenDict_LAPI.cleanup ////////////////
```
da-woods:
```suggestion
/////////////// PyFrozenDict.cleanup ////////////////
```

? Don't think this'll get used otherwise

da-woods (commented):
Other last comment.

PR #7546, opened 2026-02-26
Control string compression via compiler directive

dalcinl:
The new string compression is great for modules with large number or long strings, but not all modules fall in that category. Moreover, decompression depends on stdlib modules that may not be available if Python is specially configured/built. This forces users to define `CYTHON_COMPRESS_STRING=0`, and how to do that depends on the details of the build system (setuptools, cmake, meson, etc). It would be great to allow module authors to turn off the string compression feature via compiler directive.

### DIFF
--- a/Cython/Compiler/Code.py
+++ b/Cython/Compiler/Code.py
@@ -2156,6 +2156,10 @@ class GlobalState:
 
         has_if = False
         for algo_number, algo_name, compressed_bytes in reversed(compressions):
+            if not has_if:
+                w.putln('#ifndef CYTHON_COMPRESS_STRINGS')
+                w.putln('#define CYTHON_COMPRESS_STRINGS 1')
+                w.putln('#endif')
             if algo_name == 'zlib':
                 # Use zlib as fallback if the selected compression module is not available.
                 assert algo_number == 1, f"Compression algorithm no. 1 must be 'zlib' to be used as fallback."
--- a/Cython/Utility/ModuleSetupCode.c
+++ b/Cython/Utility/ModuleSetupCode.c
@@ -433,10 +433,6 @@
   #endif
 #endif
 
-#ifndef CYTHON_COMPRESS_STRINGS
-  #define CYTHON_COMPRESS_STRINGS 1
-#endif
-
 #ifndef CYTHON_FAST_PYCCALL
 #define CYTHON_FAST_PYCCALL  CYTHON_FAST_PYCALL
 #endif


dalcinl:
There is still a minor nit I'm not sure about. If string compression is turned off, Cython would still emit 
```
#ifndef CYTHON_COMPRESS_STRINGS
  #define CYTHON_COMPRESS_STRINGS 1
#endif
```
Would it make sense to move that definition from `Utility/ModuleSetupCode.c` to `DecompressString.proto`?

```patch
index 89efbd8ae..f1dda61b8 100644
@@ -433,10 +433,6 @@
   #endif
 #endif
 
-#ifndef CYTHON_COMPRESS_STRINGS
-  #define CYTHON_COMPRESS_STRINGS 1
-#endif
-
 #ifndef CYTHON_FAST_PYCCALL
 #define CYTHON_FAST_PYCCALL  CYTHON_FAST_PYCALL
 #endif
diff --git a/Cython/Utility/StringTools.c b/Cython/Utility/StringTools.c
index 3e4be44d0..b13f1e354 100644
--- a/Cython/Utility/StringTools.c
+++ b/Cython/Utility/StringTools.c
@@ -53,6 +53,10 @@ static CYTHON_INLINE size_t __Pyx_Py_UNICODE_strlen(const Py_UNICODE *u)
 
 //////////////////// DecompressString.proto ////////////////////
 
+#ifndef CYTHON_COMPRESS_STRINGS
+  #define CYTHON_COMPRESS_STRINGS 1
+#endif
+
 static PyObject *__Pyx_DecompressString(const char *s, Py_ssize_t length, int algo); /*proto*/
 
 //////////////////// DecompressString ////////////////////
```

scoder:
The directive needs to be added to `Shadow.py` as well, there's a test for that.

> decompression depends on stdlib modules that may not be available if Python is specially configured/built

The choice of compression is very conservative, exactly for that reason. You really have to set `CYTHON_COMPRESS_STRINGS` yourself if you want anything but `zlib` (or "no compression", which is chosen automatically if `zlib` isn't effective). Are you suggesting that even `zlib` might be missing on some platforms?

> Would it make sense to move that definition from `Utility/ModuleSetupCode.c` to `DecompressString.proto`?

Hmm, if that works? There might be ordering issues. But it would also allow users to set it from inside of the module code, e.g. through an include.

However, it's not used in the utility code but only in line generated code (in `Code.py`). We could move it there, though, in order to get the late-most evaluation.

dalcinl:
On a second though, should this be a compiler directive or rather a global option?

> Are you suggesting that even `zlib` might be missing on some platforms?

I believe Python can be configured without zlib.

> > Would it make sense to move that definition from `Utility/ModuleSetupCode.c` to `DecompressString.proto`?
> 
> Hmm, if that works? There might be ordering issues. But it would also allow users to set it from inside of the module code, e.g. through an include.

There are no ordering issues, ~but your second observation makes total sense. Let's keep it as it is~. Indeed, what I suggested would allow to set the value via include. I'll go for it,

> However, it's not used in the utility code but only in line generated code (in `Code.py`). We could move it there, though, in order to get the late-most evaluation.

I think I understood what you meant. I'll take a look.

dalcinl:
> The directive needs to be added to `Shadow.py` as well, there's a test for that.

No, no, I think that's not the right. I missed to specify that the directive is only module scope.

dalcinl:
@scoder The gcc-13 package from PPA in ubuntu-22.04 has broken dependency. I propose to upgrader the runner OS to ubuntu-24.04, keeping GCC at version 13. Are you OK with me pushing a hotfix here?

da-woods:
> The gcc-13 package from PPA in ubuntu-22.04 has broken dependency.

I think that's probably been fixed now so no need to

scoder:
Hmm, I'm wondering now, if `zlib` not being available is really a threat, then isn't the right solution to change the default to "no compression" ?

I chose `zlib` as a default because it's really ubiquitous, small, free of patents, and all that. And it's a dependency that is used so widely that it seems more effort to safely exclude it from CPython than to just keep it.

But I can imagine scenarios where e.g. a high risk of compression bombs is worth that effort. Something that is not there at all cannot interfere with what is there.

I would expect the very large majority of Python installations out there to include zlib. So many, that package authors will not usually run into a system that doesn't have it. Thus, they're likely to unknowingly distribute packages that depend on `zlib`, and that will fail to import on systems that don't have it.

However uncommon that failure case is, it's something that we impose on extension authors to resolve, and many of them can't because they won't even be aware of it. That feels the wrong way round.

We can change the default and instead issue a performance warning when the estimated benefit of compression surpasses a certain threshold, so that users can decide to enable it. What do you think?

dalcinl:
> I would expect the very large majority of Python installations out there to include zlib. 

That's my expectation, too. However, I'm also concern with special builds of Python that may not have it. Additionally, there are extension modules where the size gain on compressing strings is rather marginal (e.g. 2% in may case) so I would rather disable string compression at all, definitely not worth the risk, even if remote.

> Thus, they're likely to unknowingly distribute packages that depend on `zlib`, and that will fail to import on systems that don't have it.

Indeed.

BTW, this (withdrawn) PEP is relevant https://peps.python.org/pep-0775/
https://discuss.python.org/t/pep-775-make-zlib-required-to-build-cpython/82672

> However uncommon that failure case is, it's something that we impose on extension authors to resolve, and many of them can't because they won't even be aware of it. That feels the wrong way round.

Look, I noticed this new feature by inspecting generated C code. The Cython changelog has no mention to the introduction of this thing. I was definitely not aware of it.

> We can change the default and instead issue a performance warning when the estimated benefit of compression surpasses a certain threshold, so that users can decide to enable it. What do you think?

IMHO, I think any optimization that changes how things are usually done should be opt-in. I believe string compression definitely falls in that category (this is not something standard in CPython extensions). So yes, maybe I would rather prefer that Cython emits a "performance warning" telling users about the benefits of string compression, instructing users to explicitly set the directive `# cython: compress_strings=True/False`. That being said, I'm not sure how we should define the threshold. I guess it should not only be about relative compresses/uncompressed size, but also some absolute uncompressed string size threshold (50 KiB? 100 KiB? no idea really).

scoder:
If we change the default, then I'm not sure if we should keep the current fallback on `zlib`. The way I designed it was that users can chose e.g. `zstd` compression but if that's not available, it will use `zlib` instead, assuming that that's always available. If we change the default to "no compression", then does an "if you want compression, you get at least `zlib`" fallback still makes sense?

At the time, I decided against making an unfulfillable wish an error since the package maintainers might not be the ones building the C code. Also, different modules might end up with different compression options due to varying efficiency, but configuring each module separately seemed cumbersome. So, you can globally say at C compile time "I want `zstd` compression", and if `zlib` happens to be more effective for one of your modules, you'd get `zlib` compression for it but `zstd` for the modules for which it performs better.

It looks like we need to open up the whole user interface of this and think it through once more.

> `# cython: compress_strings=True/False`

BTW, should we also allow specifying a list of compression algorithms here? Like

`# cython: compress_strings=zlib,zstd`

Seems reasonable, but might be excessive and could still be added later if requested.

dalcinl:
> It looks like we need to open up the whole user interface of this and think it through once more.

I agree. However, I would like to request to merge this PR before such new interface is decided. Note that I'm not changing current defaults, just providing a compiler directive to turn the feature on/off. The current default is still `True`, unless you agree right now to change it to `False` for the sake of maximum safety. Again, although I have a preference for this to be off by default, I'm fine with not changing the default as long as the new compiler directive is made available.

BTW, should this new directive be made available in next release  `3.2.x`? It is technically a new feature, but I would consider the lack of sort of a bug. If yes, then how do we handle this? Update/close this PR and target the release branch?

dalcinl:
@scoder Another nit: Should the new compiler directive be namespaced as `optimize.compress_strings`?

scoder:
> should this new directive be made available in next release 3.2.x?

I actually liked the fact that the whole compression thing was configurable at C compile time with a macro. If you define a macro that is not used in your C code, that's fine. If you set a directive and your Cython version doesn't support it, you get an error.

Regardless of whether we change the directive default in 3.2.x or not, users will end up trying to set it either way and need to require the latest 3.2.x release for that. I'm not sure that's an improvement.

You're right that the feature wasn't really advertised in any way, though. Most users won't know that they are making use of it. Changing the default in a point release will lead to unexpected jumps in package sizes and will trigger users to want to change the setting. Or to report bugs first. :)

I think we can add the directive to 3.2.5, keeping the behaviour as it is now, and just let users deal with their needs. We shouldn't reconsider the default setting before 3.3.

scoder:
BTW, I found this:
https://github.com/ariya/FastLZ

It's a very small C text compression library for LZ77, MIT licensed. We could use it as a new default compression in 3.3 and just embed the decompressor in the module. Obviously, then we'd want to increase the "is it worth it" threshold by the expected size of the decompression code, but it would probably still be helpful for modules with a medium to high usage of strings.

dalcinl:
> If you set a directive and your Cython version doesn't support it, you get an error.

Yes, if set in the command line, but not if set as a `# cython: directive=value` comment.

> I think we can add the directive to 3.2.5, keeping the behaviour as it is now, and just let users deal with their needs. We shouldn't reconsider the default setting before 3.3.

That's what I meant, sorry for not clarifying. Any change in default value should definitely be in 3.3.

> https://github.com/ariya/FastLZ

Great finding. The decompresor seems trivial to modify as Cython utility code. I'm not sure about the compressor that  Cython would need.


BTW, you  OK with using a namespaced `optimize.compress_strings` directive name? https://github.com/cython/cython/pull/7546#issuecomment-3990094637

da-woods:
My views (none of which I feel particularly strongly about):

* Embedding our own decompression algorithm in the code seems like it'd be more trouble than it's worth. It's probably fairly hard to reliably estimate the compiled-size to make it worth it, and it's probably only a fairly large module that would hit that size.
* I don't have a huge opinion on "directive or macro". Macros are easy enough to control in most build systems that they should be a problem, but as @dalcinl says unknown comment-style macros don't cause a compile error.
* I'm not practically not too worried about zlib being unavailable. And anyone using such constrained systems is probably building packages themselves rather than using distributed packages so they'd be in a position to fix it themselves.
* I think that the error on compression failure https://github.com/cython/cython/blob/e055fef181084b5fb1630e606e51948b35d19303/Cython/Utility/StringTools.c#L129-L133 could be a bit more explicit about suggesting that users rebuild with compression off.

dalcinl:
> * And anyone using such constrained systems is probably building packages themselves rather than using distributed packages so they'd be in a position to fix it themselves.

I don't see the connection between a constrained system (one with striped-down dependencies, perhaps running in a container) and the use of pre-built wheels distributed on PyPI.

scoder:
> Embedding our own decompression algorithm in the code seems like it'd be more trouble than it's worth. It's probably fairly hard to reliably estimate the compiled-size to make it worth it, and it's probably only a fairly large module that would hit that size.

The two decoders are literally just this:
https://github.com/ariya/FastLZ/blob/b1342dabcf5257ab303743c9332fe75e9147a011/fastlz.c#L283-L320
https://github.com/ariya/FastLZ/blob/b1342dabcf5257ab303743c9332fe75e9147a011/fastlz.c#L429-L480

We can decide at code generation time which provides sufficient savings. Compression can often save a few hundred bytes for even simple modules, so net wins don't seem far off. If the gain isn't large enough to be a sure win across different systems when adding the decompression code, we simply don't compress by default. We don't need to care about 50 bytes of code up or down.

> you OK with using a namespaced `optimize.compress_strings` directive name? [#7546 (comment)](https://github.com/cython/cython/pull/7546#issuecomment-3990094637)

The `optimize.*` options differ a bit in their current meanings, but they usually influence the generated C code in one way or another, including the code size. So the prefix seems to fit this purpose as well.

scoder:
https://github.com/cython/cython/pull/7577 implements LZSS as proposed default string compression, which has a tiny decompressor footprint while providing a decent level of compression.

dalcinl:
#7577 Looks great. What's your plan for this PR? Close now or merge later?

scoder:
> plan for this PR? Close now or merge later?

Do we still need it? The new default compression won't go into 3.2.x, but a new directive would also only go into 3.2.5+ and would appear mostly useless in 3.3 if the LZSS PR goes in. Can you live with setting the macro in 3.2.x ?

dalcinl:
> Do we still need it?

Maybe not the directive, but there are other two commits in this PR that I think are good. Either I wait until you merge #7577 and I update this PR, or maybe faster, you could try to cherry-pick my commits into #7577?  Of course, you'll have to update e7c447b6f1103742a3a52373952e2570161a322d for the default value change (1 -> 90).

scoder:
I think https://github.com/cython/cython/pull/7546/changes/0506061f795ccf5dfe178e580f27696fb12a1d44 is included, but I'll see that I integrate https://github.com/cython/cython/pull/7546/changes/e7c447b6f1103742a3a52373952e2570161a322d

dalcinl:
> I think [0506061](https://github.com/cython/cython/commit/0506061f795ccf5dfe178e580f27696fb12a1d44) is included

Are you sure? You are unconditionally loading utility code, no matter where compression is actually used. My commit loads utility code only if truly needed. Anyway, this is probably mostly inconsequential, a bunch of lines of (unlikely) unused C code will not harm. I can take a second look after your PR is merged, and implement any marginal improvements if worthwhile.

scoder:
> You are unconditionally loading utility code

Loading, not inserting. But you're right, why load the code if we may not need it? It doesn't really cost anything to load the cached version a couple of times in the loop.

scoder:
I merged the LZSS compression PR. Can we close this one?

PR #7547, opened 2026-02-28
Try to fix gcc tests

da-woods:
For those tests, bump the ubuntu image 24.04 (so we have access to more recent GCC versions). Also bump the GCC version to GCC14 (since I think the idea was to use something up-to-date rather than gcc13 specifically).

I've also updated the extra_hash to be a bit more descriptive.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,7 @@ jobs:
           - os: ubuntu-24.04
             python-version: "3.13"
             backend: cpp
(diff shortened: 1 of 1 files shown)


da-woods:
It looks like:
* the initial issue that this is working around has resolved itself spontaneously
* There's still Pythran issues, and also linking issues with libblas (which I'm sure are solveable)

So I'll close this because I don't think it's needed right now

PR #7548, opened 2026-02-28
Speed up isolated_limited test

da-woods:
We were unsetting the CFLAGS to get control of Py_LIMITED_API so it wasn't overridden by the global flags. However this also meant that we were losing "-Og/-O0" and so it was compiling slower than needed.

This adds a way of unsetting only some of the CFLAGS so we can control the Cython-specific stuff without affecting our general CFLAGS.

On my laptop, this runs in just under half the time.


da-woods:
It looks like this knocks off ~1/3 of the time on the CI. So less than I hoped, but still worthwhile.

Note that the time is quite dependent on which Python version it's running on (just because the isolated limited API tests run a much wide range of tests on recent Python)

PR #7554, opened 2026-03-06
Tests: Read current builtins from a fresh Python process in TestBuiltin.py

scoder:
We compare the expected (known) set of builtins against what Python provides at runtime, but that risks depending on the installed third-party packages, specifically IPython and potentially others.

Thus, we start a fresh Python process and read the builtins from that, to avoid any interference with the imported test dependencies of the test runner.


scoder:
3.2.x commit: https://github.com/cython/cython/commit/8cdea4a9e5c0c75342b3bf180f4a70cce92ba534

PR #7555, opened 2026-03-07
Cancel sanitizers/benchmarks on new push

da-woods:
It doesn't look like the sanitizers and benchmarks jobs get cancelled when you update a PR. I don't think that's intended. Probably relates to https://github.com/cython/cython/commit/b774671ac075dae5818937ae0472f723c099e155

### DIFF
--- a/Cython/Build/Cache.py
+++ b/Cython/Build/Cache.py
@@ -152,7 +152,7 @@ class Cache:
             dirname = os.path.dirname(c_file)
             with zipfile.ZipFile(cached) as z:
                 for artifact in z.namelist():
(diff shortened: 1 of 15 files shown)


da-woods:
I can't merge this because it doesn't trigger the main tests to run and so the they're never shown as a success. I think it's probably pretty uncontroversial though.

scoder:
I had considered it at the time and hand-wavingly decided against it, assuming that we might be interested in intermediate results also for PRs. However, it turned out to be more of a nuisance than I had thought, so I'm happy to change it now.

PR #7556, opened 2026-03-08
Add cy_function and fused_function to the shared utility module

da-woods:
Implementation notes:
* The struct definitions are always available (even if the type is generated in the shared utility module). This is because we occasionally access into the struct directly and I see no reason to de-optimize that.
* The `__pyx_[Cython/Fused]Function_init` functions, which are used to initialize/fetch the type object, differ depending on whether it is locally defined or defined in the shared module. This is done fairly crudely by just using different utility code because it didn't seem worth a complicated mechanism.
* I had to define a new block to call the `_init` functions because they needed to be called after the shared module is imported.

Sizes:

Shared module before: 154440 bytes
Shared module with cyfunc: 172256 bytes (+17816 bytes)
Shared module with fusedfunc+cyfunc: 172704 bytes (+448 bytes from cyfunc, +18264 bytes from before)

So adding fusedfunc is a pretty negligable difference (which is good, because it's used relatively infrequently compared to cyfunc)

Module that uses CyFunc drops 17880 bytes
Module that uses CyFunc and fusedfunc drops 18328 bytes

(All sizes are stripped)

----------------------------------------------------

The other shared types are probably less obvious about whether to add, but could be handled in much the same way. My view is that they're infrequently used so you'd probably want to be able to configure whether they're included rather than always doing it. I'd like to keep this PR to just the function types.

### DIFF
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -1,22 +1,22 @@
-//////////////////// CythonFunctionAlways.module_state_decls ////////////////////
+//////////////////// CythonFunctionPerModule.module_state_decls ////////////////////
 
 PyTypeObject *__pyx_CyFunctionType;
 
-//////////////////// CythonFunctionAlways.module_state_traverse ///////////////////
+//////////////////// CythonFunctionPerModule.module_state_traverse ///////////////////
 
 Py_VISIT(traverse_module_state->__pyx_CyFunctionType);
 
-//////////////////// CythonFunctionAlways.module_state_clear ///////////////////
+//////////////////// CythonFunctionPerModule.module_state_clear ///////////////////
 
 Py_CLEAR(clear_module_state->__pyx_CyFunctionType);
 
-//////////////////// CythonFunctionAlways.init //////////////////
+//////////////////// CythonFunctionPerModule.init //////////////////
 //@substitute: naming
 //@init_block: init_after_shared_utility
 
 if (likely(__pyx_CyFunction_init($module_cname) == 0)); else
 
-//////////////////// CythonFunctionAlways.proto //////////////////////////
+//////////////////// CythonFunctionPerModule.proto //////////////////////////
 // This section always gets included whether we're using CythonFunction through
 // shared utility code or not.
 
@@ -128,7 +128,7 @@ static int __pyx_CyFunction_init(PyObject *module);
 #endif
 #endif
 
-//////////////////// CythonFunctionAlways /////////////////////
+//////////////////// CythonFunctionPerModule /////////////////////
 //@requires: ObjectHandling.c::CachedMethodType
 
 #if CYTHON_COMPILING_IN_LIMITED_API
@@ -205,7 +205,7 @@ static CYTHON_INLINE void __Pyx_CyFunction_SetAnnotationsDict(PyObject *func, Py
 }
 
 //////////////////// CythonFunctionFromSharedModule /////////////////
-//@requires: CythonFunctionAlways
+//@requires: CythonFunctionPerModule
 //@requires: CythonFunction
 //@requires: CommonStructures.c::VerifyCachedType
 //@substitute: naming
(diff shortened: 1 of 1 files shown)


Cython/Utility/CythonFunction.c
```diff
@@ -1,21 +1,24 @@
...
+//////////////////// CythonFunctionAlways.init //////////////////
 //@substitute: naming
+//@init_block: init_after_shared_utility
 
 if (likely(__pyx_CyFunction_init($module_cname) == 0)); else
 
-//////////////////// CythonFunctionShared.proto ////////////////////
+//////////////////// CythonFunctionAlways.proto //////////////////////////
+// This section always gets included whether we're using CythonFunction through
+// shared utility code or not.
```
da-woods: I'm not sure this is the best bit of naming ever, but a better name wasn't leaping out at me

matusvalo:
~~I tried to compile scikit-image with PR and without and this PR increases the resulting binary size (sum of all .so files) only few bytes bigger so +1 for me.~~

da-woods:
> I tried to compile scikit-image with PR and without and this PR increases the resulting binary size (sum of all .so files) only few bytes bigger so +1 for me.

I guess that comment was really for https://github.com/cython/cython/pull/7557?

(This PR here is a big disappointment if it increases sizes)

Cython/Utility/CythonFunction.c
scoder: Maybe hint at the intention to keep it in the user modules?
- `CythonFunctionModuleCode`
- `CythonFunctionPerModule`
- `CythonFunctionNonShared`
- `CythonFunctionInlineable`

matusvalo:
> > I tried to compile scikit-image with PR and without and this PR increases the resulting binary size (sum of all .so files) only few bytes bigger so +1 for me.
> 
> I guess that comment was really for #7557?
> 
> (This PR here is a big disappointment if it increases sizes)

Yes correct. I am sorry. I had two tabs open and wrote in wrong one :-(. Lately, I am kind of messy in GH... :-(

Cython/Utility/CythonFunction.c
```diff
@@ -1384,18 +1433,48 @@ typedef struct {
...
+static int __pyx_FusedFunction_init(PyObject *module) {
+    $modulestatetype_cname *mstate = __Pyx_PyModule_GetState(module);
+
+    PyTypeObject *tp = __Pyx_Get_FusedFunction_Type();
+    if (!tp) return -1;
+
+    if (__Pyx_VerifyCachedType((PyObject*)tp,
+            __PYX_TYPE_MODULE_PREFIX "fused_cython_function",
+            __PYX_SHARED_SIZEOF(__pyx_FusedFunctionObject)) != 0) {
+        Py_DECREF(tp);
```
mattip, line 1458: Is this right? I think it needs to cast to a `PyObject *` (and also in `__Pyx_Get_FusedFunction_Type`)

This came up in astropy/astropy#19422

da-woods: You're right (at least in the Limited API).

I'm a bit puzzled how astropy are getting here though because I don't think they're using the shared utility code feature.

I'll investigate and make sure that nothing else is going wrong

da-woods: > I'm a bit puzzled how astropy are getting here though because I don't think they're using the shared utility code feature.

Actually `__Pyx_Get_FusedFunction_Type` is universal so that's fine.

Not sure why this doesn't break every limited API test on Cython though

mattip: Does it matter that they are using CPython3.11?

da-woods: No - it's just that our Limited API testing is currently invisibly turned off by mistake.

da-woods: Will be fixed in #7576

Issue #7521, opened 2026-02-11
Performance degradation when shared utility is enabled


matusvalo:
### Describe your issue

I have ran performance tests of cython and it seems that enabling shared utility module causes performance degradation. Here are most affected (> 10% performance degradation):

```
2026-02-11 14:12:01  ### Benchmark 'bm_slice_memoryview' (min/median/max/ ±% of median):
2026-02-11 14:12:01      Cython 'master'           = 168.789 nsec, 169.037 nsec, 169.417 nsec
2026-02-11 14:12:01      S-Cython 'master'         = 265.030 nsec, 265.359 nsec, 265.745 nsec  (   +57.0 %)
2026-02-11 14:12:01  ### Benchmark 'bm_create_inner_func_plain' (min/median/max/ ±% of median):
2026-02-11 14:12:01      Cython 'master'           = 178.644 nsec, 179.003 nsec, 179.423 nsec
2026-02-11 14:12:01      S-Cython 'master'         = 209.291 nsec, 209.592 nsec, 214.308 nsec  (   +17.1 %)
2026-02-11 14:12:01  ### Benchmark 'fused_args_1' (min/median/max/ ±% of median):
2026-02-11 14:12:01      Cython 'master'           =   2.360 usec,   2.381 usec,   2.400 usec
2026-02-11 14:12:01      S-Cython 'master'         =   3.770 usec,   3.782 usec,   3.806 usec  (   +58.9 %)
2026-02-11 14:12:01  ### Benchmark 'fused_args_2' (min/median/max/ ±% of median):
2026-02-11 14:12:01      Cython 'master'           =   1.257 usec,   1.262 usec,   1.269 usec
2026-02-11 14:12:01      S-Cython 'master'         =   1.449 usec,   1.451 usec,   1.455 usec  (   +15.0 %)
2026-02-11 14:12:01  ### Benchmark 'bm_getitem_tuple' (min/median/max/ ±% of median):
2026-02-11 14:12:01      Cython 'master'           =  36.994 usec,  37.552 usec,  38.014 usec
2026-02-11 14:12:01      S-Cython 'master'         =  42.142 usec,  42.312 usec,  42.718 usec  (   +12.7 %)
2026-02-11 14:12:01  ### Benchmark 'bm_getitem_ext' (min/median/max/ ±% of median):
2026-02-11 14:12:01      Cython 'master'           = 202.649 usec, 204.566 usec, 204.844 usec
[...]

PR #7557, opened 2026-03-08
Inline slice_memviewslice

da-woods:
Partial fix for #7521.

This came up in a particular scikit-image microbenchmark so almost certainly isn't the only thing. But it looks like this particular function relies heavily on unused branches being eliminated so really wants to be inlined.

On the scikit-image microbenchmark, it's now actually faster with the shared utility code than without. I'm not sure why but possibly just the effect of the `inline` keyword.

The shared utility module itself doesn't change size at all with this fix.

I don't really like the code duplication but I think it's worth it at least for now.

### DIFF
--- a/Cython/Utility/MemoryView.pxd
+++ b/Cython/Utility/MemoryView.pxd
@@ -10,7 +10,11 @@ cdef extern from *:
     typedef int (*__pyx_memoryview_to_dtype_func_type)(char*, PyObject*);
     """
     ctypedef struct {{memviewslice_name}}:
-        pass
+        __pyx_memoryview *memview
+        char *data
+        Py_ssize_t shape[{{max_dims}}]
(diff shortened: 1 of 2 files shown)


da-woods:
> I don't really like the code duplication but I think it's worth it at least for now.

Thinking about this more, I wonder if this is better solved by moving this function to the C utility code. It's all very simple and were aren't really using Cython to do any Python object handling or anything. So moving it to C would just give better control of how it's included.

da-woods:
#7558 is an alternative version of this translated to C rather than duplicated. Performance-wise it's basically identical I think

scoder:
> I wonder if this is better solved by moving this function to the C utility code

Agreed. Closing this PR since it's superseded by https://github.com/cython/cython/pull/7558

PR #7558, opened 2026-03-09
Inline __pyx_memoryview_slice_memviewslice as C function

da-woods:
Partial fix for #7521.

This came up in a particular scikit-image microbenchmark so almost certainly isn't the only thing. But it looks like this particular function relies heavily on unused branches being eliminated so really wants to be inlined.

On the scikit-image microbenchmark, it's now actually faster with the shared utility code than without. I'm not sure why but possibly just the effect of the inline keyword.

The shared utility module itself doesn't change size at all with this fix (despite moving it to C)

Alternative to https://github.com/cython/cython/pull/7557. I prefer this version because it has less duplication but there's some risk of errors in the translation to C

### DIFF
--- a/Cython/Compiler/MemoryView.py
+++ b/Cython/Compiler/MemoryView.py
@@ -336,6 +336,7 @@ class MemoryViewSliceBufferEntry(Buffer.BufferEntry):
                     util_name = "SimpleSlice"
                 else:
                     util_name = "ToughSlice"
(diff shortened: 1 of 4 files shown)


scoder:
3.2.x commit is https://github.com/cython/cython/commit/db550170fe01445b55762adca82f968cb6a4ac04

Issue #7559, opened 2026-03-10
[BUG] Cython cache seems to extracts files to wrong folder when finding zip cache files


raulcd:
### Describe the bug

I am trying to enable Cython cache (https://github.com/cython/cython/pull/6091) at PyArrow and have found the following error when re-running compilation a second time hence when hitting the cache:
```
  FAILED: [code=1] CMakeFiles/arrow_python.dir/pyarrow/src/arrow/python/pyarrow.cc.o
  /usr/bin/x86_64-linux-gnu-g++ -DARROW_HAVE_RUNTIME_AVX2 -DARROW_HAVE_RUNTIME_AVX512 -DARROW_HAVE_RUNTIME_BMI2 -DARROW_HAVE_RUNTIME_SSE4_2 -DARROW_HAVE_SSE4_2 -DARROW_PYTHON_EXPORTING -Darrow_python_EXPORTS -I/home/raulcd/code/arrow/python/pyarrow/src -I/tmp/tmp7015iv4v/build/pyarrow/src -I/usr/include/python3.13 -I/home/raulcd/code/dist/include -I/home/raulcd/code/pyarrow-dev/lib/python3.13/site-packages/numpy/_core/include -Wno-noexcept-type -Wno-self-move  -Wall -fno-semantic-interposition -msse4.2  -fdiagnostics-color=always  -fno-omit-frame-pointer -Wno-unused-variable -Wno-maybe-uninitialized -g -O0  -std=c++20 -fPIC -MD -MT CMakeFiles/arrow_python.dir/pyarrow/src/arrow/python/pyarrow.cc.o -MF CMakeFiles/arrow_python.dir/pyarrow/src/arrow/python/pyarrow.cc.o.d -o CMakeFiles/arrow_python.dir/pyarrow/src/arrow/python/pyarrow.cc.o -c /home/raulcd/code/arrow/python/pyarrow/src/arrow/python/pyarrow.cc
  In file included from /home/raulcd/code/arrow/python/pyarrow/src/arrow/python/pyarrow.cc:32:
  /home/raulcd/code/arrow/python/pyarrow/src/arrow/python/pyarrow_api.h:19:10: fatal error: arrow/python/lib_api.h: No such file or directory
     19 | #include "arrow/python/lib_api.h"
        |          ^~~~~~~~~~~~~~~~~~~~~~~~
  compilation terminated.

```
After some investigation I think this has to do with:
[...]

PR #7561, opened 2026-03-12
Fix cache extraction for multiple files

raulcd:
Closes: https://github.com/cython/cython/issues/7559

As shared on the issue when having a multi file output on the cache in the form of a zip file the extraction extracts into an incorrect folder structure having the name of the file as part of the path:
```
├── lib_api.h
│   └── lib_api.h
├── lib.cpp
│   └── lib.cpp
├── lib.h
│   └── lib.h
```
or `/tmp/tmp7015iv4v/build/lib_api.h/lib_api.h`

I've updated a test and validated the test failed without the change (it was a dir instead of a file before).

### DIFF
--- a/Cython/Build/Tests/TestCyCache.py
+++ b/Cython/Build/Tests/TestCyCache.py
@@ -149,7 +149,7 @@ class TestCyCache(CythonTest):
 
         for output in expected:
             self.assertTrue(os.path.exists(output), output)
-            self.assertTrue(os.path.isfile(output))
+            self.assertTrue(os.path.isfile(output), output)
 
     def test_multi_file_output_cythonize(self):
         self._test_multi_file_output(self.fresh_cythonize)


raulcd:
@da-woods I've submitted a PR and validated the updated test was hitting the scenario and added an assertion. Thanks!

da-woods:
This looks good to me - thanks. I don't know a lot about the cache so I'll wait a few days at least to see if anyone else wants to review it. But if no-one does then I'll merge it.

Cython/Build/Tests/TestCyCache.py
```diff
@@ -149,6 +149,7 @@ def _test_multi_file_output(self, compilation_method):
 
         for output in expected:
             self.assertTrue(os.path.exists(output), output)
+            self.assertTrue(os.path.isfile(output))
```
matusvalo:
```suggestion
            self.assertTrue(os.path.isfile(output), output)
```

matusvalo:
Let's wait for for green CI/CD. Then we can merge and create back port PR to 3.2.5

PR #7567, opened 2026-03-13
Remove unused code in Tempita.

loqs:
This removes use of pkg_resources removed in setuptools 82 and paste which currently in maintenance mode.


scoder:
Thanks, but I don't see a reason to remove the whole CLI if it's easy to fix. (I'm aware that it was broken before due to a missing function.)

PR #7571, opened 2026-03-15
Documentation improvements for shared utility module

da-woods:
* Removed outdated comment about it being *only* memoryviews.

* Add explanation about our expectations for binary compatibility.

### DIFF
--- a/docs/src/userguide/source_files_and_compilation.rst
+++ b/docs/src/userguide/source_files_and_compilation.rst
@@ -471,8 +471,9 @@ There are a few things to bear in mind when deciding whether to use the shared u
   module rather than using a global system module.
 * The shared utility module should be compiled with the same
   :ref:`C macro defines<C_macro_defines>`  as the rest of your Cython modules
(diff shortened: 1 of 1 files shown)


docs/src/userguide/source_files_and_compilation.rst
```diff
@@ -459,7 +459,21 @@ Cython modules, this can result in a larger total size of compiled extensions.
...
+* The contents and interfaces of the shared utility module are not designed to be
+  stable across Python versions.  That means that you must build the shared utility
+  module with the same version of Cython as you build you extension modules.
+  In practice that means that each package will ship its own shared utility
+  module rather than using a global system module.
+* The shared utility module should be compiled with the same
+  :ref:`C macro defines<C_macro_defines>`  as the rest of your Cython modules
+  (essentially, any macro that starts with ``CYTHON`` or the ``Py_LIMITED_API``
+  macro). This is likely to be correct provided you use a shared utility module
+  shipped within your own package.
```
scoder, line 477: Should we be a bit more specific and list some macros that are most or least unlikely to be problematic in the future? At least listing the most problematic ones may make this section easier to find for users.

PR #7572, opened 2026-03-15
Fix "table not aligned" warning in docs

da-woods:


da-woods:
This is trivial so I'm just merging it on success

PR #7573, opened 2026-03-15
Improve documentation of alternative build methods

da-woods:
I've created little sub-pages under source files and compilation to document different possible build backends. Obviously setuptools is by far the most detailed because that's what most of our documentation focused on.

I think they're better on sub-pages because "source files and compilation" was already far too long.

In the short-term this gives minimal examples and a bit of a starting point for people to understand their options. In the long-term it hopefully provides a place for people who actually understand this stuff to fill in the details.


docs/src/userguide/compilation_setuptools.rst
```diff
@@ -0,0 +1,372 @@
...
+    version = "0.1"
+
+    [tool.setuptools]
+    ext-modules = [
+        {name = "cython_example.example", sources = ["example.pyx"]}
+    ]
+
+.. _basic_setup.py:
+
+Basic :file:`setup.py`
```
da-woods, line 39: From here downwards is just moved and is unchanged so shouldn't need extensive review.

docs/src/userguide/compilation_scikit_build.rst
```diff
@@ -0,0 +1,77 @@
...
+reimplement :mod:`scikit-build` with "lessons learned"-type fixes.
+
+Currently :mod:`scikit-build` has Cython helpers while :mod:`scikit-build-core`
+requires manual invocation.
+The main disadvantage of :mod:`scikit-build` is that
+it works with :mod:`setuptools` rather than completely replacing it,
+which :mod:`scikit-build-core` is independent.
+
+Both are based around CMake so are a good choice if you're wrapping
+or using a C/C++ library that uses CMake heavily.
```
da-woods: This does remind me that I should go back to https://github.com/cython/cython/issues/5486 and see if we can get some native cmake support added into Cython.

docs/src/userguide/compilation_scikit_build.rst
```diff
@@ -0,0 +1,77 @@
+***************************
+Compiling with scikit-build
```
rgommers: I wouldn't add new content for `scikit-build`, and certainly not above `scikit-build-core`. The former is legacy, and the latter is essentially a modern rewrite of it. If users are looking for a new build backend to use with CMake, they should only consider `scikit-build-core`. 

So I'd either remove this part, or move it to the bottom with a clear "legacy" warning.

docs/src/userguide/compilation_meson.rst
```diff
@@ -0,0 +1,50 @@
+********************
+Compiling with Meson
+********************
+
+Meson is a build system mainly focused on building compiled languages
+such as C and C++.  However the `meson-python` package
```
rgommers:
```suggestion
such as C and C++. The `meson-python` package
```

docs/src/userguide/compilation_meson.rst
```diff
@@ -0,0 +1,50 @@
...
+allows it to work as a PEP517 backend to build Python wheels.
```
rgommers:
```suggestion
allows it to work as a build backend to build Python wheels.
```

It's discouraged to use PEP numbers to explain or refer to something, and "build backend" is used in other sections.

docs/src/userguide/compilation_meson.rst
```diff
@@ -0,0 +1,50 @@
...
+
+* `The meson-python docs <https://mesonbuild.com/meson-python/>`_ which
+  describe how to use Meson as a build backend. 
+* `The section on their built-in Cython support <https://mesonbuild.com/Cython.html>`_
+  provides examples of how to build a Cython module.
+* `A complete simple example <https://github.com/oscarbenjamin/meson_simple>`_.
+
+You need two files (plus your Cython source).
+First is :file:`pyproject.toml` which specifies the build
+backend and dependences as well as giving details of your project::
```
rgommers:
```suggestion
backend and dependencies, and contains the your project-level metadata::
```

docs/src/userguide/compilation_meson.rst
```diff
@@ -0,0 +1,50 @@
...
+    requires = ['meson-python', 'cython']
+
+    [project]
+    name = 'example'
+    version = '1.0.0'
+
+Secondly you need a :file:`meson.build`, which describes how your module will be
+built::
+
+    project('my project', 'cython')
```
rgommers: It's not required but I'd say it is recommended to explicitly add `'c'` here, rather than keeping that implicit. 

You may also want to explain that if you target C++ as the intermediate language, it should be `'cpp'`.

docs/src/userguide/compilation_meson.rst
```diff
@@ -0,0 +1,50 @@
...
+       'foo',
+       'foo.pyx',
+       install: true
+    )
+
+Additional Cython arguments can be passed to ``py.extension_module`` e.g.::
+
+    cython_args : ['-Xboundscheck=False'],
+
+(these are essentially command-line arguments to the `cython` executable).
```
rgommers:
```suggestion
(these are essentially command-line arguments to the ``cython`` executable).
```

docs/src/userguide/compilation_scikit_build.rst
da-woods: Thanks - I'll do that.

I wasn't sure whether `scikit-build-core` should be considered "intended to replace `scikit-build`" or "has actually replaced `scikit-build`". But it sounds like the latter so can rewrite this.

docs/src/userguide/source_files_and_compilation.rst
```diff
@@ -58,9 +58,13 @@ compiled with the C compiler using whatever options are appropriate on your
...
 official Python documentation.
 
-The other, and probably better, way is to use the :mod:`setuptools` extension
-provided with Cython. The benefit of this method is that it will give the
-platform specific compilation options, acting like a stripped down autotools.
+If you want to compile Cython modules in an automated, reproducible way,
+especially as part of a largely package, it is usually better to use one
+of the Python build backends to compile your modules.  The benefit of this
+method is that it will give the platform specific compilation options,
+acting like a stripped down autotools.  To compile a single module for
```
rgommers: I'm not sure exactly how to phrase a rewrite, but I'd say that you should recommend using Meson or CMake as modern build systems. They're way beyond autotools, let alone a "stripped down autotools". For `setuptools` I'd say that is a reasonable characterization, which is why it should not be recommended anymore in 2026 - I'd call it legacy here.

docs/src/userguide/compilation_scikit_build.rst
rgommers: It is indeed. Let's ping @henryiii here as the main author of `scikit-build-core` - Henry, you may be interested in reviewing this (and maybe the `hatchling` part as well)?

matusvalo:
In separate PR we can move section `compiling-shared-module-using-setuptools` to `setuptools`  and replace it just with link. As followup we can add something similar in meson, we can use what was done in scipy/scikit-learn.

da-woods:
> In separate PR we can move section `compiling-shared-module-using-setuptools` to `setuptools` and replace it just with link. As followup we can add something similar in meson, we can use what was done in scipy/scikit-learn.

Definitely worth documenting if you have a good recipie. I don't think we need to be ultra-thorough about covering all the documentation in all build systems though.

docs/src/userguide/compilation_scikit_build.rst
henryiii: Sure, I can. Agree, and scikit-build-core does have a cython helper module, the `cython-cmake` package. (I'll go ahead and do a small update on that).

docs/src/userguide/compilation_scikit_build.rst
```diff
@@ -0,0 +1,70 @@
...
+            "${CMAKE_CURRENT_SOURCE_DIR}/example.pyx" --output-file example.c
+    DEPENDS example.pyx
+    VERBATIM)
+
+    python_add_library(example MODULE example.c WITH_SOABI)
+
+    install(TARGETS example DESTINATION .)
+
+Currently the Cython command is invoked manually as a custom CMake command,
+which is a bit verbose but fairly simple.
```
isuruf: I would suggest using `cython-cmake` to make this a bit simpler. See https://github.com/scikit-build/cython-cmake

docs/src/userguide/compilation_scikit_build.rst
```diff
@@ -0,0 +1,70 @@
...
+
+.. code-block:: cmake
+
+    cmake_minimum_required(VERSION 3.15...3.26)
+    project(${SKBUILD_PROJECT_NAME} LANGUAGES C)
+
+    find_package(
+    Python
+    COMPONENTS Interpreter Development.Module
+    REQUIRED)
```
isuruf: Make this a little bit easier to read
```suggestion
    find_package(
      Python
      COMPONENTS Interpreter Development.Module
      REQUIRED)
```

docs/src/userguide/compilation_hatchling.rst
```diff
@@ -0,0 +1,31 @@
+************************
+Compiling with hatchling
+************************
+
+Hatchling is a Python build backend. It does not
+appear to focus much of extension modules, however the
+`hatch-cython plugin <https://github.com/joshua-auchincloss/hatch-cython>`_
+does provide Cython support.
```
henryiii:
```suggestion
Hatchling is a Python build backend. It does not natively
build extension modules, however it supports plugins, and the
`hatch-cython plugin <https://github.com/joshua-auchincloss/hatch-cython>`_
does provide Cython support.
```

docs/src/userguide/compilation_scikit_build.rst
```diff
@@ -0,0 +1,70 @@
...
+:mod:`scikit-build-core` is a modern and more fully-featured replacement.
+
+Both are based around CMake and so are a good choice if you're wrapping
+or using a C/C++ library that uses CMake heavily.
+
+If you want to use one of these build backends, our recommendation is
+to use :mod:`scikit-build-core`.  However, the main remaining advantage
+of :mod:`scikit-build` is that it comes with builtin utilities to
+handle invoke Cython while for :mod:`scikit-build-core` you must
+invoke Cython manually through CMake.
```
henryiii:
```suggestion
CMake is the most popular build system, with support for multiple languages,
including C, C++, and Cuda. :mod:`scikit-build-core` is the official adaptor
for using CMake for a Python package.
```

docs/src/userguide/compilation_scikit_build.rst
```diff
@@ -0,0 +1,70 @@
+********************************
+Compiling with scikit-build-core
```
henryiii:
```suggestion
Compiling with CMake
```

docs/src/userguide/compilation_scikit_build.rst
```diff
@@ -0,0 +1,70 @@
...
+of :mod:`scikit-build` is that it comes with builtin utilities to
+handle invoke Cython while for :mod:`scikit-build-core` you must
+invoke Cython manually through CMake.  
+
+:mod:`scikit-build-core`
+========================
+
+If you want to use `scikit-build-core` then
+`their docs <https://scikit-build-core.readthedocs.io/en/latest/guide/getting_started.html>`_
+provides a simple, minimal example of how to build a Cython module that has been adapted here.
```
henryiii:
```suggestion
If you want to use `scikit-build-core` then
`their docs <https://scikit-build-core.readthedocs.io/en/latest/guide/getting_started.html>`_
provides a simple, minimal example of how to build a Cython module. There is an official Cython plugin, `cython-cmake <https://github.com/scikit-build/cython-cmake>`_.
```

docs/src/userguide/compilation_scikit_build.rst
```diff
@@ -0,0 +1,70 @@
...
+
+Like most Python build backends, you need a :file:`pyproject.toml`::
+
+    [build-system]
+    requires = ["scikit-build-core", "cython"]
```
henryiii:
```suggestion
    requires = ["scikit-build-core", "cython", "cython-cmake"]
```

docs/src/userguide/compilation_scikit_build.rst
henryiii:
```suggestion
    cmake_minimum_required(VERSION 3.21...4.0)
    project(${SKBUILD_PROJECT_NAME} LANGUAGES C)

    find_package(
      Python
      COMPONENTS Interpreter Development.Module
      REQUIRED)
    include(UseCython)
    
    cython_transpile(example.pyx LANGUAGE C OUTPUT_VARIABLE example_c)

    python_add_library(example MODULE ${example_c} WITH_SOABI)
    install(TARGETS example DESTINATION .)
```

docs/src/userguide/compilation_scikit_build.rst
```diff
@@ -0,0 +1,70 @@
...
+which is a bit verbose but fairly simple.
+
+Install your package by running ``pip install .``.
+
+:mod:`scikit-build`
+===================
+
+If you do want to use the legacy :mod:`scikit-build` then
+A good place to learn from is
+`the minimal Cython example in their tests <https://github.com/scikit-build/scikit-build/tree/main/tests/samples/hello-cython>`_.
```
henryiii:
```suggestion
```

docs/src/userguide/source_files_and_compilation.rst
```diff
@@ -109,343 +115,27 @@ that CPython generates for disambiguation, such as
...
-        ext_modules = cythonize("package/*.pyx")
-    )
+If you are building a larger package that includes Cython modules then it is
+recommended to use a "build backend" which can take care of all stages of
+the compilation and packaging.  Historically that has mostly meant
+:mod:`setuptools`, however other options are increasingly popular and
+should be strongly recommended when starting new projects: a lot
+of the Scientific Python ecosystem uses `meson` as their first choice.
+Most of the Cython documentation assumes :mod:`setuptools` however we
+have tried to provide simple examples for other more modern options.
```
ngoldbaum, line 130: Would you be open to a contributor updating the rest of the docs to add versions for other build backends using a tab picker UI?

docs/src/userguide/compilation_scikit_build.rst
da-woods: Thanks. I've taken most of your other suggestions, but I think this cuts it down a bit more than I'd like so I'll have a second go at this section myself.

docs/src/userguide/compilation_meson.rst
```diff
@@ -0,0 +1,53 @@
...
+
+    cython_args : ['-Xboundscheck=False'],
+
+(these are essentially command-line arguments to the ``cython`` executable).
+Arguments to the C compiler are passed as::
+
+    c_args: ['-DCYTHON_USE_TYPE_SPECS=1'],
+
+To use C++ as an intermediate language instead of C, change the ``"c"`` in
+the ``project`` line at the start of the file to ``"cpp"``.
```
ngoldbaum, line 53: I recently updated the NumPy docs to include a section on using the NumPy C API from Cython: https://numpy.org/devdocs/user/c-info.python-as-glue.html#cython

Maybe it makes sense to crosslink? The new NumPy docs aren’t yet in a released version.

da-woods: I'll add a link to them.

docs/src/userguide/source_files_and_compilation.rst
da-woods: I wasn't sure if it needed doing thoroughly (i.e. I'm not necessarily opposed but I don't think it necessarily needs doing rigorously).

I wonder if the most useful place would be the tutorials (especially the first ones https://cython.readthedocs.io/en/latest/src/quickstart/build.html and https://cython.readthedocs.io/en/latest/src/tutorial/cython_tutorial.html).

da-woods:
I'll merge this on the basis that it's more of a starting point than an end point so I think it's more useful if it's in place. Thanks everyone that gave feedback.

PR #7574, opened 2026-03-16
Fix "special" CI runs

da-woods:
Currently the `env` argument is ignored rather than used. This affects the Limited API, sharedlib, different GCC versions and a few others.

I can't see an easy way of passing env in, so I've replaced it with a whole bunch of optional parameters.


da-woods:
This is expected to fail because of a Limited API issue which I'll fix separately - I actual want to see it fail though to know it's working.

scoder:
Your PR description looks incomplete. Apart from that, I'm fine with the change. It's as explicit as it gets.

PR #7575, opened 2026-03-16
Simplify codestyle on CI and split off docs build

da-woods:
A while ago I added a standalone file to just run the code style test (just to make it easier that manually disabling all other tests). Use that on the CI run instead.

Also fix the docs build (by actually building it, rather than just trying to upload it), and split it off into a separate job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
 
 
   ci-all:
-    needs: [codespell, codestyle, ci-early]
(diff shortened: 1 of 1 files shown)


da-woods:
This isn't working right now. But actually the code style tests look to silently being skipped on master because of the missing dependency so it's no worse....

I'll have to come back to it later though

.github/workflows/ci.yml
da-woods, line 321: This appears to upload the docs, but not build the docs

PR #7577, opened 2026-03-17
Implement LZSS-like compression as default string compression

scoder:
The decompressor is tiny enough to become part of the modules, thus making them stand-alone despite any potentially unavailable stdlib modules.

See discussion in https://github.com/cython/cython/pull/7546

### DIFF
--- a/Cython/Utility/StringTools.c
+++ b/Cython/Utility/StringTools.c
@@ -153,6 +153,7 @@ bad:
 static PyObject *__Pyx_DecompressString_LZSS(const char *s, size_t compressed_length, size_t uncompressed_length); /*proto*/
 
 //////////////////// DecompressString_LZSS ////////////////////
+//@requires: IncludeStringH
 
 #ifndef __Pyx_DecompressString_LZSS_UNUSED
 // Depends on <stdint.h>, which is globally included in the module preamble.


scoder:
The reported module sizes in the benchmark look roughly the same as for the master branch, some slightly larger, some slightly smaller, mostly in the range of 100-200 bytes difference. That's great because it means that the compression is close enough to zlib's to be a useful replacement.

Cython/Utility/StringTools.c
```diff
@@ -135,8 +143,85 @@ static PyObject *__Pyx_DecompressString(const char *s, Py_ssize_t length, int al
...
+//////////////////// DecompressString_LZSS.proto ////////////////////
+// TODO: Move to the shared module if we can import that before initialising the strings.
+
+static PyObject *__Pyx_DecompressString_LZSS(const char *s, size_t compressed_length, size_t uncompressed_length); /*proto*/
+
+//////////////////// DecompressString_LZSS ////////////////////
+
+#ifndef __Pyx_DecompressString_LZSS_UNUSED
+CYTHON_UNUSED
+static CYTHON_SMALL_CODE size_t __pyx_lzss_decompress(const uint8_t *src, uint8_t *dst, size_t dst_len) {
```
da-woods: In principle I think you need to include stdint.h for uint8_t - it doesn't look to be part of the headers that Python.h includes automatically. I suspect it usually ends up included indirectly of course, but it's not guaranteed

da-woods (commented):
I can't see any obvious issues although this type of code does tend to end up pretty opaque

Cython/Utility/StringTools.c
scoder: We include it in `ModuleSetupCode.c`, so it's already there.
https://github.com/cython/cython/blob/d0835eabed53d79134a17959e49e359c19aa31c2/Cython/Utility/ModuleSetupCode.c#L547

Arguably, that's a little fragile because the usage there seems local and it's easy to accidentally remove the include when changing the code that seems to use it, but OTOH, it's such a basic header these days that I doubt that we'd consider removing the include without a bit of looking around if we depend on it elsewhere. I can add a comment that would make the dependency easier to spot.

da-woods: Fair enough - I did a `git grep` for `stdint` but missed it in ModuleSetupCode.c

PR #7580, opened 2026-03-18
Fix issue with profiling and non-convertable ctuple

da-woods:
`CCodeWriter.put_trace_return` used the presence of a `to_py_function` to work out if something could be converted to Python. This means that setting it in the constructor of `CTupleType` was misleading it and was the wrong thing to do. Instead, only create the name when asked to.

### DIFF
--- a/Cython/Compiler/PyrexTypes.py
+++ b/Cython/Compiler/PyrexTypes.py
@@ -4689,6 +4689,8 @@ class CTupleType(CType):
 
     _convert_to_py_code = None
     _convert_from_py_code = None
+    to_py_function = None
+    from_py_function = None
 
     def __init__(self, cname, components):
         from .Builtin import tuple_type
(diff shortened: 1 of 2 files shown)


Cython/Compiler/PyrexTypes.py
```diff
@@ -4696,8 +4696,8 @@ def __init__(self, cname, components):
         self.components = components
         self.equivalent_type = tuple_type
         self.size = len(components)
-        self.to_py_function = f"{Naming.convert_func_prefix}_to_py_{self.cname}"
-        self.from_py_function = f"{Naming.convert_func_prefix}_from_py_{self.cname}"
+        self.to_py_function = None
+        self.from_py_function = None
```
scoder: Can these become general class attributes of `CType` ?

da-woods:
3.2.x commit: 9260918d5

PR #7582, opened 2026-03-19
Add thread-safe cache initialization section to free-threading docs

ngoldbaum:
Based on my experience working on https://github.com/scipy/scipy/pull/24799.

### DIFF
--- a/docs/examples/userguide/freethreading/buggy_init.pyx
+++ b/docs/examples/userguide/freethreading/buggy_init.pyx
@@ -1,0 +1,16 @@
+import time
+
+def expensive_function():
+    time.sleep(5)
+    return {"hello": "world"}
+
+cdef class A:
+    def __init__(self):
+        self._py_obj = None
+
+    @property
+    def obj(self):
+        if not self._py_obj:
+            self._py_obj = expensive_function()
+        return self._py_obj
+
--- a/docs/examples/userguide/freethreading/buggy_init.pyx
+++ b/docs/examples/userguide/freethreading/buggy_init.pyx
@@ -1,3 +1,5 @@
+# distutils: language=c++
+
 import time
 
 def expensive_function():
@@ -5,6 +7,8 @@ def expensive_function():
     return {"hello": "world"}
 
 cdef class A:
+cdef class A:
+    cdef object _py_obj
     def __init__(self):
         self._py_obj = None
 
--- a/docs/examples/userguide/freethreading/buggy_init.pyx
+++ b/docs/examples/userguide/freethreading/buggy_init.pyx
@@ -7,7 +7,6 @@ def expensive_function():
     return {"hello": "world"}
 
 cdef class A:
-cdef class A:
     cdef object _py_obj
     def __init__(self):
         self._py_obj = None
--- a/docs/examples/userguide/freethreading/correct_init.pyx
+++ b/docs/examples/userguide/freethreading/correct_init.pyx
@@ -1,3 +1,5 @@
+# distutils: language=c++
+
 from libcpp.mutex cimport py_safe_call_once, py_safe_once_flag
 
 import time
@@ -11,6 +13,7 @@ def expensive_function():
     return {"hello": "world"}
 
 cdef class A:
+    cdef object _py_obj
     cdef py_safe_once_flag flag
 
     def __init__(self):
--- a/docs/examples/userguide/freethreading/correct_init.pyx
+++ b/docs/examples/userguide/freethreading/correct_init.pyx
@@ -1,0 +1,24 @@
+from libcpp.mutex cimport py_safe_call_once, py_safe_once_flag
+
+import time
+
+cdef void init_py_obj(void *void_instance):
+    cdef A instance = <A>void_instance
+    instance._py_obj = expensive_function()
+
+def expensive_function():
+    time.sleep(5)
+    return {"hello": "world"}
+
+cdef class A:
+    cdef py_safe_once_flag flag
+
+    def __init__(self):
+        self._py_obj = None
+
+    @property
+    def obj(self):
+        cdef void *void_self = <void *>self
+        py_safe_call_once(self.flag, init_py_obj, void_self)
+        return self._py_obj
+
--- a/docs/examples/userguide/freethreading/correct_py_init.pyx
+++ b/docs/examples/userguide/freethreading/correct_py_init.pyx
@@ -23,5 +23,6 @@ cdef class A:
         if not self.cache_flag.load():
             def closure():
                 self._py_obj = expensive_function()
+                self.cache_flag.store(True)
             py_safe_call_object_once(self.flag, closure)
         return self._py_obj
--- a/docs/examples/userguide/freethreading/correct_py_init.pyx
+++ b/docs/examples/userguide/freethreading/correct_py_init.pyx
@@ -1,3 +1,5 @@
+# distutils: language=c++
+
 from libcpp.atomic cimport atomic
 from libcpp.mutex cimport py_safe_call_object_once, py_safe_once_flag
 
@@ -8,6 +10,7 @@ def expensive_function():
     return {"hello": "world"}
 
 cdef class A:
+    cdef object _py_obj
     cdef py_safe_once_flag flag
     cdef atomic[int] cache_flag
 
--- a/docs/examples/userguide/freethreading/correct_py_init.pyx
+++ b/docs/examples/userguide/freethreading/correct_py_init.pyx
@@ -1,0 +1,24 @@
+from libcpp.atomic cimport atomic
+from libcpp.mutex cimport py_safe_call_object_once, py_safe_once_flag
+
+import time
+
+def expensive_function():
+    time.sleep(5)
+    return {"hello": "world"}
+
+cdef class A:
+    cdef py_safe_once_flag flag
+    cdef atomic[int] cache_flag
+
+    def __init__(self):
+        self._py_obj = None
+        self.cache_flag.store(0)
+
+    @property
+    def obj(self):
+        if not self.cache_flag.load():
+            def closure():
+                self._py_obj = expensive_function()
+            py_safe_call_object_once(self.flag, closure)
+        return self._py_obj
--- a/docs/src/userguide/freethreading.rst
+++ b/docs/src/userguide/freethreading.rst
@@ -192,11 +192,11 @@ Consider the following code::
             self.key = key
             self._py_obj = None
 
-        property obj:
-            def __get__(B self):
-                if not self._py_obj:
-                    self._py_obj = A(create_c_object(self.key))
-                return self._py_obj
+        @property
+        def obj(self):
+            if not self._py_obj:
+                self._py_obj = A(create_c_object(self.key))
+            return self._py_obj
 
 Concurrently accessing the ``obj`` property of the ``A`` class from multiple
 threads could cause data races when one thread does ``if not self._py_obj``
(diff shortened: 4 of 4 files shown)


docs/src/userguide/freethreading.rst
```diff
@@ -158,6 +158,109 @@ modules with the Limited API (since ``PyMutex`` is unavailable in the
...
+
+          def __init__(self, int key):
+              self.key = key
+              self._py_obj = None
+
+          property obj:
+              def __get__(B self):
+                  def closure():
+                      self._py_obj = A(create_c_object(self.key))
+                      self.cache_flag.store(True)
```
da-woods: This I think belongs to your later example

docs/src/userguide/freethreading.rst
```diff
@@ -158,6 +158,109 @@ modules with the Limited API (since ``PyMutex`` is unavailable in the
...
+      cdef class B:
+          cdef:
+              int key
+              py_safe_once_flag flag
+
+          def __init__(self, int key):
+              self.key = key
+              self._py_obj = None
+
+          property obj:
```
da-woods: I'd probably recommend just using the Python syntax:

```
@property
def obj(self):
    ...
```

It should end up as exactly the same code, and we general try to steer people to Python-type syntax where available.

da-woods:
One more comment (which is maybe more a suggestion to "how to do it in SciPy" than this documentation PR):

you could avoid the closure by using the "args" version of `py_safe_call_once`. Unfortunately it's a little fiddly to pass `self` in because we don't allow `object` to varargs. But I think you could cast it to/from `void*`. There's no lifetime issues because it's called immediately or not at all. Untested example:

```
cdef void _initialize_obj(void *self_B):
    (<B>self_B.)_py_obj = A(create_c_object(self.key))


cdef class B:
    ...

    @property
    def obj(self):
        py_safe_call_once(self.flag, _initialize_obj, <void*>self)
        return self._py_obj
```

I'll have a proper look at the proposed documentation tomorrow

ngoldbaum:
The C++ compiler complains:

```
scipy/spatial/_ckdtree.cpython-314-darwin.so.p/_ckdtree.cpp:11183:5: error: no matching function for call to '__pyx_cpp_py_safe_call_once'
 11183 |     __pyx_cpp_py_safe_call_once<void *>(__pyx_v_self->flag, ((void *)(&__pyx_f_5scipy_7spatial_8_ckdtree_init_pytree)));
       |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
scipy/spatial/_ckdtree.cpython-314-darwin.so.p/_ckdtree.cpp:3780:10: note: candidate function template not viable: expects an lvalue for 2nd argument
 3780 |     void __pyx_cpp_py_safe_call_once(std::once_flag& flag, Callable& callable, Args&&... args) {
      |          ^                                                 ~~~~~~~~~~~~~~~~~~
1 error generated.
```

docs/src/userguide/freethreading.rst
```diff
@@ -158,6 +158,107 @@ modules with the Limited API (since ``PyMutex`` is unavailable in the
...
+
+          @property
+          def obj(self):
+              def closure():
+                  self._py_obj = A(create_c_object(self.key))
+              py_safe_call_object_once(self.flag, closure)
+              return self._py_obj
+
+If you want to avoid the need to define a closure or wrapper object in every
+call to ``__get__``, you can also use an atomic boolean flag to indicate whether
```
ngoldbaum: should be obj now

da-woods:
> The C++ compiler complains:

So I think that's because you're casting the function to `<void*>` rather than the `self` arg. However... casting the `self` arg also fails with a different compiler error. I'll do a fix for it and make sure we're testing passing arguments correctly

da-woods:
Fixup to make my proposed example work is in https://github.com/cython/cython/pull/7585 (together with a small amount of working code that demonstrates it)

ngoldbaum:
> So I think that's because you're casting the function to <void*> rather than the self arg

Yes, that's wrong. Sorry for the confusion, I was at the end of a long day...

I can confirm that with your Cython branch I'm able to get rid of the closure in SciPy. Are you planning to backport that for 3.2.5? And if so, when were you planning to cut a new bugfix release?

ngoldbaum:
It probably makes sense to discuss `py_safe_call_once` before `py_safe_call_object_once`. I only used it in these docs because I couldn't figure out how to pass `self`. The docs should also try to explain when you'd prefer `py_safe_call_object_once` (e.g. if you need to single-initialize a resource managed by Python code or a Python singleton or something like that.

da-woods:
> It probably makes sense to discuss `py_safe_call_once` before `py_safe_call_object_once`. I only used it in these docs because I couldn't figure out how to pass `self`. The docs should also try to explain when you'd prefer `py_safe_call_object_once` (e.g. if you need to single-initialize a resource managed by Python code or a Python singleton or something like that.

Yeah agree with all of this. I suspect it probably also isn't work having the `cache_flag` version either (it's really only a workaround for "making a closure is a bit expensive" and is really duplicating some of the logic in `once_flag`).

----------------------------

I've been trying to think of a way of making passing Python arguments in a bit nicer, but unfortunately I'm not coming up with many good ideas (especially trying to keep it as a library feature rather than adding anything to the language). `void*` is a bit hacky but I think it's fine if documented and so people don't have to find it independently.

docs/src/userguide/freethreading.rst
```diff
@@ -158,6 +158,147 @@ modules with the Limited API (since ``PyMutex`` is unavailable in the
...
+Consider the following code::
+
+    from mylib cimport c_object, create_c_object
+
+    class A:
+        cdef c_object *obj
+        cdef public int count
+
+        def __init__(A self, c_object *obj):
+            self.obj = c_object
```
da-woods: I don't think this'll work because you can't pass C pointers to `def` functions

docs/src/userguide/freethreading.rst
```diff
@@ -158,6 +158,147 @@ modules with the Limited API (since ``PyMutex`` is unavailable in the
...
+
+
+    cdef class B:
+        cdef:
+            int key
```
da-woods: I think we're missing the declaration for `_py_obj` here

docs/src/userguide/freethreading.rst
```diff
@@ -158,6 +158,147 @@ modules with the Limited API (since ``PyMutex`` is unavailable in the
...
+
+          @property
+          def obj(self):
+              cdef void *void_self = <void *>self
+              py_safe_call_once(self.flag, init_py_obj, void_self)
+              return self._py_obj
+
+Since ``py_safe_call_once`` cannot accept a Python object as an argument, this
+example casts to ``void *`` to work around that. This is safe so long as the
+``cdef`` initialization function doesn't need any data managed by Python objects.
```
da-woods: I'm not quite sure about the "This is safe..." section.

From my point of view it's safe because we hold an actual reference to `self` for the entire duration of the `call_once` call

docs/src/userguide/freethreading.rst
```diff
@@ -158,6 +158,147 @@ modules with the Limited API (since ``PyMutex`` is unavailable in the
...
+
+In cases where the expensive calculation happens in Python, or where you need to
+initialize a Python singleton, you can use ``py_safe_call_object_once``::
+
+
+  from libcpp.mutex cimport py_safe_call_object_once, py_safe_once_flag
+
```
da-woods: Indentation looks wrong here

da-woods (commented):
There's a few bits where I think the code won't quite compile correctly. We've typically put our documentation examples as separate files in the `docs/examples` directory, and then they at least get compiled (although not run) to verify correctness.

I wonder if the example could be cut down further partly for the sake of simplicity and partly to get it to something compilable without too many extra details.

How about omitting class `A` and just saying

```
from somewhere import expensive_python_function

cdef void init_py_obj(void *void_instance):
          cdef B instance = <B>void_instance
          instance._py_obj = expensive_python_function(instance.key)
```

and then the details of what you're initializing `_py_obj` with disappear completely.

ngoldbaum:
Thanks for the pointer about the examples directory. I moved the code examples to there. I also decided to delete one of the `py_safe_call_object_once` examples, leaving only the version with an atomic flag because I think that's a useful pattern to document.

docs/src/userguide/freethreading.rst
ngoldbaum: I adopted a slight tweak of this language, thanks!

docs/examples/userguide/freethreading/buggy_init.pyx
```diff
@@ -0,0 +1,16 @@
+import time
```
da-woods, line 3:
```suggestion
# distutils: language=c++

import time
```

I think you might need to add this to the three files to get the examples to build with C++ mode. But it's probably worth waiting to see if that's actually the case

docs/examples/userguide/freethreading/correct_init.pyx
```diff
@@ -0,0 +1,24 @@
+from libcpp.mutex cimport py_safe_call_once, py_safe_once_flag
```
da-woods, line 3:
```suggestion
# distutils: language=c++

from libcpp.mutex cimport py_safe_call_once, py_safe_once_flag
```

(probably)

docs/examples/userguide/freethreading/correct_py_init.pyx
```diff
@@ -0,0 +1,24 @@
+from libcpp.atomic cimport atomic
```
da-woods, line 3:
```suggestion
# distutils: language=c++

from libcpp.atomic cimport atomic
```

(probably)

docs/examples/userguide/freethreading/buggy_init.pyx
```diff
@@ -0,0 +1,16 @@
...
+
+def expensive_function():
+    time.sleep(5)
+    return {"hello": "world"}
+
+cdef class A:
+    def __init__(self):
```
da-woods, line 11:
```suggestion
cdef class A:
    cdef object _py_obj
    def __init__(self):
```

This'll be needed for it to actually run - it isn't usually possible to add attributes to `cdef classes` (unless they're special-cased to have a dict)

docs/examples/userguide/freethreading/correct_init.pyx
```diff
@@ -0,0 +1,24 @@
...
+cdef void init_py_obj(void *void_instance):
+    cdef A instance = <A>void_instance
+    instance._py_obj = expensive_function()
+
+def expensive_function():
+    time.sleep(5)
+    return {"hello": "world"}
+
+cdef class A:
+    cdef py_safe_once_flag flag
```
da-woods, line 17:
```suggestion
    cdef object _py_obj
    cdef py_safe_once_flag flag
```

docs/examples/userguide/freethreading/correct_py_init.pyx
```diff
@@ -0,0 +1,24 @@
...
+from libcpp.mutex cimport py_safe_call_object_once, py_safe_once_flag
+
+import time
+
+def expensive_function():
+    time.sleep(5)
+    return {"hello": "world"}
+
+cdef class A:
+    cdef py_safe_once_flag flag
```
da-woods, line 14:
```suggestion
    cdef object _py_obj
    cdef py_safe_once_flag flag
```

docs/examples/userguide/freethreading/buggy_init.pyx
ngoldbaum: Seems to be needed. I just went ahead and applied all your suggestions.

ngoldbaum:
OK, I think tests are passing again. Anything else needed here?

docs/examples/userguide/freethreading/correct_py_init.pyx
```diff
@@ -0,0 +1,27 @@
...
+
+    def __init__(self):
+        self._py_obj = None
+        self.cache_flag.store(0)
+
+    @property
+    def obj(self):
+        if not self.cache_flag.load():
+            def closure():
+                self._py_obj = expensive_function()
```
da-woods, line 25:
```suggestion
                self._py_obj = expensive_function()
                self.cache_flag.store(True)
```

da-woods:
@seberg These are the docs that Nathan added on call_once

da-woods:
and the relevant SciPy change is https://github.com/scipy/scipy/pull/24799

PR #7588, opened 2026-03-22
Escape module string when importing class from unicode module

da-woods:
I don't think this is is an important fix, because most compilers do handle unicode characters in strings. However, we do generally try to generate ASCII C code for maximum compatibility.

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -512,8 +512,7 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
         h_code.putln("")
         h_code.putln("static int %s(void) {" % self.api_name("import", env))
         h_code.putln("PyObject *module = 0;")
-        h_code.putln(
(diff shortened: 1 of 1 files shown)


scoder:
Looks right, but it seems we're not always `EncodedString` wrapping all module names. That would need fixing as well, in the right spot.

Cython/Compiler/ModuleNode.py
```diff
@@ -512,7 +512,8 @@ def put_utility_code(name, src_file, include_requires=True):
         h_code.putln("")
         h_code.putln("static int %s(void) {" % self.api_name("import", env))
         h_code.putln("PyObject *module = 0;")
-        h_code.putln('module = PyImport_ImportModule(%s);' % env.qualified_name.as_c_string_literal())
+        h_code.putln(
+            'module = PyImport_ImportModule(%s);' % env.qualified_name.as_c_string_literal())
```
scoder: There is no functional change here, so no need to touch this line.
```suggestion
        h_code.putln('module = PyImport_ImportModule(%s);' % env.qualified_name.as_c_string_literal())
```

PR #7592, opened 2026-03-27
Pin Github action versions

da-woods:
... to exact commits.  This seems to be recommended as a security measure to avoid problems if someone malicious manages to retag an existing release to change github actions underneath existing users.

The possible flaw in the system is that we're still likely to accept dependabot's recommendations for updates, and it just makes the version used a little more opaque.

### DIFF
--- a/.github/workflows/nightly-wheels.yml
+++ b/.github/workflows/nightly-wheels.yml
@@ -40,9 +40,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Cython
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      - uses: deadsnakes/action@v3.2.0
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: "3.13"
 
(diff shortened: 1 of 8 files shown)


.github/workflows/nightly-wheels.yml
```diff
@@ -40,9 +40,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Cython
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      - uses: deadsnakes/action@v3.2.0
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
```
da-woods, line 45: This is changed from deadsnakes because I don't think there's a good reason to be using deadsnakes here

scoder:
This is safer and doesn't hurt, given that dependabot takes care of the updates for us.

We can also go one step further and add hashes to our pip dependencies:
https://pip.pypa.io/en/stable/topics/repeatable-installs/#hash-checking
But I haven't found out yet if dependabot supports that format in our `cibuildwheel-requirement.txt`.

.github/workflows/nightly-wheels.yml
scoder: There probably was before we dropped support for anything older than Py3.9.

da-woods:
I think it's only the cibuildwheel dependency that's updated by depandabot. It's probably easier just to add the hash to that and revert if it doesn't work. 

I'll have a look at the rest of the pip dependencies. It's probably worth pinning the ones used for the wheel build (twine, setuptools, maybe abi3audit).

I guess maybe the docs build too.

I'm not convinced that hashes of the test dependencies are too important.

scoder:
I'll merge this PR as it's a self-contained change. We can try the rest independently.

PR #7593, opened 2026-03-27
Escape github label name

da-woods:
... by sending it through the environment.

This is a potential "code injection via template expansion" that zizmor (zizmor.sh) identifies.  It does identify a few others too however they're all from input variables that we directly control (e.g. `${{ inputs.compiler }}`) so I believe are genuinely fine, while this one is from something somewhat external. I don't think it's a real issue
but it seemed like it might be worth fixing while I was looking at it


da-woods:
I can't merge this unfortunately because it didn't trigger the full set of tests so doesn't show as "passed"

scoder:
3.2.x commit is https://github.com/cython/cython/commit/535051276c4ceec81176751e6b436bcdefcf4667

PR #7594, opened 2026-03-27
Fix a couple of cyfunction ref-counting errors

da-woods:
1. The annotations dict wasn't visited. Normally it's just strings, so this shouldn't matter. However, it is modifiable so it can participate in reference cycles if people try hard.
2. Classobj wasn't freed in the Limited API.
3. Classobj wasn't visited in the Limited API.

@devdanzin sent me an automated analysis (I think parsing the C code to find potentially suspicious bits and then putting them into AI to try to identify problems, but I could be misrepresenting it) that pointed these out. More fixes to follow based on that...


scoder:
3.2.x commit is https://github.com/cython/cython/commit/9c9501dcd62f30d58c5a35da6ecaee393030c00d

PR #7595, opened 2026-03-27
Remove redundant check in __Pyx_MergeKeywords_any

da-woods:
It's not doing any harm, but it's always true (even if marked unlikely)


scoder:
3.2.x commit is https://github.com/cython/cython/commit/5d1d888c3a26098c993107db6a97d438861a9162

PR #7597, opened 2026-03-28
Fix reference leak in Generator_Replace_StopIteration

da-woods:
`PyErr_SetObject` doesn't steal a reference.

Another thing flagged by @devdanzin's analysis.


da-woods:
3.2.x commit [66846de](https://github.com/cython/cython/commit/66846de4a81587b80738d4b1708927d7f07aa631)

PR #7599, opened 2026-03-28
Fix error handling for !SAFE_MACROS cyfunc.is_coroutine

da-woods:
PyList_SetItem always steals a reference even on error.


da-woods:
3.2.x commit e9242dfdb

PR #7600, opened 2026-03-28
Try to avoid unguarded PyErr_Clear

da-woods:
Another thing raised by @devdanzin's automated analysis...

There's quite a few places where we discard all errors with PyErr_Clear. That's usually close to what's intended, but it also discards "BaseException" which are typically things that we don't want to ignore like MemoryError, KeyboardInterrupt.

I've usually adopted the pattern or "discard regular exception; keep base exception". One exception is `__Pyx_PyDict_GetItemStr`
- this minics the interface for `PyDict_GetItem` so in that case I've just followed their lead and done "write unraisable".

There's a few examples which I haven't addressed here because they looked like they needed a bit more thought:

* MergeVTables
* __Pyx_setup_reduce and __Pyx_setup_reduce_is_named (I have a branch for this https://github.com/da-woods/cython/tree/setup-reduce but it depends on this PR)
* `__Pyx_Py_UNICODE_is_*`
* `__Pyx_PyType_GetFullyQualifiedName` (in this case it should probably be failable, and there's a better option in recent Python versions).

### DIFF
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -614,7 +614,7 @@ __Pyx_CyFunction_get_is_coroutine_value(__pyx_CyFunctionObject *op) {
             return is_coroutine_value;
         }
 ignore_or_error:
-        if (!__Pyx_IgnoreException(NULL, PyExc_Exception)) {
+        if (!__Pyx_IgnoreException(PyExc_Exception)) {
             return NULL;
         }
     }
--- a/Cython/Utility/ModuleSetupCode.c
+++ b/Cython/Utility/ModuleSetupCode.c
@@ -2171,7 +2171,7 @@ static int __Pyx_RegisterCleanup(void) {
         ret = PyList_Insert(reg, 0, args);
     } else {
         if (!reg) {
-            if (!__Pyx_IgnoreException(NULL, PyExc_AttributeError)) {
+            if (!__Pyx_IgnoreException(PyExc_AttributeError)) {
               goto bad;
             }
         } else {
@@ -2372,7 +2372,7 @@ static void __Pyx_FastGilFuncInit(void) {
   struct __Pyx_FastGilVtab* shared = (struct __Pyx_FastGilVtab*)PyCapsule_Import(__Pyx_FastGIL_PyCapsule, 1);
   if (shared) {
     __Pyx_FastGilFuncs = *shared;
-  } else if (__Pyx_IgnoreException(NULL, PyExc_Exception)) {
+  } else if (__Pyx_IgnoreException(PyExc_Exception)) {
     return __Pyx_FastGilFuncInit0();
   }
 }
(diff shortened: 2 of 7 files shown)


Cython/Utility/ModuleSetupCode.c
```diff
@@ -957,7 +957,11 @@ static CYTHON_INLINE void *__Pyx__PyModule_GetState(PyObject *op)
 #define __Pyx_PyDict_GetItemStrWithError(dict, name)  _PyDict_GetItem_KnownHash(dict, name, ((PyASCIIObject *) name)->hash)
 static CYTHON_INLINE PyObject * __Pyx_PyDict_GetItemStr(PyObject *dict, PyObject *name) {
     PyObject *res = __Pyx_PyDict_GetItemStrWithError(dict, name);
-    if (res == NULL) PyErr_Clear();
+    if (res == NULL) {
+        // Like PyDict_GetItem, this is a bit indiscriminate about catching *all* errors.
+        PyErr_WriteUnraisable(NULL);
+        PyErr_Clear();
```
scoder: `PyErr_WriteUnraisable()` clears the exception after printing it.

```suggestion
```

Cython/Utility/CythonFunction.c
```diff
@@ -604,14 +604,20 @@ __Pyx_CyFunction_get_is_coroutine_value(__pyx_CyFunctionObject *op) {
...
             return is_coroutine_value;
         }
-ignore:
-        PyErr_Clear();
+ignore_or_error:
+        if (PyErr_ExceptionMatches(PyExc_Exception)) {
+            PyErr_Clear();
+        } else {
+            return NULL; // BaseException - don't ignore.
+        }
```
scoder: This seems worth moving into a utility function, like `if (!__Pyx_ClearException()) return NULL;`.

Or maybe `if (!__Pyx_IgnoreException(PyExc_Exception)) return NULL;` to cover more use cases. A bit more verbose but might be more explicit in what it does.

da-woods: I  did a slight variation with an optional first argument for the exception. Just because sometimes you can avoid looking up the exception state twice.

Cython/Utility/CythonFunction.c
```diff
@@ -613,10 +614,8 @@ __Pyx_CyFunction_get_is_coroutine_value(__pyx_CyFunctionObject *op) {
             return is_coroutine_value;
         }
 ignore_or_error:
-        if (PyErr_ExceptionMatches(PyExc_Exception)) {
-            PyErr_Clear();
-        } else {
-            return NULL; // BaseException - don't ignore.
+        if (!__Pyx_IgnoreException(NULL, PyExc_Exception)) {
```
scoder: Hmm. It seems annoying to have an opaque `NULL` value as a required argument here. Could you add a macro that hides it? Python adds `Given` (`PyErr_GivenExceptionMatches`) when an exception reference is provided. It seems that the normal case is that we don't have a reference to an exception, so I'd reserve `__Pyx_IgnoreException` for that.

PR #7601, opened 2026-03-28
Improve __Pyx_PyType_GetFullyQualifiedName error handling

da-woods:
1. On newer Limited API we can move all the logic into PyErr_Format by using the `N` format code. This is unlikely to be worse than whatever we do.
2. Allow `__Pyx_PyType_GetFullyQualifiedName` to fail on older Limited API (although only do that if no other options are available or a BaseException is raised).
3. Preserve exceptions inside `__Pyx_PyType_GetFullyQualifiedName`. Since it's usually used in exception handlers so it's quite reasonable that something might already be raised.
4. Don't cache an obscure 3rd-tier string just for a fallback.
5. Adapt other code for the possibility it might fail.

Follow up to #7600

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -2260,9 +2260,11 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
 
         def handle_not_supported(op_name):
             code.putln("__Pyx_TypeName o_type_name = __Pyx_PyType_GetFullyQualifiedName(Py_TYPE(o));")
+            code.putln("if (!__Pyx_Typename_ErrorCheck(o_type_name)) {")
             code.putln("PyErr_Format(PyExc_NotImplementedError,")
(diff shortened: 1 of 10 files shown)


scoder:
There is a lot of opaque macro magic going on around type names by now, and this adds even more to it.

A very large chunk deals with error reporting. Maybe it could help to factor out a few type related dedicated error reporting functions that do the whole "report error on an object's type" (or two object's types) to avoid repeating the same kind of code all over the place? And in most cases, the result is a `TypeError`, that could just be passed in by a C macro alias. I wouldn't even be surprised if that [such a refactoring] had a (slight) effect on the module size.

scoder:
> Maybe it could help to factor out a few type related dedicated error reporting functions

I'll give that a try.

scoder:
See https://github.com/cython/cython/pull/7680

I think I got all of your changes integrated. It ended up with only three utility functions for the exception raising and a couple of singular special cases remaining.

PR #7602, opened 2026-03-28
Allow Py_UCS4 unicode predicates to fail

da-woods:
Outside of Limited API, try to write them in a way that makes it fairly clear to the C compiler that -1 can never happen

In the Limited API, allow them to fail rather than swallowing any errors and returning a false value.

Part of fixes related to #7600


da-woods:
I've cherry-picked this to 3.2.x as f464cb01b.

It was a bit borderline whether it was significant enough to justify that. However, it's very straightforward so I decided to do it. I probably won't do with the other "swallowing `BaseException`" PRs.

PR #7614, opened 2026-04-09
Enable vectorcall on PyPy and GraalPy

da-woods:
I think this should be supported, and would be beneficial if it was.


da-woods:
I haven't tested this locally so it's definitely possible that it fails dramatically. But if it works I think it's worthwhile.

da-woods:
As far as I can tell this is largely performance-neutral on PyPy (e.g. `bm_richards_*` which contain a lot of Python function calls are basically the same). So it may not really be worth it, but does make PyPy and GraalPy more similar to the CPython code path

da-woods:
I've put this as draft pending https://github.com/cython/cython/pull/7616 (since I think it'll conflict). A quick test suggests that enabling vectorcall for both calling and CyFunction makes it a little more worthwhile for PyPy (although probably not a huge win still)

da-woods:
On PyPy 3.11:

```
2026-04-12 08:16:22  ### Benchmark 'bm_richards_cclass' (min/median/max/ ±% of median):
2026-04-12 08:16:22      Cython 'master'           = 388.763 msec, 395.898 msec, 563.004 msec
2026-04-12 08:16:22      Cython 'HEAD'             = 308.405 msec, 325.294 msec, 342.071 msec  (   -17.8 %)
2026-04-12 08:16:22  ### Benchmark 'bm_richards_pyclass' (min/median/max/ ±% of median):
2026-04-12 08:16:22      Cython 'master'           =    1.390 sec,    1.419 sec,    1.512 sec
2026-04-12 08:16:22      Cython 'HEAD'             =    1.202 sec,    1.253 sec,    1.345 sec  (   -11.7 %)
```

I think the rough conclusions is that calling regular functions with vectorcall doesn't make a huge difference on PyPy but calling Cython functions with vectorcall enabled on both ends is worth it.

da-woods:
@Mattip do you have any opinions on this for PyPy

@msimacek do you have any opinions in this for GraalPy?

In both cases it seems to work (i.e. the tests still pass). I did a little bit of profiling for PyPy and it seems like a slight benefit. I failed to do useful profiling for GraalPy (the results just seemed too variable and it was too slow to run generally).

Intuitively it seems like vectorcall shouldn't be worse, and might end up being better (either by avoiding intermediate objects or moving the creation of those objects into code you control). However - it's just a flag for us so I'm happy to set it whichever way I'm told.

msimacek:
Sounds good to enable on GraalPy, it could be a slight improvement due to avoiding the arg tuple.

mattip:
PyPy implements all the flags for vectorcall, but has not optimized it yet. The feeling is that any savings in packing the arguments without a tuple is shadowed by the pyobject - rpython conversions.

Bottom line: more cpython compatibility is a good thing.

da-woods:
Thanks both. I was mostly just checking that there wasn't a good reason why I shouldn't do this...

PR #7616, opened 2026-04-09
Cleanup of vectorcall macros

da-woods:
Essentially, we have 4 and I don't see any justification to have more than 1 any more. They were useful in the past for when vectorcall was gradually introduced, but now it's really just one feature.

The macros removed where:
* `CYTHON_FAST_PYCCALL` - wasn't really used for anything any more; at one point was used for calling into C functions I think.
* `CYTHON_FAST_PYCALL` - only used to turn on `CYTHON_VECTORCALL`.
* `CYTHON_METH_FASTCALL` - current usage was to enable vectorcall on CyFunctions. This is arguably separate from the decision about whether to use vectorcall when calling other functions, but I don't think it's a useful distinction any more.

Arguably this may affect anyone that's purposefully changing these macros to affect Cython code generation.


scoder:
The only question would be about legacy settings on user side. We allowed setting either `CYTHON_FAST_PYCCALL` or `CYTHON_FAST_PYCALL` (or both) with an impact on `CYTHON_VECTORCALL`. If I read the current code correctly, it is possible to set `CYTHON_VECTORCALL` directly, before and with this change, and that seems an easy way for users of the old options to migrate.

I think it's fine to remove the legacy options completely.

da-woods:
My inclination is that we can probably treat all the macros that are hidden by default in https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#c-macro-defines as "semi-private". i.e. we shouldn't worry too much about changing them where it makes sense.

> it is possible to set CYTHON_VECTORCALL directly, before and with this change, and that seems an easy way for users of the old options to migrate.

Yes it should be.  I don't really see why you'd want to but it's there for people who really like creating tuples.

PR #7617, opened 2026-04-09
Introduce `restrict` qualifier

### LINKED ISSUE #7566: [ENH] Introduce `restrict` qualifier
### Describe your issue

When `restrict` type qualifier helps the compiler to vectorise to loop [1]. It is can be useful for projects using Cython. E.g. I was able to rewrite scikit-image _extrema_cy.pyx to use `restrict` keyword (using patched Cython)

Original version:
```cython
cdef inline void _mark_candidates_in_last_dimension(
        dtype_t[::1] image, unsigned char[::1] flags) noexcept nogil:
    cdef Py_ssize_t i, i_ahead
    i = 1
    while i < image.shape[0]:
        if image[i - 1] < image[i] and flags[i] != BORDER_INDEX:
            # Potential maximum (in last dimension) is found, find
            # other edge of current plateau or "edge of dimension"
            i_ahead = i + 1
            while (
                image[i] == image[i_ahead] and
                flags[i_ahead] != BORDER_INDEX
            ):
                i_ahead += 1
            if image[i] > image[i_ahead]:
                # Found local maximum (in one dimension), mark all
                # parts of the plateau as potential maximum
                flags[i:i_ahead] = CANDIDATE
            i = i_ahead
        else:
            i += 1
```

version using `restrict` keyword for which I am dirty PoC:

```cython
cdef inline void _mark_candidates_in_last_dimension(
        dtype_t[::1] image, unsigned char[::1] flags) noexcept nogil:
    cdef Py_ssize_t i, i_ahead
    # Cast data pointers to C pointers with restrict
    cdef restrict dtype_t* im_ptr = <dtype_t*> &image[0]
    cdef restrict unsigned char* fl_ptr = <unsigned char*> &flags[0]

    i = 1
    while i < image.shape[0]:
        if im_ptr[i - 1] < im_ptr[i] and fl_ptr[i] != BORDER_INDEX:
            i_ahead = i + 1
            while im_ptr[i] == im_ptr[i_ahead] and fl_ptr[i_ahead] != BORDER_INDEX:
                i_ahead += 1
            if im_ptr[i] > im_ptr[i_ahead]:
                # mark plateau as candidate
                for j in range(i, i_ahead):
                    fl_ptr[j] = CANDIDATE
            i = i_ahead
        else:
            i += 1
```

Ideal version I would like to have (but can be tricky to implement):

```cython
cdef inline void _mark_candidates_in_last_dimension(
        restrict dtype_t[::1] image, restrict unsigned char[::1] flags) noexcept nogil:
    cdef Py_ssize_t i, i_ahead

    i = 1
    while i < image.shape[0]:
        if image[i - 1] < image[i] and flags[i] != BORDER_INDEX:
            i_ahead = i + 1
            while image[i] == image[i_ahead] and flags[i_ahead] != BORDER_INDEX:
                i_ahead += 1
            if image[i] > image[i_ahead]:
                # mark plateau as candidate
                for j in range(i, i_ahead):
                    flags[j] = CANDIDATE
            i = i_ahead
        else:
            i += 1
```

New version of code gave me ~2% speedup in following test:

```python
In [1]: import skimage as ski
   ...: import numpy as np
   ...: img = np.random.randint(0, 255, (8000, 8000), dtype=np.uint8)
In [2]: %timeit ski.morphology.local_maxima(img)
```

* No restrict

```
519 ms ± 1.49 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
517 ms ± 731 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)
525 ms ± 19 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

* restrict

```
506 ms ± 586 μs per loop (mean ± std. dev. of 7 runs, 1 loop each)
510 ms ± 5.56 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
509 ms ± 1.43 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Downsides of the proposal:
* Using `restrict` can be dangerous. The memory behind `restrict` pointer must be accessed only via given pointer. Otherwise behaviour is undefined, most probably nasty crash.
* When `restrict` is used properly, the compiler uses SIMD instruction for parallel execution. This can lead to worse performance when the loop is looping over smaler dataset
* The syntax is cumbersome for memoryviews. It would be great to introduce syntax sugar `restrict <TYPE>[::1]` which will cause that underlying `.data` field is access via restrict pointer. But not sure how complex it would be to implement it.

The syntax is up to discussion. C language gives restrict keyworkd after `*` so there are more possibilities how to implement it.
```
cdef restrict TYPE *a = ...
cdef TYPE * restrict a = ...
```
The Cython implementation should be fairly limited, in my oppinion we should focus only on limited usages of `restrict` keyword (e.g. only in function parameters)

[1] https://en.cppreference.com/w/c/language/restrict.html
[2] https://github.com/scikit-image/scikit-image/blob/main/src/skimage/morphology/_extrema_cy.pyx

matusvalo:
This PR introduce for support restricted pointers in Cython.

Note: Restricted support of memoryviews described in linked bug are out of scope if this PR.

Partially fixes #7566


Cython/Compiler/Nodes.py
```diff
@@ -923,16 +923,20 @@ def declare_optional_arg_struct(self, func_type, env, fused_cname=None):
         func_type.op_arg_struct = PyrexTypes.c_ptr_type(op_args_struct.type)
 
 
-class CConstDeclaratorNode(CDeclaratorNode):
+class CQualifierDeclaratorNode(CDeclaratorNode):
```
da-woods, line 926: I sort of wonder why this node doesn't handle `volatile` too (apart from that nobody really uses it so people possibly haven't spotted that it doesn't parse where const does)

Cython/Compiler/Nodes.py
```diff
@@ -923,16 +923,20 @@ def declare_optional_arg_struct(self, func_type, env, fused_cname=None):
...
     # base     CDeclaratorNode
+    # is_const      boolean
+    # is_restrict   boolean
 
     child_attrs = ["base"]
 
     def analyse(self, base_type, env, nonempty=0, visibility=None, in_pxd=False):
+        if isinstance(self.base, CFuncDeclaratorNode):
+            error(self.pos, "Restrict qualifier cannot be applied to function type")
```
da-woods: `if isinstance(self.base, CFuncDeclaratorNode) and self.is_restrict:`?

Cython/Compiler/Nodes.py
```diff
@@ -923,16 +923,20 @@ def declare_optional_arg_struct(self, func_type, env, fused_cname=None):
...
         if base_type.is_pyobject:
             error(self.pos,
                   "Const base type cannot be a Python object")
```
da-woods, line 938: Either the error message or the check should be updated?

In principle we could have a `restrict` PyObject but in practice I think we probably shouldn't.

Cython/Compiler/Nodes.py
```diff
@@ -1518,7 +1522,7 @@ def analyse(self, env, could_be_name=False):
         if base.is_pyobject:
             error(self.pos,
                   "Const/volatile base type cannot be a Python object")
```
da-woods, line 1524: Update the error message?

Cython/Compiler/ParseTreeTransforms.py
```diff
@@ -3178,6 +3178,8 @@ def visit_DefNode(self, node):
         return node
 
     def visit_FuncDefNode(self, node):
+        if self.directives.get('target_clones', False):
+            node.modifiers.append('target_clones')
```
da-woods: This feels unneeded (in that `target_clones` isn't used elsewhere in Cython so I don't think this can do anything)

da-woods (commented):
On a quick look through this looks sensible

Cython/Compiler/ParseTreeTransforms.py
matusvalo: Yes this was added by mistake.

Cython/Compiler/Nodes.py
matusvalo: Good catch. Fixed!

Cython/Compiler/Nodes.py
matusvalo: Because there are two Nodes:
* `CConstDeclaratorNode` (`CQualifierDeclaratorNode`) - it handles qualifiers **after** `*` which does not support `volatile`: `int * const bar`
* `CConstOrVolatileTypeNode` (`CQualifierTypeNode`) - it handles qualifiers **before** `*` which supports only `const` and `volatile` no `restrict`: `const int * bar`

da-woods: > `CConstDeclaratorNode` (`CQualifierDeclaratorNode`) - it handles qualifiers **after** `*` which does not support `volatile`: `int * const bar`

I don't think that's right from a C point of view. `int * volatile bar` is just saying "what `bar` points to can change unexpectedly". That might be because it gets changed in a signal handler for example.

I don't think there's any need to change it in this PR though (or possibly ever...)

matusvalo: Hmm yes you are right. `volatile` can be on both sides of pointer - https://en.cppreference.com/w/c/language/volatile.html

> I don't think there's any need to change it in this PR though (or possibly ever...)

I think I agree...

Cython/Compiler/PyrexTypes.py
```diff
@@ -1915,38 +1915,42 @@ def __hash__(self):
...
 
-    def cv_string(self):
+    def cv_string(self, for_display=0):
         cvstring = ""
         if self.is_const:
             cvstring = "const " + cvstring
         if self.is_volatile:
             cvstring = "volatile " + cvstring
+        if self.is_restrict:
+            cvstring = "restrict " if for_display else "CYTHON_RESTRICT " + cvstring
```
scoder, line 1962: Since we're requiring C99, I think we can get rid of the `CYTHON_RESTRICT` macro all together. That removes the need for passing `for_display` into this method.

… OTOH, I wonder if there's any value in allowing users to override `CYTHON_RESTRICT` from outside. That could be a reason to keep it.

scoder (approved):
Looks good to me overall. Not sure what to do with the `CYTHON_RESTRICT` macro, though.

Cython/Compiler/PyrexTypes.py
matusvalo: Is there any real need/motivation for overriding this macro?

matusvalo: Oh, we cannot get rid of `CYTHON_RESTRICT ` since C++ does not support in directly in syntax but compilers does via non-standard keywords - see discussion https://github.com/cython/cython/issues/7566#issuecomment-4058215313

scoder:
I think the error messages should be updated, then this can be merged.

matusvalo:
> I think the error messages should be updated, then this can be merged.

Error messages were updated. I will wait few days and if there won't be other comments I will close this PR. Thank you @da-woods  and @scoder for review!

PR #7618, opened 2026-04-11
Fix marking async gen as not running in athrow_throw

da-woods:
This was another thing marked as dodgy in @devdanzin's analysis. The description is

> When athrow_throw encounters StopAsyncIteration/GeneratorExit in aclose() mode, it doesn't reset ag_running_async=0 or agt_state=CLOSED, unlike the identical path in athrow_send_impl (line 884-885). The generator is permanently stuck in "running_async" state, causing subsequent operations to fail with "asynchronous generator is already running."

This looks correct, but I don't really understand async generators well enough to really know.
Essentially, I've copied the code from https://github.com/python/cpython/commit/7d369d471cf2b067c4d795d70b75201c48b46f5b I've also taken as much as possible to of the tests added by that PR (and confirmed that they do fail without the patch). In both cases I've taken the version in the current main branch of CPython rather than using that PR exactly. Some of the `TestUnawaitedWarnings` tests are disabled because they don't currently work - I think we need other changes from elsewhere.

(So treat this PR suspiciously because it's a bit of a blind copy-paste)

### DIFF
--- a/Cython/Utility/AsyncGen.c
+++ b/Cython/Utility/AsyncGen.c
@@ -916,7 +916,12 @@ __Pyx_async_gen_athrow_throw(__pyx_PyAsyncGenAThrow *o, PyObject *args)
 
     retval = __Pyx_Coroutine_Throw((PyObject*)o->agt_gen, args);
     if (o->agt_args) {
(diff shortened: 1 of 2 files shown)


scoder:
Seems reasonable enough to go in.

PR #7619, opened 2026-04-11
Improve error handling in GetVtable

da-woods:
Distinguish between "not found"-type errors, which are sometimes unimportant and which we may want to ignore, and "other" errors which usually do want to be propagated.

The main difference in behaviour is that `BaseException` (e.g. `MemoryError` etc) will now be propagated out of `__Pyx_MergeVtables`. It also means that if a base class has an attribute `__pyx_vtable__` which isn't a capsule that's now an error that is propagated rather than silently ignored.


Cython/Utility/ImportExport.c
```diff
@@ -686,27 +686,36 @@ static int __Pyx_SetVtable(PyTypeObject *type, void *vtable) {
...
     ptr = PyCapsule_GetPointer(ob, 0);
     if (!ptr && !PyErr_Occurred())
         PyErr_SetString(PyExc_RuntimeError, "invalid vtable found for imported type");
     Py_DECREF(ob);
-    return ptr;
-bad:
+    *table = ptr;
+    return 1;
+no_attr:
+    if (PyErr_ExceptionMatches(PyExc_Exception)) {
```
scoder: This is certainly better than what's there now, but looking at where the exception originates, we _could_ be more specific and only allow `KeyError` and `AttributeError`.

scoder:
I think merging this uncovered a hidden issue with Py3.9 LAPI (3.10 should be fine). It doesn't allow calling `PyType_GetSlot()` on non-heap types and fails with a `SystemError` in `__Pyx_MergeVtables()` when we reach `PyBaseObject` going up the base ladder here:
https://github.com/cython/cython/blob/e131a06096e245cce57fd2f9623ab210ecb69fb1/Cython/Utility/ImportExport.c#L738

We could check for heap types ourselves, only if we're in the Limited API (because otherwise we probably won't use heap types), but I wonder if that's safe and we can make sure that we're not accidentally missing relevant base types, e.g. from external non-abi3 modules.

I also wonder if we can't always skip the last entry in `bases`. That should always be `PyBaseObject`, which never has a vtable.

da-woods:
> We could check for heap types ourselves, only if we're in the Limited API (because otherwise we probably won't use heap types), but I wonder if that's safe and we can make sure that we're not accidentally missing relevant base types, e.g. from external non-abi3 modules.

I think any extern non-heaptype bases will end up being unusable because `tp_new` needs `PyType_GetSlot` too.

Therefore I don't think it's worth doing more than the minimal fix here. I'll do that, and add a better guard inside `tp_new` so people get a more reasonable error message.

da-woods:
I think https://github.com/cython/cython/pull/7763 should fix this as well as is possible.

PR #7621, opened 2026-04-11
Rename _obj_to_string to fit with our general naming scheme

da-woods:


da-woods:
I'll merge this because it's clearly uncontraversial

PR #7624, opened 2026-04-13
Clean up after #7543

matusvalo:


Cython/Compiler/PyrexTypes.py
```diff
@@ -295,6 +295,13 @@ class PyrexType(BaseType):
     is_pybytearray_type = False
     is_pymemoryview_type = False
 
+    # is_bytes_or_str is True if bytes or str
+    is_bytes_or_str = False
+    # is_builting_sequence is True if bytes or str or bytearray or memoryview or tuple or list
+    is_builtin_sequence = False
+    # is_bytes_or_str_or_bytearray is True if bytes or str or bytearray
+    is_bytes_or_str_or_bytearray = False
```
scoder: Hmm, the comments seem mostly redundant. And the sequence comment risks getting forgotten on future updates. I'd remove them all.
```suggestion
    # Combined type group flags, based on common needs in the code base.
    is_bytes_or_str = False
    is_builtin_sequence = False
    is_bytes_or_str_or_bytearray = False
```

Cython/Compiler/PyrexTypes.py
```diff
@@ -295,6 +295,13 @@ class PyrexType(BaseType):
     is_pybytearray_type = False
     is_pymemoryview_type = False
 
+    # is_bytes_or_str is True if bytes or str
+    is_bytes_or_str = False
```
scoder: If I'm not mistaken, this is only used once, in a case where `is_bytes_or_str_or_bytearray` might actually be more appropriate. It can probably be removed, but we'd then want to add a test for the bytearray case in that corner.
https://github.com/cython/cython/blob/a94adc71cf1aa7d77aeae3f021daacfb56a1c804/Cython/Compiler/ExprNodes.py#L191-L194

Cython/Compiler/PyrexTypes.py
```diff
@@ -1514,19 +1506,19 @@ class BuiltinObjectType(PyObjectType):
...
+        'complex': ['is_pycomplex_type'],
+        'list': ['is_pylist_type', 'is_builtin_sequence'],
+        'dict': ['is_pydict_type'],
+        'set': ['is_pyset_type'],
+        'tuple': ['is_pytuple_type', 'is_builtin_sequence'],
+        'frozenset': ['is_pyfrozenset_type'],
+        'bytes': ['is_pybytes_type', 'is_builtin_sequence', 'is_bytes_or_str', 'is_bytes_or_str_or_bytearray'],
+        'str': ['is_pystr_type', 'is_builtin_sequence', 'is_bytes_or_str', 'is_bytes_or_str_or_bytearray'],
+        'bytearray': ['is_pybytearray_type', 'is_builtin_sequence', 'is_bytes_or_str_or_bytearray'],
+        'memoryview': ['is_pymemoryview_type', 'is_builtin_sequence'],
```
scoder, line 1518: To prevent typos, I suggest adding a test that
- checks that all attributes named here exist and default to False on the base type
- validates that the expected types are in the groups (e.g. by reversing the mapping)

That would give us the safety that a test needs to be changed explicitly if we make changes to the groups.

matusvalo:
I saw that in the code there is often type test based on type name. I simplify it to use new attributes...

Cython/Compiler/PyrexTypes.py
matusvalo: Fixed.

Cython/Compiler/PyrexTypes.py
matusvalo: After some investigation, I decided to simply remove `.is_bytes_or_str` and keep the logic of the code as is.

matusvalo:
Should be ready for next round of review.

Cython/Compiler/Tests/TestTypes.py
```diff
@@ -16,6 +18,41 @@ def assert_widest(type1, type2, widest):
...
+
+    def test_set_builtin_type_flags(self):
+        builtin_flags = set(chain.from_iterable(PT.BuiltinObjectType._builtin_type_flag_mapping.values()))
+        builtin_types = set(chain.from_iterable(self.BUILTIN_FLAGS_MAPPING.values()))
+
+        self.assertSetEqual(builtin_flags, set(self.BUILTIN_FLAGS_MAPPING))
+        self.assertSetEqual(set(PT.BuiltinObjectType._builtin_type_flag_mapping), builtin_types)
+
+        for attr in builtin_flags:
+            self.assertFalse(getattr(PT.PyrexType, attr))
```
scoder:
```suggestion
            self.assertIs(getattr(PT.PyrexType, attr), False)
```

Cython/Compiler/Tests/TestTypes.py
```diff
@@ -16,6 +18,41 @@ def assert_widest(type1, type2, widest):
...
+            for type_name in self.BUILTIN_FLAGS_MAPPING[attr]:
+                self.assertTrue(
+                    getattr(PT.BuiltinObjectType(type_name, f'c_{type_name}'), attr),
+                    f"{attr} should be set for {type_name}"
+                )
```
scoder:
```suggestion
                self.assertIs(
                    getattr(PT.BuiltinObjectType(type_name, f'c_{type_name}'), attr),
                    True,
                    f"{attr} should be set for {type_name}"
                )
```

scoder (approved):

Issue #5892, opened 2023-12-04
[BUG] Link error with final class


skunkyevil:
### Describe the bug

I get link error ```error C2129: static function "PyObject __pyx_f_11final_class_10FinalClass_method(__pyx_obj_11final_class_FinalClass *)" is declared but not defined```

If I remove @cython.final - everything is OK

### Code to reproduce the behaviour:

setup.py:
```
from setuptools import setup
from Cython.Build import cythonize

setup(
    name='Hello',
    ext_modules=cythonize('*.pyx', language_level = 3)
)

```

final_class.pyx:
```
# distutils: language = c++
# distutils: define_macros=NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION

cimport cython


@cython.final
cdef class FinalClass:
    def __init__(self):
        pass

    cdef method(self):
        pass 
```

main.pyx:
```
# distutils: language = c++
# distutils: define_macros=NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION

from final_class cimport FinalClass


def func():
    cdef FinalClass final_class = FinalClass()
    final_class.method() 
```

### Expected behaviour

It's expected to get compiled and linked

### OS

Windows 10

### Python version

3.9.7

### Cython version

3.0.6

PR #7626, opened 2026-04-15
Auto cpdef extension to class methods

jairov4:
## Summary

This PR extends the `auto_cpdef` feature to automatically promote eligible methods in `@cclass`-decorated classes to `cpdef` functions. 
This enables big performance wins with C-level optimization of large codebases with minimal changes, while still allowing selective exclusion of methods that are incompatible or error-prone.

### How it works

- **Automatic method promotion**: Methods in classes with `auto_cpdef=True` directive are automatically promoted to `cpdef` when compatible, including final classes where optimization is most effective
- **Fine-grained control**: New directives `no_ccall` allow excluding specific methods from promotion

This brings the same optimization benefits previously available only for module-level functions to instance methods within `@cclass`-decorated classes.

### Test Cases:

New test file `tests/run/auto_cpdef_classes.py` validates:
- Method promotion in base classes


da-woods:
Unless I'm missing something, I don't think the use of `public` is right here.

When applied to a class or a module-level function it means "put it in a header file". This is pretty unrelated to `auto_cpdef` and I don't think should be linked.

When applied to a class it also requires the objstruct name to be specified (e.g. `cdef public class C[object C_Obj]`) so just decorating a class with `@cython.public` doesn't make sense.

`public` applied to a `cdef` attribute of a class creates property accessors for it. That doesn't apply to methods of a class though.

--------------------------------------

It probably would be nice to be able to specify `public` in pure-Python mode, but I don't think it should be tied to auto_cpdef and it doesn't quite work like you think.

Cython/Compiler/Nodes.py
```diff
@@ -2791,7 +2793,7 @@ def analyse_declarations(self, env):
         self.args = declarator.args
 
         opt_arg_count = self.cfunc_declarator.optional_arg_count
-        if (self.visibility == 'public' or self.api) and opt_arg_count:
+        if (self.visibility == 'public' or self.api) and opt_arg_count and not self.promoted:
             error(self.cfunc_declarator.pos,
                   "Function with optional arguments may not be declared public or api")
```
da-woods, line 2798: I also think this error message is something that *can't* be ignored for `public` functions. If it has optional arguments then it doesn't have a simple C signature that can be put in a header file.

Which should be fine because `auto_cpdef` shouldn't require public.

jairov-apptio: Hi, I have updated and it do not require public anymore

jairov-apptio:
@da-woods Please review again

jairov4:
I noticed that this also fixes #5892

jairov4:
Fixes #7692
It was required to make the optimization of super() works cross module

da-woods:
I'll have a more detailed look at the weekend when I have more time.

However, I remain suspicious of the changes to visibility.

I also think that the fixes for https://github.com/cython/cython/issues/7692 are likely small and uncontroversial and would likely be better as part of a separate PR rather than this one which seems like a large combination of different changes.

tests/run/cimport_from_pyx_final.srctree
```diff
@@ -0,0 +1,40 @@
+PYTHON setup.py build_ext --inplace
+PYTHON -c "import b; import a"
```
da-woods: This is the fundamental, massive problem with the proposed change.

If I change this to just `import a` I get:

```
undefined symbol: __pyx_f_1b_1B_m
```

`public` (for classes and functions) is not how we shared things between Cython modules. `public` is a mechanism for people writing their own C code who want to be able to call functions defined in Cython or to use classes defined in Cython.  It puts everything onto the dll-loader/operating system, and so it requires the symbols used by `a` to be available at the point you import `a`.

The mechanism for sharing things between Cython modules is to use the `cimport` mechanism. In this case Cython controls the loading of symbols so can avoid strict restrictions on the import order.

tests/run/auto_cpdef_classes.py
```diff
@@ -0,0 +1,116 @@
...
+    True
+    >>> test_vector(Vector(2, 3).__eq__(Vector(4, 5)))  # special method semantics preserved
+    False
+    >>> test_vector(Vector(2, 3).__eq__(Vector(2, 3)))
+    True
+    """
+    return v
+
+
+@public
```
da-woods: In pyx mode, `cdef public class` declarations look like

https://github.com/cython/cython/blob/fe937cc07680cdd75714b04be3a1ad965ce3c03f/tests/compile/a_capi.pyx#L11-L12

Specifying the cname of the object and the type is not optional. This is because they are designed for people writing their own C code who will want to use the generated header files and who will want to know what the name is.

It is not acceptable that `public` works differently in pyx mode than in does in py mode.

Cython/Compiler/ParseTreeTransforms.py
```diff
@@ -1,4 +1,6 @@
 import cython
+from .. import auto_cpdef
```
da-woods: ?

da-woods (commented):
As I've said before: there's too much going on here.

The changes to `auto_cpdef` to cover extension types are probably useful.

The changes to `public` are a mostly broken way of sharing things between modules that contradicts the existing behaviour and doesn't use Cython's existing mechanism for sharing.  They aren't mergeable now and are unlikely to become mergeable.

tests/run/cimport_from_pyx_final.srctree
jairov4: @da-woods  Thanks for taking the time of reviewing this change, I will remove the public directive, I added it because I suspected it was required to make auto_cpdef work for class methods since the simpler surgical change was suffering of methods not being public when they got optimized. Then, I realize it is not required due the imported method pointers approach at module level.

tests/run/auto_cpdef_classes.py
jairov4: right, as I said I realize it is not required but didnt change the test.
I will remove this directive.

jairov4:
@da-woods I have removed all the non-related code and also removed everything related to change the visibility.

Cython/Compiler/Nodes.py
```diff
@@ -3317,6 +3320,27 @@ def is_cdef_func_compatible(self):
...
+        is_property = False
+        from . import ExprNodes
+        if self.decorators:
+            for decorator in self.decorators:
+                func = decorator.decorator
+                if func.is_name:
+                    self.is_classmethod |= func.name == 'classmethod'
+                    self.is_staticmethod |= func.name == 'staticmethod'
+                    is_property |= func.name == 'property'
+                elif isinstance(func, ExprNodes.AttributeNode):
```
da-woods, line 3337:
```suggestion
                elif func.is_attribute:
```

Cython/Compiler/Nodes.py
```diff
@@ -3317,6 +3320,27 @@ def is_cdef_func_compatible(self):
...
             return False
+        if self.name.startswith('__') and self.name.endswith('__'):
+            return False
+        if self.num_required_args != len(self.args):
+            return False
+
+        is_property = False
+        from . import ExprNodes
+        if self.decorators:
+            for decorator in self.decorators:
```
da-woods, line 3331: I'd assume most decorators should make it incompatible with being transformed.

I'm not sure what wouldn't.

Cython/Compiler/Nodes.py
```diff
@@ -3317,6 +3320,27 @@ def is_cdef_func_compatible(self):
...
+                func = decorator.decorator
+                if func.is_name:
+                    self.is_classmethod |= func.name == 'classmethod'
+                    self.is_staticmethod |= func.name == 'staticmethod'
```
da-woods, line 3335: I'd probably suggest not modifying `self` in this function.

Cython/Compiler/Pipeline.py
```diff
@@ -199,10 +199,9 @@ def create_pipeline(context, mode, exclude_classes=()):
         InterpretCompilerDirectives(context, context.compiler_directives),
         ParallelRangeTransform(context),
         WithTransform(),
-        AdjustDefByDirectives(context),
-        _align_function_definitions,
         MarkClosureVisitor(context),
-        AutoCpdefFunctionDefinitions(context),
+        _align_function_definitions,
+        AdjustDefByDirectives(context),
```
da-woods, line 204: I'm a little wary of this change. It seems like it at least has the potential to change how compiler directives and pxd files work together.

(i.e. I'd rather have the least invasive version of this overall change)

Cython/Compiler/ModuleNode.py
```diff
@@ -554,10 +554,20 @@ def generate_cclass_include_code(self, type, i_code):
             type.module_name, type.name))
         i_code.indent()
         var_entries = type.scope.var_entries
-        if var_entries:
+        method_entries = [
```
da-woods, line 557: Back in 2022 I was apparently in favour of removing generate pxi completely: https://github.com/cython/cython/issues/5047. I think I still agree with my past self here. It certainly doesn't seem part of auto_cpdef

Cython/Compiler/PyrexTypes.py
```diff
@@ -3520,7 +3520,7 @@ def declaration_code(self, entity_code,
             cc = ''
         else:
             cc = self.calling_convention_prefix()
-            if (not entity_code and cc) or entity_code.startswith("*"):
+            if (not entity_code and cc) or (entity_code and entity_code.startswith("*")):
```
da-woods, line 3523: This feels like it might be related to "generate_pxi"?

Cython/Compiler/Options.py
```diff
@@ -362,6 +362,7 @@ class DEFER_ANALYSIS_OF_ARGUMENTS:
     'binding' : bool,
     'cfunc' : None,  # decorators do not take directive value
     'ccall' : None,
+    'no_ccall': None,
```
da-woods, line 365: Do we need a separate directive here? Would `auto_cpdef(False)` work as a local directive?

da-woods (commented):
I'm a bit confused what this is actually doing.

On current master, if I take a very simple file:

```
# cython: auto_cpdef=True

import cython

@cython.cclass
class C:
    def f(self):
        pass

def tst(c: C):
    c.f()
```

then the call to `c.f()` becomes

```
  __pyx_t_1 = ((struct __pyx_vtabstruct_4acpd_C *)__pyx_v_c->__pyx_vtab)->f(__pyx_v_c, 0); if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 11, __pyx_L1_error)
```

To me that looks like it's already being transformed into a `cpdef` function. Which means the PR description doesn't really describe what the *actual* changes being made are.

PR #7629, opened 2026-04-21
Relax libc_math sin doctest

msimacek:
We're running Cython's tests in GraalPy's CI and this test fails on darwin-aarch64 because on that platform `java.Math.sin` (and even `java.StrictMath.sin`)  is one bit off from `libm`'s `sin`. From what I've gathered, these tiny platform differences are normal, so I've modified the test to be a bit (pun intended) more permissive.

### DIFF
--- a/tests/run/libc_math.pyx
+++ b/tests/run/libc_math.pyx
@@ -30,9 +30,12 @@ def test_sin(x):
     """
     >>> test_sin(0)
     0.0
-    >>> from math import sin
-    >>> [sin(k) == test_sin(k) for k in range(10)]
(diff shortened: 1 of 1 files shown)


scoder:
It's a "works and does something reasonable" kind of test rather than a mathematics test, so I think we can be even more relaxed here and use a fixed tolerance. The ulp-values here seem to be in the order of `1e-16`, so I guess something like `1e-12` would be fine?

msimacek:
Done

scoder:
3.2.x commit: https://github.com/cython/cython/commit/5bcdc8ec1a4238dc5674c942d9e69ea26a681240

PR #7630, opened 2026-04-21
Implement some form of stdatomics for C

Vizonex:
I wanted to try implementing c11-std-atomics into cython since free-threading is going to become extremely common in the future and users likewise should have some way of easily access to atomic apis in C and not just in C++. 

My other suggestion or workaround would be to make a header file similar to pyatomics but for Cython that could have unified support for gcc, msvc intrinsics, and std.

> [!WARNING]
> I have no idea if fused templating is going to work or not but it was worth a shot so for now this pr is a draft.

### DIFF
--- a/Cython/Includes/libc/stdatomic.pxd
+++ b/Cython/Includes/libc/stdatomic.pxd
@@ -26,43 +26,43 @@ cdef extern from "<stdatomic.h>" nogil:
     int ATOMIC_LLONG_LOCK_FREE
     int ATOMIC_POINTER_LOCK_FREE
 
-    ctypedef bint atomic_bool "_Atomic(_Bool)"
-    ctypedef char atomic_char "_Atomic(char)"
-    ctypedef signed char atomic_schar "_Atomic(signed char)"
-    ctypedef unsigned char atomic_uchar "_Atomic(unsigned char)"
-    ctypedef short atomic_short "_Atomic(short)"
-    ctypedef unsigned short atomic_ushort "_Atomic(unsigned short)"
-    ctypedef int atomic_int "_Atomic(int)"
-    ctypedef unsigned int atomic_uint "_Atomic(unsigned int)
-    ctypedef long atomic_long "_Atomic(long)"
-    ctypedef unsigned long atomic_ulong "_Atomic(unsigned long)"
-    ctypedef long long atomic_llong "_Atomic(long long)"
-    ctypedef unsigned long long atomic_ullong "_Atomic(unsigned long long)"
-    ctypedef char16_t atomic_char16_t "_Atomic(char16_t)"
-    ctypedef char32_t atomic_char32_t "_Atomic(char32_t)"
-    ctypedef wchar_t atomic_wchar_t "_Atomic(wchar_t)"
-    ctypedef char atomic_int_least8_t "_Atomic(int_least8_t)" 
-    ctypedef unsigned char atomic_uint_least8_t "_Atomic(uint_least8_t)"
-    ctypedef short atomic_int_least16_t "_Atomic(int_least16_t)"
-    ctypedef unsigned short atomic_uint_least16_t "_Atomic(uint_least16_t)"
-    ctypedef int atomic_int_least32_t "_Atomic(int_least32_t)"
-    ctypedef unsigned int atomic_uint_least32_t "_Atomic(uint_least32_t)"
-    ctypedef long long atomic_int_least64_t "_Atomic(int_least64_t)"
-    ctypedef unsigned long long atomic_uint_least64_t "_Atomic(uint_least64_t)"
-    ctypedef char atomic_int_fast8_t "_Atomic(int_fast8_t)" 
-    ctypedef unsigned char atomic_uint_fast8_t "_Atomic(uint_fast8_t)"
-    ctypedef short atomic_int_fast16_t "_Atomic(int_fast16_t)"
-    ctypedef unsigned short atomic_uint_fast16_t "_Atomic(uint_fast16_t)"
-    ctypedef int atomic_int_fast32_t "_Atomic(int_fast32_t)" 
-    ctypedef unsigned int atomic_uint_fast32_t "_Atomic(uint_fast32_t)" 
-    ctypedef long long atomic_int_fast64_t "_Atomic(int_fast64_t)" 
-    ctypedef unsigned long long atomic_uint_fast64_t "_Atomic(uint_fast64_t)" 
-    ctypedef intptr_t atomic_intptr_t "_Atomic(intptr_t)"
-    ctypedef uintptr_t atomic_uintptr_t "_Atomic(uintptr_t)"
-    ctypedef size_t atomic_size_t "_Atomic(size_t)" 
-    ctypedef ptrdiff_t atomic_ptrdiff_t "_Atomic(ptrdiff_t)"
-    ctypedef intmax_t atomic_intmax_t "_Atomic(intmax_t)"
-    ctypedef uintmax_t atomic_uintmax_t "_Atomic(uintmax_t)" 
+    ctypedef bint atomic_bool "atomic_bool"
+    ctypedef char atomic_char "atomic_char"
+    ctypedef signed char atomic_schar "atomic_schar"
+    ctypedef unsigned char atomic_uchar "atomic_uchar"
+    ctypedef short atomic_short "atomic_short"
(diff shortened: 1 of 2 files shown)


da-woods:
I don't think that fused types are going to work here unfortunately.  We don't really have a good way of handling C generic macros.

For most of the functions you can change the first (atomic) argument to `volatile void*` (possibly skipping the volatile). That will tell Cython that it accepts any pointer, which is probably good enough for here.  You might just be able to replace the numeric arguments with `double` - i.e. this is good enough for Cython to know that it accepts a number.

The return type (for `load` and `exchange`) is a bit awkward, because Cython uses it for type inference if you don't explicitly specify all the types. I don't think there's a great solution here. Probably `double` (or `int64_t`?) with a big warning not to rely on it for type inference is still the best option.  I think all we can really do is generate valid-looking code and let the C compiler deal with the problem.

Cython/Includes/libc/stdatomic.pxd
```diff
@@ -0,0 +1,213 @@
...
+    int ATOMIC_CHAR16_T_LOCK_FREE
+    int ATOMIC_CHAR32_T_LOCK_FREE
+    int ATOMIC_WCHAR_T_LOCK_FREE
+    int ATOMIC_SHORT_LOCK_FREE
+    int ATOMIC_INT_LOCK_FREE
+    int ATOMIC_LONG_LOCK_FREE
+    int ATOMIC_LLONG_LOCK_FREE
+    int ATOMIC_POINTER_LOCK_FREE
+
+    ctypedef bint atomic_bool "_Atomic(_Bool)"
```
da-woods: I don't think you really need the `"_Atomic(_Bool)"` cnames here. Since the definitions for `atomic_bool` are already in the stdatomic header

da-woods:
This other thing this could use is some tests to show that it at least compiles sensibly (e.g. like https://github.com/cython/cython/blob/master/tests/run/cpp_stl_atomic.pyx for the C++ ones). We aren't looking to test the C standard library - just the idea that it's usable from Cython.

scoder:
> You might just be able to replace the numeric arguments with `double` - i.e. this is good enough for Cython to know that it accepts a number.

That's risky, though. There are cases where we move function call arguments into temporary variables to assure their expected ordering of evaluation. It'd be subtly incorrect and very unexpected if we ended up changing integers to floats here. And as @da-woods noted, it won't solve the return type problem.

We might end up having to declare different function names per type with the same explicit `"cname"`. The combinations can be generated from a script without much hassle, to make sure we don't forget any. That'd be a bit annoying to use, but I don't currently know a good way to deal with this otherwise.

(EDIT: It would get in the way of calling these functions from within fused typed code, though, which would then still require a separate declaration, probably on user side.)

da-woods:
> There are cases where we move function call arguments into temporary variables to assure their expected ordering of evaluation.

I think the big problem is likely to be any coercion from `object`, which will definitely be coerced to the wrong type.

> but I don't currently know a good way to deal with this otherwise.

For the function arguments you could declare it as C variadic arguments (`...`) which stops Cython thinking about it at all, and just prevents you passing Python objects. That's probably a better solution. 

> We might end up having to declare different function names per type with the same explicit "cname". 

I think that might well be worth doing for the functions with a generic return type. I think it'd be enough to have an integer version and a floating version, but not to worry much about the width of what's returned.

scoder:
> I think it'd be enough to have an integer version and a floating version, but not to worry much about the width of what's returned.

We'd then still run into the temporary variable problem, both for coercion back to Python and when using the return value in an expression. The return types are really the biggest issue here.

I wonder if we should allow "type in, type out" style generic declarations also for C, which would be helpful for both compiler builtins and macros, like
```cython
cdef extern from "somewhere.h":
    T cmacro[T](int x, T v)
```
But I guess that's not for _right_ now… and it comes with the caveat that the example above is very different from a similar looking (IIRC) C declaration of an array of function pointers like below…
```cython
T (cfunc[123])(int x, T v)
```

For the time being, a declaration list of separate Cython names, one per type, all cname-mapping to the same C function, definitely seems like the easiest way out.

da-woods:
> We'd then still run into the temporary variable problem, both for coercion back to Python and when using the return value in an expression. The return types are really the biggest issue here.

Provided you used sufficiently wide types I don't think that matters - you just end up with slightly more storage than you need. You'd probably also need to distinguish between signed and unsigned ints too though.

There's a lot of corner cases here that it's easy to miss something though.

> I wonder if we should allow "type in, type out" style generic declarations also for C, which would be helpful for both compiler builtins and macros [...]

I think the existing C++ ones end up explicitly calling the exact template. So a call would look like `cmacro<int>(args)` which wouldn't be valid C. Otherwise the just relaxing it to allow some of the existing C++ would *almost* do.

Vizonex:
I knew this would be a problem going into stdatomics which is why I am glad I made this a draft. 

> We might end up having to declare different function names per type with the same explicit `"cname"`. The combinations can be generated from a script without much hassle, to make sure we don't forget any. That'd be a bit annoying to use, but I don't currently know a good way to deal with this otherwise.

I actually wouldn't mind seeing a new directive to get around these problems and on windows `intrin.h` would be a good workaround for any older compilers. When we talk about cnames my best assumption would be this, I'll use frozenlist as an example here since I am also a maintainer of some of aiohttp's tools and dependencies.

```cython
cdef class FrozenList:
    atomic bint frozen
```

Or maybe this?

```cython
cdef class FrozenList:
    atomic<bint> frozen
```

We would just need to add in a few new cython nodes, I remember using cython nodes before so this shouldn't be too tricky to relearn.

Vizonex:
@scoder 
> ```cython
> cdef extern from "somewhere.h":
>     T cmacro[T](int x, T v)
> ```
The idea for enabling different C types using generic types could be considered an acceptable addition to cython. I have pretty good example involving the use of a random number generator such as romu-random. Let's just say we hypothetically need the RTOL macro to accept both 32 or 64 integers.
```cython
cdef extern from *:
    """
/* this example comes from romu-random's source code and
is here to give an example of needing to accept different 
C types such as a 32 or 64 bit integer. */
#define ROTL(d,lrot) ((d<<(lrot)) | (d>>(8*sizeof(d)-(lrot))))
    """
    A ROTL[A, B](d A, lrot B)
  ```

**Comment by da-woods** [comment]
> ```
> cdef class FrozenList:
>    atomic bint frozen
> ```

Unless there's a real need, I think I'd rather add a fixed set of atomic typedefs rather than add `atomic` as a whole as a new type modifier. Just because it's a fairly obscure feature and doesn't really translate particularly well to C++ code.

C generic types could definitely be useful. I think we'd want to keep them as close as possible to the existing C++ template syntax.

**Comment by Vizonex** [comment]
@da-woods I'll see about creating a separate PR for C Macro templates it shouldn't be too difficult for me to figure out since I know the Cython Node structure pretty well. 

Edit: Ok I need help, I can't find where in the code generator I need to change is, I'm looking for something like `<...>` So I can replace it with parenthesis `(...)` if the compiler is compiling to C.

```cython
# cython: language = c 
from libc.stdint cimport uint64_t

cdef extern from *:
    """
#define macro_test(t) t + 1
    """
    T macro_test[T](T t)


def test_macro(uint64_t t):
    return macro_test(t)
```

```c
 /* "x.pyx":12
 * 
 * def test_macro(uint64_t t):
 *     return macro_test(t)             # <<<<<<<<<<<<<<
 * 
*/
  __Pyx_XDECREF(__pyx_r);                        /* vvvvvvvvv */
  __pyx_t_1 = __Pyx_PyLong_From_uint64_t(macro_test<uint64_t>(__pyx_v_t)); if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 12, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_1);
  __pyx_r = __pyx_t_1;
  __pyx_t_1 = 0;
  goto __pyx_L0;
  ```

scoder:
> I need help, I can't find where in the code generator I need to change is

I think you end up with C++ template types here, mostly handled by the `CppClassType` in `PyrexTypes.py`.

We need to prevent them from getting into the code generation phase all together. In non-C++ mode, they must be replaced by the appropriate C type in the `infer_type` and `analyse_types` phases (read: methods). (EDIT: That's when calling such a function in `SimpleCallNode` in `ExprNodes.py`.)

There might(!) also be things to do at the `analyse_declarations` phase already, can't say yet.

And I guess we'd also have to prevent code from taking a function pointer of such a template "function", which cannot really be a function in C.

PR #7631, opened 2026-04-22
Fix dict iteration with CYTHON_AVOID_BORROWED_REFS

msimacek:
In the dict iteration code, some of the code paths used `CYTHON_AVOID_BORROWED_REFS` and some used `CYTHON_COMPILING_IN_PYPY`, resulting in broken state on GraalPy where one part of the exact-dict fast-path was active and one was not. It would end up calling `PyIter_Next` with a dict object instead of an iterator.

The existing test `iterdict` tests this AFAICS. Unfortunately, it's currently disabled on GraalPy. I had to fix one more thing on our side to make it pass. So I'm not adding any new test right now with the plan that after we release, we can enable the existing test.

### DIFF
--- a/Cython/Utility/Optimize.c
+++ b/Cython/Utility/Optimize.c
@@ -314,7 +314,7 @@ static CYTHON_INLINE int __Pyx_dict_iter_next(PyObject* dict_or_iter, Py_ssize_t
 //@requires: ObjectHandling.c::IterFinish
 //@requires: ObjectHandling.c::PyObjectCallMethod0
 
-#if CYTHON_COMPILING_IN_PYPY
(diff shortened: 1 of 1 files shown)


da-woods:
That looks right to me. I'm fairly sure the failures are unrelated (probably just changes in CPython `main`) but I'd like to hold off on merging until that's a bit better investigated.

msimacek:
I'm just exchaning one macro for another and on CPython, both of the macros are defined the same, so it can't be this PR. It seems it's just that CPython changed when they emit tracing events a little bit and that fails the tracing test.

da-woods:
3.2.x commit 0b0a8ddf6

PR #7632, opened 2026-04-22
Fix __Pyx__Coroutine_get_frame critical section

da-woods:
See https://groups.google.com/g/cython-users/c/Z-qVaNk8wRI.

get_frame could return early and thus forget to close the critical section.

Also, change to using our internal macros rather than guarding them.


da-woods:
3.2.x commit

PR #7634, opened 2026-04-24
Use fixed relative dstdir instead of the unpredictable tmpdir for sharedmodule

ZhixiongChi:
When the option "--generated-shared" is used to generated __cyutility as the
link https://github.com/scikit-learn/scikit-learn/pull/31151/files, the path
for filename_table in the generated pyx/c file contains tmp dir which is not
predictable though it has been updated to the relative path, and it caused
the generated output file is not stable at each build and made the generated
library is not reproducible [1] between builds.

example as python3_pandas:

vim build/_cyutility.c
......
/* #### Code section: filename_table ### */

static const char* const __pyx_f[] = {
  "../../../../../../../../../../../../tmp/tmpXXXXXX/_cyutility.pyx",
  "<stringsource>",
};

Replace with the fixed path based dest directory provided by the option
"--generated-shard", and don't use the tmpdir anymore
.
After applied this commit, vim build/_cyutility.c
 ......
/* #### Code section: filename_table ### */

static const char* const __pyx_f[] = {
  "shared_buildpath/_cyutility.pyx",
  "<stringsource>",
};

### DIFF
--- a/Cython/Build/SharedModule.py
+++ b/Cython/Build/SharedModule.py
@@ -71,23 +71,27 @@ def generate_shared_module(options):
     Errors.reset()
 
     dest_c_file = options.shared_c_file_path
+    dest_c_dir = os.path.dirname(dest_c_file)
+    dest_tmp_buildpath = os.path.join(dest_c_dir, "shared_buildpath")
     module_name = os.path.splitext(os.path.basename(dest_c_file))[0]
 
     context = Main.Context.from_options(options)
     scope = Symtab.ModuleScope('MemoryView', parent_module = None, context = context, is_package=False)
 
-    with tempfile.TemporaryDirectory() as tmpdirname:
-        pyx_file = os.path.join(tmpdirname, f'{module_name}.pyx')
-        c_file = os.path.join(tmpdirname, f'{module_name}.c')
-        with open(pyx_file, 'w'):
-            pass
-        source_desc = FileSourceDescriptor(pyx_file)
-        comp_src = Main.CompilationSource(source_desc, EncodedString(module_name), os.getcwd())
-        result = Main.create_default_resultobj(comp_src, options)
+    os.makedirs(dest_tmp_buildpath, exist_ok=True)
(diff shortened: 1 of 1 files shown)


da-woods:
I'm not convinced by this as a solution.  Using a fixed symlink to an randomly generated temp path seems like a complicated solution that loses the advantages of the temp path.  I do understand the desire for reproducible builds, but this doesn't seem like the way to do it.

rossburton:
A hardcoded fixed path is going to break anything where two builds are done at the same time, so yeah that isn't realistic.

ZhixiongChi:
How about using the tmpdir under destination directory provided by the caller with the option '--generated-shared' instead of the original temp directory in the host, finally the fixed relative path will be generated in the shared module code.
Meanwhile the parallel building is not a problem anymore, because the real path is decided by the builddir of the other modules who call this shared module function.

da-woods:
@matusvalo Any thoughts here?

My thoughts:
* it's a real issue and that building in the temp directory isn't a great idea.
* Using the fixed `"shared_buildpath"` within seems reasonable. But: there's some possibility that it conflicts with a user-chosen name.
* Could we add some predictable collision avoidance? i.e. if `"shared_buildpath"` already exists then try `"shared_buildpath_0"` or something like that?

matusvalo:
I think I agree. Current approach using shared module not ideal. But on the other hand I don't like much approach using new "special" directory. I also don't like that we show .pyx file which is just "byproduct" of .c file generation.

I created PR #7723 which uses alternative approach. PR uses in-memory .pyx file instead. This simplifies the code and ensures that resulting `__pyx_f` is constant across the builds. Let me know what do you think.

da-woods:
I think "in memory" is probably the best solution if it works.

da-woods:
We've gone for https://github.com/cython/cython/pull/7723 instead so I'll close this. 

Thanks for pointing out the issue though - it's definitely the sort of thing we'd like to fix.

PR #7635, opened 2026-04-24
Small cleanup of memoryview assertion

da-woods:
Cast via `void*` was unnecessary and caused a warning about casting away atomicness.

C-division directive avoids a divide by 0 check. I'm sure the C compiler would trivially remove it, but it seems worth doing while changing ths anyway.

### DIFF
--- a/Cython/Utility/MemoryView.pyx
+++ b/Cython/Utility/MemoryView.pyx
@@ -385,7 +385,8 @@ cdef class memoryview:
         else:
             self.dtype_is_object = dtype_is_object
 
-        assert <uintptr_t><void*>(&self.acquisition_count) % sizeof(__pyx_atomic_int_type) == 0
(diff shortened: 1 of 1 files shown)


scoder:
Out of curiosity (and not questioning this PR), I looked up if `sizeof()` can actually ever return 0 (which seems generally undesired), and it turns out that there are such cases, although by C compiler extension only:

https://stackoverflow.com/questions/2632021/can-sizeof-return-0-zero

Thus, while `sizeof() != 0` seems a tempting assumption, users could still trick us into making it wrong for the C compiler. Whether it's then worth finding safe cases and treating them specially, I can't really say, but I doubt it.

scoder:
Slightly adapted 3.2.x commit is https://github.com/cython/cython/commit/a4bae70c15eddc81387535a08fa57beab6b87e33

PR #7637, opened 2026-04-24
Remove static PyObject* from PyPy/GraalPy __Pyx_dict_iterator

da-woods:
My reasoning is: with module-state / freethreading, static storage in functions makes me increasingly unhappy. Neither of those apply to PyPy or GraalPy yet so it's more future-proofing. I also think the number of people using the Python 2 iteritems/iterkeys/itervalues is increasingly low, so I don't really care about caching a string for their speed.

Also refactor the string comparison a bit to try to skip the common chunks.

### DIFF
--- a/Cython/Compiler/Optimize.py
+++ b/Cython/Compiler/Optimize.py
@@ -1089,6 +1089,17 @@ class IterationTransform(Visitor.EnvTransform):
         temps.append(temp)
         pos_temp = temp.ref(node.pos)
 
+        legacy_method = False
+        if method and method.startswith('iter'):
+            assert method in ('iterkeys', 'itervalues', 'iteritems'), method
+            if dict_obj.type.is_pydict_type:
+                method = EncodedString(method[4:])
+            elif dict_obj.type.is_pyfrozendict_type:
+                # Don't apply legacy Python 2 behaviour to Python 3.15 types
+                return node
+            else:
+                legacy_method = True
+
         key_target = value_target = tuple_target = None
         if keys and values:
             if node.target.is_sequence_constructor:
@@ -1131,9 +1142,6 @@ class IterationTransform(Visitor.EnvTransform):
         body.stats[0:0] = [iter_next_node]
 
         if method:
-            if dict_obj.type.is_pyanydict_type and method.startswith('iter'):
-                assert method in ('iterkeys', 'itervalues', 'iteritems'), method
-                method = EncodedString(method[4:])
             method_node = ExprNodes.IdentifierStringNode(dict_obj.pos, value=method)
             dict_obj = dict_obj.as_none_safe_node(
                 "'NoneType' object has no attribute '%{}s'".format('.30' if len(method) <= 30 else ''),
@@ -1156,7 +1164,7 @@ class IterationTransform(Visitor.EnvTransform):
                 lhs = dict_temp,
                 rhs = ExprNodes.PythonCapiCallNode(
                     dict_obj.pos,
-                    "__Pyx_dict_iterator",
+                    f"__Pyx_dict_iterator{'_legacy' if legacy_method else ''}",
                     self.PyDict_Iterator_func_type,
                     utility_code = UtilityCode.load_cached("dict_iter", "Optimize.c"),
                     args = [dict_obj, is_dict, method_node, dict_len_temp_addr, is_dict_temp_addr],
--- a/Cython/Compiler/Optimize.py
+++ b/Cython/Compiler/Optimize.py
@@ -1131,7 +1131,7 @@ class IterationTransform(Visitor.EnvTransform):
         body.stats[0:0] = [iter_next_node]
 
         if method:
-            if dict_obj.type.is_pydict_type and method.startswith('iter'):
+            if dict_obj.type.is_pyanydict_type and method.startswith('iter'):
                 assert method in ('iterkeys', 'itervalues', 'iteritems'), method
                 method = EncodedString(method[4:])
             method_node = ExprNodes.IdentifierStringNode(dict_obj.pos, value=method)
--- a/Cython/Compiler/Optimize.py
+++ b/Cython/Compiler/Optimize.py
@@ -1166,7 +1166,9 @@ class IterationTransform(Visitor.EnvTransform):
                     dict_obj.pos,
                     "__Pyx_dict_iterator_legacy" if legacy_method else "__Pyx_dict_iterator",
                     self.PyDict_Iterator_func_type,
-                    utility_code = UtilityCode.load_cached("dict_iter", "Optimize.c"),
+                    utility_code = UtilityCode.load_cached(
+                        "dict_iter_legacy" if legacy_method else "dict_iter",
+                        "Optimize.c"),
                     args = [dict_obj, is_dict, method_node, dict_len_temp_addr, is_dict_temp_addr],
                     is_temp=True,
                 )),
--- a/Cython/Compiler/Optimize.py
+++ b/Cython/Compiler/Optimize.py
@@ -1164,7 +1164,7 @@ class IterationTransform(Visitor.EnvTransform):
                 lhs = dict_temp,
                 rhs = ExprNodes.PythonCapiCallNode(
                     dict_obj.pos,
-                    f"__Pyx_dict_iterator{'_legacy' if legacy_method else ''}",
+                    "__Pyx_dict_iterator_legacy" if legacy_method else "__Pyx_dict_iterator",
                     self.PyDict_Iterator_func_type,
                     utility_code = UtilityCode.load_cached("dict_iter", "Optimize.c"),
                     args = [dict_obj, is_dict, method_node, dict_len_temp_addr, is_dict_temp_addr],
--- a/Cython/Compiler/Optimize.py
+++ b/Cython/Compiler/Optimize.py
@@ -1132,6 +1132,7 @@ class IterationTransform(Visitor.EnvTransform):
 
         if method:
             if dict_obj.type.is_pydict_type and method.startswith('iter'):
+                assert method in ('iterkeys', 'itervalues', 'iteritems'), method
                 method = EncodedString(method[4:])
             method_node = ExprNodes.IdentifierStringNode(dict_obj.pos, value=method)
             dict_obj = dict_obj.as_none_safe_node(
--- a/Cython/Compiler/Optimize.py
+++ b/Cython/Compiler/Optimize.py
@@ -1131,6 +1131,8 @@ class IterationTransform(Visitor.EnvTransform):
         body.stats[0:0] = [iter_next_node]
 
         if method:
+            if dict_obj.type.is_pydict_type and method.startswith('iter'):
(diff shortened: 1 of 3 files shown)


Cython/Utility/Optimize.c
```diff
@@ -330,20 +331,16 @@ static CYTHON_INLINE PyObject* __Pyx_dict_iterator(PyObject* iterable, int is_di
 #else
         // On PyPy3/GraalPy, we need to translate manually a few method names.
         // This logic is not needed on CPython thanks to the fast case above.
```
scoder: Since we're only targeting Py3 these days, can't we translate the names at compile time?

da-woods: Partly I think?

We don't necessarily *know* it's a `dict`. At least some of the time the function is called with things that we've guessed is a `dict` but might not be.

If it isn't a `dict` then we want to use the untranslated name - i.e. a user class might genuinely have `iteritems`.

We could add a `py3_name` argument that is translated at compile-time though, and the section could just be `name = py3_name`

scoder: Ok, let's *not* do that then. It never really bothered anyone that we're building a Python string here.

da-woods: I think I'll do the translation in the cases where we know it's a `dict` because there's no reason not to.

Cython/Compiler/Optimize.py
```diff
@@ -1131,6 +1131,8 @@ def _transform_dict_iteration(self, node, dict_obj, method, keys, values):
         body.stats[0:0] = [iter_next_node]
 
         if method:
+            if dict_obj.type.is_pydict_type and method.startswith('iter'):
+                method = EncodedString(method[4:])
```
scoder: I know, we're taking care of this elsewhere, but that's elsewhere.
```suggestion
            if dict_obj.type.is_pydict_type and method.startswith('iter'):
                assert method in ('iterkeys', 'itervalues', 'iteritems'), method
                method = EncodedString(method[4:])
```

Cython/Utility/Optimize.c
```diff
@@ -330,20 +331,16 @@ static CYTHON_INLINE PyObject* __Pyx_dict_iterator(PyObject* iterable, int is_di
...
+            if (strncmp(name, "iter", 4) == 0 && (
+                    strcmp(name+4, "items") == 0 ||
+                    strcmp(name+4, "keys") == 0 ||
+                    strcmp(name+4, "values") == 0)) {
+                method_name = PyUnicode_FromString(name + 4);
+                if (unlikely(!method_name))
+                    return NULL;
+                owned_method_name = 1;
             }
         }
```
scoder: Instead of coding the name back and forth, we might actually get away with this:
```C
if (method_name && PyUnicode_ReadChar(method_name, 3) == 'r') {
    // This is enough to distinguish the "iter*" names from "items", "values", "keys".
    method_name = PyUnicode_Substring(method_name, 4, 10);  // longest is "itervalues" (len=10)
}
```
Plus error handling for both calls, plus refcount cleanup for `method_name`.
(I'm aware that UTF-8 encoding should be cheap in PyPy, but unpacking the string buffer for an undetermined lifetime might not be.)

Cython/Utility/Optimize.c
```diff
@@ -330,20 +331,16 @@ static CYTHON_INLINE PyObject* __Pyx_dict_iterator(PyObject* iterable, int is_di
...
-                if (!*pp) {
-                    *pp = PyUnicode_FromString(name + 4);
-                    if (!*pp)
-                        return NULL;
-                }
-                method_name = *pp;
+            if (strncmp(name, "iter", 4) == 0 && (
+                    strcmp(name+4, "items") == 0 ||
+                    strcmp(name+4, "keys") == 0 ||
+                    strcmp(name+4, "values") == 0)) {
```
scoder: If we want to keep the UTF-8 comparison, at least we shouldn't be getting here if the method name wasn't one of (iter)?(keys|values|items), so I think we can avoid the complete name comparisons at runtime.
```suggestion
            if (strncmp(name, "iter", 4) == 0) {
```

Cython/Utility/Optimize.c
da-woods: I've gone for checking the length since that's a bit more obvious to me than the 4th character.

Cython/Compiler/Optimize.py
```diff
@@ -1131,7 +1131,7 @@ def _transform_dict_iteration(self, node, dict_obj, method, keys, values):
         body.stats[0:0] = [iter_next_node]
 
         if method:
-            if dict_obj.type.is_pydict_type and method.startswith('iter'):
+            if dict_obj.type.is_pyanydict_type and method.startswith('iter'):
```
scoder: Functionally questionable since `frozendict` never had `iter*` methods, but since users won't run into attribute lookup errors in CPython, failures in PyPy&friends might easily go unnoticed. Thus, leaving a functional difference between CPython and PyPy would be even worse.

OTOH, since starting to use `frozendict` requires code changes, wouldn't that be the right time to also replace all left-over `.iter*()` usages with the modern `dict` view based methods? I'd rather actively push users into updating their code *while they are changing it* than silently making `frozendict.iterkeys()` work.

Cython/Utility/Optimize.c
```diff
@@ -331,17 +331,12 @@ static CYTHON_INLINE PyObject* __Pyx_dict_iterator(PyObject* iterable, int is_di
...
-            if (strncmp(name, "iter", 4) == 0 && (
-                    strcmp(name+4, "items") == 0 ||
-                    strcmp(name+4, "keys") == 0 ||
-                    strcmp(name+4, "values") == 0)) {
-                method_name = PyUnicode_FromString(name + 4);
-                if (unlikely(!method_name))
-                    return NULL;
-                owned_method_name = 1;
-            }
+        if (method_name && PyUnicode_GET_LENGTH(method_name) > 6) {
```
scoder: This is implicitly assuming that the Limited API (and its users) will never enable `CYTHON_AVOID_BORROWED_REFS`, which I don't think is a safe bet. I guess we'll have to live with `__Pyx_PyUnicode_GET_LENGTH()` and its potential exception return here, much as I regularly bemoan it.

Cython/Compiler/Optimize.py
da-woods: I agree it'd be good not to support `iter*` on `frozendict`. And also that PyPy and GraalPy should do the same as CPython.

The current situation (without this change) is that `frozendict.iteritems` works in Cython.  The change above makes it work a little faster, but doesn't alter any behaviour.

But we could definitely modify is so that `frozendict` doesn't work - that'd probably be better

da-woods:
I've banned the `iter*` methods on frozendict. I've created a new function to handle the `iter*` methods since a single function that handled both seemed quite fiddly.

Cython/Utility/Optimize.c
```diff
@@ -306,6 +306,9 @@ static CYTHON_INLINE int __Pyx_PyDict_Pop_ignore(PyObject *d, PyObject *key, PyO
 
 static CYTHON_INLINE PyObject* __Pyx_dict_iterator(PyObject* dict, int is_dict, PyObject* method_name,
                                                    Py_ssize_t* p_orig_length, int* p_is_dict);
+// "legacy" handles old-style iter* methods
+static CYTHON_INLINE PyObject* __Pyx_dict_iterator_legacy(PyObject* dict, int is_dict, PyObject* method_name,
+                                                   Py_ssize_t* p_orig_length, int* p_is_dict);
```
scoder:
```suggestion
static CYTHON_INLINE PyObject* __Pyx_dict_iterator_legacy(PyObject* dict, int is_dict, PyObject* method_name,
                                                          Py_ssize_t* p_orig_length, int* p_is_dict);
```

Cython/Compiler/Optimize.py
```diff
@@ -1153,7 +1164,7 @@ def _transform_dict_iteration(self, node, dict_obj, method, keys, values):
                 lhs = dict_temp,
                 rhs = ExprNodes.PythonCapiCallNode(
                     dict_obj.pos,
-                    "__Pyx_dict_iterator",
+                    f"__Pyx_dict_iterator{'_legacy' if legacy_method else ''}",
```
scoder: Spelling out both names makes them easier to grep for.
```suggestion
                    "__Pyx_dict_iterator_legacy" if legacy_method else "__Pyx_dict_iterator",
```

Cython/Utility/Optimize.c
```diff
@@ -319,49 +322,73 @@ static CYTHON_INLINE int __Pyx_dict_iter_next(PyObject* dict_or_iter, Py_ssize_t
...
+    }
+#endif
+    *p_orig_length = 0;
+    if (method_name) {
+        iterable = __Pyx_dict_call_to_get_iterable(iterable, method_name);
+        if (unlikely(!iterable)) {
+            return NULL;
+        }
+    }
+    return PyObject_GetIter(iterable);
```
scoder: The method call function already calls `PyObject_GetIter()` on the result.
```suggestion
    if (method_name) {
        return __Pyx_dict_call_to_get_iterable(iterable, method_name);
    } else {
        return PyObject_GetIter(iterable);
    }
```

Cython/Utility/Optimize.c
```diff
@@ -319,49 +322,73 @@ static CYTHON_INLINE int __Pyx_dict_iter_next(PyObject* dict_or_iter, Py_ssize_t
...
     if (method_name) {
-        PyObject* iter;
-        iterable = __Pyx_PyObject_CallMethod0(iterable, method_name);
-        if (!iterable)
+        iterable = __Pyx_dict_call_to_get_iterable(iterable, method_name);
+        if (owned_method_name) {
+            Py_DECREF(method_name);
+        }
+        if (unlikely(!iterable)) {
             return NULL;
```
scoder:
```suggestion
        return iterable;
```

Cython/Utility/Optimize.c
```diff
@@ -319,49 +322,73 @@ static CYTHON_INLINE int __Pyx_dict_iter_next(PyObject* dict_or_iter, Py_ssize_t
...
-#if !CYTHON_AVOID_BORROWED_REFS
-        if (PyTuple_CheckExact(iterable) || PyList_CheckExact(iterable))
-            return iterable;
-#endif
-        iter = PyObject_GetIter(iterable);
-        Py_DECREF(iterable);
-        return iter;
+        }
     }
     return PyObject_GetIter(iterable);
```
scoder:
```suggestion
} else {
    return PyObject_GetIter(iterable);
}
```

scoder (commented):
Yes, a separate function is better after all. Most people won't use this feature. We could even move it into a separate utility code section to avoid including it completely. (You could replace the `legacy_method` flag with `legacy_utility_code`.)

da-woods:
> We could even move it into a separate utility code section to avoid including it completely. (You could replace the legacy_method flag with legacy_utility_code.)

They're in a separate section now

PR #7640, opened 2026-04-25
Don't trigger full CI on docs-only changes

da-woods:
Exclude all markdown and rst files. Exclude the docs directory (but not docs/examples).


matusvalo:
+1

PR #7641, opened 2026-04-25
Try to fix GraalPy CI memory usage

da-woods:
We're seeing periodic test failures of the GraalPy run with

> The hosted runner lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.

I don't really know what's causing it, but my initial assumption is memory.

The is an uninformed attempt to fix that. To do so I've:
* force the GC to run before each compilation. (We already force it after each "run" test, but let's assume that cleaning up after the Cython compilation might be the problem here).
* Make the `ProcessPoolExecutor` periodically restart the processes (on the assumption that we're gradually accumulating memory leaks somewhere).

I've got no idea if this will work but it at least seems worth a try.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -2729,14 +2729,12 @@ def main():
             # Make sure subprocesses can print() Unicode text.
             os.environ["PYTHONIOENCODING"] = sys.stdout.encoding or sys.getdefaultencoding()
         from concurrent.futures import ProcessPoolExecutor, as_completed
-        pool = ProcessPoolExecutor(
-            options.shard_count,
-            # GraalPy seems to like using all the memory and crashing the GitHub workflow runner.
-            # This is a shot-in-the dark attempt to keep it down on the assumption that cleanup of
-            # Cython modules is poor. Because of this, always run the GraalPy tests in the executor
-            # even with -j1.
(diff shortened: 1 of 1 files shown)


da-woods:
I was able to run the GraalPy test 3 times on this PR without it breaking, so I'm going to assume it helps (even if not absolute proof). I may try to see if that lets us up the parallelism on the GraalPy job a little, but I'll do that separately.

da-woods:
> I may try to see if that lets us up the parallelism on the GraalPy job a little, but I'll do that separately.

It looks like I can't

runtests.py
```diff
@@ -1119,6 +1119,8 @@ def setUp(self):
         ]
         Options.warning_errors = self.warning_errors
 
+        gc.collect()
```
scoder, line 1122: Can we make this conditional? A full GC run can take a considerable amount of time in CPython, given the whole compiler, build and test machinery that is imported at this point.

da-woods: https://github.com/cython/cython/pull/7643

This could probably be dropped all together - I suspect that keeping the executor process lifetime short is the more useful change of the two. But making it conditional is easier than testing if that's true.

PR #7642, opened 2026-04-27
Get rid of specialized print statement utility code

da-woods:
In favour of generating the call to the 'print' function directly through nodes.

My reasoning was:
* The utility code is not thread-safe in the freethreading build because it lazily caches the 'print' function and kwargs for a common argument combination.
* While looking at that, I decided it would benefit from vectorcall too.

I initially thought about rewriting it, but it seemed easier to use our existing vectorcall code generation and just do the transformation into Cython nodes.

I think the only likely performance degredation is earlier limited api (without vectorcall) using the "no newline" syntax. This will now generate a keyword dict each time rather than caching it.


da-woods:
I was mainly interesting in fixing the thread-safety bit, so I'm happy to do a smaller change using the existing utility code with the caching on startup rather than lazily. I still prefer this version because going via vectorcall seems worthwhile, but it is probably more involved than the minimal change that would do what I wanted.

Cython/Compiler/Nodes.py
```diff
@@ -6816,61 +6816,40 @@ class PrintStatNode(StatNode):
...
+            if self.stream:
+                key_value_pairs.append((
+                    UnicodeNode(self.pos, value=EncodedString("file")),
+                    self.stream
+                ))
+            if not self.append_newline:
+                key_value_pairs.append((
+                    UnicodeNode(self.pos, value=EncodedString("end")),
+                    UnicodeNode(self.pos, value=EncodedString(" "))
+                ))
```
scoder, line 6845: Isn't it rather the empty string?

Cython/Compiler/Nodes.py
```diff
@@ -6816,61 +6816,40 @@ class PrintStatNode(StatNode):
...
+        func = NameNode(
+            self.pos,
+            name="print",
+            entry=env.builtin_scope().lookup_here("print")
+        )
+        if self.stream or not self.append_newline:
+            key_value_pairs = []
+            if self.stream:
+                key_value_pairs.append((
+                    UnicodeNode(self.pos, value=EncodedString("file")),
```
```suggestion
                    IdentifierStringNode(self.pos, value=EncodedString("file")),
```

Cython/Compiler/Nodes.py
```diff
@@ -6816,61 +6816,40 @@ class PrintStatNode(StatNode):
...
+                    self.stream
+                ))
+            if not self.append_newline:
+                key_value_pairs.append((
+                    UnicodeNode(self.pos, value=EncodedString("end")),
```
```suggestion
                    IdentifierStringNode(self.pos, value=EncodedString("end")),
```

Cython/Compiler/Nodes.py
```diff
@@ -6816,61 +6816,40 @@ class PrintStatNode(StatNode):
...
-            self.stream.generate_evaluation_code(code)
-            stream_result = self.stream.py_result()
-        else:
-            stream_result = '0'
-        if len(self.arg_tuple.args) == 1 and self.append_newline:
-            arg = self.arg_tuple.args[0]
-            arg.generate_evaluation_code(code)
+        from .ExprNodes import (
+            DictNode, UnicodeNode, SimpleCallNode, GeneralCallNode,
+            NameNode
```
```suggestion
            NameNode, IdentifierStringNode
```

scoder (commented):
It is a bit of a surprising side-effect to build this tree in `analyse_expressions()`.
You might have tried to put it in a transform first and failed because of the special GIL handling? We could stretch the purpose of `EarlyReplaceBuiltinCalls` a bit and move it there. Or … keep everything together in `PrintStatNode`. It's entirely localised and thus not too bad to do it there. Might even be clearer than moving the transformation elsewhere.

Cython/Compiler/Nodes.py
da-woods: That matches what the old code did. I'm not sure I currently have a version of Python2 around to actually test the behaviour it's replicating. But it leaves a space so the next thing you print is separated.

da-woods:
> You might have tried to put it in a transform first and failed because of the special GIL handling?

No. But I think it easier to do it after `InjectGilHandling`

> Or … keep everything together in PrintStatNode. It's entirely localised and thus not too bad to do it there.

You mean something like:

```
- PrintStatNode(...)
  - print_function_call: SimpleCallNode()
     - .... 
```

so the `PrintStatNode` remains, but just ends up evaluating the "implementation" node?

Both that and `EarlyReplaceBuiltinCalls` would probably be simpler.

scoder:
> You mean something like:

It wasn't actually what I meant (I was referring to the current state of the PR, sorry for being unclear), but I think it is worth keeping the `PrintStatNode` alive. It should still be simple enough to do but would simplify the existing handling of print statements a lot (however important that still is these days…).

da-woods:
I've kept everything in `PrintStatNode` which lets me revert most of the changes to the test, but kept the transformation in the same place.

Cython/Compiler/Nodes.py
da-woods: This is tested in https://github.com/cython/cython/blob/c52acc09740e4729dafcd804e6820a9ab00b8f81/tests/run/print.pyx#L15 and https://github.com/cython/cython/blob/c52acc09740e4729dafcd804e6820a9ab00b8f81/tests/run/print.pyx#L41

scoder: > it leaves a space so the next thing you print is separated

… sort of. Py2 had something called "soft space" for which it didn't print the space just yet but remembered that there was a space yet to come and injected it when more was printed in the same line. I don't think that would be easy to emulate.

EDIT: See https://peps.python.org/pep-3105/#specification – there's intentionally no equivalent to it in Py3.

scoder: … and since it was never available in Py3, the original C replacement that mapped `print` to `print()` also didn't have it. We're not breaking anything by replacing it with  Python implementation. Problem solved, keep walking, nothing to see here…

scoder:
Oh, I just noticed one issue with this: code size. If we replicate the entire implementation for each `print` statement, depending on the number of usages, this can make a difference.

EDIT: … although, using vectorcall might mitigate that. Not sure. Might be acceptable, but should at least be looked into.

da-woods:
With:

```
# cython: language_level=2

def f1(a):
    print a
    print a,
    print a, "hello"

def f2(a, b):
    print a, b
    print a, b,
    print a, b, "hello"

def f3(a, b, c):
    print a, b, c
    print a, b, c,
    print a, b, c, "hello"

def f4(*args):
    print args
    print *args
    print args,
    print *args
    print args, "hello"
    print *args, "hello"

def f5():
    print "hello"
    print "hello",
```

I get stripped sizes of
before: 49504
after: 49568
(on Py3.13, where vectorcall will be being applied)

You get some variation in the difference depending on which lines you leave in. For example, if you have `print "hello"` adds an extra static tuple in the old version but not the new version. And `print *args, "hello"` adds a bigger chunk to the new version but not the old version because it can't easily go through vectorcall.

Generally it looks like the new version is slightly bigger.

scoder:
Thanks for bringing in the numbers. I think we can easily afford that tiny difference.

PR #7644, opened 2026-04-27
Base multi-type inference for sequences on spanning_type() instead of set deduplication

scoder:


da-woods:
Without having looked at anything but the title, I suspect https://github.com/cython/cython/issues/5986 is relevant

matusvalo:
I cannot comment to PR by itself (I believe that it is improvement), but it is not solving the issue described in comment https://github.com/cython/cython/pull/7288#discussion_r3145917884

PR #7645, opened 2026-04-27
Avoid cleanup of unused temp in function call

da-woods:
At the moment this only really affects calls to builtin functions (since for most other cases Cython does generate unpacking code). Practically it probably only makes a difference in the Limited API (where DECREF is opaque).

Before

```
  __pyx_t_2 = NULL;
  __pyx_t_3 = 1;
  {
    PyObject *__pyx_callargs[2] = {__pyx_t_2, __pyx_v_a};
    __pyx_t_1 = __Pyx_PyObject_FastCall((PyObject*)__pyx_builtin_property, __pyx_callargs+__pyx_t_3, (2-__pyx_t_3) | (__pyx_t_3*__Pyx_PY_VECTORCALL_ARGUMENTS_OFFSET));
    __Pyx_XDECREF(__pyx_t_2); __pyx_t_2 = 0;
    if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 4, __pyx_L1_error)
    __Pyx_GOTREF(__pyx_t_1);
  }
```

After

```
__pyx_t_2 = NULL;
  __pyx_t_3 = 1;
  {
    PyObject *__pyx_callargs[2] = {__pyx_t_2, __pyx_v_a};
    __pyx_t_1 = __Pyx_PyObject_FastCall((PyObject*)__pyx_builtin_property, __pyx_callargs+__pyx_t_3, (2-__pyx_t_3) | (__pyx_t_3*__Pyx_PY_VECTORCALL_ARGUMENTS_OFFSET));
    if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 4, __pyx_L1_error)
    __Pyx_GOTREF(__pyx_t_1);
  }
```

The difference is to `__pyx_t_2`

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -7136,7 +7136,8 @@ class PyMethodCallNode(CallNode):
 
         # Clean up.
 
-        code.put_xdecref_clear(self_arg, py_object_type)
+        if self.unpack:
+            code.put_xdecref_clear(self_arg, py_object_type)
         for tmp in [self_arg, space_for_selfarg]:
             code.funcstate.release_temp(tmp)
 
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -7136,7 +7136,7 @@ class PyMethodCallNode(CallNode):
 
         # Clean up.
 
-        if self.unpack:
+        if self.unpack or self.use_method_vectorcall:
             code.put_xdecref_clear(self_arg, py_object_type)
         for tmp in [self_arg, space_for_selfarg]:
             code.funcstate.release_temp(tmp)


scoder:
Hmm, interesting. Yes, the decref seems incorrect because the `__Pyx_PY_VECTORCALL_ARGUMENTS_OFFSET` basically hands over the ownership of the first array entry to CPython. OTOH, CPython shouldn't leave dangling pointers in there. If it needs the space, it should (and probably does) clean it up after the call, so this should not be an issue. And `XDECREF` checks for `NULL` in the C code, so it's a quick check that's just redundant.

However, your change is mixing different levels of abstraction here. It lets `generate_evaluation_code` know when it needs to clean up from conditions that lower level function calls made. It's not difficult to imagine changes in the calling code that leave us with a reference leak due to the now duplicated condition. I'd consider `self.use_method_vectorcall` a (sufficiently) local condition that's fine to use here but `self.unpack` appears too remote.

da-woods:
It's only for a tiny bit of efficiency (rather than fixing anything incorrect) and I think it's slightly less complicated than you think.

> Hmm, interesting. Yes, the decref seems incorrect because the `__Pyx_PY_VECTORCALL_ARGUMENTS_OFFSET` basically hands over the ownership of the first array entry to CPython. OTOH, CPython shouldn't leave dangling pointers in there. If it needs the space, it should (and probably does) clean it up after the call, so this should not be an issue. And `XDECREF` checks for `NULL` in the C code, so it's a quick check that's just redundant.

This is just the variable in `__pyx_t_2` I'm worried about rather than the array. The array we pass to vectorcall is reference-neutral (i.e. CPython may mess with the storage space, but doesn't modify any of the reference counting in there). We hold references for everything we put in the array, but the references we hold are are the temporary variables outside the array.

> However, your change is mixing different levels of abstraction here. It lets generate_evaluation_code know when it needs to clean up from conditions that lower level function calls made. It's not difficult to imagine changes in the calling code that leave us with a reference leak due to the now duplicated condition. I'd consider self.use_method_vectorcall a (sufficiently) local condition that's fine to use here but self.unpack appears too remote.

Yeah - I agree with that.

I'll see if I can move it somewhere clearer and less abstracted. If I can't then I'll just drop the change because it's only a pointless check on something that's always `NULL` but there's no actual harm done.

scoder:
Looking through the existing code, I get the impression that it would generally benefit from extracting the vectorcall code paths. Several methods have an entirely separate special case for it. Separating the vectorcall and method unpacking paths seems a good way to cleanly resolve issues like this.

scoder:
Thinking one step further – is it worth separating out an entirely new node class for vectorcalls? They really seem to use mostly separate code paths. We could add a factory function that builds a suitable node instance for a given kind of pre-analysed call. There's already `PyMethodCallNode.from_node()`, which seems the One Way to create such an instance right now, but a separate function would help with the class split (and would then decide and call the right node class's `.from_node()` method).

da-woods:
I think I'll close this for the moment.

I've had a more detailed look at `Py_XDECREF` in the Limited API and the NULL check looks like it won't be opaque, so I'm pretty confident that any decent C compiler would be able to optimize it all out.

It'd be nice to drop it and make it easy for the compiler. But I'm not that enthusiastic about a bigger refactoring here rght now.

PR #7648, opened 2026-05-01
Add missing inputs for ci-job.yml

da-woods:
They were in the matrix but weren't being passed to ci-job so weren't doing anything.


da-woods:
This should have been part of https://github.com/cython/cython/pull/7574 but I clearly missed the second use of ci-job.yml

da-woods:
I don't think this was passing last time I looked. The changes here were correct, but they'd exposed some other changes that needed to be made.

The runs using a custom C compiler were falling (possibly because the version we requested wasn't on Ubuntu).

PR #7649, opened 2026-05-01
Extend `dict` optimisations to `frozendict`

### LINKED ISSUE #7647: Allow frozendict where we handle dict
### Describe your issue

There are several special cases and optimisations for `dict`s, e.g. iteration or faster lookup code. Many of them equally apply to `frozendict` and should be triggered in the same way, but were not anticipated because previously, there was only one `dict` type.

I'm marking this as a "good first issue" because much of the work is finding the special cases and, at least in some places, changing type checks (Cython' `dict_type` builtin type, `isinstance()` or runtime `PyDict_Check()` / `PyDict_CheckExact()`) to include `frozendict` as well. We can discuss the more involved code changes if we find any along the way.

Remember that the difference between `dict` and `frozendict` is the latter being immutable, so not all usages and specialisations of `dict` apply to it.

Simple approach 1:
- grep for usages of the above patterns in the code base.
- Try to get an idea of what the code does that you find and come up with a (failing) test that uses `frozendict` instead of `dict` to exercise it.
- Fix the test by changing the code to also handle `frozendict`.

Simple approach 2:
- Duplicate the `dict` related test files in `tests/run/` for `frozendict`, see if they work or fail (`dict.pyx` is already duplicated as `frozendict.pyx`, to give an example).
- Also find tests that (partly) use dicts, e.g. any tests related to item lookup, get/set/delitem, by `# tag:` line, file name or usage patterns), decide (or ask) if they seem worth duplicating for `frozendict`, and if so, add the corresponding tests (preferably in the same file).
- See if the tests fail and where, then either try to fix them or list them here and ask for further guidance.

When you add tests, it might be a good idea to open a pull request to discuss them there, even if the new tests fail.

omkar-334:
Related Issue - #7647
@scoder 

As given in the issue, I started with the simple approach 2 - I ran three commands which gave me `dict`-related test files - 
```bash                                             
ls tests/run/dict*.pyx tests/run/*dict*.pyx                                                                                                                 
grep -l '^# tag:.*\bdict\b' tests/run/*.pyx tests/run/*.py                                                                                                  
grep -lE 'def [a-zA-Z_]+\(dict |PyDict_|cdef dict ' tests/run/*.pyx                                                                                         
```  
This gave me 45 unique files. Out of these some were unrelated or already covered in `frozendict.pyx`. So I chose 12 test files in the end . I duplicated the `dict` tests in these files and replaced them with `frozendict`.   
So Running these tests against the unmodified compiler reported `Ran 318 tests, FAILED (failures=4, errors=8)`. The 8 ERRORs are compile-time path-assertion violations; the 4 FAILs are doctest-output mismatches.

| # | Module | Backend | Type  | Where | What didn’t happen | Fix |
|---|--------|---------|-------|--------|--------------------|-----|
| 1 | assigned_builtin_methods | c | ERROR | line 49: `bound_frozendict_get` | `//ReturnStatNode//PythonCapiCallNode` was expected (bound-method optimisation lowering `fd.get` to a C API call). Compiler still treats `frozendict` as generic Python object, so `fd.get(1)` remains attribute access | In `Cython/Compiler/ExprNodes.py` (AttributeNode analysis), replace `is_pydict_type` with `is_pyanydict_type` |
| 2 | builtin_len | c | ERROR | line 108: `len_frozendict` | `//PythonCapiCallNode` expected (`len(fd)` → `PyDict_Size`). `_map_to_capi_len_function` only knows `Builtin.dict_type` | Add mapping: `Builtin.frozendict_type → "PyDict_Size"` in `Cython/Compiler/Optimize.py` |
| 3 | dict_get | c | ERROR | lines 98, 125: `get_frozendict`, `get_default_frozendict` | `//PythonCapiCallNode` expected (`fd.get(...)` lowering). `MethodDispatcherTransform` lacks handler for frozendict | Alias `_handle_simple_method_frozendict_get = _handle_simple_method_dict_get` in `Cython/Compiler/Optimize.py` |
| 4 | iterdict | c | ERROR | lines 566, 583: `iterfrozendict`, `iterfrozendict_listcomp` | `//WhileStatNode//DictIterationNextNode` expected (dict iteration optimisation). Checks only apply to `is_pydict_type` | Replace those checks with `is_pyanydict_type` in `Optimize.py` |
| 5 | inop | c | FAIL | `q_frozendict` doctest | Error message mismatch: `TypeError: argument of type 'NoneType' is not iterable` vs expected dict-style wording | Relax doctest using `# doctest: +ELLIPSIS` and match `TypeError: ...NoneType...` |
| 6 | notinop | c | FAIL | `q_frozendict` doctest | Same mismatch as `inop` but for `not in` | Same fix as above |

The other 6 were the same issues, but on the cpp backend.   

So to fix these tests, 
1. ``Cython/Compiler/PyrexTypes.py`` - I added `is_pyfrozendict_type` (per-type flag) and `is_pyanydict_type` (composite "dict or frozendict") to the `PyrexType` defaults
2. `Cython/Compiler/Tests/TestTypes.py` - I also updated the `BuiltinObjectType._builtin_type_flag_mapping` to reflect this. 
3. The remaining two files - `Cython/Compiler/ExprNodes.py` and `Cython/Compiler/Optimize.py` - I updated the `if` conditions

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -15109,7 +15109,7 @@ class CoerceToBooleanNode(CoercionNode):
             return '__Pyx_PyTuple_GET_SIZE'
         if typ.is_pyset_type or typ.is_pyfrozenset_type:
             return '__Pyx_PySet_GET_SIZE'
-        if typ.is_pydict_type or typ.is_pyfrozendict_type:
+        if typ.is_pyanydict_type:
             return '__Pyx_PyDict_GET_SIZE'
         if typ.is_pybytes_type:
             return '__Pyx_PyBytes_GET_SIZE'
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -7727,9 +7727,11 @@ class MergedDictNode(ExprNode):
         args = iter(self.keyword_args)
         item = next(args)
         item.generate_evaluation_code(code)
-        if not item.type.is_pydict_type:
+        if not item.type.is_pyanydict_type:
             # CPython supports calling functions with non-dicts, so do we
-            code.putln('if (likely(PyDict_CheckExact(%s))) {' %
+            code.globalstate.use_utility_code(UtilityCode.load_cached(
+                "PyFrozenDict", "Builtins.c"))
+            code.putln('if (likely(__Pyx_PyAnyDict_CheckExact(%s))) {' %
                        item.py_result())
 
         if item.is_dict_literal:
@@ -7757,7 +7759,7 @@ class MergedDictNode(ExprNode):
             if item.result_in_temp():
                 code.putln("}")
 
-        if not item.type.is_pydict_type:
+        if not item.type.is_pyanydict_type:
             code.putln('} else {')
             code.globalstate.use_utility_code(UtilityCode.load_cached(
                 "PyObjectCallOneArg", "ObjectHandling.c"))
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -4448,14 +4448,14 @@ class IndexNode(_IndexingBaseNode):
 
     def analyse_as_pyobject(self, env, is_slice, getting, setting):
         base_type = self.base.type
-        if self.index.type.is_unicode_char and not base_type.is_pydict_type:
+        if self.index.type.is_unicode_char and not base_type.is_pyanydict_type:
             # TODO: eventually fold into case below and remove warning, once people have adapted their code
             warning(self.pos,
                     "Item lookup of unicode character codes now always converts to a Unicode string. "
                     "Use an explicit C integer cast to get back the previous integer lookup behaviour.", level=1)
             self.index = self.index.coerce_to_pyobject(env)
             self.is_temp = 1
-        elif self.index.type.is_int and not base_type.is_pydict_type:
+        elif self.index.type.is_int and not base_type.is_pyanydict_type:
             if (getting
                     and not env.directives['boundscheck']
                     and (base_type.is_pybytearray_type or base_type.is_pylist_type or base_type.is_pytuple_type)
@@ -4494,7 +4494,7 @@ class IndexNode(_IndexingBaseNode):
                 # Infer homogeneous item type when looping over container literals.
                 item_type = infer_sequence_item_type(env, self.base, seq_type=base_type)
 
-            if base_type.is_pylist_type or base_type.is_pytuple_type or base_type.is_pydict_type:
+            if base_type.is_pylist_type or base_type.is_pytuple_type or base_type.is_pyanydict_type:
                 # Do the None check explicitly (not in a helper, and regardless of 'nonecheck') to allow optimising it away.
                 self.base = self.base.as_none_safe_node("'NoneType' object is not subscriptable")
 
@@ -4831,7 +4831,7 @@ class IndexNode(_IndexingBaseNode):
                     function = "__Pyx_GetItemInt"
                 utility_code = TempitaUtilityCode.load_cached("GetItemInt", "ObjectHandling.c")
             else:
-                if base_type.is_pydict_type:
+                if base_type.is_pyanydict_type:
                     function = "__Pyx_PyDict_GetItem"
                     utility_code = UtilityCode.load_cached("DictGetItem", "ObjectHandling.c")
                 elif base_type is py_object_type and self.index.type.is_pystr_type:
@@ -10101,7 +10101,7 @@ class SortedDictKeysNode(ExprNode):
 
     def analyse_types(self, env):
         arg = self.arg.analyse_types(env)
-        if arg.type.is_pydict_type:
+        if arg.type.is_pyanydict_type:
             arg = arg.as_none_safe_node(
                 "'NoneType' object is not iterable")
         self.arg = arg
@@ -10112,7 +10112,7 @@ class SortedDictKeysNode(ExprNode):
 
     def generate_result_code(self, code):
         dict_result = self.arg.py_result()
-        if self.arg.type.is_pydict_type:
+        if self.arg.type.is_pyanydict_type:
             code.putln('%s = PyDict_Keys(%s); %s' % (
                 self.result(), dict_result,
                 code.error_goto_if_null(self.result(), self.pos)))
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -14085,7 +14085,7 @@ class CmpNode:
                          _) = result
                         return True
         elif self.operator in ('in', 'not_in'):
-            if self.operand2.type.is_pydict_type:
+            if self.operand2.type.is_pyanydict_type:
                 self.operand2 = self.operand2.as_none_safe_node("'NoneType' object is not iterable")
                 self.special_bool_cmp_utility_code = UtilityCode.load_cached("PyDictContains", "ObjectHandling.c")
                 self.special_bool_cmp_function = "__Pyx_PyDict_ContainsTF"
@@ -14565,7 +14565,7 @@ class CascadedCmpNode(Node, CmpNode):
 
     def coerce_operands_to_pyobjects(self, env):
         self.operand2 = self.operand2.coerce_to_pyobject(env)
-        if self.operand2.type.is_pydict_type and self.operator in ('in', 'not_in'):
+        if self.operand2.type.is_pyanydict_type and self.operator in ('in', 'not_in'):
             self.operand2 = self.operand2.as_none_safe_node("'NoneType' object is not iterable")
         if self.cascade:
             self.cascade.coerce_operands_to_pyobjects(env)
--- a/Cython/Compiler/Optimize.py
+++ b/Cython/Compiler/Optimize.py
@@ -2588,7 +2588,7 @@ class OptimizeBuiltinCalls(Visitor.NodeRefCleanupMixin,
 
     PyDict_Copy_func_type = PyrexTypes.CFuncType(
         Builtin.dict_type, [
-            PyrexTypes.CFuncTypeArg("dict", Builtin.dict_type, None)
+            PyrexTypes.CFuncTypeArg("dict", PyrexTypes.py_object_type, None)
             ])
 
     def _handle_simple_function_dict(self, node, function, pos_args):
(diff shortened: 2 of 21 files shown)


omkar-334:
the 
```python
def make_frozendict(d):
    return frozendict(d)
```
function is required because `frozendict` is a builtin only from python 3.15. For the versions below this it cant run directly i think.

Cython/Compiler/Tests/TestTypes.py
```diff
@@ -27,6 +27,8 @@ class TestBuiltinTypes(TimedTest):
         'is_pycomplex_type': ['complex'],
         'is_pylist_type': ['list'],
         'is_pydict_type': ['dict'],
+        'is_pyfrozendict_type': ['frozendict'],
+        'is_pyanydict_type': ['dict', 'frozendict'],
```
scoder, line 31: This will conflict a tiny bit with the upcoming https://github.com/cython/cython/pull/7288 but it's totally the right way to do it.

Cython/Compiler/ExprNodes.py
```diff
@@ -4809,7 +4809,7 @@ def generate_result_code(self, code):
                     function = "__Pyx_GetItemInt"
                 utility_code = TempitaUtilityCode.load_cached("GetItemInt", "ObjectHandling.c")
             else:
-                if base_type.is_pydict_type:
+                if base_type.is_pyanydict_type:
                     function = "__Pyx_PyDict_GetItem"
                     utility_code = UtilityCode.load_cached("DictGetItem", "ObjectHandling.c")
```
scoder, line 4836: For this, I think we have to adapt the `__Pyx_PyObject_Dict_GetItem()` macro in `ObjectHandling.c` to allow `frozendict`. There's a runtime decision that we cannot really test for.

Cython/Compiler/Optimize.py
scoder, line 1158: I think this utility code (`dict_iter` in `Optimize.c`) also needs to be adapted.

tests/run/dict.pyx
```diff
@@ -177,3 +182,19 @@ def from_keys_bound(dict d, val):
...
+
+    >>> sorted(from_keys_bound_frozendict(make_frozendict({}), 100).items())
+    [('a', 100), ('b', 100)]
+    >>> sorted(from_keys_bound_frozendict(make_frozendict({}), None).items())
+    [('a', None), ('b', None)]
+    """
+    if val is not None:
+        return fd.fromkeys(("a", "b"), val)
+    else:
+        return fd.fromkeys(("a", "b"))
```
scoder: This should go into `frozendict.pyx`.
(The other – special – ones like `dict_get.pyx` seem fine to keep in one file because they test for certain method/operator optimisation features, not the `dict` type specifically.)

tests/run/dictintindex.pyx
```diff
@@ -243,3 +247,45 @@ def test_del_ulonglong_big():
...
     return d[key]
+
+
+def test_get_char_frozendict():
+    """
+    >>> test_get_char_frozendict()
+    1
+    """
+    cdef char key = 0
+    cdef frozendict fd = make_frozendict({0: 1})
```
scoder: This is in Cython code, so you can just use `frozendict()` directly. (Same for the others below.)
```suggestion
    fd = frozendict({0: 1})
```

tests/run/dictintindex.pyx
```diff
@@ -1,3 +1,7 @@
+def make_frozendict(d):
```
scoder: This file was probably generated at the time, at least partly. I think it uses dict literals intentionally (although that's not the main thing it tests) and `frozendict` doesn't have these literals, so I'm not sure what's better – duplicate the file or just add a few sample tests.

But I think we can get away with adding only samples. It's unlikely that we'd add special code for a dedicated `frozendict` implementation at some point that would make the entire range of tests necessary. Let's keep it as you have it.

tests/run/inop.pyx
```diff
@@ -317,6 +322,29 @@ def q(a):
...
 
+
+def p_frozendict(a):
+    """
+    >>> p_frozendict(1)
+    0
+    >>> p_frozendict('a')
+    1
+    """
+    cdef frozendict fd = make_frozendict({u'a': 1, u'b': 2})
```
scoder: I don't think we need the helper function in this file.
```suggestion
    cdef frozendict fd = frozendict({u'a': 1, u'b': 2})
```

tests/run/iterdict.pyx
```diff
@@ -1,8 +1,14 @@
...
 cimport cython
 
+
+def make_frozendict(d):
+    return frozendict(d)
+
+
 dict_size = 4
 d = dict(zip(range(10,dict_size+10), range(dict_size)))
+fd = make_frozendict(d)
```
scoder:
```suggestion
fd = frozendict(d)
```

tests/run/notinop.pyx
```diff
@@ -309,3 +314,26 @@ def q(a):
...
+
+
+def p_frozendict(a):
+    """
+    >>> p_frozendict('a')
+    0
+    >>> p_frozendict(1)
+    1
+    """
+    cdef frozendict fd = make_frozendict({u'a': 1, u'b': 2})
```
scoder: No need for `make_frozendict`.
```suggestion
    cdef frozendict fd = frozendict({u'a': 1, u'b': 2})
```

scoder (commented):
Cool, that looks good.

What's still missing is that some of the utility code in `Cython/Utility/*.c` also does C-level type checks for the `dict` type as well as some more involved C-API usages that may not all work for `frozendict`. Could you look into that as well?

Cython/Compiler/Tests/TestTypes.py
matusvalo: PR is merged to master.

omkar-334: Resolved this conflict.... kept `is_pyanydict_type`.. thanks!

omkar-334:
hey @scoder there were 4 failed tests at commit https://github.com/cython/cython/pull/7649/commits/487bb6e06ec933fcd2d2c33d28f92018a183f25a, something related to gcc version 13... is it related to code? or just flaky tests?
Screenshot 2026-05-02 at 11 43 32 PM

tests/run/dictintindex.pyx
omkar-334: got it , changed this and others as well

da-woods:
Ubuntu is having problems with their PPA servers and that's why the tests are failing - it's nothing to do with this change.

Cython/Compiler/ExprNodes.py
omkar-334: @scoder https://github.com/cython/cython/commit/20f1138d00a3046c0e7bc8febbe200744b2de6dd -  implements what you refer to - `__Pyx_PyAnyDict_CheckExact`

I replaced three instances of `PyDict_CheckExact` with `__Pyx_PyAnyDict_CheckExact` - 1 in `Optimize.C` and 2 in `ObjectHandling.c`

omkar-334: is this correct?

omkar-334:
thanks @da-woods.... I just added 2 more commits, can you approve the tests again?

scoder:
> Ubuntu is having problems with their PPA servers

It's probably an effect of the DDoS attacks that they current face. Not their fault at all in that case.

Cython/Utility/ObjectHandling.c
```diff
@@ -397,12 +397,15 @@ static PyObject *__Pyx_PyObject_GetItem(PyObject *obj, PyObject *key) {
 
 
 /////////////// DictGetItem.proto ///////////////
+//@requires: Builtins.c::PyFrozenDict
 
 #if !CYTHON_COMPILING_IN_PYPY
 static PyObject *__Pyx_PyDict_GetItem(PyObject *d, PyObject* key);/*proto*/
 
+// `PyDict_GetItem` accepts any dict subclass at the C level, so frozendict
+// (which is a true dict subclass on Python 3.15+) takes the fast path here too.
```
da-woods: I don't think the subclass bit is true either practically or in principle

```
>>> frozendict.__bases__
(<class 'object'>,
```

(frozendict shouldn't be a subclass because it can't be substituted anywhere that a dict can be).

I actually don't think the comment is needed - it's only "non-obvious" because frozendict is new

Cython/Utility/ObjectHandling.c
```diff
@@ -551,7 +555,8 @@ static CYTHON_INLINE PyObject *__Pyx_GetItemInt_Fast(PyObject *o, Py_ssize_t i,
 #endif
 #if CYTHON_USE_TYPE_SLOTS && !CYTHON_COMPILING_IN_PYPY
     // This dict check is so cheap after the other type checks above that it would be costly *not* to do it.
-    if (PyDict_CheckExact(o)) {
+    // Frozendict shares dict's mp_subscript layout, so the same fast path applies.
```
da-woods: Same here. In this case the comment isn't wrong, but I don't think we need to explain it.

omkar-334: understood

scoder:
From a quick look, I might have found one more spot:
https://github.com/cython/cython/blob/87f82b7d7c84886396e15f0c09910011188047af/Cython/Utility/FunctionArguments.c#L880-L883
Here, `source_mapping` should allow any dict but that needs testing (I think that would be passing a `frozendict` as `**kwargs` together with other keyword arguments)

(EDIT: With "needs testing", I meant that we are lacking a test that uses `frozendict` for this, and that such a test should be added in order to validate the respective code.)

scoder:
The additional test for `MergeKeywords` can go into `tests/run/callargs.pyx`. It uses a `kwargs` global variable that is passed into some function calls as `**kwargs`, and you can put a `frozendict()` copy of the dict next to it and duplicate the function calls in the test functions to make use of that instead of `**kwargs`.

matusvalo:
After mergeing #7662, `_special_builtin()` can be updated to use `is_pyanydict_type` instead of `typ.is_pydict_type or typ.is_pyfrozendict_type`

omkar-334:
@scoder @matusvalo I've made the requested changes (`MergeKeywords` tests and replacing in `_special_builtin`) along with 2 other occurrences I've noticed.

scoder:
Looks great. Happy to merge this once CI is green.

scoder:
Thanks for your contribution.

scoder:
It turned out after the fact that two of the changes made it run into `PyDict_Copy()` with a frozendict, which isn't allowed in the now released Py3.15.0b1. I've fixed that in
https://github.com/cython/cython/commit/33db04f9550eb8dadc3f8e54d48e724018ffbfa8

PR #7652, opened 2026-05-02
Fully specify CI matrix

da-woods:
According to https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/run-job-variations it doesn't look like the matrix looks quite like we've been assuming it does.

Essentially anything extra in "include" that doesn't contradict something existing is added everywhere. So the `gcc_version: 13` aren't being added as extra subjobs but are being applied to everything that the other fields match.

I think the gist is that we at least need to specify `extra_hash` to the original matrix so that they definitely contradict that and get added as as extra job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
       matrix:
         os: [ubuntu-22.04, windows-2022, macos-latest]
         backend: [c, cpp]
+        extra_hash: ['']
 
         include:
           # Limited API
@@ -112,6 +113,7 @@ jobs:
           - "3.14t"
           - "3.15-dev"
           - "3.15t-dev"
+        extra_hash: ['']
 
         exclude:
           # Windows and free-threading is difficult to build right now
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,12 +135,6 @@ jobs:
           # ================
           # GCC 13 (with broad language standards)
           - os: ubuntu-24.04
-            python-version: "3.9"
-            backend: c
-            gcc_version: 13
-            extra_cflags: "-std=c99"
-            extra_hash: "-c99"
-          - os: ubuntu-24.04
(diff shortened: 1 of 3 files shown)


da-woods:
I think the failures are just because ubuntu is having ppa server problems. But I think this is working much closer to what is intended.

da-woods:
I've decided to bump the Ubuntu version to 24.04. That has all the GCC versions that we want by default so we don't need to use any extra ppa repositories.  That seems to resolve the immediate test failures. And is probably a good thing anyway

.github/workflows/ci.yml
```diff
@@ -132,35 +134,29 @@ jobs:
           # Ubuntu sub-jobs:
           # ================
           # GCC 13 (with broad language standards)
-          - os: ubuntu-22.04
-            python-version: "3.9"
-            backend: c
-            gcc_version: 13
-            extra_cflags: "-std=c99"
-            extra_hash: "-c99"
```
scoder, line 140: I think this test was not only that it needs C99 but also that it doesn't need more than C99 by accident. I'd keep it for that.

scoder (commented):
I already tried switching to 24.04 a couple of times and failed, for changing reasons. Happy to see that it works now.

.github/workflows/ci.yml
scoder: Alternatively, we could generally pass this option everywhere in our CFLAGS now.

da-woods: Yeah happy to revert that one

da-woods:
> I already tried switching to 24.04 a couple of times and failed, for changing reasons. Happy to see that it works now.

I think the only real barrier (this time) was the missing blas library.

There were some enormous, hard to read, C++ template warnings about unused arguments from Pythran.  I think last time I tried switching to 24.04 I thought those were the problem and couldn't face trying to debug the C++ templates.  But actually they're really minor.

PR #7655, opened 2026-05-03
Separate wheel check step from wheel build step

da-woods:
Thinking about trying to improve supply chain security and pinning our build dependencies.

The dependencies that we use during the wheel build step are:
* cibuildwheel (which is pinned through the github action, so is hopefully relative secure)
* setuptools (which has no dependencies)
* abi3audit and its dependencies
* twine and its dependencies.

This PR moves abi3audit and twine to a separate step so there's no way they can influence the built wheels.

That leaves us with only setuptools, and managing a single pinned dependency seems pretty reasonable.

It's probably more paranoid than it needs to be (and no-one else seems to really be worried about test dependencies like this). However it's also fairly simple to do given Cython's limited dependencies.

`twine` may be a lost cause though because I suspect it's ultimately used to upload to PyPI. But I can't see that so I'm ignoring it.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -241,7 +241,7 @@ jobs:
           cd ./wheel_check
           for filename in *.whl; do
             twine check "$filename"
-            if [[ "$filename" == *"abi3"*]]; then
+            if [[ "$filename" == *"abi3"* ]]; then
               abi3audit --strict --report "$filename"
             fi
           done
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -123,11 +123,6 @@ jobs:
         with:
           only: ${{ matrix.only }}
 
-      - name: Check wheel
-        run: |
-          python -m pip install twine
-          python -m twine check ./wheelhouse/*.whl
-
       - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
         with:
           name: ${{ matrix.only }}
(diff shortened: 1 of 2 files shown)


.github/workflows/wheels.yml
```diff
@@ -226,9 +215,40 @@ jobs:
           name: pure-wheel
           path: ./dist/*.whl
 
+  # Keep abi3audit and twine checks out of the the wheel build step
+  # to minimize the dependencies that need to be installed (ideally to
+  # steptools, cibuildwheel and cibuildwheel's dependencies only).
```
scoder: Typo, but I'd actually remove the last part which easily risks getting outdated in this place.
```suggestion
  # to minimize the dependencies that need to be installed.
```

.github/workflows/wheels.yml
```diff
@@ -226,9 +215,40 @@ jobs:
...
+
+      - name: Run tools on wheels
+        run: |
+          cd ./wheel_check
+          for filename in *.whl; do
+            twine check "$filename"
+            if [[ "$filename" == *"abi3"* ]]; then
+              abi3audit --strict --report "$filename"
+            fi
+          done
```
scoder, line 247: I doubt that this fails on validation errors. It should.
```suggestion
      - name: Run tools on wheels
        run: |
          cd ./wheel_check
          ERRORS=0
          for filename in *.whl; do
            twine check "$filename" || ERRORS=$((ERRORS + 1))
            if [[ "$filename" == *"abi3"* ]]; then
              abi3audit --strict --report "$filename" || ERRORS=$((ERRORS + 1))
            fi
          done
          exit $ERRORS
```

scoder (approved):
Very good idea. We should definitely do this, also in the 3.2.x branch, if possible.

da-woods:
It didn't backport perfectly cleanly to 3.2.x so I'll go via a PR (https://github.com/cython/cython/pull/7660) just to check that it works before merging it.

da-woods:
3.2.x commit https://github.com/cython/cython/commit/074a1b05413e86860f793eda3dd4e3a803fd1e54

Issue #7603, opened 2026-03-29
Document the binary interface of cimported modules (`.pxd` ABI rules, capsule signature checks)


rgommers:
## Background

While investigating a compatibility breakage between SciPy nightlies and scikit-learn ([scikit-learn#33616](https://github.com/scikit-learn/scikit-learn/issues/33616), a recurrence of [scipy#18377](https://github.com/scipy/scipy/issues/18377)), I found it necessary to read the Cython compiler source in detail to understand exactly what constitutes a binary-incompatible change to a `.pxd` file. That information does not appear to be documented anywhere. This issue proposes adding it.

## What actually happens, and where in the source

The Cython documentation describes `.pxd` files as working "like C header files". This is not actually the case when it comes to binary compatibility of `cimport`'ing across packages.

There are **two separate checking mechanisms**, operating at different times and with different rules.

### 1. Compile-time check: `.pxd` vs `.pyx` declaration mismatch

When a `.pxd` and a `.pyx` file share the same name, Cython merges them during compilation. It calls `same_c_signature_as_resolved_type` (and relatives `compatible_signature_with_resolved_type`, `narrower_c_signature_than_resolved_type`) in [`Cython/Compiler/PyrexTypes.py`](https://github.com/cython/cython/blob/master/Cython/Compiler/PyrexTypes.py) to decide whether the declarations are compatible. Crucially, type comparison at this stage calls `resolve()` on both sides, which **strips typedefs** down to the underlying primitive - so `ctypedef int blas_int` and `int` are considered the same type at compile time, and no error is raised.

The merge and error reporting happens in [`Symtab.ModuleScope.declare_cfunction`](https://github.com/cython/cython/blob/master/Cython/Compiler/Symtab.py), with the compile-time error `"Function signature does not match previous declaration"` emitted there.

### 2. Runtime check: `cimport` across pre-compiled modules

This is the check that causes the `TypeError: C function X has wrong signature` error
at import time. The mechanism lives in [`Cython/Utility/ImportExport.c`](https://github.com/cython/cython/blob/master/Cython/Utility/ImportExport.c):

[...]

PR #7656, opened 2026-05-04
Improve names in fused `__pyx_api__`

da-woods:
So that they're based on the types of the argument rather than the slightly arbitrary order of the types within a fusd type. This should mean they keep being right if the fused type is expanded.

For an example fused function:

```
# pxd file
cdef fused IntOrFloat:
    int
    float

cdef fused FloatOrObject:
    float
    object

cdef IntOrFloat fused_func(IntOrFloat x, FloatOrObject y)
```

the old names in the `__pyx_capi__` dictionary (which are retained as aliases for now) are:

```
__pyx_fuse_1_1fused_func
__pyx_fuse_1_0fused_func
__pyx_fuse_0_1fused_func
__pyx_fuse_0_0fused_func
```

while the new names are:

```
fused_func[float,object]
fused_func[float,float]
fused_func[int,object]
fused_func[int,float]
```

The old values are kept for compatibility, and will be used if the new values aren't found (ensuring compatibility with old modules). I'd intend to remove this in the end but likely not any time soon.

Related to https://github.com/cython/cython/issues/7603 (vaguely)


scoder:
Could you try to clarify in the PR description what this PR adds and how the list of exported and imported signature names changes? Remember that the description will become a stand-alone commit comment that should help future readers understand what happened and why it was done.

da-woods:
I've updated the description.

(and converted to draft because I've spotted that `cpdef` functions generate some unwanted names in the module dictionary and need to remove that)

scoder:
Thanks, that is much clearer. And it makes it obvious that we should do this.

scoder:
Postponing for the first alpha but would be good to have ready for the first beta.

da-woods:
My view is this is probably ready now. It includes https://github.com/cython/cython/pull/7778 which I think makes sense as a separate change.

scoder:
Really good idea. That's a much better naming scheme.

PR #7659, opened 2026-05-04
Adapt array module declarations Python 3.15 changes

da-woods:
In https://github.com/python/cpython/pull/148676 the format of the `arraydescr` struct get changed, which meant that our representation was outdated and crashed.

It isn't easy to adapt it perfectly for both (if users are manually using the `arraydescr` type anyway) but we can make our `array.array.__getbuffer__` function not crash by hiding some of the details in C code.

### DIFF
--- a/Cython/Includes/cpython/array.pxd
+++ b/Cython/Includes/cpython/array.pxd
@@ -104,34 +104,6 @@ cdef extern from *:  # Hard-coded utility code hack.
         cdef inline __data_union data(self) noexcept nogil:
             return __Pyx_PyArray_Data(self)
 
-        def __getbuffer__(self, Py_buffer* info, int flags):
-            # This implementation of getbuffer is geared towards Cython
-            # requirements, and does not yet fulfill the PEP.
-            # In particular strided access is always provided regardless
-            # of flags
-            item_count = Py_SIZE(self)
-
-            info.suboffsets = NULL
-            info.buf = self.data.as_chars
-            info.readonly = 0
-            info.ndim = 1
-            info.itemsize = self.ob_descr.itemsize   # e.g. sizeof(float)
-            info.len = info.itemsize * item_count
-
-            info.shape = <Py_ssize_t*> PyObject_Malloc(sizeof(Py_ssize_t) + 2)
-            if not info.shape:
-                raise MemoryError()
-            info.shape[0] = item_count      # constant regardless of resizing
-            info.strides = &info.itemsize
-
-            info.format = <char*> (info.shape + 1)
-            info.format[0] = self.ob_descr.typecode
-            info.format[1] = 0
-            info.obj = self
-
-        def __releasebuffer__(self, Py_buffer* info):
-            PyObject_Free(info.shape)
-
     array newarrayobject(PyTypeObject* type, Py_ssize_t size, arraydescr *descr)
 
     __data_union __Pyx_PyArray_Data(array self) noexcept nogil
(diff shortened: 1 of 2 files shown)


Cython/Utility/arrayarray.h
```diff
@@ -23,13 +23,36 @@
 // below.  That's defined later because the appropriate get and set
 // functions aren't visible yet.
 typedef struct arraydescr {
-    int typecode;
```
da-woods, line 26: I think we were only getting away with using `int` for padding reasons, and the correct declaration had been `char` for the last 15 years

Cython/Utility/arrayarray.h
```diff
@@ -23,13 +23,36 @@
...
+#else
+    return 1;
+#endif
+}
+
+static CYTHON_INLINE void __Pyx_PyArrayDescr_copy_typecode(arraydescr *desc, char *dst) {
+#if PY_VERSION_HEX <= 0x030F00a8
+    strcpy(dst, desc->typecode);
+#else
+    dst[0] = arrayarray->typecode;
```
da-woods:
```suggestion
    dst[0] = desc->typecode;
```

da-woods:
The change doesn't actually help until beta1 is out and the version number gets bumped of course

scoder:
If https://github.com/python/cpython/pull/149455 goes in, the struct field will be changed to `char[3]` instead of a pointer, basically keeping the ABI intact (for the beginning of the struct, `formats` is still removed, which is probably fine).

Cython/Includes/cpython/array.pxd
```diff
@@ -70,11 +70,14 @@ cdef extern from *:  # Hard-coded utility code hack.
     ctypedef object GETF(array a, Py_ssize_t ix)
     ctypedef object SETF(array a, Py_ssize_t ix, object o)
     ctypedef struct arraydescr:  # [object arraydescr]:
-        char typecode
+        char typecode   # This is untrue in Python 3.15+ but it isn't easy to expose both
```
scoder: We can at least make it _possible_ for users to access the variants:
```suggestion
        char typecode   # backwards compatibility only
        char typecode_char "typecode"      # Python <= 3.14
        char typecode_array[3] "typecode"  # Python 3.15+
```

da-woods:
I wonder why we're reimplenting the buffer protocol at all here? This feels like a relic from some early version of Python before `array.array` implemented it themselves. 

My thought is that if we remove our implementation then the immediate problem (the CI crash...) goes away, and we can wait for Python to settle a bit before updating the structs.

I'm sure I'll find out why I'm wrong when I try it...

da-woods:
> I wonder why we're reimplenting the buffer protocol at all here? This feels like a relic from some early version of Python before `array.array` implemented it themselves.

I think we definitely can get rid of them, but that doesn't fix the crash.

da-woods:
This includes https://github.com/cython/cython/pull/7668 which simplifies it a lot because we no longer have to adapt `__getbuffer__`.  I think they're best merged separately though.

scoder:
Ah, yes, the C `union` is a good idea.

The release of Py3.15b1 is close, so let's wait for it to become available and rerun the tests against it.

da-woods:
> The release of Py3.15b1 is close, so let's wait for it to become available and rerun the tests against it.

I think with the changes to Python we no longer need this to avoid a crash, so it's just keeping the declarations in sync.  So it makes sense to wait.

scoder:
Py3.15b1 should get installed by `setup-python` now. I've merged the latest master to start the test runs.

scoder:
3.2.x commit: https://github.com/cython/cython/commit/a470483da2274ca272e70280fa27b5f7e90c9e09

PR #7660, opened 2026-05-05
[3.2.x] Separate wheel check step from wheel build step

da-woods:
Thinking about trying to improve supply chain security and pinning our build dependencies.

The dependencies that we use during the wheel build step are:
* cibuildwheel (which is pinned through the github action, so is hopefully relative secure)
* setuptools (which has no dependencies)
* abi3audit and its dependencies
* twine and its dependencies.

This PR moves abi3audit and twine to a separate step so there's no way they can influence the built wheels.

That leaves us with only setuptools, and managing a single pinned dependency seems pretty reasonable.

It's probably more paranoid than it needs to be (and no-one else seems to really be worried about test dependencies like this). However it's also fairly simple to do given Cython's limited dependencies.

`twine` may be a lost cause though because I suspect it's ultimately used to upload to PyPI. But I can't see that so I'm ignoring it.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -117,12 +117,7 @@ jobs:
         with:
           only: ${{ matrix.only }}
 
-      - name: Check wheel
-        run: |
-          python -m pip install twine
-          python -m twine check ./wheelhouse/*.whl
-
-      - uses: actions/upload-artifact@v7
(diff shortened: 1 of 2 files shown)


da-woods:
The wheels build successfully for me on https://github.com/da-woods/cython/actions/runs/25391461563 so I'll merge this

PR #7665, opened 2026-05-05
Pin build requirements versions

da-woods:
... and manage them with dependabot.

I've intentionally left the version in project.toml open and then installed it separately with a file used only on our CI (so as not to put artificial barriers on future users trying to install it somewhere we hadn't anticipated).

I tried using the argument to `build` `--dependency-constraints-txt`. That's unfortunately too new so didn't work.

Setting an environmental variable to `PIP_BUILD_CONSTRAINT` seemed like it might also work, but to me looked a bit more opaque than what I've done here.

For Cython 3.2.x we'll need a slightly more complicated requirements file with different versions for Python 3.8.


.github/wheel_build_constraints.txt
```diff
@@ -0,0 +1 @@
+setuptools == 82.0.1
```
da-woods, line 1: I'm not set on the path I've used here (.wheel_build_requirements). The main thing is that it should be unique because you can only point dependabot to a directory.

(Although we could just bundle it into the .github directory)

pyproject.toml
```diff
@@ -19,6 +19,12 @@ test-command = [
     "cython {project}/Cython/Compiler/Scanning.py",
     "cython {project}/Cython/Compiler/Parsing.py",
 ]
+# For cibuildwheel, use strict version constraints to ensure it's as reproducible as possible.
+# This means not installing from pyproject.toml but instead using a separate, more constrained file.
+# However, keep pyproject.toml build-system.requires open to not artificially constrain users
+# who may be trying to install on future versions of Python with unknown setuptools support. 
+build-frontend = "build; args: --no-isolation"
```
scoder: I understand the intention to not break existing build setups, but if you want reproducible builds then `--no-isolation` is really contrary.

It does look like `dependabot` understands `pyproject.toml` by now. Did you try that?
https://github.com/dependabot/dependabot-core/issues/3290

da-woods: > Did you try that?

I didn't try it (because I don't think there's a good way to test it in advance) but I believe it does.

I was trying to avoid putting a hard version pin in pyproject.toml because that forces the strict version check onto user when they probably don't care.

I agree about the no-isolation though.  That sounds like `PIP_BUILD_CONSTRAINT` might be the better solution here.

da-woods: The current version uses a separate file but doesn't break the isolation in the build process.

That may still be over-complicated, but I'd still prefer not to force a version on future users in pyproject.toml (just because it's always a pain to workaround if it becomes a problem in future).

.github/wheel_build_constraints.txt
scoder: > Although we could just bundle it into the .github directory

Yes, we should do that. We already have `cibuildwheel-requirement.txt` in there, we can just put this next to it. Ok, it can be more broadly used via `pyproject.toml`, but in the rare cases when users or distributions build their own Cython wheel, they can probably live with that place.

scoder:
The wheel builds failures might be unrelated. It seems that we didn't actually compile Cython as abi3 binaries in the Limited API CI jobs, thus missing a recent regression. I (hopefully) fixed the job config in
https://github.com/cython/cython/commit/cb07b17ce7f7d8f64942e60bb36ad16b23fa7699

EDIT: the failure is also in master.

PR #7668, opened 2026-05-06
Remove buffer protocol from array.array

da-woods:
I think this is a relic from when array didn't implement the buffer protocol itself.  Currently the functions look to be generated but unused.


scoder:
3.2.x commit: https://github.com/cython/cython/commit/31c7957d8fd062197c7cc38a07e4addfc0a11247

PR #7670, opened 2026-05-07
Fix walrus operator re-evaluated multiple times in isinstance() tuple optimization

P403n1x87:
When isinstance() is called with a walrus operator as the first argument and a tuple of types, e.g. ``isinstance(ni := next(it), (int, float, str, bytes))``, Cython's isinstance-tuple optimization expands this into a short-circuit OR of individual type checks. Each branch of that OR tree independently calls ``generate_evaluation_code`` on the first argument, causing the walrus expression to be re-evaluated once per type — calling ``next(it)`` four times per loop iteration instead of once.

Root cause: ``_handle_simple_function_isinstance`` wraps the arg in a ``ResultRefNode`` (evaluated-once cache) only when ``not arg.is_simple()``. A walrus operator reaches the optimizer wrapped in a ``CoerceToTempNode``, which reports ``is_simple() = True`` (because ``is_temp = True``), so the caching is incorrectly skipped.

Fix: unwrap any ``CoercionNode`` layers before the ``is_simple()`` check and force ``ResultRefNode`` caching when the underlying expression is an ``AssignmentExpressionNode``.


da-woods:
```
cdef class C:
    cdef object x

def f(x):
    return isinstance(C().x, (bool, list))
```

This looks to trigger the same bug to me. I suspect the approach you describe would fix it (but haven't actually checked). So it isn't entirely specific to the walrus operator.

Cython/Compiler/Optimize.py
```diff
@@ -2977,7 +2977,14 @@ def _handle_simple_function_isinstance(self, node, function, pos_args):
...
         if types and (allowed_none_node or len(types) > 1):
-            if arg.is_attribute or not arg.is_simple():
+            # CoerceToTempNode wrapping a walrus operator (AssignmentExpressionNode) reports
+            # is_simple()=True (is_temp=True), but each call to generate_evaluation_code()
+            # re-evaluates the underlying expression. Unwrap any CoercionNode layers to detect
+            # this case and force the arg into a ResultRefNode so it is evaluated only once.
+            inner = arg
+            while isinstance(inner, ExprNodes.CoercionNode):
+                inner = inner.arg
+            if arg.is_attribute or not arg.is_simple() or isinstance(inner, ExprNodes.AssignmentExpressionNode):
```
da-woods: I wonder if it would be easier to make a shorter list of what's allowed. Maybe

```
if not (arg.is_literal or (arg.is_name and not arg.is_temp)):
    # wrap in result-ref-node
```

It's doing the wrong this in quite a few cases. Even for a global name it's looking it up twice, which works but is not really optimal.

`is_simple` isn't really the right check here because it does catch most nodes that evaluate to temps (except call nodes which are special-cased)

scoder: It seems generally fuzzy what that positive list should look like, so I created https://github.com/cython/cython/issues/7671 for coming up with a general scheme to mark such nodes.

P403n1x87: would it make sense to always assign to a temp then, even when not strictly needed?

scoder: Unnecessary temp assignment is harmful for Python objects because it requires reference counting.

P403n1x87: > Unnecessary temp assignment is harmful for Python objects because it requires reference counting.

I see, so I guess there are cases where the temp can outlive the original reference?

da-woods: > I see, so I guess there are cases where the temp can outlive the original reference?

There are certain cases in general - and this is one of the reasons we use temps. I'm not sure if there's any case here.

But our temporary handling is built around the idea that the temporary owns a reference. Usually this is correct anyway (e.g. the result of function call is an owned reference).  But for something simple like a function local variable it would add an incref and decref.

da-woods:
I think we should accept this PR, and then worry about the general problem in https://github.com/cython/cython/issues/7671 later.  Provided we do the right thing with `NameNode` I think the most useful thing remains optimized.

scoder:
3.2.x commit is https://github.com/cython/cython/commit/80d9e7e307e751e2b4da4948edec5626d380896e

PR #7676, opened 2026-05-09
Fix functools.wraps on cyfunctions on 3.14+

### LINKED ISSUE #7675: [BUG] Python 3.14: typing.get_type_hints() returns different dict for cythonized decorated async functions
### Describe the bug

Source code running result (python 3.12 and python 3.14):

```python
import test_cython_annotation
111 {'path': <class 'int'>, 'body': <class 'str'>}
222 {'path': <class 'int'>, 'body': <class 'str'>}
```

Compile and run (python 3.12):

```python
import test_cython_annotation
111 {'path': <class 'int'>, 'body': <class 'str'>}
222 {'path': <class 'int'>, 'body': <class 'str'>}
```

Compile and run (python 3.14):

```python
import test_cython_annotation
111 {'a': <class 'int'>, 'b': <class 'str'>}
222 {'path': <class 'int'>, 'body': <class 'str'>}
```

**setup.py**

```python
import sys

if len(sys.argv) == 1:
    sys.argv.extend(["build_ext", "--inplace", "-j8"])

extensions = [
    Extension(
        "test_cython_annotation",
        ["test_cython_annotation.py"],
    )
]

setup(
    name="setup",
    ext_modules=cythonize(
        module_list=extensions,
        compiler_directives={'language_level': "3"},
        build_dir="build/c",
        force=True
    ),
    zip_safe=False
)


```

**test_cython_annotation.py**

```python
from typing import get_type_hints
from functools import wraps


def deco(func):
    @wraps(func)
    def wrapper(a:int, b:str):
        print(222, get_type_hints(func))
        return func(a, b)

    return wrapper


@deco
def test(path: int, body: str):
    pass


print(111, get_type_hints(test))
test(1, "2")
```

### Expected behaviour

The output result of python3.14 is the same as python3.12.

### OS

windows

### Python version

3.12 &  3.14

### Cython version

3.2.4

### Additional context

_No response_

da-woods:
... by providing basic `__annotate__` implementation.

Fixes #7675.

The issue is that in 3.14, `wraps` reassigns `__annotate__` rather than `__annotations__`. The basic implementation I provide is a function that just returns the `__annotations__` dictionary (stored at the time it was created),
while setting `__annotate__` just calls the function and then sets the `__annotations__` to the result. This should also stand a decent chance of working with non-cyfunctions.

My view is that `wraps` working correctly is a desirable property and should be tested.


scoder:
3.2.x commit is here: https://github.com/cython/cython/commit/98635e5e8058197bab6868002990b474304e818c

PR #7677, opened 2026-05-09
[3.2.x] Fix fastcall nargs in coroutine throw

da-woods:
I believe nargs to vectorcall is taken as fact, and it doesn't generally handle `NULL` values as part of the array.

Targeted at 3.2.x because I have a slightly bigger change aimed at 3.3 which includes the same fix.

### DIFF
--- a/Cython/Utility/Coroutine.c
+++ b/Cython/Utility/Coroutine.c
@@ -1311,9 +1311,10 @@ static PyObject *__Pyx__Coroutine_Throw(PyObject *self, PyObject *typ, PyObject 
             if (likely(args)) {
                 ret = __Pyx_PyObject_Call(meth, args, NULL);
             } else {
-                // "tb" or even "val" might be NULL, but that also correctly terminates the argument list
+                // "tb" or even "val" might be NULL.
                 PyObject *cargs[4] = {NULL, typ, val, tb};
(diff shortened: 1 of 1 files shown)


Cython/Utility/Coroutine.c
```diff
@@ -1311,9 +1311,10 @@ static PyObject *__Pyx__Coroutine_Throw(PyObject *self, PyObject *typ, PyObject
             if (likely(args)) {
                 ret = __Pyx_PyObject_Call(meth, args, NULL);
             } else {
-                // "tb" or even "val" might be NULL, but that also correctly terminates the argument list
+                // "tb" or even "val" might be NULL.
                 PyObject *cargs[4] = {NULL, typ, val, tb};
-                ret = __Pyx_PyObject_FastCall(meth, cargs+1, 3 | __Pyx_PY_VECTORCALL_ARGUMENTS_OFFSET);
+                Py_ssize_t nargs = 1 + (val != NULL) + (tb != NULL);
+                ret = __Pyx_PyObject_FastCall(meth, cargs+1, nargs | __Pyx_PY_VECTORCALL_ARGUMENTS_OFFSET);
```
scoder, line 1317: I was wondering if this is safe or if `val` could be `NULL` without `tb` being `NULL`, so I looked through the usages and found that all usages of `__Pyx__Coroutine_Throw()` are directly or indirectly preceded by a call to `PyArg_UnpackTuple()` to get the exception info, which unpacks left-to-right.

Looks good then.

scoder (approved):

PR #7678, opened 2026-05-09
Switch to vectorcall for coroutine methods where possible

da-woods:
It's likely faster most of the time by avoiding creating a tuple without good reason.

Also switch a cyfunction method too (although in this case to NOARGS, again to avoid a tuple).

### DIFF
--- a/Cython/Utility/Coroutine.c
+++ b/Cython/Utility/Coroutine.c
@@ -1363,7 +1363,6 @@ static PyObject *__Pyx_Coroutine_Throw(PyObject *self,
     if (unlikely(!PyArg_UnpackTuple(args, "throw", 1, 3, &typ, &val, &tb)))
         return NULL;
 #else
-    nargs = PyVectorcall_NARGS(nargs);
     switch (nargs) {
     case 3:
         tb = args[2];
@@ -1939,7 +1938,7 @@ static PyObject *__Pyx_Coroutine_await(PyObject *coroutine) {
 static PyMethodDef __pyx_Coroutine_methods[] = {
     {"send", (PyCFunction) __Pyx_Coroutine_Send, METH_O,
      PyDoc_STR("send(arg) -> send 'arg' into coroutine,\nreturn next iterated value or raise StopIteration.")},
-    {"throw", (PyCFunction)(void (*)(void))__Pyx_Coroutine_Throw,
+    {"throw", __PYX_REINTERPRET_FUNCION(PyCFunction, __Pyx_Coroutine_Throw),
 #if CYTHON_COMPILING_IN_LIMITED_API && __PYX_LIMITED_VERSION_HEX < 0x030A0000
         METH_VARARGS,
 #else
--- a/Cython/Utility/Coroutine.c
+++ b/Cython/Utility/Coroutine.c
@@ -1874,7 +1874,7 @@ static PyObject *__Pyx_CoroutineAwait_no_new(PyTypeObject *type, PyObject *args,
 static PyMethodDef __pyx_CoroutineAwait_methods[] = {
     {"send", (PyCFunction) __Pyx_CoroutineAwait_Send, METH_O,
      PyDoc_STR("send(arg) -> send 'arg' into coroutine,\nreturn next yielded value or raise StopIteration.")},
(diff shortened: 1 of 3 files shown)


Cython/Utility/Coroutine.c
```diff
@@ -1310,12 +1329,16 @@ static PyObject *__Pyx__Coroutine_Throw(PyObject *self, PyObject *typ, PyObject
...
                 ret = __Pyx_PyObject_Call(meth, args, NULL);
-            } else {
-                // "tb" or even "val" might be NULL, but that also correctly terminates the argument list
+            }
+#endif
+            else {
+                // "tb" or even "val" might be NULL
                 PyObject *cargs[4] = {NULL, typ, val, tb};
-                ret = __Pyx_PyObject_FastCall(meth, cargs+1, 3 | __Pyx_PY_VECTORCALL_ARGUMENTS_OFFSET);
+                Py_ssize_t nargs = 1 + (val != NULL) + (tb != NULL);
```
da-woods, line 1324: I believe this is a bugfix and the 3.2.x PR is https://github.com/cython/cython/pull/7677

Cython/Utility/Coroutine.c
```diff
@@ -1263,7 +1269,12 @@ __Pyx_Coroutine_Close(PyObject *self, PyObject **retval) {
 }
 
 static PyObject *__Pyx__Coroutine_Throw(PyObject *self, PyObject *typ, PyObject *val, PyObject *tb,
-                                        PyObject *args, int close_on_genexit) {
+#if CYTHON_COMPILING_IN_LIMITED_API && __PYX_LIMITED_VERSION_HEX < 0x030A0000
+                                        PyObject *args,
+                                        // On higher versions, args are in typ, val, tb
+                                        // and there's no optimization to be gained by passing the tuple through.
+#endif
```
da-woods: Thinking about it more, I might revert this change because it might still be useful in one place or higher versions

Cython/Utility/Coroutine.c
```diff
@@ -1891,7 +1939,12 @@ static PyObject *__Pyx_Coroutine_await(PyObject *coroutine) {
 static PyMethodDef __pyx_Coroutine_methods[] = {
     {"send", (PyCFunction) __Pyx_Coroutine_Send, METH_O,
      PyDoc_STR("send(arg) -> send 'arg' into coroutine,\nreturn next iterated value or raise StopIteration.")},
-    {"throw", (PyCFunction) __Pyx_Coroutine_Throw, METH_VARARGS,
+    {"throw", (PyCFunction)(void (*)(void))__Pyx_Coroutine_Throw,
```
scoder:
```suggestion
    {"throw", __PYX_REINTERPRET_FUNCION(PyCFunction, __Pyx_Coroutine_Throw),
```

scoder (commented):
Don't we need to guard all of this with `#if CYTHON_VECTORCALL`?

da-woods:
> Don't we need to guard all of this with `#if CYTHON_VECTORCALL`?

You can declare methods with the `METH_FASTCALL` flag from Python 3.10 in the Limited API (and in every version we care about outside the Limited API).

What you can't do is use `PyObject_Vectorcall` until 3.12 (and this is what `CYTHON_VECTORCALL` guards). I need to remove the uses of `PyVectorcall_NARGS` though to make it build which `METH_FASTCALL` doesn't need. 

I think it still makes sense to do even without the ability to do vectorcall calls because there's enough paths through `__Pyx__Coroutine_Throw` that don't end in just forwarding the call.

Cython/Utility/Coroutine.c
```diff
@@ -1341,15 +1348,45 @@ static PyObject *__Pyx__Coroutine_Throw(PyObject *self, PyObject *typ, PyObject
...
+) {
     PyObject *typ;
     PyObject *val = NULL;
     PyObject *tb = NULL;
 
+#if CYTHON_COMPILING_IN_LIMITED_API && __PYX_LIMITED_VERSION_HEX < 0x030A0000
     if (unlikely(!PyArg_UnpackTuple(args, "throw", 1, 3, &typ, &val, &tb)))
         return NULL;
+#else
+    nargs = PyVectorcall_NARGS(nargs);
```
da-woods:
```suggestion
```

(Breaks limited API, not needed)

PR #7682, opened 2026-05-11
Fix issue with GIL being enabled in abi3t extensions

ngoldbaum:
In my local tests this is needed to prevent the GIL from being enabled in a test extension.


da-woods:
Thanks. I think that was originally in there but probably got lost in the giant merge mess to update that branch

da-woods:
I've removed the 3.3 milestone because this went into my "freethreading-limited-api-preview" branch so isn't currently in 3.3

PR #7686, opened 2026-05-12
Infer subscripted types of initialised variables 

matusvalo:
This PR implements following:
* allow to infer types of variables initialized by `tuple()`, `frozenset()` and `frozendict()` functions
* Allow to infer types variables initialized by tuple `(...)`
* Mutable container types are not implicitly inferred (must be explicitly annotated):
  ```python
  a = [1, 2, 3]  # a has type list
  ```
* Nested mutable container types are not implicitly inferred too:
  ```python
  a = (1, 2, ['a', 'b'])  # a has type tuple[int, int, list]
  ```


matusvalo:
There is one additional issue. This PR compiles `Code.py` with following line:
```c
__pyx_t_3 = PyObject_Length(__pyx_v_int_constants_by_bytesize);
```
instead of what we have in master:
```c
__pyx_t_3 = __Pyx_PyList_GET_SIZE(__pyx_v_int_constants_by_bytesize);
```

The affected line is following: https://github.com/cython/cython/blob/7dda0262a7954a076129d2b4f105696e75eb3bce/Cython/Compiler/Code.py#L2391

matusvalo:
> There is one additional issue. This PR compiles `Code.py` with following line:
> 
> ```c
> __pyx_t_3 = PyObject_Length(__pyx_v_int_constants_by_bytesize);
> ```
> 
> instead of what we have in master:
> 
> ```c
> __pyx_t_3 = __Pyx_PyList_GET_SIZE(__pyx_v_int_constants_by_bytesize);
> ```
> 
> The affected line is following:
> 
> https://github.com/cython/cython/blob/7dda0262a7954a076129d2b4f105696e75eb3bce/Cython/Compiler/Code.py#L2391

OK culprit is not related to this PR but here: https://github.com/cython/cython/blob/7dda0262a7954a076129d2b4f105696e75eb3bce/Cython/Compiler/Optimize.py#L2885-L2895

Cython/Compiler/ExprNodes.py
```diff
@@ -9251,8 +9251,12 @@ def type_dependencies(self, env):
         return ()
 
     def infer_type(self, env):
-        # TODO: Infer non-object list arrays.
-        return list_type
+        typ = list_type
+        if len(self.args) == 0:
+            return typ
+        item_type = reduce(
+            lambda a, b: a if a == b else None, (arg.type for arg in self.args))
```
scoder: I think this should rather look for common types than doing a simple equality test.

Also, since this is used a couple of times, it seems worth factoring out a function into the respective corner of `PyrexTypes.py`.

scoder: I merged https://github.com/cython/cython/pull/7644 which adds `PyrexTypes.reduce_spanning_types()`.

matusvalo:
Coercion of constants in list was implemented. We must coerce constatns to python object, otherwise the assigning to the list is failing during C compilation.

tests/run/pep526_variable_annotations_cy.pyx
```diff
@@ -210,9 +210,154 @@ def common_container_type(cond, list[float] a, list[int] b):
...
+    ('set[int object] object', 'int object')
+    ('set[int object] object', 'int object')
+    ('set[int object] object', 'int object')
+    Testing set:
+    set object
+    Testing set with mixed types:
+    ('set object', 'Python object')
+    ('set object', 'Python object')
+    ('set object', 'Python object')
+    Testing set with inferrable function call:
```
scoder:
```suggestion
    Testing set with inferable function call:
```

tests/run/pep526_variable_annotations_cy.pyx
```diff
@@ -210,9 +210,154 @@ def common_container_type(cond, list[float] a, list[int] b):
...
+    for i1 in s1:
+        print(cython.typeof(s1), cython.typeof(i1))
+    print("Testing set:")
+    s2 = set()
+    print(cython.typeof(s2))
+    print("Testing set with mixed types:")
+    s3 = {1, 3.0, "5"}
+    for i3 in s3:
+        print(cython.typeof(s3), cython.typeof(i3))
+    print("Testing set with inferrable function call:")
```
scoder:
```suggestion
    print("Testing set with inferable function call:")
```

tests/run/pep526_variable_annotations_cy.pyx
```diff
@@ -210,9 +210,154 @@ def common_container_type(cond, list[float] a, list[int] b):
...
+    (5, 'list[int object] object', 'int object')
+    [1, 0, 5]
+    Testing list:
+    list object
+    Testing list with mixed types:
+    (1, 'list object', 'Python object')
+    (3.0, 'list object', 'Python object')
+    ('5', 'list object', 'Python object')
+    [1, 0, '5']
+    Testing list with inferrable function call:
```
scoder:
```suggestion
    Testing list with inferable function call:
```

tests/run/pep526_variable_annotations_cy.pyx
```diff
@@ -210,9 +210,154 @@ def common_container_type(cond, list[float] a, list[int] b):
...
+    print("Testing list:")
+    l2 = []
+    print(cython.typeof(l2))
+    print("Testing list with mixed types:")
+    l3 = [1, 3.0, "5"]
+    for i3 in l3:
+        print(i3, cython.typeof(l3), cython.typeof(i3))
+    l3[1] = 0
+    print(l3)
+    print("Testing list with inferrable function call:")
```
scoder:
```suggestion
    print("Testing list with inferable function call:")
```

tests/run/pep526_variable_annotations_cy.pyx
```diff
@@ -210,9 +210,154 @@ def common_container_type(cond, list[float] a, list[int] b):
...
+    ('dict[Python object,int object] object', 'Python object', 'int object')
+    Testing dict with mixed value types:
+    ('dict[int object,Python object] object', 'int object', 'Python object')
+    ('dict[int object,Python object] object', 'int object', 'Python object')
+    ('dict[int object,Python object] object', 'int object', 'Python object')
+    Testing dict with mixed key and value types:
+    ('dict object', 'Python object', 'Python object')
+    ('dict object', 'Python object', 'Python object')
+    ('dict object', 'Python object', 'Python object')
+    Testing dict with inferrable function call:
```
scoder:
```suggestion
    Testing dict with inferable function call:
```

tests/run/pep526_variable_annotations_cy.pyx
```diff
@@ -210,9 +210,154 @@ def common_container_type(cond, list[float] a, list[int] b):
...
+    d4 = {1: 1.0, 3: "b", 5: "c"}
+    for k4 in d4:
+        print(cython.typeof(d4), cython.typeof(k4), cython.typeof(d4[k4]))
+
+    print("Testing dict with mixed key and value types:")
+    d5 = {1: 1.0, 3.0: "b", "5": "c"}
+    for k5 in d5:
+        print(cython.typeof(d5), cython.typeof(k5), cython.typeof(d5[k5]))
+
+    print("Testing dict with inferrable function call:")
```
scoder:
```suggestion
    print("Testing dict with inferable function call:")
```

Cython/Compiler/ExprNodes.py
matusvalo: `==` replaced with `reduce_spanning_types()`

matusvalo: I moved the logic to shared function, but not to PyrexTypes.py though because I the new function's logic is working with nodes. The name is **bad**. But I am bad at naming thinks. Please give me a hint hot o call the function...

matusvalo:
OK I implemented the comments. But there will be needed some cosmetics though. Otherwise the PR just waits for #7689 in order to have CI/CD green.

Cython/Compiler/ExprNodes.py
```diff
@@ -9252,7 +9258,10 @@ def type_dependencies(self, env):
         return ()
 
     def infer_type(self, env):
-        # TODO: Infer non-object list arrays.
```
scoder, line 9376: This comment seems still valid.

Cython/Compiler/ExprNodes.py
```diff
@@ -9910,7 +9926,11 @@ def type_dependencies(self, env):
         return ()
 
     def infer_type(self, env):
-        # TODO: Infer struct constructors.
```
scoder, line 10254: This comment seems still valid.

Cython/Compiler/ExprNodes.py
```diff
@@ -325,6 +325,13 @@ def translate_double_cpp_exception(code, pos, lhs_type, lhs_code, rhs_code, lhs_
     code.putln(code.error_goto(pos))
     code.putln('}')
 
+def infer_container_type(env, container_nodes):
+    for arg in container_nodes:
+        arg.type = arg.infer_type(env)
+    return PyrexTypes.reduce_spanning_types(
+        arg.type if arg.type.is_pyobject else arg.coerce_to_pyobject(env).type for arg in container_nodes
+    )
```
scoder: It's not inferring the type of the container but that of its items, so a better name is `infer_common_container_item_type`.

```suggestion
def infer_common_container_item_type(env, item_nodes):
    for node in item_nodes:
        node.type = node.infer_type(env)
    return PyrexTypes.reduce_spanning_types(
        arg.type if arg.type.is_pyobject else arg.coerce_to_pyobject(env).type for arg in item_nodes
    )
```

However, it also doesn't just infer the types but assign them back to the nodes, so it's more of an analysis than pure inference function.

And it generates intermediate throw-away coercion nodes just to get at their type. I don't think we should take that route. Creating the (then used) coercion nodes is fine for `analyse_types()`, but seeing it in `infer_type()` makes me frown. I also don't see why it should be needed. `spanning_type()` looks at the `equivalent_type` internally for C-to-Python coercion, that should be enough.

In the end, I think we're left with the following, and I'm not sure that's worth a separate function. I'd just fold it back into the places where it's used. The code seems more obvious in what it does without the indirection of a function name.

```suggestion
def infer_common_container_item_type(env, item_nodes):
    return PyrexTypes.reduce_spanning_types(
        [node.infer_type(env) for node in item_nodes])
```

scoder (commented):
The tests that you added do not test variable annotations but type inference. Please move them to a type inference test module. If there isn't one yet that tests specifically subscripted containers, I think we can afford to add a new one.

Cython/Compiler/ExprNodes.py
matusvalo: Fixed.

Cython/Compiler/ExprNodes.py
matusvalo: Fixed.

Cython/Compiler/ExprNodes.py
matusvalo: > However, it also doesn't just infer the types but assign them back to the nodes, so it's more of an analysis than pure inference function.

Assigning is needed because otherwise following code:

```python
l1 = [1, 3, 5]
l1[1] = 0
```

Generates C code which compiling fails with following:
```
/home/dev/cython/test/a.c: In function ‘__pyx_pf_1a_main’:
/home/dev/cython/test/a.c:2437:3: error: ‘None’ undeclared (first use in this function)
 2437 |   None = 0;
      |   ^~~~
/home/dev/cython/test/a.c:2437:3: note: each undeclared identifier is reported only once for each function it appears in
error: command '/usr/bin/cc' failed with exit code 1
```

Maybe we need to fix code elsewhere but just noting about that...

scoder: Regardless, the assignment shouldn't happen during type inference. Type inference should never change node status.

During type analysis or coercion, you can coerce each item to the expected (previously inferred) item type of the container. That might be enough to resolve this issue.

I still think that my second proposal should be right, then you can continue from there.

matusvalo: > I still think that my second proposal should be right, then you can continue from there.

Currently, I was mainly working on finishing the test suit and some refactor of the code. As next step, I plan to revisit your comments ...

tests/run/type_inference_generics.py
```diff
@@ -111,3 +113,271 @@ def test_dict_builtin_methods():
...
+    s7 = frozenset(arg_b)
+    print(cython.typeof(s7) + (' object' if not cython.compiled else ''))
+
+    print("* nested container:")
+    # FIXME: the result should be:
+    # frozenset[tuple[int object, int object] object] object tuple[int object, int object] object
+    s8 = frozenset({(1, 2), (3, 4)})
+    for i8 in s8:
+        print(
+            cython.typeof(s8) + ('[tuple object] object' if not cython.compiled else ''),
```
matusvalo: Inference of ctuples is not 100% ideal. Imrpovement here could be inferring subscripted types of tuple: `tuple[int, int] object` instead of `tuple object` but, currently, I would not complicate more this PR...

Cython/Compiler/ExprNodes.py
matusvalo: > I still think that my second proposal should be right, then you can continue from there.

Implemented and it works with charm. Even it fixed other pending issues! ❤️

tests/run/type_inference_generics.py
matusvalo: now fixed with [f17b418](https://github.com/cython/cython/pull/7686/commits/f17b4188bb107d33e904e9fbe25866b2597598ab)

matusvalo:
OK marking as ready for review.

Cython/Compiler/PyrexTypes.py
```diff
@@ -5402,12 +5417,18 @@ class BuiltinTypeConstructorObjectType(BuiltinObjectType, PythonTypeConstructorM
...
+    _immutable_types = {'tuple', 'frozenset', 'frozendict'}
+
     def __init__(self, name, cname, objstruct_cname=None, **kwargs):
         super().__init__(
             name, cname, objstruct_cname=objstruct_cname)
         self.set_python_type_constructor_name(self.get_container_type().name)
         for attr_name, value in kwargs.items():
             setattr(self, attr_name, value)
+        if name in self._immutable_types:
+            self.is_immutable = True
```
scoder: -> `_builtin_type_flag_mapping`

Cython/Compiler/ExprNodes.py
```diff
@@ -9207,8 +9232,7 @@ def infer_type(self, env):
         if self.mult_factor or not self.args:
             return tuple_type
         arg_types = [arg.infer_type(env) for arg in self.args]
-        if any(type.is_pyobject or type.is_memoryviewslice or type.is_unspecified or type.is_fused
-               for type in arg_types):
+        if len(arg_types) > 0:
```
```suggestion
        if arg_types:
```

Cython/Compiler/ExprNodes.py
```diff
@@ -9368,14 +9392,20 @@ class ListNode(SequenceNode):
...
 
     def infer_type(self, env):
-        # TODO: Infer non-object list arrays.
+        if len(self.args) > 0 and self.read_only:
+            item_type = PyrexTypes.reduce_spanning_types(
+                [node.infer_type(env) for node in self.args]
+            )
+            if item_type is not py_object_type:
+                return list_type.specialize_here(self.pos, env, [item_type])
         return list_type
```
scoder: This is repeated three times and seems worth a helper function like
`def specialise_builtin_container_type(pos, env, container_type, items)`
called as
```python
if self.read_only and self.args:
    return specialise_builtin_container_type(self.pos, list_type, self.args)
return list_type
```
EDIT: I think `specialise_builtin_container_type` is a clearer name than just `specialise_builtin_container`.

tests/run/type_inference_generics.py
```diff
@@ -111,3 +113,261 @@ def test_dict_builtin_methods():
     print(cython.typeof(di) + ('[tuple[str object,int object] object] object' if not cython.compiled else ''))
     for k in di:
         print(cython.typeof(k) + ('[str object,int object] object' if not cython.compiled else ''), k)
+
+
+def bar():
+    return object()
```
scoder: `bar` doesn't say anything about its purpose in the test.
```suggestion
def new_object():
    return object()
```

Cython/Compiler/ExprNodes.py
matusvalo: fixed.

tests/run/type_inference_generics.py
matusvalo: fixed.

Cython/Compiler/PyrexTypes.py
matusvalo: Yes makes sense. But I wonder should we move `is_immutable` attribute move under `BuiltinObjectType` (currently it is under `BuiltinTypeConstructorObjectType` and set immutable flag also for `bytes`, `str`, `int`, etc. since they are immutable too. Currently, it is not used in Cython code but it conceptually makes sense...

Cython/Compiler/ExprNodes.py
matusvalo: I implemented the change, but I would propose following:
1. name the function `specialise_builtin_sequence_type` instead of `specialise_builtin_container_type`
2. create additional function  `specialise_builtin_mapping_type` (the logic of `DictNode.infer_type()`)
3. ~~Move both functions from 1. and 2. to `PyrexTypes.BuiltinTypeConstructorObjectType` class as class/static methods.~~

the goal of 1.-3. is just place code in better places. What do you think?

Update: Never mind step 3 is working with nodes too so it makes no sense to move it to PyrexTypes I suppose.
Update 2: Without step 3, step2 not sure if it is worth it if it is only on one single place... Maybe ignore my whole comment.

Cython/Compiler/PyrexTypes.py
matusvalo: OK after moving to `_builtin_type_flag_mapping` I was forced to add `is_immutable` flat to `PyrexType` class hence, there is no reason not to make the flag consistent across all types...

Cython/Compiler/PyrexTypes.py
```diff
@@ -1866,18 +1868,21 @@ class BuiltinObjectType(PyObjectType):
...
-        'frozendict': ['is_pyfrozendict_type', 'is_pyanydict_type', 'supports_container_type'],
+        'frozendict': ['is_pyfrozendict_type', 'is_pyanydict_type', 'supports_container_type', 'is_immutable'],
         'set': ['is_pyset_type', 'is_pyanyset_type', 'supports_container_type', 'has_uniform_element_type'],
-        'frozenset': ['is_pyfrozenset_type', 'is_pyanyset_type', 'supports_container_type', 'has_uniform_element_type'],
-        'bytes': ['is_pybytes_type', 'is_builtin_sequence', 'is_bytes_or_str_or_bytearray'],
-        'str': ['is_pystr_type', 'is_builtin_sequence', 'is_bytes_or_str_or_bytearray'],
+        'frozenset': [
+            'is_pyfrozenset_type', 'is_pyanyset_type', 'supports_container_type', 'has_uniform_element_type',
+            'is_immutable'
+        ],
```
scoder, line 1884:
```suggestion
        'frozenset': [
            'is_pyfrozenset_type', 'is_pyanyset_type', 'supports_container_type', 'has_uniform_element_type',
            'is_immutable',
        ],
```

Cython/Compiler/ExprNodes.py
```diff
@@ -6422,6 +6423,39 @@ class SimpleCallNode(CallNode):
     analysed = False
     overflowcheck = False
 
+    def infer_type(self, env):
+        infered_type = super().infer_type(env)
```
```suggestion
        inferred_type = super().infer_type(env)
```

Cython/Compiler/ExprNodes.py
```diff
@@ -6422,6 +6423,39 @@ class SimpleCallNode(CallNode):
...
+                function.entry.type and
+                function.entry.type.supports_container_type
+        ):
+            pass
+        else:
+            return infered_type
+
+        if infered_type.supports_container_type and infered_type.is_immutable and self.args and len(self.args) == 1:
+            if isinstance(self.args[0], (SetNode, DictNode, ListNode)):
+                self.args[0].read_only = True
```
scoder: As I wrote before, `.infer_type()` should not have side-effects.

I think we can avoid this attribute. It's fine to _always_ infer the specialised result type. As long as there is no assignment, the compile time known type won't change and specialised types can happily make their way through arbitrarily long expressions. `frozenset([1,2]) | frozenset([3, 4])` is definitely a `frozenset[int]`. But `x = (the above) ; x += frozenset(['a'])` invalidates the item type, and that is fine. It's the assignment that matters.

I'm not sure but we might have to change the algorithm in `TypeInference.py` to strip the specialisations off from types when inferring specialised mutable container types for variables, because from the assignment on, the item types are no longer safe assumptions (as long as we only look at variable assignments during type inference and not also item assignments, appends, calls, etc., which is really complex).

That said, I don't think it's a safe assumption that the type specialisation coming out of an arbitrary call is the same as what went it. Imagine this:
```python
def append_one(t) -> tuple:
    return tuple(t) + (1,)

def inferred():
    x = append_one(['a', 'b'])
```
If I understand correctly (and I might be misreading your code or intent), your implementation would infer `tuple[str, ...]` for `x`, which is incorrect.

matusvalo: > If I understand correctly (and I might be misreading your code or intent), your implementation would infer tuple[str, ...] for x, which is incorrect.

No it the code you provided it infers as `Python object` (for `def` function return annotation is ignored) and if I change `append_one()` to `cfunc`, `x` is of type `tuple`.

Let me add this case to test suite to prove my point.

EDIT: voila here is commit adding the test: cfae86644ec1395f659f6bc35fe24098d56aa3fd

matusvalo: > As I wrote before, .infer_type() should not have side-effects.

My apologies that you need to repeat yourself. I will try to fix that.

matusvalo: >  It's fine to always infer the specialised result type. As long as there is no assignment, the compile time known type won't change and specialised types can happily make their way through arbitrarily long expressions.

Hmm not sure about it. I did based on your proposal - see diff below (I also removed all occurences of `read_only` attribute I added before):

```diff
diff --git a/Cython/Compiler/TypeInference.py b/Cython/Compiler/TypeInference.py
index 7b2c5c816..75cfef502 100644
--- a/Cython/Compiler/TypeInference.py
+++ b/Cython/Compiler/TypeInference.py
@@ -540,6 +540,11 @@ def simply_type(result_type):
     result_type = PyrexTypes.remove_cv_ref(result_type, remove_fakeref=True)
     if result_type.is_array:
         result_type = PyrexTypes.c_ptr_type(result_type.base_type)
+    if result_type.supports_container_type and not result_type.is_immutable:
+        # When inferring type of mutable container, we must not infer
+        # subscripted types since container can have added objects with different
+        # type later during execution of code.
+        result_type = result_type.get_container_type()
     return result_type

 def aggressive_spanning_type(types, might_overflow, scope):
```

For assignment it works with charm:

```python
x = [1, 2, 3]  # x has type `list` 
```

But the issue is with **nested** types like:

```python
x = ([1], [2])
```
In this case `x` should have type `tuple[list, list]` but it has type `tuple[list[int], list[int]]` and this is dangerous because later the nested object can be modified - see following code:

```python
x = ([1], [2])
x[0].append('a')
x[1].append('b')
print(cython.typeof(x))
print(cython.typeof(x[0]))
print(cython.typeof(x[1]))
print(x)
```
produces following:
```
tuple[list[long] object,list[long] object] object
list[long] object
list[long] object
([1, 'a'], [2, 'b'])
```

Issue #7684, opened 2026-05-12
[ENH] Infer item type of comprehensions


scoder:
### Describe your issue

Cython should infer the item type of comprehensions, if possible, as a specialisation of the container type. For example:
```cython
x: list[int] = …
result = [ (a, a+1) for a in x ]
print(cython.typeof(result)
```
should print `list[tuple[int, int]]` in the best case, but should at least show that `result` is a list of tuples.

omkar-334:
hey @scoder can i work on this?

scoder:
Sure. You need to change `.infer_type()` for the comprehension nodes in `ExprNodes.py` to infer the type of the appended items. I suggest adding a new test file for this, e.g. `tests/run/comprehension_type_inference.pyx` since it's testing a specific intersection of features and needs a number of tests that cover the different comprehension types and different types of items (different Python types, C numbers, …). Ask here if you need more advice.

scoder:
Actually, let's make it a convention to move the test category first (if there is one), i.e. `type_inference_comprehensions.pyx`. That sorts better in the file system.

PR #7687, opened 2026-05-13
Infer item type of comprehensions 

omkar-334:
This PR fixes the issue #7684 where `cython.typeof([(a, a+1) for a in x])` reports a bare `list object` even when the items have a known shape.   and Adds tests for this in  `tests/run/type_inference_comprehensions.pyx`.  

`ComprehensionNode.infer_type` was a one-line stub that returned `self.type`. Now it looks at the body expression (`self.append.expr` for list/set, `key_expr`/`value_expr` for dict), infers each item's type, and calls `specialize_here` on the container. Falls back to the bare container whenever any item is unresolvable, so no path can regress.  

After the change:  

```python
[(a, a + 1) for a in x]      # list[tuple object] object   (was: list object)
[42 for _ in range(3)]        # list[long] object
{str(i): 42 for i in range(2)}  # dict[str object,long] object
```


omkar-334:
Testing script:  
```bash
#!/usr/bin/env bash
# Demo: comprehension item-type inference. Run from the repo root.

T=$(mktemp -d)

cat > $T/issue_example.pyx <<'EOF'
# cython: language_level=3, infer_types=True
from cython cimport typeof

def issue_example():
    cdef list[int] x = [1, 2, 3, 4]
    result = [(a, a + 1) for a in x]
    return result, typeof(result)
EOF

PYTHONPATH=$(pwd) /tmp/cy313venv/bin/python \
    -m Cython.Build.Cythonize -i $T/issue_example.pyx >/dev/null 2>&1

/tmp/cy313venv/bin/python - <<PYEOF
import sys
sys.path.insert(0, "$T")
from issue_example import issue_example
result, type_str = issue_example()
print(result)
print(type_str)
PYEOF

```  
Output  
Screenshot 2026-05-13 at 2 05 58 PM

omkar-334:
Do we need to add more tests for this?  I removed a few from my original commit because i thought they were not needed.

omkar-334:
Also, to achieve the best case (getting `list[tuple[int, int]]`) -do I need to change the `infer_type` method of `TupleNode`, `ListNode`, `SetNode` and `DictNode`?

FOr `ListNode` - `Cython/Compiler/ExprNodes.py:9251:` - There is already a TODO: `Infer non-object list arrays.`

> You need to change .infer_type() for the comprehension nodes in ExprNodes.py to infer the type of the appended items. 

I think you meant the same thing with this right

scoder:
> do I need to change the `infer_type` method of `TupleNode`, `ListNode`, `SetNode` and `DictNode`?

That's already on its way in https://github.com/cython/cython/pull/7686/

Cython/Compiler/ExprNodes.py
```diff
@@ -9481,7 +9480,32 @@ class ComprehensionNode(ScopedExprNode):
...
 
     def infer_type(self, env):
-        return self.type
+        # Bail early for container types that cannot be parameterised
+        # (e.g. unspecialised builtins added later); avoids analysing the body.
+        if not self.type.supports_container_type:
+            return self.type
+        append = self.append
+        # DictComprehensionAppendNode is defined later in this module, so
+        # dispatch by attribute shape to avoid a forward reference.
```
scoder: Forward references inside of the same module are just fine (inside of methods), but here, I think we should rather use recursion and ask the append node to infer its type.

omkar-334: So in `DictComprehensionAppendNode` and `ComprehensionAppendNode` I can add the `infer_type(env)` method and then `ComprehensionNode.infer_type` can just collect these right?

omkar-334:
so with these changes im getting this
image  
The last case - it ends at tuple, even though the tuple contains further items inside it
There seems to be some issue with `PyTupleObject` and `CTupleType`

omkar-334:
For `[(a, a + 1) for a in xs]` with `xs: cdef list[int]`, typeof prints`list[(int, long)] object`. But the list actually stores Python tuples  at runtime, not C tuples (comprehensions always box items). The displayed  type is misleading.

```python
  r = [(a, a + 1) for a in xs]
  print(typeof(r))      # list[(int, long)] object   <- claims CTuple items
  print(type(r[0]))     # <class 'tuple'>            <- actual: Python tuple
  ```

**Comment by scoder** [comment]
> There seems to be some issue with `PyTupleObject` and `CTupleType`

Tuple items are not type-inferred yet, pretty much for that reason.

**Comment by scoder** [comment]
> the list actually stores Python tuples at runtime, not C tuples

It's more concrete to infer the underlying C types rather than the Python types. A ctuple seems as good as it gets for that. Note that the item types only become relevant when requesting the items (and potentially unpacking them). They don't need to represent the actual storage used inside of the list/set/dict.

It'll be a tradeoff in the end, though. If we automatically unpack the tuple values into ctuples that are then passed back into some Python API as Python tuples, we end up wasting time and space converting back and forth for no good reason. But if the user code then unpacks the items into C variables and processes them as C values further on, we can have big wins. Overall, I would expect the latter case to be more relevant, so let's keep the ctuples until we find a problem with it.

**Comment by omkar-334** [comment]
@scoder Agreed, keeping the ctuples. The implementation already does that. `[(a, a+1) for a in x]` infers `list[(int, long)]`... what do you think about the current implementation?

**Comment by omkar-334** [comment]
@scoder there were 10 tests failing , so i added a fix related to sliced/nested objects,.... and now 5 are failing..

**Comment by omkar-334** [comment]
@scoder any suggestions?

**Comment by da-woods** [comment]
> and now 5 are failing..

Looking at this in hindsight, I think most of the failures were problems we were having with the master branch at the time. So hopefully it's better now.... We shall see.

**Comment by omkar-334** [comment]
thanks @da-woods , looks like all tests have passed

File: Cython/Compiler/ExprNodes.py [hunk 2]
```diff
@@ -9483,7 +9482,47 @@ class ComprehensionNode(ScopedExprNode):
...
+            return self.type
+        # Resolve names in the comprehension's own scope. NameNode.infer_type
+        # caches its entry, so using the outer env here would steal the inner
+        # loop variable's binding (regression: list_comp_in_closure_T598).
+        body_env = self.expr_scope or env
+        # Propagate the iterator's item type into the loop target before
+        # inspecting the body. Without this, `[a for a in xs]` where
+        # xs: list[int] sees `a` as py_object because ControlFlowAnalysis
+        # has not yet typed the comprehension's loop variable.
+        self._propagate_loop_target_type(env)
```
scoder, line 9496: It's difficult to spot the code between the long comments.
```suggestion
        if not self.type.supports_container_type:
            return self.type

        body_env = self.expr_scope or env
```
I've removed the target type assignment. See my other comment.

Cython/Compiler/ExprNodes.py
```diff
@@ -9483,7 +9482,47 @@ class ComprehensionNode(ScopedExprNode):
...
+        item_types = []
+        for item_type in self.append.infer_item_types(body_env):
+            if (item_type is None
+                    or item_type is py_object_type
+                    or item_type.is_error
+                    or item_type.is_unspecified
+                    or item_type.is_fused):
+                return self.type
+            item_types.append(item_type)
```
scoder, line 9505: I don't think unspecified types should ever originate from a type inference method, so we shouldn't handle them specially here. I also don't think the error type needs to be handled specially. It probably won't hurt to specialise the comprehension type with it.

Why are fused types an issue here?

If possible, I'd just do the following:
```suggestion
        item_types = self.append.infer_item_types(body_env)
```

Cython/Compiler/ExprNodes.py
```diff
@@ -9483,7 +9482,47 @@ class ComprehensionNode(ScopedExprNode):
...
+        return self.type.specialize_here(self.pos, env, item_types)
+
+    def _propagate_loop_target_type(self, env):
+        if not isinstance(self.loop, Nodes._ForInStatNode):
+            return
+        target = self.loop.target
+        if (target is None or not getattr(target, 'is_name', False)
+                or target.entry is None):
```
scoder, line 9513: It's a node, there should be an `.is_name` attribute.
```suggestion
        if target is None or not target.is_name or target.entry is None:
```

Cython/Compiler/ExprNodes.py
```diff
@@ -9483,7 +9482,47 @@ class ComprehensionNode(ScopedExprNode):
...
+        if target.entry.type not in (py_object_type, unspecified_type):
+            return
+        sequence_type = self.loop.iterator.sequence.infer_type(env)
+        if not sequence_type.supports_container_type:
+            return
+        item_type = sequence_type.infer_iterator_type()
+        if (item_type is None
+                or item_type is py_object_type
+                or item_type.is_unspecified):
+            return
```
scoder, line 9524:
```suggestion
        item_type = sequence_type.infer_iterator_type()
        if item_type is None or item_type is py_object_type:
            return
```

Cython/Compiler/ExprNodes.py
scoder, line 9496: Hmmm. Looking at this now, I think this is the wrong place to assign a type because it drops a side-effect into `.infer_type()`. That type analysis step shouldn't modify the tree, it should really only analyse it. I think this part needs to go. I've removed the method call from the reordering above. We can do it in `.analyse_types()` later, maybe even rather in the for-loop node than here.

Although … why is this an issue at all? The item values that we collect in the comprehension are necessarily Python objects. Does it matter whether they are typed as anything but `py_object_type` ?

PR #7689, opened 2026-05-13
Fix type inference of subscripted list slice

### LINKED ISSUE #7688: [BUG] Type error for range index of subscripted list
### Describe the bug

Cython is not taking into account range index. range index of list of `list[T]` is `list`. Cython expects `T`.

```cython
def main():
    l: list[tuple] = [(1,2,3), (1,2,3)]
    ll = l[::-1]
    print(ll)

main()
```

### Expected behaviour

Cython correctly infers `list` type when range index is used.

### OS

Linux

### Python version

3.14

### Cython version

master

### Additional context

_No response_

matusvalo:
Fixes #7688


matusvalo:
I am taking liberty and merge this change without code review because:
* it fixes nasty bug which causes CI fail in other PRs
* it affects only subscripted types
* change is simple and limited

If there are any comments/improvements can be implemented in another PR...

da-woods:
I don't think any of the test failures are particularly relevant here (they all affect every other branch right now and I think aren't issues that we can fix). I don't fully understand ones on the Python development branch; I've reported the pypu one to pypy.

matusvalo:
Yeah I ser. Unfortunately it blocks my automerge ...

da-woods:
One extra check I'd add here:

```
d = { slice(1, 5): "some_value" }
cython.typeof(d[1:5])
```

In this case the return type should be a single element (or just `object`).

I've never seen this used in real code so I think the main thing is that it doesn't infer something incorrect like a `dict` return type. But it's fine if it just says "object"

matusvalo:
> One extra check I'd add here:
> 
> ```
> d = { slice(1, 5): "some_value" }
> cython.typeof(d[1:5])
> ```
> 
> In this case the return type should be a single element (or just `object`).
> 
> I've never seen this used in real code so I think the main thing is that it doesn't infer something incorrect like a `dict` return type. But it's fine if it just says "object"

I think already cython in master contains error your code produces `dict object`. Python on the other hand:

```python
>>> d = { slice(1, 5): "some_value" }
>>> d[1:5]
'some_value'

**Comment by scoder** [comment]
> I don't fully understand ones on the Python development branch

See https://github.com/python/cpython/pull/149583#issuecomment-4428338802

**Comment by scoder** [comment]
> ```cython
> d = { slice(1, 5): "some_value" }
> cython.typeof(d[1:5])
> ```
We could add the following to the start of the `analyse_as_pyobject` method:
```python
if is_slice and base_type.is_pyanydict_type:
    is_slice = False
```
For a dict, a slice lookup is never special.

matusvalo:
> > ```cython
> > d = { slice(1, 5): "some_value" }
> > cython.typeof(d[1:5])
> > ```
> 
> We could add the following to the start of the `analyse_as_pyobject` method:
> 
> ```python
> if is_slice and base_type.is_pyanydict_type:
>     is_slice = False
> ```
> 
> For a dict, a slice lookup is never special.

The fix is totally unrelated to this PR so I fixed it in separate PR: #7702

PR #7694, opened 2026-05-16
Enable most of cyfunction tests on PyPy

da-woods:
There's one small defect that still remains skipped but otherwise they work.


da-woods:
One issue with this is that the PyPy tests are currently failing for reasons that I don't think are our fault https://github.com/pypy/pypy/issues/5472 so we might want to wait before enabling these.

Cython/Utility/CommonStructures.c
```diff
@@ -85,6 +85,7 @@ static PyObject* __Pyx_PyType_FromMetaclass(PyTypeObject *metaclass, PyObject *m
         Py_INCREF((PyObject*)metaclass);
         Py_SET_TYPE(result, metaclass);
         Py_DECREF(old_tp);
+        PyType_Modified((PyTypeObject*)result);
```
da-woods, line 88: I was hoping that this would fix the remaining bug. It didn't, but I've left it in because it feels like the right thing to do.

PR #7696, opened 2026-05-16
Optimize `len()` for subscripted types

### LINKED ISSUE #7695: Optimize `len()` for subscripted types
Cython optimize `len()` based on container passed as argument. It currently does not work for subscripted types (`list[...], ...`). See comments below

> There is one additional issue. This PR compiles `Code.py` with following line:
> 
> ```c
> __pyx_t_3 = PyObject_Length(__pyx_v_int_constants_by_bytesize);
> ```
> 
> instead of what we have in master:
> 
> ```c
> __pyx_t_3 = __Pyx_PyList_GET_SIZE(__pyx_v_int_constants_by_bytesize);
> ```
> 
> The affected line is following:
> 
> https://github.com/cython/cython/blob/7dda0262a7954a076129d2b4f105696e75eb3bce/Cython/Compiler/Code.py#L2391

OK culprit is not related to this PR but here: https://github.com/cython/cython/blob/7dda0262a7954a076129d2b4f105696e75eb3bce/Cython/Compiler/Optimize.py#L2885-L2895

_Originally posted by @matusvalo in https://github.com/cython/cython/issues/7686#issuecomment-4435345854_

matusvalo:
Fixes #7695


Cython/Compiler/Optimize.py
```diff
@@ -2882,18 +2883,26 @@ def _handle_simple_function_memoryview(self, node, function, pos_args):
...
+            return "__Pyx_PyTuple_GET_SIZE"
+        if typ.is_pyset_type or typ.is_pyfrozenset_type:
+            return "__Pyx_PySet_GET_SIZE"
+        if typ.is_pydict_type or typ.is_pyfrozendict_type:
+            return "PyDict_Size"
+        if ((typ.is_extension_type or typ.is_builtin_type)
+                and typ.entry.qualified_name in self._ext_types_with_pysize):
+            return 'Py_SIZE'
+        return None
+
```
da-woods:
```suggestion

```

da-woods:
Looks good to me.

I think in this case I prefer the old `map` version to chained `if`s, but clearly it doesn't work any more with subscripted types so chained `if`s is the way to go.

Cython/Compiler/Optimize.py
```diff
@@ -2882,18 +2883,25 @@ def _handle_simple_function_memoryview(self, node, function, pos_args):
...
-        Builtin.tuple_type:      "__Pyx_PyTuple_GET_SIZE",
-        Builtin.set_type:        "__Pyx_PySet_GET_SIZE",
-        Builtin.frozenset_type:  "__Pyx_PySet_GET_SIZE",
-        Builtin.dict_type:       "PyDict_Size",
-        Builtin.frozendict_type: "PyDict_Size",
-    }.get
-
+    def _find_special_capi_len_function(self, typ) -> Optional[str]:
+        if typ.is_pystr_type:
+            return "__Pyx_PyUnicode_GET_LENGTH"
```
scoder: I suggest adding two major cases to avoid going through the whole if-chain for everything that doesn't match at all. We can also drop the "Py_SIZE() for builtins" case, I think, because the set that it uses for the lookup does not currently contain any builtin types.
```suggestion
    def _find_special_capi_len_function(self, typ) -> Optional[str]:
        if typ.is_extension_type:
            if typ.entry.qualified_name in self._ext_types_with_pysize:
                return 'Py_SIZE'
        elif typ.is_builtin_type:
            if typ.is_pystr_type:
                …
            …
        return None

```
Oh, and please move the `_ext_types_with_pysize` set definition above the new method. It's too hidden where it is now.

scoder (commented):
Worth fixing.

Cython/Compiler/Optimize.py
matusvalo: Fixed.

Cython/Compiler/Optimize.py
```diff
@@ -2882,20 +2883,29 @@ def _handle_simple_function_memoryview(self, node, function, pos_args):
...
+                return "__Pyx_PyUnicode_GET_LENGTH"
+            if typ.is_pybytes_type:
+                return "__Pyx_PyBytes_GET_SIZE"
+            if typ.is_pybytearray_type:
+                return "__Pyx_PyByteArray_GET_SIZE"
+            if typ.is_pylist_type:
+                return "__Pyx_PyList_GET_SIZE"
+            if typ.is_pytuple_type:
+                return "__Pyx_PyTuple_GET_SIZE"
+            if typ.is_pyset_type or typ.is_pyfrozenset_type:
```
```suggestion
            if typ.is_pyanyset_type:
```

PR #7697, opened 2026-05-17
Allow type constructors `list()`, `set()`, ... to infer subscripted type

matusvalo:
This PR supports type inference of builtin type container constructors. Hence, subscripted type is correctly propagated:

```cython
l: list[int] = [1,2,3]
s = set(l)   # s variable is of type set[int]
```
The PR consists of two parts:
1. implementation and tests
2. migration type inference tests to separate file based on https://github.com/cython/cython/issues/7729#issuecomment-4583410709

This PR needs to be adjusted when #7704 is merged


matusvalo:
~~Current state is just raw PoC. Allows to compile and infer type in following example:~~

```python
    a = frozenset({1,2,3})
    print(cython.typeof(a))
```

Cython/Compiler/ExprNodes.py
```diff
@@ -6226,7 +6226,15 @@ def infer_type(self, env):
                         return PyrexTypes.c_double_type
                     elif func_name == 'bool':
                         return PyrexTypes.c_bint_type
-                    elif func_name in Builtin.types_that_construct_their_instance:
+                    elif func_name in [name for name, typ in Builtin.builtin_types.items() if typ.supports_container_type]:
```
matusvalo: I would like to carve out this code somehow in separate function to increase readability...

Cython/Compiler/ExprNodes.py
```diff
@@ -6226,6 +6226,19 @@ def infer_type(self, env):
                         return PyrexTypes.c_double_type
                     elif func_name == 'bool':
                         return PyrexTypes.c_bint_type
+                    elif (
+                        hasattr(self, 'args') and self.args and len(self.args) == 1 and
+                        func_name in [name for name, typ in Builtin.builtin_types.items() if typ.supports_container_type]
+                    ):
```
scoder: Isn't this just
```suggestion
                    elif result_type.supports_container_type and len(self.args) == 1:
```
I haven't tried it but I don't think we need the `hasattr()` check here. That might be needed for the C function case a bit up, but here we're dealing with a call to a builtin type. That should always have `self.args`.

tests/run/type_inference_generics.py
```diff
@@ -0,0 +1,352 @@
...
+    print(cython.typeof(b1) + ("[int] object" if not cython.compiled else ""))
+    print(cython.typeof(b2) + ("[int] object" if not cython.compiled else ""))
+    print(cython.typeof(b3) + (" object" if not cython.compiled else ""))
+    print(cython.typeof(c) + ("[Python object] object" if not cython.compiled else ""))
+
+if sys.version_info >= (3, 11) or cython.compiled:
+    # This part of the test is failing in Python 3.9 and 3.10 with the following exception in Shadow.py:
+    # isinstance() argument 2 cannot be a parameterized generic
+
+    def test_casting_subscripted_types():
```
scoder: It would be better to use a decorator (`@min_plain_pyversion(3,11)`) for this kind of skipping. Once we drop support for Py<3.10 or fix/work around the issue, it'll be easier to remove the decorator line than to realign the whole function, thus losing its VCS line history.

Cython/Compiler/ExprNodes.py
```diff
@@ -6226,6 +6226,19 @@ def infer_type(self, env):
...
+                        func_arg_type = self.args[0].infer_type(env)
+                        if func_arg_type.supports_container_type:
+                            # FIXME: tuples needs to be handled here when supporting subscripts
+                            # list(tuple[int]) -> list[int] but list(tuple[int, float] -> list[common_type(int, float)])
+                            if func_name in ['dict', 'frozendict']:
+                                subscripted_types = func_arg_type.subscripted_types
```
scoder: From a quick look, I couldn't find a test for something like this:
```python
items: list[tuple[int, str]] = [(1, 'xaz')]
assert cython.typeof(dict(items)) == "dict[int, str]"
```
Is that covered here?

matusvalo: This I missed to implement. I will add support of it.

Cython/Compiler/ExprNodes.py
matusvalo: Not sure I added it because some example was failing but maybe I messed up...

Cython/Compiler/ExprNodes.py
```diff
@@ -6245,6 +6245,27 @@ def infer_type(self, env):
...
+                        func_arg_type = self.args[0].infer_type(env)
+                        if func_arg_type.supports_container_type:
+                            if func_name in ['dict', 'frozendict']:
+                                # We infer subscripted type of new dictionary only if type of the argument
+                                # is CONTAINER[CONTAINER[...]]
+                                st = func_arg_type.subscripted_types[0] if len(func_arg_type.subscripted_types) == 1 else None
+                                if func_arg_type.is_pyanydict_type:
+                                    subscripted_types = func_arg_type.subscripted_types
+                                elif st and st.is_pytuple_type and len(st.subscripted_types) == 2:
+                                    subscripted_types = (st.subscripted_types[0], st.subscripted_types[1])
```
scoder: This should be the same as
```suggestion
                                    subscripted_types = st.subscripted_types
```
(Although, if you need a tuple, then it would be `subscripted_types = tuple(st.subscripted_types)`).

matusvalo:
OK good news is that the new logic seems to be complete now. The bad news is that it is a mess :-(. I am inclined to carve it somewhere else...

Cython/Compiler/ExprNodes.py
```diff
@@ -6245,6 +6245,27 @@ def infer_type(self, env):
...
+                                elif st and len(st.subscripted_types) > 0:
+                                    # The argument type does not reveal number of types like list[list[...]]
+                                    # Hence, we just take first type and hope for the best. In any case,
+                                    # the runtime will error out when there are more than two elements.
+                                    subscripted_types = (st.subscripted_types[0], st.subscripted_types[0])
```
scoder: Should we require homogeneous item types here?

Cython/Compiler/ExprNodes.py
```diff
@@ -6245,6 +6245,27 @@ def infer_type(self, env):
                         return PyrexTypes.c_double_type
                     elif func_name == 'bool':
                         return PyrexTypes.c_bint_type
+                    elif result_type.supports_container_type and len(self.args) == 1:
+                        func_arg_type = self.args[0].infer_type(env)
+                        if func_arg_type.supports_container_type:
+                            if func_name in ['dict', 'frozendict']:
```
scoder: Why don't you call `.infer_iterator_type()` for dicts as well ?

matusvalo: Dicts are special e.g. `dict(d: dict[str, int])` -> `dict[str, int]`

`.infer_iterator_type()` will for `d` in example above is just `str` which is wrong. Not to mention fun when dict is constructed from list of tuples or list of lists...

Cython/Compiler/ExprNodes.py
matusvalo: Well it is questionable how to proceed. Python supports following:

```python
l: list[list[int]] = [[1,2], [3, 4]]
d = dict(list)
```
In the end d will contain: `{1: 2, 3: 4}`

But we can just make `d` as plain `dict` it is possible

matusvalo: Long story short we can make the logic much simpler I am open to suggestions what is best here...

matusvalo:
I moved the tests from pep526_variable_annotations.py...

Cython/Compiler/ExprNodes.py
```diff
@@ -6245,6 +6245,27 @@ def infer_type(self, env):
...
+                                    # Hence, we just take first type and hope for the best. In any case,
+                                    # the runtime will error out when there are more than two elements.
+                                    subscripted_types = (st.subscripted_types[0], st.subscripted_types[0])
+                                else:
+                                    subscripted_types = ()
+                            else:
+                                subscripted_types = [func_arg_type.infer_iterator_type()]
+                            return result_type.specialize_here(self.pos, env, subscripted_types)
                     elif func_name in Builtin.types_that_construct_their_instance:
                         return result_type
```
scoder: Ok, so, this is untested, but I came up with the following:
```suggestion
                    if func_name == 'bool':
                        return PyrexTypes.c_bint_type
                    posargs = getattr(self, 'args', getattr(self, 'positional_args', ()))
                    if result_type.supports_container_type and len(posargs) == 1 and not getattr(self, 'keyword_args', None):
                        arg = posargs[0]
                        arg_type = arg.infer_type(env)
                        if arg_type.supports_container_type:
                            if result_type.is_pyanydict_type:
                                if arg_type.is_pyanydict_type:
                                    return result_type.specialize_here(self.pos, env, arg_type.subscripted_types)
                                if arg_type.is_builtin_sequence:
                                    return self._infer_dict_container_type_from_sequence(env, result_type, arg_type, arg)
                            if len(arg_type.subscripted_types) == 1:
                                return result_type.specialise_here(self.pos, env, arg_type.subscripted_types[0])
                            # FIXME: Normally, "arg.infer_type()" should have called "infer_sequence_item_type()"
                            # already if appropriate, but it looks like that's not yet implemented, e.g. for ListNode?
                            item_type = infer_sequence_item_type(env, arg, seq_type=arg_type)
                            if item_type is not None:
                                return result_type.specialize_here(self.pos, env, [item_type])
                    if func_name in Builtin.types_that_construct_their_instance:
                        return result_type

```
and then `self._infer_dict_container_type_from_sequence()` is:
```python
    def _infer_dict_container_type_from_sequence(self, env, dict_type, sequence_type, sequence):
        """Infer the item types of a (frozen)dict construction from a key-value sequence.

        This will usually be "list[tuple[x,y]]", but Python allows arbitrary sequence types.
        """
        def unpack_keyvalue_types(items_type):
            if items_type.is_pytuple_type:
                if len(items_type.subscripted_types) == 2:
                    # Simple 2-tuple.
                    return items_type.subscripted_types
            elif items_type.is_builtin_sequence and len(items_type.subscripted_types) == 1:
                # Homogeneous sequence. If the length is != 2 at runtime, the dict creation will fail,
                # so it's safe to infer the item type from it.
                return items_type.subscripted_types * 2
            return None

        key_value_type = None
        if len(sequence_type.subscripted_types) == 1:
            key_value_type = unpack_keyvalue_types(sequence_type.subscripted_types[0])

        if key_value_type is None and sequence.is_sequence_constructor:
            # Try to infer common key-value types.
            key_types, value_types = set(), set()
            for item in sequence.args:
                kv_type = unpack_keyvalue_types(item.infer_type(env))
                if kv_type is None:
                    break
                key_types.add(kv_type[0])
                value_types.add(kv_type[1])
            else:
                key_value_type = (
                    PyrexTypes.reduce_spanning_types(key_types),
                    PyrexTypes.reduce_spanning_types(value_types)
                )

        if key_value_type is None or key_value_type == (py_object_type, py_object_type):
            return dict_type

        return dict_type.specialize_here(self.pos, env, key_value_type)
```
I think this handles a lot of cases, can't say if they are all covered by tests or what other interesting cases I've forgotten to include here. But you've done most of the thinking for this so far, so you're in the best position to validate it.

EDIT: `len(self.args)` doesn't work for `GeneralCallNode`. I wonder, though, if that one's special enough to merit its own extended `infer_type()` method. It allows `dict(**x)`, for example. (And as you may guess, I dislike that `getattr()` usage for this but couldn't quickly come up with something better. Happy to see this improved.)

Cython/Compiler/ExprNodes.py
scoder: > Python supports following:
> ```python
> l: list[list[int]] = [[1,2], [3, 4]]
> d = dict(list)
> ```

I don't think that is an issue. Even if we do not know the length of the non-tuple items at compile time, if it's a homogeneous sequence, then either the key-value pairs both have that type, or the dict creation will fail at runtime if it doesn't find key-value pairs. So we can always infer the key-value types from a homogeneous sequence.

scoder:
Funny enough, I now had to change this test in `type_inference_new.pyx`:
```python
def test_python_objects(cond):
    """
    >>> test_python_objects(0)
    """
    if cond == 1:
        a = [1, 2, 3]
        o_list = a
    elif cond == 2:
        a = set([1, 2, 3])
        o_set = a
    else:
        a = {1:1, 2:2, 3:3}
        o_dict = a
    assert typeof(a) == "Python object", typeof(a)
    assert typeof(o_list) == "list object", typeof(o_list)
    assert typeof(o_dict) == "dict object", typeof(o_dict)
    assert typeof(o_set) == "set[long] object", typeof(o_set)
```
We infer the item type of the set because it actually calls `set()`, but we do not infer the sort-of much more obvious item types of the literals. And this inference is actually dangerous because nothing keeps users from adding non-`long` integers (and arbitrary other Python types) to the set in `o_set`, which would render Cython's inferred type dangerous. I think we should only infer the item type for immutable types, and possibly for mutable containers if `infer_types=True` is enabled, ~not~ but that's up for debate.

matusvalo:
>  we do not infer the sort-of much more obvious item types of the literals.

Yes this is goal of #7686

>  I think we should only infer the item type for immutable types, and possibly for mutable containers if infer_types=True is enabled, not that's up for debate.

I think it makes sense... And the same holds for #7686

Cython/Compiler/ExprNodes.py
```diff
@@ -6243,15 +6243,74 @@ def infer_type(self, env):
                     func_name = function.entry.name
                     if func_name == 'float':
                         return PyrexTypes.c_double_type
-                    elif func_name == 'bool':
+                    if func_name == 'bool':
                         return PyrexTypes.c_bint_type
-                    elif func_name in Builtin.types_that_construct_their_instance:
+                    # Handle both SimpleCallNode (.args) and GeneralCallNode (.positional_args / .keyword_args).
+                    posargs = getattr(self, 'args', getattr(self, 'positional_args', ()))
+                    if result_type.supports_container_type and len(posargs) == 1 and not getattr(self, 'keyword_args', None):
```
matusvalo, line 6251: It is a slighly mess:

```python
 frozendict()
```

yields `posargs` as a `tuple` => `len(posargs)` ✅ 

On the other hand:

```python
frozendict(parrot=u"resting", answer=42)
```

yields `posargs` as `TupleNode` => `len(posargs)` raises an exception ❌  (one must do additional `posargs.args`)

matusvalo: For no I added a simple fix to make CI/CD 🟢

da-woods:
Is this too strict?

For example

```
import cython

def f(a: dict[str, list]):
    b = dict(a)
    b['a'] = set()
    x = b[1]
    print(cython.typeof(x), len(x))
```

This currently fails to compile for me with `Cannot assign type 'set object' to 'list object'`.  I think that would be a reasonable error if I did `a['a'] = set()` but I don't see where I've promised to keep that restriction for `b`.

-------------------------------------------

Obviously for immutable types it's completely safe, but for mutable types I think you need to look at what the user is doing with it (and probably anything that gets used after being passed to an unknown function also becomes suspect)

scoder:
> Is this too strict?

Yes, see https://github.com/cython/cython/pull/7697#issuecomment-4758354444 :

>> I think we should only infer the item type for immutable types, and possibly for mutable containers if `infer_types=True` is enabled, but that's up for debate.

My underinformed feeling is that even `infer_types=True` is too weak a signal here to infer mutable container types.

Either way, at least for the timeframe of 3.3, it seems reasonable to enable this inference for `frozenset` and `frozendict` but not for mutable types. That's a rather small win, though, given how little immutable types are used.

Cython/Compiler/ExprNodes.py
```diff
@@ -6243,15 +6243,75 @@ def infer_type(self, env):
...
                         return PyrexTypes.c_double_type
-                    elif func_name == 'bool':
+                    if func_name == 'bool':
                         return PyrexTypes.c_bint_type
-                    elif func_name in Builtin.types_that_construct_their_instance:
+                    # Handle both SimpleCallNode (.args) and GeneralCallNode (.positional_args / .keyword_args).
+                    posargs = getattr(self, 'args', getattr(self, 'positional_args', ()))
+                    posargs = posargs.args if hasattr(posargs, 'args') else posargs
+                    if result_type.supports_container_type and len(posargs) == 1 and not getattr(self, 'keyword_args', None):
+                        arg = posargs[0]
```
scoder, line 6252: I just noticed that we're excluding `GeneralCallNode` already by testing for kwargs, so let's exclude it earlier.
```suggestion
                    if result_type.supports_container_type and not isinstance(self, GeneralCallNode) and len(self.args) == 1:
                        arg = self.args[0]
```

scoder (commented):
By constraining this (and https://github.com/cython/cython/pull/7686) to only immutable types, these become a lot less important. We can postpone them and take our time to get them properly designed for their limited purpose.

matusvalo:
Closing in favor of #7686

PR #7698, opened 2026-05-17
Enable vectorcall for cdef class creation

da-woods:
Take a fairly cautious approach here and only enable vectorcall for classes where we have complete visibility of the object creation (so disable it for anything external and anything from a pxd).

In the situations where we think it's worthwhile, I make the vectorcall-style versions the main implementation. Therefore the "args, kwds" `tp_new` and `tp_init` slots just defer to the vectorcall implementation.  This is probably a minor pessimization in the case where they are needed (mainly when inheriting from them with something that can't use vectorcall).

Some small microbenchmarking (of creating classes from Python) suggests it might make class creation between ~30-50% faster so is very likely worthwhile.


da-woods:
> This involves a reasonable amount of code duplication:
> 
>     * all the `tp_new` slots are duplicated with a vectorcall version.
> 
>     * all the `__cinit__` and `__init__` functions in suitable cdef classes have an extra wrapper attached taking vectorcall arguments.

As I typed this I began to think that I should probably redo this bit of it. I suspect we can make the vectorcall versions the "true" version and then delegate the `tp_new` calls to them. I think this would clean up a lot, but it's also likely a problem for another day.

docs/src/userguide/source_files_and_compilation.rst
```diff
@@ -1189,6 +1189,9 @@ hidden by default since most users will be uninterested in changing them.
         ``CYTHON_VECTORCALL``
             Enable the vectorcall calling mechanism.
 
+        ``CYTHON_VECTORCALL_NEW``
+            Enable the vectorcall calling mechanism for cdef class creation.
```
scoder: "NEW" is too general for a specialised feature.
```suggestion
        ``CYTHON_VECTORCALL_TPNEW``
            Enable the vectorcall calling mechanism for cdef class creation.
```

Cython/Compiler/TypeSlots.py
```diff
@@ -502,6 +519,40 @@ def generate_dynamic_init_code(self, scope, code):
...
+            # If we don't have the second overloaded alternative for __cinit__ or __init__
+            # this means the signature was wrong and so we haven't generated code for the
+            # vectorcall wrappers in DefNode.analyse_declarations
+            if (entry := tp.scope.lookup_here("__cinit__")):
+                if not entry.is_special or not entry.tp_new_can_be_vectorcall:
+                    return "0"
+            if not seen_init and (entry := tp.scope.lookup_here("__init__")):
+                seen_init = True
+                if not entry.is_special or not entry.tp_new_can_be_vectorcall:
+                    return "0"
```
scoder: The reuse of the `entry` variable in assignment expressions makes it difficult for me to trace where it was assigned.
```suggestion
            cinit_entry = tp.scope.lookup_here("__cinit__")
            if cinit_entry and not (cinit_entry.is_special and cinit_entry.tp_new_can_be_vectorcall):
                return "0"
            if not seen_init and (init_entry := tp.scope.lookup_here("__init__")):
                if not (init_entry.is_special and init_entry.tp_new_can_be_vectorcall):
                    return "0"
                seen_init = True
```

Cython/Compiler/TypeSlots.py
```diff
@@ -605,6 +656,18 @@ def __init__(self, signature, slot_name, left_method, method_name_to_slot, **kar
...
 
 
+
+class InitSlot(MethodSlot):
+    def slot_code(self, scope):
+        super_slot_code = super().slot_code(scope)
+        if (super_slot_code != "0" and (entry := scope.lookup(self.method_name))
+                and entry.is_special and entry.signature.use_fastcall):
+            # we need a wrapper
+            return InternalMethodSlot.slot_code(self, scope)
```
scoder:
```suggestion
        if super_slot_code != "0":
            entry = scope.lookup(self.method_name)
            if entry and entry.is_special and entry.signature.use_fastcall:
                # we need a wrapper
                return InternalMethodSlot.slot_code(self, scope)
```

Cython/Compiler/TypeSlots.py
```diff
@@ -90,7 +91,12 @@ class Signature:
     }
 
     # Use METH_FASTCALL instead of METH_VARARGS
-    use_fastcall = False
+    class FastcallType(enum.IntEnum):
+        NO = 0
+        YES = 1
+        TP_NEW = 2
+
+    use_fastcall = FastcallType.NO
```
scoder: I think the comment only applies to the last line. But the enum might deserve one, too.

Since the enum name is spelled out as prefix wherever its values are used, maybe `FastcallUsed` is a better name? Like `use_fastcall = FastcallUsed.YES`. Yes/no doesn't read well as a "type".

Cython/Compiler/ParseTreeTransforms.py
```diff
@@ -2420,7 +2420,7 @@ def _handle_def(self, decorators, env, node):
         self.visitchild(node, 'py_func')
         node.update_fused_defnode_entry(env)
         # For the moment, fused functions do not support METH_FASTCALL
-        node.py_func.entry.signature.use_fastcall = False
+        node.py_func.entry.signature.use_fastcall = node.py_func.entry.signature.FastcallType.NO
```
```suggestion
        call_signature = node.py_func.entry.signature
        call_signature.use_fastcall = call_signature.FastcallType.NO
```

Cython/Compiler/TypeSlots.py
```diff
@@ -859,6 +922,13 @@ def get_slot_by_name(slot_name, compiler_directives):
     for slot in get_slot_table(compiler_directives).slot_table:
         if slot.slot_name == slot_name:
             return slot
+    # Special-case for hacky dummy slot used to implement tp_vectorcall.
+    # It has the same rules about when to generate it as tp_new.
+    if slot_name == "tp_newv":
+        slot = get_slot_by_name("tp_new", compiler_directives)
+        slot = copy.copy(slot)
+        slot.slot_name = "tp_newv"
+        return slot
```
scoder: Yes, this looks hacky.  Maybe change `ConstructorSlot` to take the decision and generate an appropriate `slot_name` ?

Cython/Compiler/Nodes.py
```diff
@@ -3799,6 +3803,12 @@ def analyse_declarations(self, env):
         self.np_args_idx = self.target.np_args_idx
 
     def prepare_argument_coercion(self, env):
+        if (self.target.entry.is_special and self.target.entry.name in ("__cinit__", "__init__")
+                and TypeSlots.TpVectorcallSlot("tp_vectorcall").slot_code(env) != "0"):
+            # At this stage we know enough about both __cinit__, __init__ and the class scopes
+            # to know if we should prefer vectorcall for class creation.
+            self.signature = self.target.entry.signature = self.signature.with_fastcall(self.signature.FastcallType.TP_NEW)
```
scoder: Seems worth extracting a variable here.
```suggestion
    def prepare_argument_coercion(self, env):
        target_entry = self.target.entry
        if (target_entry.is_special and target_entry.name in ("__cinit__", "__init__")
                and TypeSlots.TpVectorcallSlot("tp_vectorcall").slot_code(env) != "0"):
            # At this stage we know enough about both __cinit__, __init__ and the class scopes
            # to know if we should prefer vectorcall for class creation.
            self.signature = target_entry.signature = self.signature.with_fastcall(self.signature.FastcallType.TP_NEW)
```

da-woods:
It looks like I didn't write get the file filters (relating to docs) right for when to run the CI jobs. Will try to fix it later...

Cython/Compiler/Nodes.py
```diff
@@ -3971,9 +3981,10 @@ def generate_function_header(self, code, with_pymethdef, proto_only=0):
             if sig.use_fastcall:
                 fastcall_args = "PyObject *const *%s, Py_ssize_t %s, PyObject *%s" % (
                         Naming.args_cname, Naming.nargs_cname, Naming.kwds_cname)
+                fastcall_guard = sig.fastcall_guard
                 arg_code_list.append(
-                    "\n#if CYTHON_VECTORCALL\n%s\n#else\n%s\n#endif\n" % (
-                        fastcall_args, varargs_args))
+                    "\n#if %s\n%s\n#else\n%s\n#endif\n" % (
+                        fastcall_guard, fastcall_args, varargs_args))
```
scoder: f-string? The template string is full of special characters either way, but it may (or may not) be a little more readable still.

scoder (commented):
I'm very much in favour of this feature.

docs/src/userguide/source_files_and_compilation.rst
da-woods: `NEW` definitely meant to mean `tp_new` rather than "recent". I'll change it.

Cython/Compiler/TypeSlots.py
da-woods: I've added a `to_vectorcall_slot` to `ConstructorSlot` which I think does the same thing a little less opaquely.  I guess an alternative would be a `vectorcall_slot_code` method on `ConstructorSlot` to skip the dummy slot and just generate the code. It's all fairly similar though I think.

The reason I can't _just_ change `ConstructorSlot` is that we do need to generate `tp_new` functions too in case it gets called directly.

Cython/Compiler/ModuleNode.py
```diff
@@ -1659,6 +1661,13 @@ def generate_new_function(self, scope, code, cclass_entry):
         if tp_slot.slot_code(scope) != slot_func:
             return  # never used
 
+        format_vectorcall_if = "\n#if CYTHON_VECTORCALL_TPNEW\n    {}\n#else\n    {}\n#endif\n".format
```
scoder: I think this might be slightly clearer as `format_vectorcall_conditional`, `with_vectorcall_guard` or just `vectorcall_choice` than `format_vectorcall_if`. It's not only formatting something *if* a condition is met, it's creating the condition.

Also, if we assume that vectorcall-tpnew is quite common and will become even more common in the future, this generation code might benefit from always creating both string variants right next to each other, and then dropping the unused one by making this formatting function `lambda vc, non_vc: non_vc` if we do not generate the vectorcall variant.

Cython/Compiler/ModuleNode.py
```diff
@@ -1659,6 +1661,13 @@ def generate_new_function(self, scope, code, cclass_entry):
...
+
+        vectorcall_tp_slot = TypeSlots.get_slot_by_name("tp_vectorcall", scope.directives)
+        vectorcall_slot_func = scope.mangle_internal("tp_vectorcall")
+        is_vectorcall = vectorcall_tp_slot.slot_code(scope) == vectorcall_slot_func
+        tp_newv_slot = TypeSlots.ConstructorSlot("tp_newv", "__cinit__")
```
scoder: The `newv` looks a bit too much like a typo to me. I think we can afford to spell out `tp_new_vectorcall` etc. instead.

Cython/Compiler/ModuleNode.py
```diff
@@ -1816,6 +1845,101 @@ def generate_new_function(self, scope, code, cclass_entry):
...
+                    "CallSlotAsVectorcall", "ExtensionTypes.c",
+                    context=dict(ret_type="PyObject *", name="tpnew", obj_type="PyTypeObject*", error_value="NULL"))
+            )
+            code.putln(f"return __Pyx_CallTpnewAsVectorcall({scope.mangle_internal('tp_newv')}, t, a, k);")
+            code.putln("}")
+            code.exit_cfunc_scope()
+            code.globalstate['decls'].putln("#else")
+            code.globalstate['decls'].putln(f"#define {scope.mangle_internal('tp_new')} {scope.mangle_internal('tp_newv')}")
+            code.putln("#endif")
+            code.globalstate['decls'].putln("#endif")
```
scoder: It's unclear here why we need to put the guard into `decls` as well. I'd rather add a `guard` argument to `start_slotfunction` and let that put it where needed, than remotely duplicating the decision where it will be needed.

I think we can also split the two cases and use the guard twice to untangle it more.

The same applies to other two generation methods further down.

```suggestion
        if is_vectorcall:
            code.globalstate.use_utility_code(
                TempitaUtilityCode.load_cached(
                    "CallSlotAsVectorcall", "ExtensionTypes.c",
                    context=dict(ret_type="PyObject *", name="tpnew", obj_type="PyTypeObject*", error_value="NULL"))
            )

            code.start_slotfunc(
                scope, PyrexTypes.py_objptr_type,
                "tp_new",
                f"PyTypeObject *t, PyObject *a, PyObject *k",
                needs_prototype=True,
                guard="CYTHON_VECTORCALL_TPNEW",
            )
            code.putln(f"return __Pyx_CallTpnewAsVectorcall({scope.mangle_internal('tp_newv')}, t, a, k);")
            code.putln("}")
            code.exit_cfunc_scope()
            code.putln("#endif")  # CYTHON_VECTORCALL_TPNEW

            decls = code.globalstate['decls']
            decls.putln("#if !CYTHON_VECTORCALL_TPNEW")
            decls.putln(f"#define {scope.mangle_internal('tp_new')} {scope.mangle_internal('tp_newv')}")
            decls.putln("#endif")
```

Cython/Utility/ExtensionTypes.c
```diff
@@ -966,3 +966,180 @@ int __Pyx_ApplySequenceOrMappingFlag(PyTypeObject *tp, int is_sequence) {
...
+        if (unlikely(!args)) {
+            PyErr_NoMemory();
+            return {{error_value}};
+        }
+    }
+    Py_ssize_t i = 0;
+    for (; i < a_size; ++i) {
+        args[i] = __Pyx_PyTuple_GET_ITEM(a, i);
+#if !CYTHON_ASSUME_SAFE_MACROS
+        if (unlikely(!args[i])) return {{error_value}};
```
scoder: May leak the allocated memory.

Cython/Utility/ExtensionTypes.c
```diff
@@ -966,3 +966,180 @@ int __Pyx_ApplySequenceOrMappingFlag(PyTypeObject *tp, int is_sequence) {
...
+        goto end;
+    }
+    {
+        initproc tp_init;
+        if (PyObject_TypeCheck(result, t) && (tp_init = __Pyx_PyObject_GetSlot(result, tp_init, initproc))) {
+            if (unlikely(tp_init(result, args_tuple, kwds_dict) < 0)) {
+                Py_CLEAR(result);
+            }
+        }
+    }
```
```suggestion
    if (PyObject_TypeCheck(result, t)) {
        initproc tp_init = __Pyx_PyObject_GetSlot(result, tp_init, initproc);
        if (tp_init && unlikely(tp_init(result, args_tuple, kwds_dict) < 0)) {
            Py_CLEAR(result);
        }
    }
```

Cython/Utility/ExtensionTypes.c
```diff
@@ -966,3 +966,180 @@ int __Pyx_ApplySequenceOrMappingFlag(PyTypeObject *tp, int is_sequence) {
...
+////////////////////////////// CallNewInitFromVectorcall //////////////////////
+//@requires: ObjectHandling.c::TupleFromArray
+
+#if CYTHON_VECTORCALL_TPNEW
+static PyObject *__Pyx_CallNewInitFromVectorcall(PyTypeObject *t, PyObject *const *args, size_t nargsf, PyObject *kwnames) {
+    newfunc tp_new = __Pyx_PyType_GetSlot(t, tp_new, newfunc);
+    if (unlikely(!tp_new)) {
+        __Pyx_TypeName type_name = __Pyx_PyType_GetFullyQualifiedName(t);
+        PyErr_Format(PyExc_TypeError, "cannot create " __Pyx_FMT_TYPENAME " instances", type_name);
+        __Pyx_DECREF_TypeName(type_name);
```
scoder: Another case where https://github.com/cython/cython/pull/7680 would help. I've updated it to see if that fixes CI.

Cython/Utility/ExtensionTypes.c
```diff
@@ -966,3 +966,180 @@ int __Pyx_ApplySequenceOrMappingFlag(PyTypeObject *tp, int is_sequence) {
...
+#endif
+
+///////////////////////// CallSlotAsVectorcall ////////////////////////////////////
+//@requires: Synchronization.c::CriticalSections
+//@requires: CallSlotAsVectorcallUnpackDict
+
+// Possible TODO - the !ASSUME_SAFE_MACROS case is inefficient due to a lot of
+// error checking, and the inability to optimize the common no-keywords case.
+// I think it might be possible to use a trampoline type and PyVectorcall_Call to
+// push this work into the Python interpreter (where they have access to the full C API).
```
scoder, line 1053: I'm not sure why we'd need a type rather than a function like this to map the call to the `PyVectorcall` API, but it seems fine for the time being and can be considered as an optimisation later.

scoder:
I quickly sketched up https://github.com/cython/cython/pull/7746 (with obvious collisions) because it occurred to me that it would be beneficial to separate the allocation from the initialisation of the object. While we could do the allocation in different ways (including freelists and direct calls to `PyType_GenericAlloc()` for `tp.__new__(tp)`), the initialisation would rather stay the same and could be reused by different entry points.

Doing that split first might also help this PR.

Cython/Utility/FunctionArguments.c
```diff
@@ -927,6 +927,22 @@ static CYTHON_INLINE int __Pyx_MergeKeywords(PyObject *kwdict, PyObject *source_
...
+    #define __Pyx_KwValues_FASTCALL_NEW __Pyx_KwValues_FASTCALL
+    #define __Pyx_GetKwValue_FASTCALL_NEW __Pyx_GetKwValue_FASTCALL
+    #define __Pyx_KwargsAsDict_FASTCALL_NEW __Pyx_KwargsAsDict_FASTCALL
+#else
+    #define __Pyx_ArgRef_FASTCALL_NEW __Pyx_ArgRef_VARARGS
+    #define __Pyx_NumKwargs_FASTCALL_NEW __Pyx_NumKwargs_VARARGS
+    #define __Pyx_KwValues_FASTCALL_NEW __Pyx_KwValues_VARARGS
+    #define __Pyx_GetKwValue_FASTCALL_NEW __Pyx_GetKwValue_VARARGS
+    #define __Pyx_KwargsAsDict_FASTCALL_NEW __Pyx_KwargsAsDict_VARARGS
+#endif
```
scoder, line 961:
```suggestion
#if CYTHON_VECTORCALL_TPNEW
    #if !CYTHON_VECTORCALL
        #error Enabling CYTHON_VECTORCALL_TPNEW without CYTHON_VECTORCALL is not supported
    #endif
    #define __Pyx_ArgRef_FASTCALL_TPNEW __Pyx_ArgRef_FASTCALL
    #define __Pyx_NumKwargs_FASTCALL_TPNEW __Pyx_NumKwargs_FASTCALL
    #define __Pyx_KwValues_FASTCALL_TPNEW __Pyx_KwValues_FASTCALL
    #define __Pyx_GetKwValue_FASTCALL_TPNEW __Pyx_GetKwValue_FASTCALL
    #define __Pyx_KwargsAsDict_FASTCALL_TPNEW __Pyx_KwargsAsDict_FASTCALL
#else
    #define __Pyx_ArgRef_FASTCALL_TPNEW __Pyx_ArgRef_VARARGS
    #define __Pyx_NumKwargs_FASTCALL_TPNEW __Pyx_NumKwargs_VARARGS
    #define __Pyx_KwValues_FASTCALL_TPNEW __Pyx_KwValues_VARARGS
    #define __Pyx_GetKwValue_FASTCALL_TPNEW __Pyx_GetKwValue_VARARGS
    #define __Pyx_KwargsAsDict_FASTCALL_TPNEW __Pyx_KwargsAsDict_VARARGS
#endif
```

Cython/Compiler/TypeSlots.py
```diff
@@ -200,21 +208,32 @@ def method_function_type(self):
...
 
     @property
     def fastvar(self):
         # Used to select variants of functions, one dealing with METH_VARARGS
         # and one dealing with __Pyx_METH_FASTCALL
-        if self.use_fastcall:
+        if self.use_fastcall == self.FastcallUsed.YES:
             return "FASTCALL"
+        elif self.use_fastcall == self.FastcallUsed.TP_NEW:
+            return "FASTCALL_NEW"
```
scoder:
```suggestion
            return "FASTCALL_TPNEW"
```

Cython/Compiler/TypeSlots.py
```diff
@@ -605,6 +666,19 @@ def __init__(self, signature, slot_name, left_method, method_name_to_slot, **kar
         self.right_slot = MethodSlot(signature, "", right_method, method_name_to_slot, **kargs)
 
 
+
```
scoder:
```suggestion
```

Cython/Compiler/TypeSlots.py
```diff
@@ -605,6 +666,19 @@ def __init__(self, signature, slot_name, left_method, method_name_to_slot, **kar
...
+class InitSlot(MethodSlot):
+    def slot_code(self, scope):
+        super_slot_code = super().slot_code(scope)
+        if super_slot_code != "0":
+            entry = scope.lookup(self.method_name)
+            if entry and entry.is_special and entry.signature.use_fastcall:
+                # we need a wrapper
+                return InternalMethodSlot.slot_code(self, scope)
+        return super_slot_code
+
```
scoder, line 678:
```suggestion
```

Cython/Utility/ExtensionTypes.c
scoder:
```suggestion
        __Pyx_RaiseErrorWithType(PyExc_TypeError, "cannot create " __Pyx_FMT_TYPENAME " instances", t);
```

Cython/Utility/ExtensionTypes.c
```diff
@@ -966,3 +966,178 @@ int __Pyx_ApplySequenceOrMappingFlag(PyTypeObject *tp, int is_sequence) {
...
 #endif
+
+////////////////////////////// CallNewInitFromVectorcall.proto //////////////////////
+
+#if CYTHON_VECTORCALL_TPNEW
+static PyObject *__Pyx_CallNewInitFromVectorcall(PyTypeObject *t, PyObject *const *args, size_t nargsf, PyObject *kwnames); /* proto */
+#endif
+
+////////////////////////////// CallNewInitFromVectorcall //////////////////////
+//@requires: ObjectHandling.c::TupleFromArray
```
scoder, line 974:
```suggestion
//@requires: ObjectHandling.c::TupleFromArray
//@requires: ObjectHandling.c::RaiseErrorWithObjectType
```

scoder (approved):

Cython/Utility/ModuleSetupCode.c
```diff
@@ -372,6 +382,13 @@
   #ifndef CYTHON_VECTORCALL
     #define CYTHON_VECTORCALL 1
   #endif
+  #if CYTHON_USE_TYPE_SPECS && PY_VERSION_HEX < 0x030E0000
+    // Py_tp_vectorcall slot unavailable
+    #undef CYTHON_VECTORCALL_TPNEW
+    #define CYTHON_VECTORCALL_TPNEW 0
+  #elif !defined(CYTHON_VECTORCALL_TPNEW)
+    #define CYTHON_VECTORCALL_TPNEW CYTHON_VECTORCALL
+  #endif
```
mattip, line 391: I think this needs to be repeated on the PyPy branch. Should I open an issue/PR?

Edit: I run weekly PyPy tests with -DCYTHON_USE_TYPE_SPECS and they started failing. They have been passing for quite a while now.

da-woods: Ah yes - we support type slots on PyPy (but don't test it ourselves).

Yes - PR would be useful. It might be good to add it to the graalpy branch too. I know `CYTHON_USE_TYPE_SPECS` is currently turned off there but the same rules will apply if we ever change that.

PR #7700, opened 2026-05-18
Add a new type attribute "is_pyanyset_type" for "set or frozenset" type tests

scoder:


matusvalo:
After mergeing this PR needs to be updated: #7696

PR #7702, opened 2026-05-18
Fix using slice as a dictionary key

matusvalo:
- **Fix type analysis of passing slice to dict**
- **Added test**

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -5672,7 +5672,9 @@ class SliceIndexNode(ExprNode):
     def analyse_types(self, env, getting=True):
         self.base = self.base.analyse_types(env)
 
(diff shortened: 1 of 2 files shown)


scoder (approved):

scoder:
We might actually consider this a candidate for a bug fix backport to 3.2.x, but I'd rather stay conservative and keep it to 3.3 for now since it's somewhat unpredictable which side effects better type inference will have on user side.

matusvalo:
I fixed the tests, slice object is hashable only from python 3.12 [1]

@scoder if you approve the PR, you must merge it. I cannot do it since CI is red because of Sanitizers. The same holds also for other PRs.

[1] https://docs.python.org/3/whatsnew/3.12.html#other-language-changes

scoder:
> if you approve the PR, you must merge it

Sure, I was just waiting for it to be validated. Just ping me if I forget to do it.

PR #7703, opened 2026-05-19
Fix cpdef enums with negative values on Python 3.15

### LINKED ISSUE #7701: [BUG] enums with negative values become incorrect in python
### Describe the bug

enums with negative become incorrect in python. See the examples below:

`t.py`:
```cython
cpdef enum ObjectTrackingEvent:
    OBJECT_CREATED = -1
    OBJECT_DESTROYED = -2
```
with `setup.py`:
```py
from setuptools import setup

setup(
    ext_modules = cythonize("t.pyx")
)
```
then in python
```
❯ python
Python 3.15.0b1 (main, May 10 2026, 19:21:12) [Clang 22.1.3 ] on darwin
>>> import t
>>> t.ObjectTrackingEvent.OBJECT_CREATED
<ObjectTrackingEvent.OBJECT_CREATED: 0>
>>> t.ObjectTrackingEvent.OBJECT_DESTROYED
<ObjectTrackingEvent.OBJECT_CREATED: 0>
>>> t.ObjectTrackingEvent.OBJECT_DESTROYED.value
0
>>> t.ObjectTrackingEvent.OBJECT_CREATED.value
0
```

It's more strange if I combine negative and positive numbers like so
```cython
cpdef enum ObjectTrackingEvent:
    TEST1 = 1
    TEST2 = 2
    OBJECT_CREATED = -1
    OBJECT_DESTROYED = -2
    OBJECT_SURVIVED = -999
```
Then this gets me
```
❯ python
Python 3.15.0b1 (main, May 10 2026, 19:21:12) [Clang 22.1.3 ] on darwin
>>> import t
>>> t.ObjectTrackingEvent.TEST1
<ObjectTrackingEvent.TEST1: 1>
>>> t.ObjectTrackingEvent.TEST2
<ObjectTrackingEvent.TEST2: 2>
>>> t.ObjectTrackingEvent.OBJECT_CREATED
<ObjectTrackingEvent.OBJECT_CREATED: 3>
>>> t.ObjectTrackingEvent.OBJECT_SURVIVED
<ObjectTrackingEvent.TEST1: 1>
>>> t.ObjectTrackingEvent.OBJECT_DESTROYED
<ObjectTrackingEvent.TEST2: 2>
```

### Expected behaviour

_No response_

### OS

macOS

### Python version

3.15.0b1

### Cython version

3.3.0a0

### Additional context

_No response_

da-woods:
Fixes #7701

It seems like Python 3.15 has tightened the requirements for `IntFlag` and now converts negative numbers into positive numbers (because it doesn't make sense to do bitwise operations on negative numbers).

We only really use `IntFlag` because people *might* want to do bitwise operations with wrapped enums and it gives them a slightly better experience. Therefore, for negative numbers the easiest solution is just to revert to `IntEnum`. That hopefully shouldn't break any user code - bitwise operations should still work with `IntEnum` but you just lose the enum type when you do them.

It's slightly questionable how much value we're getting from `IntFlag` given that this is now the second workaround we've had to do as they've made it stricter. But I think it still makes sense to keep it where we can.


Cython/Utility/CpdefEnums.pyx
```diff
@@ -15,36 +15,46 @@ from enum import IntEnum as __Pyx_EnumBase
...
+            # (only including powers of 2 when iterating). Since we're using
+            # "flag" because C enums *might* be used as flags, not because
+            # we want strict flag behaviour, manually undo some of this.
+            result._member_names_ = list(result.__members__)
+        return result
+
+
 #################### EnumType ####################
-#@requires: EnumBase
+# requires EnumBase but this is done manually to avoid duplication
```
da-woods, line 31: It seems like `#@requires` was duplicating the code, which suddenly caused problems when I added a function to it. I've taken the lazy solution and not tried to fix the underlying problem.

tests/run/cpdef_enums.pyx
```diff
@@ -285,3 +303,41 @@ def test_special_attributes():
...
+def test_extern_negatives(ExternHasNegatives val=ExternHasNegatives.EX_NEG_A):
+    """
+    >>> val = test_extern_negatives()
+    >>> type(val) == ExternHasNegatives
+    True
+    >>> val < 0
+    True
+
+    # It's still possible to do bitwise operations of numbers and convert them too/from Cython
+    # whether Cython chooses to represent it as an IntEnum or a IntFlag.
```
scoder:
```suggestion
    # It's still possible to do bitwise operations of numbers and convert them to/from Cython
    # whether Cython chooses to represent it as an IntEnum or an IntFlag.
```

tests/run/cpdef_enums.pyx
```diff
@@ -285,3 +303,41 @@ def test_special_attributes():
...
+def test_cy_negatives(CyDefinedHasNegatives val=CyDefinedHasNegatives.CY_NEG_A):
+    """
+    >>> val = test_cy_negatives()
+    >>> type(val) == CyDefinedHasNegatives
+    True
+    >>> val < 0
+    True
+
+    # It's still possible to do bitwise operations of numbers and convert them too/from Cython
+    # whether Cython chooses to represent it as an IntEnum or a IntFlag.
```
scoder:
```suggestion
    # It's still possible to do bitwise operations of numbers and convert them to/from Cython
    # whether Cython chooses to represent it as an IntEnum or an IntFlag.
```

Cython/Compiler/Code.py
```diff
@@ -652,7 +652,7 @@ def load(cls, util_code_name, from_file, **kwargs):
         return cls(**kwargs)
 
     @classmethod
-    def load_cached(cls, utility_code_name, from_file, __cache={}):
+    def load_cached(cls, utility_code_name, from_file, *, __cache={}, **kwds):
```
scoder: If we allow arbitrary keyword arguments to be passed through, then we rather shouldn't cache the result, don't you think? Either, we could explicitly list keyword argument of which we know that they will not impact the code result (in which case I'd wonder why we need to pass them), or we'd need to include the keyword arguments in the cache key (which seems difficult for a module scope), but I think we should just use `.load()` instead of `.load_cached()` if we need additional configuration and not allow additional keyword arguments here.

Cython/Utility/CpdefEnums.pyx
```diff
@@ -15,36 +15,48 @@ from enum import IntEnum as __Pyx_EnumBase
 cdef object __Pyx_FlagBase
 from enum import IntFlag as __Pyx_FlagBase
 
+cdef inline object __Pyx_ConstructEnum(name, items, module, bint safe_flag):
+    if not safe_flag and (__PYX_LIMITED_VERSION_HEX >= 0x030F0000 or
```
scoder: To me, `safe_flag` reads like "flag that this is safe" rather than indicating a clear reference to the `IntFlag` type.
Proposal: `use_IntFlag_type`

Cython/Utility/CpdefEnums.pyx
```diff
@@ -15,36 +15,48 @@ from enum import IntEnum as __Pyx_EnumBase
...
+            (CYTHON_COMPILING_IN_LIMITED_API and __Pyx_get_runtime_version() >= 0x030B0000)):
```
scoder: I don't think the enum type should depend on the Python version. Keep it simple and predictable. If it doesn't look like an `IntFlag` enum, then don't make it an `IntFlag` enum.

Cython/Utility/CpdefEnums.pyx
```diff
@@ -101,7 +119,15 @@ cdef {{funcname}}({{name}} c_val):
...
-        return __pyx_enum(underlying_c_val)
+        try:
+            return __pyx_enum(underlying_c_val)
+        except ValueError:
+            if (__PYX_LIMITED_VERSION_HEX >= 0x030F0000 or
+                    (CYTHON_COMPILING_IN_LIMITED_API and __Pyx_get_runtime_version() >= 0x030B0000)):
+                # In Python 3.15+ we have to use IntEnum which is strict about what values to accept.
+                # In this case the best we can do is return a Python int.
+                return underlying_c_val
+            raise
```
scoder: This looks like something we should try to avoid.

scoder (commented):
ISTM that the main problem is that we're not giving users any chance to influence the type chosen. It's ok to use a heuristic as a default, but without any control, users will always run into mismatches on their side at some point.

A directive might help to cover more cases, and we could see if we can (and should) then allow something along the lines of
```cython
# cython: enum_pytype=IntFlag

cpdef enum Flaggynum: ...

with cython.enum_pytype("IntEnum"):
    cpdef enum BrzlFlf: ...
```
Once that option is in place, we should provide one default, stop guessing, and require users to make an explicit choice for everything that doesn't match the default.

I'm fine with checking for negative values and basing the choice of default on that, but then that's it, that's one simple rule to document for users, you get an `IntFlag` enum unless the values you declared prevent it.

da-woods:
I think there's two separate choices which may be a bit independent: 

* What type do you use to represent the enum (I.e. your `cython.enum_pytype`). I don't think we've seen too much demand for this as a feature in itself so there's probably no need to do this unless we have to.
* Whether you allow unknown values in the "to" (and maybe "from") Python conversion. For C enums I think we do - C treats them as pretty interchangeable with their underlying type and plenty of people so use them as flags, or to mark special values in a range.

> I don't think the enum type should depend on the Python version. Keep it simple and predictable.

Agree with this. 

------------

Thinking about all of this:

Really I've only been using `IntFlags` because it's flexible about unknown values. I'm not really too bothered about the "flag-like" behaviour. It really just gives you a slightly nicer `repr` if you do bitwise operations.

I think the right thing to do is to implement a slightly customised `IntEnum` subclass that allows you to have unknown values. I think that can be done using the documented enum API with `_missing_` and very little else. Such a class would represent the behaviour of C enums well.

(For C++ enums we'd continue to use plain `IntEnum` because the strictness about unknown values matches the c++ behaviour)

scoder:
> implement a slightly customised `IntEnum` subclass that allows you to have unknown values

Sounds best. And when moved into the shared module, it also won't hurt much in terms of module size.

Cython/Compiler/Code.py
da-woods: The arbitrary keywords wasn't needed in the end so I dropped this

da-woods:
> Sounds best. And when moved into the shared module, it also won't hurt much in terms of module size.

It seems like this version adds about 8k onto the object size compared to master.  However about 4k of that is the utility code for py3classcreate.

I don't know if cdef enums are used frequently enough to justify being in the shared utility code. But py3classcreate most likely is.

Cython/Utility/CpdefEnums.pyx
```diff
@@ -1,50 +1,61 @@
...
-        __PYX_LIMITED_VERSION_HEX
-        CYTHON_COMPILING_IN_LIMITED_API
-    int __Pyx_get_runtime_version()
+    object PyImport_Import(object)
+
+cdef object __Pyx_FlexibleEnumBase
+class __Pyx_FlexibleEnumBase(PyImport_Import("enum").IntEnum):
+    @classmethod
+    def _missing_(cls, value):
+        # This is a trimmed down version of "EnumFlag._missing_" with with
```
scoder:
```suggestion
        # This is a trimmed down version of "EnumFlag._missing_" with
```

Cython/Utility/CpdefEnums.pyx
```diff
@@ -1,50 +1,61 @@
...
+        # _value2member_map_ is an undocumented detail of enum so don't fail
+        # if we can't use it to cache pseudo-members
+        if (value2member_map := getattr(cls, '_value2member_map_')) is not None:
+            pseudo_member = value2member_map.setdefault(value, pseudo_member)
+        return pseudo_member
+
+    def __repr__(self):
+        if self._name_ is None:
+            # arbitrary value pseudo member
+            return "<%s: %s>" % (self.__class__.__name__, repr(self._value_))
```
scoder:
```suggestion
            return f"<{self.__class__.__name__}: {self._value_!r}>")
```

Cython/Utility/CpdefEnums.pyx
```diff
@@ -1,50 +1,61 @@
...
+    def _missing_(cls, value):
+        # This is a trimmed down version of "EnumFlag._missing_" with with
+        # the flag-specific details removed.
+        pseudo_member = int.__new__(cls, value)
+        if not hasattr(pseudo_member, '_value_'):
+            pseudo_member._value_ = value
+        pseudo_member._name_ = None
+        # _value2member_map_ is an undocumented detail of enum so don't fail
+        # if we can't use it to cache pseudo-members
+        if (value2member_map := getattr(cls, '_value2member_map_')) is not None:
```
scoder: Without the 3-arg `getattr()`, this would still have failed.
```suggestion
        value2member_map = getattr(cls, '_value2member_map_', None)
        if value2member_map is not None:
```

Cython/Utility/CpdefEnums.pyx
```diff
@@ -1,50 +1,61 @@
...
-    int __Pyx_get_runtime_version()
+    object PyImport_Import(object)
+
+cdef object __Pyx_FlexibleEnumBase
+class __Pyx_FlexibleEnumBase(PyImport_Import("enum").IntEnum):
+    @classmethod
+    def _missing_(cls, value):
+        # This is a trimmed down version of "EnumFlag._missing_" with with
+        # the flag-specific details removed.
+        pseudo_member = int.__new__(cls, value)
```
scoder, line 12: The direct call to `int.__new__()` skips one hierarchy level. Not sure how likely this is to pose problems in the future, but it's not impossible. We could keep this and wait for CPython to change, although it would then be good to have a way to detect it early.

Cython/Utility/CpdefEnums.pyx
```diff
@@ -1,50 +1,61 @@
...
+# requires EnumBase but this is done manually to avoid duplication
 
 cdef extern from *:
     object {{enum_to_pyint_func}}({{name}} value)
 
-# create new IntFlag() - the assumption is that C enums are sufficiently commonly
-# used as flags that this is the most appropriate base class
-{{name}} = __Pyx_FlagBase('{{name}}', [
+
+# Create new IntFlag() if possible:
```
scoder: Not sure what you mean by "is possible" here.
```suggestion
# Create new IntFlag()-like enums:
```

scoder (approved):
This seems the right kind of change for 3.3.

Cython/Utility/CpdefEnums.pyx
da-woods: I think it's unlikely to change.
`Enum.__new__` is what calls `_missing_` and so if we don't skip a level then it just ends up infinitely recursive. That feels fairly built-in to the interface of `enum`.
I'm not quite sure what would be the right thing to detect to detect it early.

PR #7704, opened 2026-05-19
Implement support of subscripted types of tuple

matusvalo:
This PR needs some love but main spirit is there. Some notes:

* This PR will not support subscript with ellipsis `tuple[int, ...]`
* This PR will not support inference when initialized: `a = (1, 'bar')`

TODO:
* [x] make CI green
* [x] tests with negative scenarios
* [x]  https://github.com/cython/cython/pull/7704#discussion_r3293964487
* [x] https://github.com/cython/cython/pull/7704#discussion_r3293977702
* [x] https://github.com/cython/cython/pull/7704#discussion_r3294163830


tests/run/pep448_extended_unpacking.pyx
```diff
@@ -614,7 +614,7 @@ def unpack_in_loop(a: list[float], b: list[float]):
     [1.0, 2.0, 3.0, [4.0, 5.0]]
     Python object
     [[1.0, 2.0, 3.0], 4.0, 5.0]
-    list[double] object
+    Python object
```
scoder: This looks like it should work, shouldn't it?

Cython/Compiler/PyrexTypes.py
```diff
@@ -4949,6 +4949,9 @@ def get_subscripted_type(self, index: int):
         except IndexError:
             return None
 
+    def get_subscripted_spanning_type(self):
```
```suggestion
    def get_common_item_type(self):
```

Cython/Compiler/PyrexTypes.py
```diff
@@ -5028,14 +5031,23 @@ def assignable_from(self, src_type):
...
 
-    def infer_indexed_type(self):
+    def infer_indexed_type(self, index=None):
         container_type = self.get_container_type()
+        if container_type.is_pytuple_type:
+            if index is None:
+                return self.get_subscripted_spanning_type()
+            if isinstance(index, int):
+                return self.get_subscripted_type(index)
+            return py_object_type
```
scoder: It took me a sleepy morning moment to understand why you're passing the index value here (not its node, type, etc.). Let's give it a clearer name. :)

And I think by moving the spanning type last, we can avoid a special case. If we can't determine which index to look at, fall back to aggregating all types.

```suggestion
    def infer_indexed_type(self, at_index=None):
        container_type = self.get_container_type()
        if container_type.is_pytuple_type:
            if isinstance(at_index, int):
                return self.get_subscripted_type(at_index)
            return self.get_common_item_type()
```

scoder (commented):
Looks good to me overall.

Cython/Compiler/PyrexTypes.py
scoder: BTW, we could evaluate the reduced spanning type once, when instantiating the type, and store it away in an attribute, say, `.common_item_type`, but for all container types. That way, we'd also have a common way to access it in one place (with the twist that `infer_iterator_type()` would still need a special case for dicts).

tests/run/pep448_extended_unpacking.pyx
matusvalo: Yes this I need to investigate why this changed.

scoder: Could you try my proposed change for `infer_indexed_type()` ? It might resolve it.

Cython/Compiler/PyrexTypes.py
matusvalo: Renamed

Cython/Compiler/PyrexTypes.py
matusvalo: I slightly refactor `infer_indexed_type()` but proposed function does not work. Have a look.

>  we could evaluate the reduced spanning type once, when instantiating the type, and store it away in an attribute, say, .common_item_type, but for all container types

This is pending

tests/run/pep448_extended_unpacking.pyx
matusvalo: unfortunatelly, it is not fixing the issue, **but** I was able to fixit 💪

scoder:
> proposed function does not work.

Oh, right. There's a bigger problem hidden here. We don't have a dedicated type for `None`, it's just `object`. For type inference, what we did before, was to exclude `None` value nodes and then construct the common type from the rest. In `reduce_spanning_type()`, we only have the types and thus cannot detect `None`. The result always ends up as `object`. :-/

matusvalo:
> We don't have a dedicated type for `None`

We can have a look on that. For now, I (somehow) fixed the highlighted problem. Now, I will try to make CI/CD green (there is still long road to go). Then I will mark this PR as ready for review and we can have a round how to do it properly...

Cython/Compiler/PyrexTypes.py
```diff
@@ -4976,14 +4981,16 @@ class BuiltinTypeConstructorObjectType(BuiltinObjectType, PythonTypeConstructorM
     def __init__(self, name, cname, objstruct_cname=None, **kwargs):
         super().__init__(
             name, cname, objstruct_cname=objstruct_cname)
+        if name in {'list', 'set', 'frozenset'}:
```
matusvalo, line 4984: Here we can do something even more smarter e.g.:
* `tuple[int, int, int]` does have uniform elements
* `tuple[int, ...]` will have uniform elements too (when supported)

matusvalo: Hmm now I am thinking. slicing inference can be done safely only when we support `...` in tuples since:

```python
l: tuple[int, int, int] = (1, 2, 3)
l[i:j] # this slice must have type either tuple or tuple[int, ...] since we don't know values of i,j
```

tests/run/pure_ctuple.py
```diff
@@ -75,6 +75,19 @@ def test_tuple(a: typing.Tuple[cython.int, cython.float], b: typing.Tuple[cython
     print(cython.typeof(c) + (" object" if not cython.compiled else ""))
     print(cython.typeof(plain_tuple) + (" object" if not cython.compiled else ""))
 
+if cython.compiled:
```
matusvalo: For this test and test below, there are several TODOs that I believe are fixed now. I think even the tests can be slightly cleaned up...

scoder: > the tests can be slightly cleaned up

Yes, could you merge the compiled and Python cases back into the `print()` lines above? And the same for the other test. It's better to see the differences in one line than in separate docstrings (which I might even miss at first sight).

Cython/Compiler/ExprNodes.py
scoder, line 9111: This should use `infer_sequence_item_type()` (in this module). That handles `None` values and starred values correctly.

Cython/Compiler/ExprNodes.py
```diff
@@ -5609,6 +5613,10 @@ def infer_type(self, env):
             return bytes_type
         elif base_type.is_pyunicode_ptr:
             return unicode_type
+        elif base_type.supports_container_type and not base_type.has_uniform_element_type:
+            # slicing a container with non-uniform element types (tuple[int, str]) depends on the slice indices,
+            # so we can't infer a more specific type here.
+            return base_type.get_container_type()
```
scoder, line 5630: I don't think this should be necessary. `self.base.infer_type()` must only return a specialised (i.e. subscripted) type if its elements are of uniform type, in which case any slice will also be uniform. If the base container is not known to be homogeneous, it must return the plain type without further specialisation. In either case, it should be correct to pass the base type through. The rest is just a potential optimisation for known slices with only uniform item types.

EDIT: The same applies to the equivalent change in `IndexNode`.

matusvalo:
OK I did another round of changes to make CI/CD green:
* I introduced `py_none_type` - currently it is just special instance of `PyObjectType`. I think, later in needs more love - to be defined what exactly, we can do separate class for that etc. Now, I would like to keep it simple and stupid.
* I changed several types in `type_inference`. The goal is to make CI/CD green. Later when CI/CD is green, we can go over it and defined what change make sense what no.

Long story short: **Now my goal is to make CI/CD green.** After that point, I will try to understand which "fixes" needs to be revisited.

Cython/Compiler/PyrexTypes.py
```diff
@@ -5028,14 +5035,22 @@ def assignable_from(self, src_type):
...
         container_type = self.get_container_type()
-        if container_type.is_pydict_type or container_type.is_pyfrozendict_type:
+        if at_index is None:
+            return self.get_common_item_type()
+        if container_type.is_pytuple_type and isinstance(at_index, int):
+            return self.get_subscripted_type(at_index)
+        if container_type.is_pyanydict_type:
             return self.get_subscripted_type(1)
-        else:
+        if container_type.is_pylist_type or container_type.is_pyanyset_type:
```
matusvalo, line 5046: Here I needed to put explicitly set check because for several tests it makes difference whether we return here `None` (via `get_subscripted_type(0)`) or `PyObject` (via `get_common_item_type()`). Currently, I don't fully understand why.

scoder:
> I introduced `py_none_type`

Hmm. When I wrote "we don't have it" I didn't mean "we need it". I think we can probably get away with not having it for this PR. See my other comments, `infer_sequence_item_type()` pretty much does what's needed here.

matusvalo:
> I think we can probably get away with not having it for this PR. See my other comments, infer_sequence_item_type() pretty much does what's needed here.

I will implement it (or at least I will try). But I would like to have green CI so I have something I can verify against my changes. Now it is stupidly simple fix. Then I will do it properly.

matusvalo:
OK. CI/CD seems to be green. I marked PR as ready for review but there are still things to be done though...

>  When I wrote "we don't have it" I didn't mean "we need it". I think we can probably get away with not having it for this PR. See my other comments, infer_sequence_item_type() pretty much does what's needed here.

I think this is fixed now...

Cython/Compiler/ExprNodes.py
matusvalo: I think this if fixed now...

tests/errors/e_typing_errors.pyx
```diff
@@ -59,6 +61,15 @@ def subscripted_types_assignments():
 
 
 def subscripted_types_assignments_to_variable():
+    ta: tuple[str, cython.float] = ('foo', 5.0)
+    tb: tuple[str, cython.int] = ('bar', 1)
+    z: cython.int = ta[1]
+    zz: cython.float = 1.0
+    tb[1] = zz
+    h: cython.int
+    for h in ta:
```
matusvalo, line 70: This test needs to be updated to get an error here.

tests/run/pep526_variable_annotations_cy.pyx
```diff
@@ -131,6 +131,55 @@ def test_tuple_assignment(i, list[int] la, set[int] sa, dict[int, int] da, froze
...
+    print(cython.typeof(ta[2]))
+    print('ta[3]:')
+    print(cython.typeof(ta[3]))
+    print(cython.typeof(ta[3][0]))
+    print(cython.typeof(ta[3][1]))
+    print(cython.typeof(ta[3][0][0]))
+    print(cython.typeof(ta[3][0][1]))
+    print(cython.typeof(ta[3][1][0]))
+    print(cython.typeof(ta[3][1][1]))
+
```
scoder, line 182:
```suggestion


```

tests/run/pep526_variable_annotations.py
```diff
@@ -341,6 +397,34 @@ def test_dict_with_subscript():
...
+    b: tuple = a
+    c: tuple[cython.float, cython.float] = b
+    print(cython.typeof(a[0]), cython.typeof(a[1]))
+    print(cython.typeof(b[0]), cython.typeof(b[1]))
+    print(cython.typeof(c[0]), cython.typeof(c[1]))
+    print(a[0], b[0], c[0])
+    print(a[1], b[1], c[1])
+
+if cython.compiled:
+    test_assignment_tuple_with_subscript.__doc__ = """
```
scoder: Please don't use this pattern. It makes it too easy to miss the doc string overwrite when reading the test function itself, and it also requires jumping back and forth across a long piece of code to compare the output of Python and Cython. And then, we are testing Cython's type system here, not Python's. The test should directly show the expected output for the Cython code, with adaptations for Python only where needed, to make it work there as well.

tests/errors/e_typing_errors.pyx
scoder: You mean a compile time error, rather than the runtime error that you'd get either way? I'm not sure it should be an error. It might just be a warning case.

… and I only just noticed that we already reject the other float to int assignments here. I'm not sure if that's a bug or feature. I understand that we cannot allow assignments of Python `float` to Python `int`, they are different objects. But C `float` to C `int`, as declared here? Why shouldn't that work?

Cython/Compiler/ExprNodes.py
```diff
@@ -3523,8 +3523,15 @@ def infer_type(self, env, iterator_type=None):
...
-        elif (sequence_type := self.iterator.sequence.infer_type(env)).supports_container_type and \
-                (iterator_type := sequence_type.infer_iterator_type()):
+        elif (
+            (sequence_type := self.iterator.sequence.infer_type(env)).supports_container_type and
+            not (
+                self.iterator.sequence.is_sequence_constructor or
+                self.iterator.sequence.is_dict_literal or
+                self.iterator.sequence.is_set_literal
+            ) and
+            (iterator_type := sequence_type.infer_iterator_type())
```
scoder: There is no need for the walrus operator here. Just split the if-elif-cases into separate if-statements and do whatever you need to do between them.

If you then reverse the first two conditions, and put them into nested if-blocks, you can avoid calling the potentially costly `infer_type()` if it's not needed and also bring the assignment and usage of `sequence_type` closer together.

I also suggest assigning `self.iterator.sequence` to a local variable to make its various usages easier to spot in the long condition.

Cython/Compiler/ExprNodes.py
```diff
@@ -4294,7 +4305,11 @@ def infer_type(self, env):
                 # TODO: Handle buffers (hopefully without too much redundancy).
                 return py_object_type
 
-        if base_type.supports_container_type and (sub_type := base_type.infer_indexed_type()):
+        if (
+            base_type.supports_container_type and
+            not (self.base.is_sequence_constructor or self.base.is_dict_literal or self.base.is_set_literal) and
+            (sub_type := base_type.infer_indexed_type(self.index.constant_result))
+        ):
             return sub_type
```
scoder: Here as well, no need for long chains of conditions. Just use nested if-blocks.

Cython/Compiler/ExprNodes.py
scoder: It's not using `infer_sequence_item_type()` but instead needlessly duplicating a part of its logic.

Cython/Compiler/PyrexTypes.py
scoder: If the interfaces of the two methods disagree about the representation of "don't know", then we need to make both consistent. You can map `py_object_type` to `None` in `get_common_item_type`.

Cython/Compiler/ExprNodes.py
matusvalo: I think this should be fixed.

Cython/Compiler/ExprNodes.py
```diff
@@ -3519,21 +3519,27 @@ def infer_type(self, env, iterator_type=None):
...
-                index=IntNode(self.pos, value='PY_SSIZE_T_MAX',
-                              type=PyrexTypes.c_py_ssize_t_type))
-            return fake_index_node.infer_type(env)
+        sequence_node = self.iterator.sequence
+        if not (
+            sequence_node.is_sequence_constructor or
+            sequence_node.is_dict_literal or
+            sequence_node.is_set_literal
+        ):
+            # Here we infer only non-literal squences. Literals are infered via special infer_sequence_item_type().
```
```suggestion
            # Here we infer only non-literal sequences. Literals are inferred via special infer_sequence_item_type().
```

Cython/Compiler/ExprNodes.py
```diff
@@ -4285,6 +4291,10 @@ def infer_type(self, env):
             elif base_type.is_pyunicode_ptr:
                 # sliced Py_UNICODE* strings must coerce to Python
                 return unicode_type
+            elif base_type.supports_container_type and not base_type.has_uniform_element_type:
+                # slicing a container with non-uniform element types (tuple[int, str]) depends on the slice indices,
+                # so we can't infer a more specific type here.
```
```suggestion
                # Slicing a container with non-uniform element types (tuple[int, str]) depends on the slice indices,
                # so we can't infer a more specific type here.
```

tests/run/pure_ctuple.py
matusvalo: Fixed.

Cython/Compiler/ExprNodes.py
```diff
@@ -3519,21 +3519,27 @@ def infer_type(self, env, iterator_type=None):
...
+        sequence_node = self.iterator.sequence
+        if not (
+            sequence_node.is_sequence_constructor or
+            sequence_node.is_dict_literal or
+            sequence_node.is_set_literal
+        ):
+            # Here we infer only non-literal sequences. Literals are inferred via special infer_sequence_item_type().
+            sequence_type = sequence_node.infer_type(env)
+            if sequence_type.supports_container_type and (iterator_type := sequence_type.infer_iterator_type()):
+                return iterator_type
```
scoder: There are certainly cases where assignment expressions make the code more compact and/or readable, but I consider it a feature that comes with downsides. It always _increases_ the code complexity by mixing a simple assignment with a simple condition into a lengthy  and less readable expression. If it can easily be avoided, as in this case, I think it's worth keeping the code simple.
```suggestion
            if sequence_type.supports_container_type:
                iterator_type = sequence_type.infer_iterator_type():
                if iterator_type:
                    return iterator_type
```

Cython/Compiler/ExprNodes.py
```diff
@@ -4294,7 +4304,11 @@ def infer_type(self, env):
                 # TODO: Handle buffers (hopefully without too much redundancy).
                 return py_object_type
 
-        if base_type.supports_container_type and (sub_type := base_type.infer_indexed_type()):
+        if (
+            base_type.supports_container_type and
+            not (self.base.is_sequence_constructor or self.base.is_dict_literal or self.base.is_set_literal) and
+            (sub_type := base_type.infer_indexed_type(self.index.constant_result))
+        ):
             return sub_type
```
```suggestion
        if base_type.supports_container_type:
            if not (self.base.is_sequence_constructor or self.base.is_dict_literal or self.base.is_set_literal):
                sub_type = base_type.infer_indexed_type(self.index.constant_result)
                if sub_type:
                    return sub_type
```

Cython/Compiler/ExprNodes.py
scoder: This method looks fine now.

scoder (commented):
Looks good overall. I wouldn't mind merging it, even though the remaining comments open some room for further improvements.

matusvalo:
Thank you @scoder  for your patience and review. I am trying to work step-by-step on your comments but I have a limited time so I pick one problem at a time try to fix it.

Cython/Compiler/ExprNodes.py
scoder: I now think that my assumption was not correct. I was thinking of literal lists/set/etc., not of tuples. For the latter, the specialised type is the correct result of type inference, so this case is needed.

jakirkham:
Thank you for your hard work here! 🙏

This fixes https://github.com/cython/cython/issues/3800

PR #7705, opened 2026-05-20
Enhance CodeWriter's current captabilities

Vizonex:
I'm thinking about write a new formatter (likely going to be a future addition to cython-lint) and in doing so I thought about ensuring that the current `CodeWriter` has it's current scope expanded by looking into some unimplemented stuff. Luckily I have already found a few and I've written another test module since so I could attempt to unit test for missing items.


Cython/CodeWriter.py
```diff
@@ -265,10 +268,38 @@ def visit_CFuncDefNode(self, node):
         self.visit(node.declarator.base)
         self.put('(')
         self.comma_separated_list(node.declarator.args)
-        self.endline('):')
+        self.put(')')
```
da-woods, line 271: Now you've removed `endline` from `CFuncDeclaratorNode` this seems identical to the contests of `visit_CFuncDeclaratorNode`.

It might be worth thinking about where else `CFuncDeclaratorNode` can be used? Possibly `cdef extern` blocks? Do your changes break that?

Cython/CodeWriter.py
```diff
@@ -265,10 +268,38 @@ def visit_CFuncDefNode(self, node):
...
+
+        if node.declarator.has_explicit_exc_clause:
```
da-woods, line 273: This exception stuff feels like it should be part of handling the declarator, rather than here

PR #7707, opened 2026-05-20
Optimize kwdless callmethod for CYTHON_VECTORCALL=0

da-woods:
Essentially to improve older Limited API.

My reasoning is that using CallMethodObjArgs does two things:
1. It avoids GetAttr and thus the need to create a bound method.
2. It allows can internally use vectorcall (at the cost of having to deal with c variadic arguments itself).

I think point 1 is the most important, and this is why I haven't bothered to do for function (rather than method) calls.

The number 5 is a bit arbitrary. It's partly to avoid having to write an excess amount of cases. It's also the number of arguments Python handles on the stack inside PyObject_CallMethodObjArgs (and above this it switches to heap allocation).


da-woods:
A microbenchmark to justify it:

```
cdef class C:
    def f(self, a=None, b=None, c=None, d=None, e=None, f=None):
        pass

from timeit import timeit

def callf0(C c):
    c.f()

def callf1(C c):
    c.f(1)

def callf2(C c):
    c.f(1, 2)

def callf3(C c):
    c.f(1, 2, 3)

def callf4(C c):
    c.f(1, 2, 3, 4)

def callf5(C c):
    c.f(1, 2, 3, 4, 5)

def callf6(C c):
    c.f(1, 2, 3, 4, 5, 6)

def fmt0(s):
    s.format()

def fmt1(s):
    s.format(1)

def fmt2(s):
    s.format(1, 2)

def fmt3(s):
    s.format(1, 2, 3)

def fmt4(s):
    s.format(1, 2, 3, 4)

def fmt5(s):
    s.format(1, 2, 3, 4, 5)

def fmt6(s):
    s.format(1, 2, 3, 4, 5, 6)

fs = frozenset()

def union0(s):
    s.union()

def union1(s):
    s.union(fs)

def union2(s):
    s.union(fs, fs)

def union3(s):
    s.union(fs, fs, fs)

def union4(s):
    s.union(fs, fs, fs, fs)

def union5(s):
    s.union(fs, fs, fs, fs, fs)

def union6(s):
    s.union(fs, fs, fs, fs, fs, fs)

c = C()

s = set()

for name in list(globals().keys()):
    if not (name.startswith("callf")):
        continue
    print(name, timeit(f"{name}(c)", globals=globals(), number=1000000))

for name in list(globals().keys()):
    if not (name.startswith("union")):
        continue
    print(name, timeit(f"{name}(s)", globals=globals(), number=1000000))


for name in list(globals().keys()):
    if not name.startswith("fmt"):
        continue
    number = int(name[3:])
    print(name, timeit(f"{name}('{'{}'*number}')", globals=globals(), number=1000000))
```

Run on Python 3.14, with `Py_LIMITED_API=0x030A0000` and `-O2`.


<html>
<head>
	
	
	<title></title>
	<meta name="generator" content="LibreOffice 24.2.4.2 (Windows)"/>
	<style type="text/css">
		body,div,table,thead,tbody,tfoot,tr,th,td,p { font-family:"Liberation Sans"; font-size:x-small }
		a.comment-indicator:hover + comment { background:#ffd; position:absolute; display:block; border:1px solid black; padding:0.5em;  } 
		a.comment-indicator { background:red; display:inline-block; border:1px solid black; width:0.5em; height:0.5em;  } 
		comment { display:none;  } 
	</style>
	
</head>

<body>

  | With | Without
-- | -- | --
callf0 | 0.1797 | 0.2179
callf1 | 0.1532 | 0.2653
callf2 | 0.2314 | 0.2255
callf3 | 0.2380 | 0.2421
callf4 | 0.2488 | 0.2463
callf5 | 0.2729 | 0.2837
callf6 | 0.2912 | 0.2807
union0 | 0.1348 | 0.1911
union1 | 0.1729 | 0.2253
union2 | 0.2200 | 0.3642
union3 | 0.2452 | 0.3164
union4 | 0.2825 | 0.3387
union5 | 0.3788 | 0.3830
union6 | 0.4265 | 0.4124
fmt0 | 0.1029 | 0.1172
fmt1 | 0.2406 | 0.2792
fmt2 | 0.2997 | 0.3004
fmt3 | 0.3594 | 0.3498
fmt4 | 0.4245 | 0.4230
fmt5 | 0.4902 | 0.4978
fmt6 | 0.5461 | 0.5736


</body>

</html>

For reference, `set.union` looks to use vectorcall internally, `str.format` doesn't, and our functions also won't in this case. It speeds up all the cases which I think indicates that the benefit mostly in not needing to create the method object.

da-woods:
> why I haven't bothered to do for function (rather than method) calls

I did a quick check to see what difference it made and it I couldn't produce a measure-able difference for function calls (either better or worse)

PR #7709, opened 2026-05-23
Fix tests with asyncio.iscoroutinefunction on Python 3.16

da-woods:
It was deprecated on Python 3.14 and removed in Python 3.16.

`inspect.iscoroutinefunction` looks to be available on all versions we care about, so just use that.

It looks like we only need `cyfunction._is_coroutine` on Python 3.9 and some early Python 3.10 builds. After that `inspect.iscoroutinefunction` will correctly identify cyfunctions without any help. Therefore I've disabled it on versions where I believe it's useless.  This isn't strictly necessary to fix the 3.16 test failures though

### DIFF
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -93,8 +93,13 @@ typedef struct {
     PyObject *(*defaults_getter)(PyObject *);
     PyObject *func_annotations; /* function annotations dict */
 
-    // Coroutine marker
+    // Coroutine marker (inspect and asyncio versions)
+#if __PYX_LIMITED_VERSION_HEX < 0x03100000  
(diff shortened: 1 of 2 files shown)


scoder:
Thanks for investigating this. I wouldn't remove the attribute in 3.2.5 but extracted the test fix part:
https://github.com/cython/cython/commit/ec152091ca7c535e04b5da96e5bedfc9b33def22

da-woods:
Yes - I think just fixing the test is reasonable for 3.2.5

da-woods:
I'll merge this into master because I think that dropping `_is_coroutine` makes sense. But if it doesn't the only way we're likely to find out is if someone complains about it. And merging it maximizes the changes of that.

PR #7710, opened 2026-05-23
CI: Split special jobs into separate groups.

scoder:
Allows configuring them independently rather than in one big matrix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,8 @@ jobs:
       matrix:
         os: [ubuntu-24.04, windows-2022, macos-latest]
         backend: [c, cpp]
+        limited_api: ['']
+        shared_utility: ['']
         extra_hash: ['']
 
         include:
@@ -68,11 +70,11 @@ jobs:
           - os: ubuntu-24.04
             backend: "c"
             shared_utility: "--shared-utility"
-            extra_hash: "--shared-utility"
+            extra_hash: "-shared-utility"
           - os: ubuntu-24.04
             backend: "cpp"
             shared_utility: "--shared-utility"
-            extra_hash: "--shared-utility"
+            extra_hash: "-shared-utility"
 
     uses: ./.github/workflows/ci-job.yml
     with:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -139,146 +148,206 @@ jobs:
...
+
+  ci-sharedmodule:
+    # compile with shared utility module
+    needs: [ci-early-success]
+
+    strategy:
+      # Allows for matrix sub-jobs to fail without canceling the rest
+      fail-fast: false
+      matrix:
+        backend: [c,cpp,"c,cpp"]
```
da-woods: What's the "c,cpp" adding here?

scoder: I was thinking that it would be good to test that the module can be shared between both variants of modules. But looking it up now, the shared module is _always_ generated as C module, so the C++ test run already validates that. I'll remove the case.

.github/workflows/ci.yml
```diff
@@ -139,146 +148,213 @@ jobs:
...
+
+  ci-compileall:
+    # compile all modules
+    needs: [ci-early-success]
+
+    strategy:
+      # Allows for matrix sub-jobs to fail without canceling the rest
+      fail-fast: false
+      matrix:
+        backend: [c, cpp]
```
da-woods, line 233:
```suggestion
        backend: [c]
```

I know this is just copying what we already have. However, I wonder if it's worth dropping `cpp` here.

In that `compile-all` failure are almost always at the "build cython" stage (which is always done in C anyway).

I don't think I've ever seen a failure here were something breaks only on the cpp tests and only on the "compile-all" Cython.

.github/workflows/ci.yml
```diff
@@ -139,146 +148,213 @@ jobs:
...
+    needs: [ci-early-success]
+
+    strategy:
+      # Allows for matrix sub-jobs to fail without canceling the rest
+      fail-fast: false
+      matrix:
+        os: [ubuntu-24.04-arm, windows-11-arm]
+        backend: [c, cpp]
+        python-version:
+          - "3.13"
```
da-woods:
```suggestion
          - "3.x"
```

Slight change to what we already have, but I wonder if "latest release version" is a good choice for the arm tests

da-woods (commented):
I think splitting it up makes sense. How `include` works here has some weird features that we've got wrong before (e.g. https://github.com/cython/cython/pull/7652) and this probably makes it less likely to happen again.

.github/workflows/ci.yml
scoder: Still, it's the only time where we test Cython with all (reasonably) compiled modules. Excluding the C++ tests would mean that we don't test that part of the code in its compiled form. I haven't looked closely, but I doubt that there is much C++ related code that gets compiled, so your question is certainly worth considering. But I'd lean towards keeping it covered regardless.

PR #7711, opened 2026-05-24
pypy fails test_nocopy_star_args

mattip:
PyPy does not pass through starargs, it creates a new tuple:

```
def f(*args):
    return args
```

See issue pypy/pypy#5486. For the meantime, let's skip the test on PyPy as an implementation detail.


da-woods:
I don't really see the need to change PyPy behaviour for this. I think just skipping the test is reasonable.

(Although obviously, what you do to PyPy is your decision and not mine :))

da-woods:
I've clicked auto-merge but I don't think it's going to let me do it (because the "sanitizers" run is required and the changes you've made don't trigger it).

Either way, thanks - seems like a sensible thing to do.

PR #7712, opened 2026-05-24
[3.2.x] Fix potential double-free in tstring code

da-woods:
This is most likely to happen if someone else using someone else's tstring backport. In this case the call with keyword arguments can fail (because it's something that's only expected to pass on our backport).


da-woods:
This is targetted at 3.2.x because I have a more drastic change to 3.3 coming.

scoder (approved):
Good catch, but bad timing right after the release. :-/

PR #7713, opened 2026-05-24
Let vectorcall use cached literal kwnames tuples

da-woods:
Most of the time, function/method calls have fixed keyword arguments, and so it's possible to skip building the `kwnames` tuple and instead used a cached literal. This can end up being a significant time saving where used.

As a side-effect of this I've eliminated my "vectorcall builder" "helper". It was a bit confusing to use and a bit hard to read so I don't think this is a great loss. I've removed the two places where it was used in the utility code.

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -7088,7 +7088,7 @@ class PyMethodCallNode(CallNode):
             code.put_error_if_neg(kwnames.pos, f"__Pyx_CheckVectorcallKwarg({kwnames_temp}, {arg_index})")
         code.putln("#else")
         code.putln("{")
-        kwnames.generate_sequence_as_array_code(code, Naming.quick_temp_cname);
+        kwnames.generate_sequence_as_array_code(code, Naming.quick_temp_cname)
         for arg_index in arg_indices_to_check:
             code.put_error_if_neg(kwnames.pos, f"__Pyx_CheckVectorcallKwarg({Naming.quick_temp_cname}, {arg_index})")
         code.putln(f"{kwnames_temp} = __Pyx_MakeKwargDict({Naming.quick_temp_cname}, "
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -7082,7 +7082,7 @@ class PyMethodCallNode(CallNode):
 
         code.putln("#if CYTHON_VECTORCALL")
         code.putln(f"{kwnames_temp} = {kwnames.result()};")
-        code.putln(f"{code.error_goto_if_null(kwnames_temp, self.pos)}")
+        code.putln(code.error_goto_if_null(kwnames_temp, self.pos))
         code.put_incref(kwnames_temp, py_object_type)
         for arg_index in arg_indices_to_check:
             code.put_error_if_neg(kwnames.pos, f"__Pyx_CheckVectorcallKwarg({kwnames_temp}, {arg_index})")
@@ -7093,7 +7093,7 @@ class PyMethodCallNode(CallNode):
             code.put_error_if_neg(kwnames.pos, f"__Pyx_CheckVectorcallKwarg({Naming.quick_temp_cname}, {arg_index})")
         code.putln(f"{kwnames_temp} = __Pyx_MakeKwargDict({Naming.quick_temp_cname}, "
                    f"{Naming.callargs_cname}+{len(args)+1}, {len(kwvalues)});")
-        code.putln(f"{code.error_goto_if_null(kwnames_temp, self.pos)}")
+        code.putln(code.error_goto_if_null(kwnames_temp, self.pos))
         code.put_gotref(kwnames_temp, py_object_type)
         code.putln("}")
         code.putln("#endif")
@@ -7163,11 +7163,7 @@ class PyMethodCallNode(CallNode):
         # To avoid passing an out-of-bounds argument pointer in the no-args case,
         # we need at least two entries, so we pad with NULL and point to that.
         # See https://github.com/cython/cython/issues/5668
-        args_and_kwargs = []
-        if args:
-            args_and_kwargs.extend(args)
-        if kwvalues:
-            args_and_kwargs.extend(kwvalues)
+        args_and_kwargs = args + (kwvalues or [])
         args_list = ', '.join(arg.py_result() for arg in args_and_kwargs) if args_and_kwargs else "NULL"
         code.putln(
             f"PyObject *{Naming.callargs_cname}[{(len(args_and_kwargs) + 1) if args_and_kwargs else 2:d}] = {{{self_arg}, {args_list}}};"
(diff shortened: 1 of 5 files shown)


Cython/Compiler/ExprNodes.py
```diff
@@ -7061,38 +7074,42 @@ def generate_runtime_method_unpacking_code(self, code, self_arg, space_for_selfa
...
-            )
+        code.putln("#if CYTHON_VECTORCALL")
+        code.putln(f"{kwnames_temp} = {kwnames.result()};")
+        code.putln(f"{code.error_goto_if_null(kwnames_temp, self.pos)}")
+        code.put_incref(kwnames_temp, py_object_type)
+        for arg_index in arg_indices_to_check:
+            code.put_error_if_neg(kwnames.pos, f"__Pyx_CheckVectorcallKwarg({kwnames_temp}, {arg_index})")
+        code.putln("#else")
+        code.putln("{")
+        kwnames.generate_sequence_as_array_code(code, Naming.quick_temp_cname);
```
```suggestion
        kwnames.generate_sequence_as_array_code(code, Naming.quick_temp_cname)
```

scoder (commented):
Seems a good idea but currently crashes.

Cython/Compiler/ExprNodes.py
```diff
@@ -6903,25 +6903,38 @@ class PyMethodCallNode(CallNode):
...
+        kwargs_key_value_pairs = kw.pop('kwargs_key_value_pairs', None)
+        if kwdict and kwdict.is_dict_literal:
+            assert not kwargs_key_value_pairs
+            kwargs_key_value_pairs = kwdict.key_value_pairs
+            del kw['kwdict']
+        if kwargs_key_value_pairs:
+            kwnames = TupleNode(pos, args=[kvp.key for kvp in kwargs_key_value_pairs])
+            kwnames = kwnames.analyse_types(env, skip_children=True)
+            kw['kwnames'] = kwnames
+            kw['kwvalues'] = [kvp.value for kvp in kwargs_key_value_pairs]
```
scoder: I can't find a place where `kwargs_key_value_pairs` is passed as keyword argument. So I think there's room for simplification here.
```suggestion
        kwdict = kw.get('kwdict', None)
        if kwdict and kwdict.is_dict_literal:
            kwargs_key_value_pairs = kwdict.key_value_pairs
            kwnames = TupleNode(pos, args=[kvp.key for kvp in kwargs_key_value_pairs])
            kw['kwnames'] = kwnames.analyse_types(env, skip_children=True)
            kw['kwvalues'] = [kvp.value for kvp in kwargs_key_value_pairs]
            del kw['kwdict']
```
And I wonder if we can't avoid passing `env` here only for that purpose. `PyMethodCallNode` is only created in two places, inside of a transform, and only one of those deals with keyword arguments. Why not create the tuple there and pass it in? A transform seems a more appropriate place to do this kind of transformation than a constructor.

Cython/Compiler/ExprNodes.py
```diff
@@ -8755,6 +8780,13 @@ def generate_sequence_packing_code(self, code, target=None, plain=False):
             code.putln('%s = %s;' % (target, Naming.quick_temp_cname))
             code.putln('}')
 
+    def generate_sequence_as_array_code(self, code, target):
+        # Currently only suitable for fixed-size sequences
+        assert not self.mult_factor
+        code.put(f"PyObject *{target}[{len(self.args)}] = {{")
+        code.put(f"{', '.join(arg.result() for arg in self.args)}")
+        code.putln("};")
```
scoder, line 8771: There's no need for an f-string in the middle line. You could either unpack the expression or use implicit string concatenation.
```suggestion
        code.putln(
            f"PyObject *{target}[{len(self.args)}] = {{"
            f"{', '.join(arg.result() for arg in self.args)}"
            f"}};"
        )
```

Cython/Compiler/ExprNodes.py
```diff
@@ -7144,16 +7163,20 @@ def generate_evaluation_code(self, code):
         # To avoid passing an out-of-bounds argument pointer in the no-args case,
         # we need at least two entries, so we pad with NULL and point to that.
         # See https://github.com/cython/cython/issues/5668
-        args_list = ', '.join(arg.py_result() for arg in args) if args else "NULL"
-        extra_keyword_args = f" + ((CYTHON_VECTORCALL) ? {len(kwargs_key_value_pairs)} : 0)" if kwargs_key_value_pairs else ""
+        args_and_kwargs = []
+        if args:
+            args_and_kwargs.extend(args)
+        if kwvalues:
+            args_and_kwargs.extend(kwvalues)
```
```suggestion
        args_and_kwargs = args + (kwvalues or [])
```

Cython/Compiler/ExprNodes.py
```diff
@@ -7061,38 +7074,42 @@ def generate_runtime_method_unpacking_code(self, code, self_arg, space_for_selfa
...
-                f"{keyvalue.key.py_result()}, "
-                f"{keyvalue.value.py_result()}, "
-                f"{kwnames_temp}, "
-                f"{Naming.callargs_cname}+{len(args) + 1}, "
-                f"{n:d}"
-                ")"
-            )
+        code.putln("#if CYTHON_VECTORCALL")
+        code.putln(f"{kwnames_temp} = {kwnames.result()};")
+        code.putln(f"{code.error_goto_if_null(kwnames_temp, self.pos)}")
```
```suggestion
        code.putln(code.error_goto_if_null(kwnames_temp, self.pos))
```

Cython/Compiler/ExprNodes.py
scoder: Alternative:

```suggestion
        code.put(f"PyObject *{target}[{len(self.args)}] = {{")
        code.put(', '.join(arg.result() for arg in self.args))
        code.putln("};")
```

Cython/Compiler/ExprNodes.py
```diff
@@ -7061,38 +7074,42 @@ def generate_runtime_method_unpacking_code(self, code, self_arg, space_for_selfa
...
+        for arg_index in arg_indices_to_check:
+            code.put_error_if_neg(kwnames.pos, f"__Pyx_CheckVectorcallKwarg({kwnames_temp}, {arg_index})")
+        code.putln("#else")
+        code.putln("{")
+        kwnames.generate_sequence_as_array_code(code, Naming.quick_temp_cname)
+        for arg_index in arg_indices_to_check:
+            code.put_error_if_neg(kwnames.pos, f"__Pyx_CheckVectorcallKwarg({Naming.quick_temp_cname}, {arg_index})")
+        code.putln(f"{kwnames_temp} = __Pyx_MakeKwargDict({Naming.quick_temp_cname}, "
+                   f"{Naming.callargs_cname}+{len(args)+1}, {len(kwvalues)});")
+        code.putln(f"{code.error_goto_if_null(kwnames_temp, self.pos)}")
```
```suggestion
        code.putln(code.error_goto_if_null(kwnames_temp, self.pos))
```

scoder (commented):
Seems worth merging for the alpha release. It's an optimisation that cleans up code, can't have it much better than that.

PR #7715, opened 2026-05-25
Fix direct call to instance methods on cross module compilation

### LINKED ISSUE #7692: [BUG] Direct call to non-final method do not work when cross modules
### Describe the bug

If you make a direct call to a non-final member it will fail due the cython do not expose the methods headers required for it, it does not even produces an error, it just generates the code pretending everything ok and fails when compiling c or c++.

It could work if cython do not perform optimizations but tbh it would make cython useless, if a simple call like this can not be correctly coded with a simple direct c-call. 

```cython
PYTHON -c "import b; import a"

######## setup.py ########

from Cython.Build.Dependencies import cythonize
import Cython.Compiler.Options
Cython.Compiler.Options.cimport_from_pyx = True

setup(
  ext_modules = cythonize("*.pyx"),
)

######## a.pyx ########
from b import B
from b cimport B
from cython import ccall, cclass

@cclass
class A(B):
    @ccall
    def s(self):
        B.mtd(self)

######## b.pyx ########

from cython import cclass, ccall

@cclass
class B:
    @ccall
    def mtd(self):
        pass

```

### Expected behaviour

Cython should generate headers for all the `__pyx_f_...` symbols so the modules can be linked without issues.
Currently it is only happening for final methods.
Or at last resort provide a flag to produce module isolated (which default this to slow python method call).

### OS

_No response_

### Python version

_No response_

### Cython version

_No response_

### Additional context

The error looks like this:

```
-- stdout:Compiling a.pyx because it changed.
Compiling b.pyx because it changed.
[1/2] Cythonizing a.pyx
[2/2] Cythonizing b.pyx

-- stderr:a.c:2773:15: error: call to undeclared function '__pyx_f_1b_1B_mtd'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
 2773 |   __pyx_t_1 = __pyx_f_1b_1B_mtd(((struct __pyx_obj_1b_B *)__pyx_v_self), 1); if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 9, __pyx_L1_error)
      |               ^
a.c:2773:15: note: did you mean '__pyx_f_1a_1A_s'?
a.c:2694:18: note: '__pyx_f_1a_1A_s' declared here
 2694 | static PyObject *__pyx_f_1a_1A_s(struct __pyx_obj_1a_A *__pyx_v_self, int __pyx_skip_dispatch) {
      |                  ^
a.c:2773:13: error: incompatible integer to pointer conversion assigning to 'PyObject *' (aka 'struct _object *') from 'int' [-Wint-conversion]
 2773 |   __pyx_t_1 = __pyx_f_1b_1B_mtd(((struct __pyx_obj_1b_B *)__pyx_v_self), 1); if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 9, __pyx_L1_error)
      |             ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2 errors generated.

```

jairov4:
Fixes #7692

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -8039,7 +8039,8 @@ class AttributeNode(ExprNode):
     def _create_unbound_cmethod_entry(self, type, entry, env):
         # Create a temporary entry describing the unbound C method in `entry`
         # as an ordinary function.
-        if entry.func_cname and entry.type.op_arg_struct is None:
+        if (entry.func_cname and entry.type.op_arg_struct is None
+                and type.module_name == env.qualified_name):
             cname = entry.func_cname
             if entry.type.is_static_method or (
                     env.parent_scope and env.parent_scope.is_cpp_class_scope):
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -8039,8 +8039,8 @@ class AttributeNode(ExprNode):
     def _create_unbound_cmethod_entry(self, type, entry, env):
         # Create a temporary entry describing the unbound C method in `entry`
         # as an ordinary function.
-        if (entry.func_cname and entry.type.op_arg_struct is None
-                and type.module_name == env.qualified_name):
+        if (entry.func_cname and entry.type.op_arg_struct is None and
+                (not type.is_extension_type or type.module_name == env.qualified_name)):
             cname = entry.func_cname
             if entry.type.is_static_method or (
                     env.parent_scope and env.parent_scope.is_cpp_class_scope):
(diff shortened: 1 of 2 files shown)


da-woods:
I think the basic cause of this issue and https://github.com/cython/cython/pull/7714 is the same: essentially that `func_cname` is set on the entry for the cfunction when it wouldn't be on an entry defined in a pxd file.

Solving that would probably be a more satisfactory solution that the two separate solutions here (and would probably also fix some other related issues).

It's essentially set due to the `defining=` in the call below

https://github.com/cython/cython/blob/c74351d8686cc8b1a6147bb1990ac745d0bf54de/Cython/Compiler/Nodes.py#L2843-L2847

(while in the pxd import the function is defined via `CVarDefNode` instead, so it's a bit different....)

da-woods:
I suspect the place to fix it might be in

https://github.com/cython/cython/blob/c74351d8686cc8b1a6147bb1990ac745d0bf54de/Cython/Compiler/Pipeline.py#L292

but that isn't based on _too_ much thought

PR #7716, opened 2026-05-25
Optimize `super().mtd(..)` method calls

jairov4:
Requires #7715 be merged


scoder:
I don't think this is necessarily correct if multiple inheritance is involved.

PR #7717, opened 2026-05-25
Fix std::vector::reverse_iterator::operator- declarations

vmarkovtsev:
This must be a typo: C++ doesn't allow subtracting a regular iterator from a reverse iterator.


da-woods:
Thanks - that looks right to me so I'll merge it when the tests pass

da-woods:
@scoder I think the required "sanitizers" jobs are causing problems - any PR that doesn't trigger them to run ends up unmergeable. https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks:

> If a workflow is skipped due to path filtering, branch filtering or a commit message, then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.

Unfortunately it doesn't really give any suggestions on how to avoid that.

vmarkovtsev:
Appreciate all the help, @da-woods :pray: Thanks.

da-woods:
I haven't _quite_ won my fight with GitHub actions yet, but nearly there...

da-woods:
3.2.x commit c5ee410b8.

Thanks

PR #7718, opened 2026-05-25
Implement boxed types allowing cythonize more codebases without changes

jairov4:
This change introduces Boxed Types.

Boxed types make Cython’s annotation system more practical for real-world APIs by allowing value types to be used where None is a valid state. This extends Cython’s type representation beyond plain native storage and Python objects, so declarations like `int | None`, `float | None`, `bint | None`, ctuples, and enum-like value types can be expressed naturally and used consistently in both function signatures and variable declarations.

This is important because allows to Cythonize **larger code bases** without requiring code changes and potentialized when used with **auto_cpdef**. Even writing interfaces with cython becomes easier with this feature even in py or pyx modes.  
Introducing boxed types keeps the native type information intact while giving Cython a clean, generalized way to support nullable value semantics.

### DIFF
--- a/Cython/Compiler/PyrexTypes.py
+++ b/Cython/Compiler/PyrexTypes.py
@@ -403,6 +403,10 @@ class PyrexType(BaseType):
         """Returns True if type can be used with typing.Optional[]."""
         return False
 
+    def boxed_type(self):
+        """Return the Python type used when this value type is boxed, or None."""
+        return None
+
     def struct_nesting_depth(self):
         # Returns the number levels of nested structs. This is
         # used for constructing a stack for walking the run-time
@@ -470,6 +474,16 @@ class PyrexType(BaseType):
 
     def needs_explicit_destruction(self, scope):
         return False
+
+
+def optional_boxed_type(tp):
+    tp = tp.resolve()
+    if tp.can_be_optional():
+        return tp
+    boxed = tp.boxed_type()
+    if boxed is not None:
+        return boxed.resolve()
+    return None
 
 
 def public_decl(base_code, dll_linkage):
@@ -703,6 +717,9 @@ class CTypedefType(BaseType):
 
     def can_coerce_from_pyobject(self, env):
         return self.typedef_base_type.can_coerce_from_pyobject(env)
+
+    def boxed_type(self):
+        return self.typedef_base_type.boxed_type()
 
     def can_coerce_to_pystring(self, env, format_spec=None):
         return self.typedef_base_type.can_coerce_to_pystring(env, format_spec)
@@ -1356,6 +1373,9 @@ class PyObjectType(PyrexType):
         """Returns True if type can be used with typing.Optional[]."""
         return True
 
+    def boxed_type(self):
+        return self
+
     def default_coerced_ctype(self):
         """The default C type that this Python type coerces to, or None."""
         return None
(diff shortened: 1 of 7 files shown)


da-woods:
I don't think this makes sense.

When people use `cython.int` they're asking for an exact C int. It's completely reasonable for Cython to error if they use it in an annotation like optional where we can no longer give an exact C int.

Note that Cython already understands `int | None`, `float | None`, `bool | None` and interprets `int/float/bool` as a Python type.

There isn't really a "compatibility with more code-bases" argument here. No-one is trying to use `Optional[cython.int]` as an annotation in a code-base that isn't aimed directly at Cython.

scoder:
I haven't looked into the details and I cannot say that I fully grasp the intention here. Some more motivating examples would help.

But note that there is already the concept of an `equivalent_type` that defines the Python type that C types convert to automatically.

jairov4:
I have two different use cases,  first one is I am using this and other local changes to compile big code bases from pure python with auto_cpdef=True and cimport_from_pyx=True with impressive performance gains. Some codebases already use optional declarations so one simple trick that unleash big wins is just override int and other builtins with native types like `from cython import int`, this trick is additive and simple to automate out of cython. As you can imagine, It drastically remove a lot of python interactions demonstrating how awsome cython is. But the pain ends being declarations of nullable types. which require manual editing of the files.
The second one is directly to write cython optimized code that allows to use nullable types, as of today we need to use other mechanisms like using a tuple[bint, T] or passing two values individually, ..., but all those escape from the simple approach that python syntax already provides. I noticed cython already supports that for python mode but not for optimized cpdef methods.
Note this change is not yet solving/optimizing deep (pe. still no custom/optimized underlying types are used yet to represent some nullable values), just allow the use of nullable to unleash those cases.

da-woods:
To me this feels like a bit of a hack to make your specific code work, rather than something that's generally the right thing to do. 

The real issue comes back to Cython's handling of `int`: We've gone for "py int" for maximum compatibility. But practically most ints are small and a lot of people would be happy with a C int. I tried to add a separate annotation mode to enable this at one point, but I think we rejected it because more configuration options just end up with too much to test. (The other option would be to use tagged integers to support both, like MyPy does. But I think that's more trouble than it's worth)

Cython/Compiler/PyrexTypes.py
```diff
@@ -476,6 +476,16 @@ def needs_explicit_destruction(self, scope):
         return False
 
 
+def optional_boxed_type(tp):
```
scoder, line 479: This should be a method, not a function.

Cython/Compiler/PyrexTypes.py
```diff
@@ -2230,6 +2235,10 @@ def can_coerce_to_pyobject(self, env):
     def can_coerce_from_pyobject(self, env):
         return True
 
+    def boxed_type(self):
+        from . import Builtin
+        return Builtin.int_type
```
scoder, line 2240: I think we can just provide a generic implementation for this that returns `self.equivalent_type`.

scoder (commented):
The implementation needs some cleanup, but I wouldn't generally rule out allowing `typing.Optional[]` on C types (IIUC, that's the main intention here?). It's like we allow `list[cython.int]`, there are no C `int`s in that Python list but the user provided declaration allows us to see the exact type that we should use when processing the items. Similarly, when processing an `Optional[cython.long]` or `cython.long | None`, we'd know that the value is a C `long` if it's not `None`. That'd be helpful in some cases.

PR #7719, opened 2026-05-26
Fix broken test in pep526_variable_annotations

matusvalo:


da-woods:
This will be another thing that ends up unmergeable because it doesn't trigger the sanitizer tests unfortunately

da-woods:
I didn't quite get the "merge when the sanitizer runs don't trigger" fixed unfortunately. I think https://github.com/cython/cython/pull/7726 should do it in the end though

Issue #7721, opened 2026-05-26
super() in Cython methods resolves __class__ from descriptor access class, not defining class (breaks Python decorator pattern)


ElliotElison:
## Summary

In pure Python, `super()` inside a method uses a `__class__` **cell variable** that is fixed at class definition time to the class in which the method is defined. In Cython 3.x, `cython_function_or_method` has no such cell variable — instead, `super()` resolves `__class__` dynamically based on the class through which the method is accessed (the `type` argument to `__get__`). This diverges from Python semantics and breaks a common Python decorator pattern.

## Python Version and Cython Version

- Python: 3.11
- Cython: 3.2.5

## Minimal Reproducer

```python
# parent.py (pure Python)
class Parent:
    def __init__(self):
        self.x = 1

# child.pyx (Cython)
from parent import Parent

class Child(Parent):
    def __init__(self):
        super().__init__()   # uses super() -- affected
        self.y = 2
```

```python
import child

# Common Python decorator pattern: create a subclass wrapper
class Wrapper(child.Child):
    pass

# Instantiate through the wrapper
obj = Wrapper()
# Expected (pure Python behaviour): super().__init__() inside Child.__init__
#   resolves to Parent.__init__, because __class__ cell is fixed to Child.
# Actual (Cython 3.x): super().__init__() resolves using Wrapper as __class__,
#   so super() looks up Wrapper in the MRO and finds Child as "next",
#   causing infinite recursion or wrong dispatch.
```

## Root Cause

Inspecting a compiled Cython class confirms the divergence:

```python
import dis

# Pure Python method that uses super():
class PyParent:
    def __init__(self): pass
class PyChild(PyParent):
    def __init__(self): super().__init__()

raw_py = PyChild.__dict__['__init__']
print(raw_py.__code__.co_freevars)   # ('__class__',)
print([c.cell_contents for c in raw_py.__closure__])  # [<class '__main__.PyChild'>]
# __class__ cell is FIXED to PyChild at definition time.

# Cython method:
raw_cy = CythonChild.__dict__['__init__']
print(raw_cy.func_closure)           # None
print(raw_cy.func_code.co_freevars)  # ()
# No __class__ cell. The func_code is a C stub (all CACHE opcodes).
# __class__ is resolved dynamically in C from the __get__ type argument.
```

[...]

PR #7722, opened 2026-05-27
Fix class cell using the decorator result

da-woods:
Fixes what I believe #7721 was hinting at before an LLM (re?)wrote it to have a reproducer that doesn't reproduce and a bunch of incorrect speculation about the cause.

The saved class cell was applied with the result of the decorator rather than the result of the class before the decorator.


Cython/Compiler/Symtab.py
```diff
@@ -2356,8 +2356,6 @@ def lookup(self, name):
 
 class PyClassScope(ClassScope):
     #  Namespace of a Python class.
-    #
-    #  class_obj_cname     string   C variable holding class object
```
da-woods, line 2362: This was just unused (only assigned to, but never read)

da-woods:
I'm setting this to merge once the tests have finished....

PR #7723, opened 2026-05-27
Shared module: Use StringSourceDescriptor instead of temporary directory.

matusvalo:
This PR uses `StringSourceDescriptor` instead of empty pyx file in temp dir. This simplifies the logic and also ensures that __pyx_f is constant across builds.

Alternative to #7634


da-woods (approved):
I think we should go for this as the solution to the non-reproducible path.

Cython/Compiler/Scanning.py
```diff
@@ -278,6 +282,10 @@ def get_relative_path(self):
 
     get_error_description = get_description
 
+    def get_file_object(self, encoding=None, error_handling=None):
+        from io import StringIO
+        return StringIO('\n'.join(self.codelines))
```
scoder: I think we intentionally left out the file methods at the time to prevent accidental mixups between the two. Not sure how large the risk of that is. If we allow both to be used interchangeably, then that would also remove that risk.

Cython/Compiler/Main.py
```diff
@@ -375,7 +375,7 @@ def find_submodule(self, name, as_package=False):
         return scope
 
     def parse(self, source_desc, scope, pxd, full_module_name):
-        if not isinstance(source_desc, FileSourceDescriptor):
+        if not isinstance(source_desc, (FileSourceDescriptor, StringSourceDescriptor)):
             raise RuntimeError("Only file sources for code supported")
```
scoder, line 379: This exception really makes no sense if we now allow both of the types that it's trying to distinguish.

scoder (commented):
I'm not 100% convinced that this is a good idea. I think it might be fine to allow string sources and files to be used interchangeably, but if it is, then we should go all the way and clean this up further.

da-woods:
> I'm not 100% convinced that this is a good idea

Is your issue with this mainly about opening the interface up to arbitrary in-memory sources?

Supporting it for the shared library only seems ok to me (in that it's one thing that's tested and with fixed comments).

I could see the argument for still disallowing it on the "public" interface

scoder:
We currently consider string sources an internal feature (we bluntly `.rstrip()` the code lines, for example), not a public feature. Unless we have a need to expose it, I think we should keep it that way and not allow `StringSourceDescriptor` in the publicly visible interfaces.

matusvalo:
> We currently consider string sources an internal feature (we bluntly .rstrip() the code lines, for example), not a public feature.

The intent of this PR is not to have it public. Only as internal instrument for shared utility

da-woods:
It looks like it'd be simple enough to have a public `parse` method that keeps the current validation and a `_parse` or `__parse` method (depending on how private you want to make it) that skips the validation.

Obviously like everything else in Python it's fairly simple to work around this kind of private. But it might make it clear that the public interface only accepts files.

scoder:
> It looks like it'd be simple enough to have a public `parse` method that keeps the current validation and a `_parse` […] that skips the validation.

Not entirely, because the shared module creation uses the public interfaces as well, and the call is somewhat indirect.

scoder:
The shared module is a very special case. It's empty and only aggregates utility code, and I don't see why that would need to change. Can't we have a dedicated file source descriptor class for that?

matusvalo:
> The shared module is a very special case. It's empty and only aggregates utility code, and I don't see why that would need to change. Can't we have a dedicated file source descriptor class for that?

OK I implemented it and seems that the code is even more simplified. I hope now should be fine.

Cython/Build/SharedModule.py
```diff
@@ -71,23 +68,17 @@ def generate_shared_module(options):
...
     module_name = os.path.splitext(os.path.basename(dest_c_file))[0]
 
     context = Main.Context.from_options(options)
     scope = Symtab.ModuleScope('MemoryView', parent_module = None, context = context, is_package=False)
 
-    with tempfile.TemporaryDirectory() as tmpdirname:
-        pyx_file = os.path.join(tmpdirname, f'{module_name}.pyx')
-        c_file = os.path.join(tmpdirname, f'{module_name}.c')
-        with open(pyx_file, 'w'):
-            pass
```
da-woods, line 83: I hadn't actually realized that the file was completely empty and only existed so there was a file...

The new version seems pretty straightforward to me, especially knowing this.

Cython/Compiler/Scanning.py
```diff
@@ -294,6 +294,18 @@ def __repr__(self):
         return "<StringSourceDescriptor:%s>" % self.name
 
 
+class SharedUtilitySourceDescriptor(FileSourceDescriptor):
+    """
+    A specialized source descriptor for shared utility code only. Not part of public API.
+    """
+
+    def __init__(self, name):
+        super().__init__(name)
```
scoder: This method isn't needed, unless you want to explicitly remove the ability to pass a `path_description` argument.

I looked through the usages to see what that `path_description` does and it seems that no-one really knew for quite a while. :) The usages pass some kind of relative path into it which the class then converts into a relative path again and stores away for display purposes. This behaviour probably emerged from the various attempts to use relative paths in C source files. Either way, it's worth cleaning up at some point (not in this PR).

scoder (approved):

Cython/Compiler/Scanning.py
matusvalo: fixed.

PR #7724, opened 2026-05-28
Avoid using paths filter for "required" github actions

da-woods:
The reason being that if jobs are skipped because of it then the PR becomes unmergeable. In contrast if jobs are skipped at job level then it's fine.

Instead move the file filters to job level and skip them there.

I've only done this for "required" jobs and no worried about anything else (e.g. coverage)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
       with:
         persist-credentials: false
       # Only needed to compare the "base" on push; not needed for PR
-      if: ${{ github.event.name == 'push' }}
+      if: ${{ github.event_name == 'push' }}
     - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d  # v4.0.1
       id: filter
       with:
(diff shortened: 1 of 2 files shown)


da-woods:
Currently this is looking like it's hasn't triggered any actions at all. But I'd normally expect it to tell me something about where I messed up....

da-woods:
The sanitizers workflow seems to work now... The ci one remains mysteriously and silently broken

da-woods:
The behaviour of the filters isn't quite like github (i.e. is `any` or `all` but not sequential. So I think would need rewriting a little

scoder:
The way Github Actions works (or doesn't work) has long started to annoy me. I think the only reason why people use it is that "it's there". :-/

Thanks for coming up with this work-around for what the infrastructure just doesn't handle itself.

da-woods:
> I think the only reason why people use it is that "it's there". :-/

Well, and "it's free", which is pretty hard to beat

da-woods:
It turns out that this isn't quite enough to solve my issue. 

It does the `if` before expanding the matrix. So I think I need to add a "gather results" job at the end, and then that needs to be the thing that's required to pass. I'll do that this evening.

PR #7726, opened 2026-05-29
Improve gathering of CI results

da-woods:
1. Skipping a job counts as success from the point of view of mergeability. This means `ci-success` isn't actually doing its job because it gets skipped if any bits of the CI fail.  This version will always run `ci-success` but with a result that tells you if all previous jobs have passed.
2. The `if` clause for a job is done before the matrix is expanded. This means that if the sanitizers job is skipped, then you still can't merge (because it doesn't work out which matrix jobs have been skipped). Add a `sanitizers-success` job to use as the new required job.

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -73,3 +73,12 @@ jobs:
       name: pydebug
     needs: [file-changes]
     if: ${{ case(github.event_name == 'push', needs.file-changes.outputs.push-files == 'true', github.event_name == 'pull_request', needs.file-changes.outputs.pr-files == 'true', 'true') }}
+
+  sanitizers-success:
+    needs: [sanitizers, pydebug]
+    if: always()
+    env:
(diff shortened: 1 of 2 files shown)


.github/workflows/sanitizers.yml
```diff
@@ -73,3 +73,12 @@ jobs:
       name: pydebug
     needs: [file-changes]
     if: ${{ case(github.event_name == 'push', needs.file-changes.outputs.push-files == 'true', github.event_name == 'pull_request', needs.file-changes.outputs.pr-files == 'true', 'true') }}
+
+  sanitizers-success:
```
da-woods, line 77: The idea is that this is the "required" job for the sanitizers runs rather than the individual jobs

scoder:
Wouldn't success() also work outside the if, to determine the overall run result without checking the needs separately?

da-woods:
> Wouldn't success() also work outside the if, to determine the overall run result without checking the needs separately?

I tried it and it said "unknown function". So apparently not, but I don't understand why.

PR #7727, opened 2026-05-30
Apply some simpler `distutils` migration advice

jakirkham:
Related to issue: https://github.com/cython/cython/issues/5751

Cleanup some of the simpler `distutils` usage cases with [their recommended replacements]( https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html#prefer-setuptools ).

There is still much more work to do here, but at least this starts moving in the right direction incrementally.

### DIFF
--- a/Cython/Build/IpythonMagic.py
+++ b/Cython/Build/IpythonMagic.py
@@ -49,7 +49,7 @@ import re
 import sys
 import time
 import copy
-import distutils.log
+import logging
 import textwrap
 
 IO_ENCODING = sys.getfilesystemencoding()
@@ -434,11 +434,11 @@ class CythonMagics(Magics):
         old_threshold = None
         try:
             if not quiet:
-                old_threshold = distutils.log.set_threshold(distutils.log.DEBUG)
+                old_threshold = logging.set_threshold(logging.DEBUG)
             build_extension.run()
         finally:
             if not quiet and old_threshold is not None:
-                distutils.log.set_threshold(old_threshold)
+                logging.set_threshold(old_threshold)
 
     def _add_pgo_flags(self, build_extension, step_name, temp_dir):
         compiler_type = build_extension.compiler.compiler_type
--- a/Cython/Debugger/Tests/TestLibCython.py
+++ b/Cython/Debugger/Tests/TestLibCython.py
@@ -8,7 +8,7 @@ import unittest
 import tempfile
 import subprocess
 #import distutils.core
-#from distutils import sysconfig
+#import sysconfig
 from distutils import ccompiler
 
 import runtests
--- a/Cython/Distutils/old_build_ext.py
+++ b/Cython/Distutils/old_build_ext.py
@@ -8,13 +8,14 @@ Note that this module is deprecated.  Use cythonize() instead.
 
 __revision__ = "$Id:$"
 
-import sys
+import logging
 import os
+import sys
+import sysconfig
+
 from distutils.errors import DistutilsPlatformError
 from distutils.dep_util import newer, newer_group
-from distutils import log
 from distutils.command import build_ext as _build_ext
-from distutils import sysconfig
 
 
 # FIXME: the below does not work as intended since importing 'Cython.Distutils' already
--- a/Demos/benchmarks/run_benchmarks.py
+++ b/Demos/benchmarks/run_benchmarks.py
@@ -33,7 +33,7 @@ if hasattr(sys, '_is_gil_enabled') and not sys._is_gil_enabled():
 
 
 try:
-    from distutils import sysconfig
+    import sysconfig
     DISTUTILS_CFLAGS = sysconfig.get_config_var('CFLAGS')
 except ImportError:
     DISTUTILS_CFLAGS = ''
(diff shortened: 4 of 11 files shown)


Cython/Build/IpythonMagic.py
```diff
@@ -49,7 +49,7 @@
 import sys
 import time
 import copy
-import distutils.log
+import logging
```
jakirkham, line 52: This alone fixes an issue with `%load_ext Cython`

Cython/Debugger/Tests/TestLibCython.py
```diff
@@ -8,7 +8,7 @@
 import tempfile
 import subprocess
 #import distutils.core
-#from distutils import sysconfig
+#import sysconfig
```
jakirkham, line 11: This was commented. So perhaps it could even be removed. However without knowing more about why it is here, just went with applying the recommended migration advice. Happy to defer to others judgment on this

Demos/benchmarks/run_benchmarks.py
```diff
@@ -33,7 +33,7 @@
 
 
 try:
-    from distutils import sysconfig
+    import sysconfig
     DISTUTILS_CFLAGS = sysconfig.get_config_var('CFLAGS')
 except ImportError:
     DISTUTILS_CFLAGS = ''
```
jakirkham, line 39: Arguably this environment variable could be renamed. Held off doing that for now, but can add if that is preferred

Edit: Should add `sysconfig` is a builtin. So the `ImportError` shouldn't be needed any more. Can clean this up as well

Tools/site_scons/site_tools/pyext.py
```diff
@@ -230,7 +230,7 @@ def exists(env):
     try:
         # This is not quite right: if someone defines all variables by himself,
         # it would work without distutils
-        from distutils import sysconfig
+        import sysconfig
         return True
     except ImportError:
         return False
```
jakirkham, line 236: I don't totally follow what this function is doing. The comment indicates there are issues with it.

Have stuck to making the simple `sysconfig` change. However `ImportError` shouldn't occur in this case as `sysconfig` is a builtin. So a little unsure on what this should be ok.

Happy to make a recommended change.

pyximport/test/test_reload.py
jakirkham, line 14: It looks like `sysconfig.set_python_build()` is not defined

```python
In [1]: import sysconfig
   ...: 
   ...: sysconfig.set_python_build()
AttributeError                            Traceback (most recent call last)
Cell In[1], line 3
      1 import sysconfig
----> 3 sysconfig.set_python_build()

AttributeError: module 'sysconfig' has no attribute 'set_python_build'
```

Will roll this back

```suggestion
if 1:
    from distutils import sysconfig
    try:
        sysconfig.set_python_build()
    except AttributeError:
        pass
    import pyxbuild
    print(pyxbuild.distutils.sysconfig == sysconfig)
```

jakirkham: Done in later changes in this PR

Cython/Build/IpythonMagic.py
da-woods: I think this needs to be

```suggestion
import logging
import distutils
```

based on the failure of coverage tests

Cython/Build/IpythonMagic.py
```diff
@@ -434,11 +434,11 @@ def _build_extension(self, extension, lib_dir, temp_dir=None, pgo_step_name=None
         old_threshold = None
         try:
             if not quiet:
-                old_threshold = distutils.log.set_threshold(distutils.log.DEBUG)
+                old_threshold = logging.set_threshold(logging.DEBUG)
```
da-woods, line 437: I think this is `setlevel` rather than `set_threshold` in the logging module

Cython/Distutils/old_build_ext.py
```diff
@@ -8,13 +8,14 @@
...
 
-import sys
+import logging
 import os
+import sys
+import sysconfig
+
 from distutils.errors import DistutilsPlatformError
 from distutils.dep_util import newer, newer_group
-from distutils import log
```
da-woods, line 15: I'm not quite sure of the status of `old_build_ext` and whether it's used any more.

I can see a couple of uses of `log` in the file though

jakirkham:
Thanks David! 🙏

Had hoped this would be simple `import` replacement. Though in looking into it more, think there is some subtlety to how the Cython code is using `distutils` here.

Am thinking a better approach would be to do some form of...

```python
try:
    import distutils
except ImportError:
    import setuptools
```

Though of course tailored to the needs of the particular usage

da-woods:
I think the main place we need to be careful is in the `Cython/Distutils` where we add our own customizations onto `distutils/setuptools`.  That's essentially a user interface and should probably keep working for the remaining users of `distutils`.

I'm completely comfortable with requiring `setuptools` for anything internal. Like Cython's own build process, run_benchmarks.py, our own tests (unless they're specifically testing something distutils only).

I don't really know what site_scons is so I couldn't comment on that. I'm also not sure about IPythonMagic and what it's reasonable to require there.

PR #7728, opened 2026-05-30
Avoid separate docstring in pure and compiled version of tests

matusvalo:
Based on comment here: https://github.com/cython/cython/pull/7704#discussion_r3323388752

### DIFF
--- a/tests/run/pep526_variable_annotations.py
+++ b/tests/run/pep526_variable_annotations.py
@@ -195,7 +195,7 @@ def test_subscripted_types():
     print(cython.typeof(b3) + (" object" if not cython.compiled else ""))
     print(cython.typeof(c) + ("[Python object] object" if not cython.compiled else ""))
 
-if sys.version_info >= (3, 11):
+if sys.version_info >= (3, 11) or cython.compiled:
     # This part of the test is failing in Python 3.9 and 3.10 with the following exception in Shadow.py:
     # isinstance() argument 2 cannot be a parameterized generic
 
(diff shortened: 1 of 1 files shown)


tests/run/pep526_variable_annotations.py
```diff
@@ -345,78 +338,51 @@ def test_assignment_list_with_subscript():
...
+    a: dict[str, cython.int] = {'a': 5}
+    b: dict = a
+    c: dict[str, cython.float] = b
+    print(cython.typeof(a['a']) if cython.compiled else 'int')
+    print(cython.typeof(b['a']) if cython.compiled else 'Python object')
+    print(cython.typeof(c['a']) if cython.compiled else 'float')
+    print(a['a'], b['a'], c['a'] if cython.compiled else float(c['a']))
 
-if sys.version_info >= (3, 15) or cython.compiled:
+if sys.version_info >= (3, 15):
```
scoder: Why was it necessary to disable the two compiled tests in older Python versions? Requiring Py3.15 skips a lot of versions that we want to support.

scoder (commented):
Thanks, I think this is better. While it might look more complex in some cases, it makes it very locally clear what the specific difference is.

tests/run/pep526_variable_annotations.py
matusvalo: I will bring it back.

matusvalo: OK fixed.

PR #7730, opened 2026-06-01
Pattern matching for mappings

da-woods:
Part of https://github.com/cython/cython/issues/4029

The most notable design decision here is that I did the "take from mapping" as a utility-code function that loops over all expected keys rather than generate code for each lookup (like what happens with sequences).  That was mostly to make it easier to write fast special cases for `dict`.


Cython/Utility/MatchCase.c
```diff
@@ -403,3 +403,336 @@ static PyObject *__Pyx_MatchCase_UnknownTypeSliceToList(PyObject *x, Py_ssize_t
...
+    raise_error:
+    #if PY_MAJOR_VERSION > 2
+    PyErr_Format(PyExc_ValueError,
+                 "mapping pattern checks duplicate key (%R)", key);
+    #else
+    // DW really can't be bothered working around features that don't exist in
+    // Python 2, so just provide less information!
+    PyErr_SetString(PyExc_ValueError,
+                    "mapping pattern checks duplicate key");
+    #endif
```
da-woods:
```suggestion
    PyErr_Format(PyExc_ValueError,
                 "mapping pattern checks duplicate key (%R)", key);
```

Cython/Compiler/Parsing.py
```diff
@@ -4628,12 +4628,16 @@ def p_mapping_pattern(s: PyrexScanner):
 
     double_star_capture_target = None
     items_patterns = []
+    double_star_set_twice = None
+    pattern_after_double_star = None
```
```suggestion
    double_star_set_twice_pos = None
    pattern_after_double_star_pos = None
```
etc.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,329 @@ def get_main_pattern_targets(self):
...
         return targets
 
+    def validate_keys(self):
+        # called after constant folding
+        seen_keys = set()
+        for k in self.keys:
+            if k.has_constant_result():
+                value = k.constant_result
+                if k.is_string_literal:
+                    value = repr(value)
```
scoder: My intuition tells me that we should always use the `repr()` here, not just for strings, because otherwise, we risk mixing cases where the values compare equal but the type disagrees. But I don't know how reasonable that case is in pattern matching. Python (3.12.3) does raise an error on the obvious case that I tried. The PEP isn't clear about this, but I would now guess that these cases are supposed to work pretty much like you implemented them here.
```python
>>> {1.0: 2}[1]
2
>>> match {1:2}:
...     case {1.0: 2, 1: 2}: print(1)
... 
  File "<stdin>", line 2
SyntaxError: mapping pattern checks duplicate key (1)
```
Might just be worth a comment that this has been thought of and is considered intentional.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,329 @@ def get_main_pattern_targets(self):
...
+                if value in seen_keys:
+                    error(k.pos, "mapping pattern checks duplicate key (%s)" % value)
+                seen_keys.add(value)
+            else:
+                self.needs_runtime_keycheck = True
+
+        if self.keys:
```
scoder: Since the sorting depends on literals, I think this constraint might be reasonable:
```suggestion
        if seen_keys and self.keys:
```
I'd also rename `seen_keys` to `literal_keys` then to make it clearer.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,329 @@ def get_main_pattern_targets(self):
...
+            return self.DictCheckResult.ExactDict
+        if type.is_pyfrozendict_type:
+            return self.DictCheckResult.ExactFrozenDict
+        if type.is_builtin_type:
+            # All other builtin types that Cython knows of are not mappings
+            # (DictProxyType is, but Cython doesn't know about that).
+            return self.DictCheckResult.NotAMapping
+        if not type.is_pyobject:
+            # for now any non-pyobject type isn't treated as a mapping
+            return self.DictCheckResult.NotAMapping
```
scoder, line 1074: What about C++ mappings?

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,329 @@ def get_main_pattern_targets(self):
...
+            utility_code=utility_code,
+            args=[subject_node, sequence_mapping_temp],
+        )
+
+        def type_check(type):
+            res = self.is_dict_type_check(type)
+            if res == self.DictCheckResult.ExactDict or res == self.DictCheckResult.ExactFrozenDict:
+                return True
+            if res == self.DictCheckResult.NotAMapping:
+                return False
```
scoder, line 1109: I think enum comparisons should generally be done with `is`.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,329 @@ def get_main_pattern_targets(self):
...
+            self.pos,
+            expr=ExprNodes.PythonCapiCallNode(
+                self.pos,
+                "__Pyx_MatchCase_CheckDuplicateKeys",
+                self.Pyx_mapping_check_duplicates_type,
+                utility_code=utility_code,
+                args=[
+                    MappingComparisonNode.make_keys_node(self.pos),
+                    ExprNodes.IntNode(self.pos, value=str(n_fixed_keys)),
+                    ExprNodes.IntNode(self.pos, value=str(len(self.keys)))
```
```suggestion
                    ExprNodes.IntNode.for_size(self.pos, n_fixed_keys),
                    ExprNodes.IntNode.for_size(self.pos, len(self.keys)),
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,329 @@ def get_main_pattern_targets(self):
...
+        )
+        func = ExprNodes.PythonCapiCallNode(
+            self.double_star_capture_target.pos,
+            "__Pyx_MatchCase_DoubleStarCapture" + tag,
+            self.Pyx_mapping_doublestar_type,
+            utility_code=utility_code,
+            args=[
+                subject_node,
+                MappingComparisonNode.make_keys_node(self.pos),
+                ExprNodes.IntNode(self.pos, value=str(len(self.keys)))
```
```suggestion
                ExprNodes.IntNode.for_size(self.pos, len(self.keys)),
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,329 @@ def get_main_pattern_targets(self):
...
+            func_name,
+            self.Pyx_mapping_extract_subjects_type,
+            utility_code=util_code,
+            args=[
+                subject_node,
+                MappingComparisonNode.make_keys_node(self.pos),
+                ExprNodes.IntNode(
+                    self.pos,
+                    value=str(len(self.keys))
+                ),
```
```suggestion
                ExprNodes.IntNode.for_size(self.pos, len(self.keys)),
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,329 @@ def get_main_pattern_targets(self):
...
+                ],
+            ),
+        )
+
+    def check_all_keys(self, subject_node):
+        # It's debatable here whether to go for individual unpacking or a function.
+        # Current implementation is a function that's loosely copied from CPython.
+        # For small numbers of keys it might be better to generate the code instead.
+        # There's three versions depending on if we know that the type is exactly
+        # a dict, definitely not or dict, or unknown.
```
```suggestion
        # a dict, definitely not a dict, or unknown.
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,329 @@ def get_main_pattern_targets(self):
...
+        # The advantages of generating a function are:
+        # * more compact code
+        # * easier to check the type once then branch the implementation
+        # * faster in the cases that are more likely to fail due to wrong keys being
+        # present than due to the values not matching the patterns
```
```suggestion
        # The advantages of generating a function are:
        # * more compact code
        # * can potentially be extracted into the shared utility module
        # * easier to check the type once then branch the implementation
        # * faster in the cases that are more likely to fail due to wrong keys being
        #    present than due to the values not matching the patterns
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,329 @@ def get_main_pattern_targets(self):
...
+            assert self.double_star_temp
+            body.stats.append(
+                # make_double_star_capture wraps itself in an if
+                self.make_double_star_capture(subject_node, test_result)
+            )
+
+        if duplicate_check or self.double_star_capture_target:
+            body = UtilNodes.TempResultFromStatNode(test_result, body)
+        else:
+            body = all_tests
```
scoder: We can get along with a bit less node-or-not juggling here.
```suggestion
        body = all_tests_node = generate_binop_tree_from_list(self.pos, "and", all_tests)
        duplicate_check = self.make_duplicate_keys_check(n_literal_keys)

        if duplicate_check or self.double_star_capture_target:
            stats = []
            if duplicate_check:
                stats.append(duplicate_check)
            test_result = UtilNodes.ResultRefNode(pos=self.pos, type=PyrexTypes.c_bint_type)
            stats.append(
                Nodes.SingleAssignmentNode(self.pos, lhs=test_result, rhs=all_tests_node))
            if self.double_star_capture_target:
                assert self.double_star_temp
                stats.append(
                    # make_double_star_capture wraps itself in an if
                    self.make_double_star_capture(subject_node, test_result)
                )
            body = UtilNodes.TempResultFromStatNode(test_result, Nodes.StatListNode(self.pos, stats=stats))
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,329 @@ def get_main_pattern_targets(self):
...
+                subjects_array=self.subject_temps
+            )
+        return LazyCoerceToBool(body.pos, arg=body)
+
+    def analyse_pattern_expressions(self, env, sequence_mapping_temp):
+        def to_temp_or_literal(node):
+            if node.is_literal:
+                return node
+            else:
+                return node.coerce_to_temp(env)
```
```suggestion
            return node if node.is_literal else node.coerce_to_temp(env)
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -1294,3 +1672,120 @@ def generate_binop_tree_from_list(pos, operator, list_of_tests):
             operand1=operand1,
             operand2=operand2
         )
+
+
+class MappingComparisonNode(ExprNodes.ExprNode):
+    """
+    Combined with MappingComparisonNodeInner this is responsible 
+    for setting up up the arrays of subjects and keys
```
```suggestion
    Combined with MappingComparisonNodeInner this is responsible 
    for setting up the arrays of subjects and keys.
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -1294,3 +1672,120 @@ def generate_binop_tree_from_list(pos, operator, list_of_tests):
...
+        return self.arg.type
+
+    def analyse_types(self, env):
+        self.arg = self.arg.analyse_types(env)
+        for n in range(len(self.keys_array)):
+            key = self.keys_array[n].analyse_types(env)
+            key = key.coerce_to_pyobject(env)
+            self.keys_array[n] = key
+        assert self.arg.type is PyrexTypes.c_bint_type
+        return self
```
scoder, line 1747: The docstring in `MappingComparisonNode` says that it owns `keys_array`, so I'd think that that class should also be responsible for analysing the elements in that list, shouldn't it? It seems that that's done twice now, here and there.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -1294,3 +1672,120 @@ def generate_binop_tree_from_list(pos, operator, list_of_tests):
...
+            arg=arg,
+            keys_array = self.keys_array,
+            subjects_array = subjects_array
+        )
+
+    def analyse_types(self, env):
+        self.inner = self.inner.analyse_types(env)
+        self.keys_array = [
+            key.analyse_types(env).coerce_to_simple(env) for key in self.keys_array
+        ]
```
scoder, line 1715: This seems enough to include the item analysis done by `MappingComparisonNodeInner`.
```suggestion
        self.keys_array = [
            key.analyse_types(env).coerce_to_pyobject(env).coerce_to_simple(env)
            for key in self.keys_array
        ]
```

scoder (commented):
Seems good overall. I mostly have code level comments.

Cython/Compiler/MatchCaseNodes.py
da-woods: I don't think it has to be in an initial implementation. I don't think C++ sequences work either for example.

I'm fine with adding C++ types later though.

scoder:
There's a test crashing with the Py3.9 fallback code:
` test_late_registration_sequence (test_patma.TestInheritance) ... Fatal Python error: Segmentation fault`

da-woods:
Yes just fixed it. Will look at other comments now

Cython/Compiler/MatchCaseNodes.py
da-woods: I'm wondering if using `repr` adds anything even for strings.

The duplicate checks like that are definitely tested in `test_patma` (although it's currently disabled... but it'll be enabled in the end).

I'll add a comment.

Cython/Compiler/MatchCaseNodes.py
da-woods: I'll update the comment, but not sure it's worth extracting to the shared utility module now... Right now we know that nobody's using it.

Cython/Compiler/MatchCaseNodes.py
da-woods: Yes. There's a small mess here where the list gets replaced in `MappingComparisonNode` so they point to different lists (which might have been the reason for the double analysis). So I've rewritten it just to point to the parent instead.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -936,6 +988,319 @@ def get_main_pattern_targets(self):
...
+        # * more compact code
+        # * can potentially be extracted into the shared utility module
+        # * easier to check the type once then branch the implementation
+        # * faster in the cases that are more likely to fail due to wrong keys being
+        #    present than due to the values not matching the patterns
+        if not self.keys:
+            return ExprNodes.BoolNode(self.pos, value=True)
+
+        is_dict = self.is_dict_type_check(subject_node.type)
+        if is_dict is self.DictCheckResult.ExactDict or is_dict == self.DictCheckResult.ExactFrozenDict:
```
```suggestion
        if is_dict is self.DictCheckResult.ExactDict or is_dict is self.DictCheckResult.ExactFrozenDict:
```

scoder:
GraalPy shows three minor test failures in the pure Python tests that are not on us. I also checked that PyPy is fine, so it's really just GraalPy that needs adaptation here.
```
File "/home/runner/work/cython/cython/tests/run/extra_patma_py.py", line 32, in extra_patma_py.test_duplicate_keys
Failed example:
    test_duplicate_keys("a", "a")
Expected:
       ...
    ValueError: mapping pattern checks duplicate key ('a')
Got:
      File "/opt/hostedtoolcache/GraalPy/25.0.3/x64/lib/python3.12/doctest.py", line 1368, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest extra_patma_py.test_duplicate_keys[1]>", line 1, in <module>
        test_duplicate_keys("a", "a")
      File "/home/runner/work/cython/cython/tests/run/extra_patma_py.py", line 42, in test_duplicate_keys
        case {Keys.KEY_1: _, Keys.KEY_2: _}:
    ValueError: mapping pattern checks duplicate key (a)

  File "tests/run/test_patma.py", line 252, in test_patma.TestInheritance.test_late_registration_mapping
    self.assertEqual(self.check_sequence_then_mapping(Parent()), "map")
AssertionError: None != 'map'

  File "tests/run/test_patma.py", line 275, in test_patma.TestInheritance.test_late_registration_sequence
    self.assertEqual(self.check_sequence_then_mapping(Parent()), "seq")
AssertionError: None != 'seq'
```

da-woods:
It looks like the `test_late_registration_*` tests are in the compiled version rather than the pure version.

I don't think this is something I can fix though. It will be that it isn't updating the type-flags for anything seen by the C API or something similar. So I've just disabled the test.

scoder:
The Windows failure looks random. Worth retrying, then this can be merged.

Cython/Utility/MatchCase.c
```diff
@@ -403,3 +391,343 @@ static PyObject *__Pyx_MatchCase_UnknownTypeSliceToList(PyObject *x, Py_ssize_t
...
+// in which case they're ignored.
+//
+// This is a specialized version for when we have an exact (frozen)dict (which is likely to be pretty common)
+
+#if CYTHON_REFNANNY
+#define __Pyx_MatchCase_Mapping_ExtractDict(...) __Pyx__MatchCase_Mapping_ExtractDict(__pyx_refnanny, __VA_ARGS__)
+#else
+#define __Pyx_MatchCase_Mapping_ExtractDict(...) __Pyx__MatchCase_Mapping_ExtractDict(NULL, __VA_ARGS__)
+#endif
+static CYTHON_INLINE int __Pyx__MatchCase_Mapping_ExtractDict(void *__pyx_refnanny, PyObject *dict, PyObject *keys[], Py_ssize_t nKeys, PyObject **subjects[]); /* proto */
```
scoder, line 527: I am now seeing warnings that `__pyx_refnanny` is an unsed argument in at least some of these functions. Is it a refactoring left-over, or is it going to be used at some later point?

da-woods: I believe it is used. I suspect if refnanny is disabled though it ends up being a pointless unused argument and should be marked as unused

da-woods: #7747 should fix the warning.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -1294,3 +1662,118 @@ def generate_binop_tree_from_list(pos, operator, list_of_tests):
...
+            "&"+subject.result() if subject is not None else "NULL" for subject in self.subjects_array
+        )
+        if not subjects_str:
+            # GCC gets worried about overflow if we pass
+            # a genuinely empty array
+            subjects_str = "NULL"
+        code.putln("PyObject **%s[] = {%s};" % (
+            MappingComparisonNode.subjects_array_cname,
+            subjects_str
+        ))
```
scoder, line 1770: @da-woods I'm seeing warnings that this array is unused in some cases where it's empty. Excluding its generation with `if not self.subjects_array` seems to resolve them without introducing failures in the patma tests.

Does that seem reasonable?

da-woods: Yeah I think so. There's certainly no secret reason to generate it when it's empty.

PR #7732, opened 2026-06-02
Pattern matching for class patterns

da-woods:
Part of https://github.com/cython/cython/issues/4029


tests/run/extra_patma.pyx
```diff
@@ -153,3 +153,56 @@ def test_ctuple_to_sequence((int, int) x):
...
+
+def class_attr_lookup(x):
+    """
+    >>> class_attr_lookup(C(5))
+    5.0
+    >>> class_attr_lookup([1])
+    >>> class_attr_lookup(None)
+    """
+    match x:
+        case C(x=y):  # This can only work with cdef attribute lookup
```
da-woods, line 169: This is the main "cython-specific" addition here, that it can do cdef attribute lookup as well as Python attribute lookup.

Cython/Compiler/ExprNodes.py
```diff
@@ -7855,6 +7855,7 @@ class AttributeNode(ExprNode):
     is_memslice_transpose = False
     is_special_lookup = False
     is_py_attr = 0
+    dont_mangle_private_names = False  # skip mangling class.__attr names
```
scoder: I'd reverse this to `mangle_private_names = True`. It's only ever used negated, making it needlessly harder to understand.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -786,13 +786,13 @@ def type_check(type):
...
-                Builtin.dict_type,
-                Builtin.set_type,
-            ]:
+            if (
+                type.is_pybytes_type or
+                type.is_pystr_type or
+                type.is_pybytearray_type or
+                type.is_pyanydict_type or
+                type.is_pyanyset_type
+            ):
```
scoder, line 793: I know, the name is somewhat unwieldy, but it's there.
```suggestion
            if (
                type.is_bytes_or_str_or_bytearray or
                type.is_pyanydict_type or
                type.is_pyanyset_type
            ):
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -1663,13 +2002,14 @@ def generate_binop_tree_from_list(pos, operator, list_of_tests):
         )
 
 
-class MappingComparisonNode(ExprNodes.ExprNode):
+class MappingOrClassComparisonNode(ExprNodes.ExprNode):
```
scoder: The combination of "mapping" and "class" seems surprising. (Why) are they the same thing?

To make a broadly underinformed additional comment, if they _are_ not the same thing but just _do_ the same thing, then we could at least misuse inheritance and give both separate names. (Either direct inheritance or a common base class.) Feel free to ignore this part of the comment if it makes no sense.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -1315,19 +1315,358 @@ class ClassPatternNode(PatternNode):
...
+            return 0
+        if self.class_known_type:
+            for type_attr in [
+                "is_pytuple_type",
+                "is_pylist_type",
+                "is_pyanydict_type",
+                "is_pyanyset_type",
+                "is_pybytes_type",
+                "is_pystr_type",
+                "is_pybytearray_type"
```
```suggestion
                "is_bytes_or_str_or_bytearray",
```
There's also `is_builtin_sequence` (+anydict+anyset would cover the entire list above), but that includes the builtin `memoryview`. Don't know if that can be matched, but if someone manages to build a match expression for it, it probably could.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -1315,19 +1315,358 @@ class ClassPatternNode(PatternNode):
...
+        ],
+        exception_value="-1",
+    )
+
+    Pyx_istype_type = PyrexTypes.CFuncType(
+        Builtin.type_type,
+        [
+            PyrexTypes.CFuncTypeArg("type", PyrexTypes.py_object_type, None),
+        ],
+    )
```
scoder: An `istype` function returns a type?

Cython/Utility/MatchCase.c
```diff
@@ -731,3 +731,233 @@ static PyObject* __Pyx_MatchCase_DoubleStarCapture{{tag}}(PyObject *mapping, PyO
...
+    return -1;
+}
+
+// Adapted from ceval.c "match_class" in CPython
+//
+// The argument match_self can equal 1 for "known to be true"
+//                                   0 for "known to be false"
+//                                  -1 for "unknown", runtime test
+// n_subjects is >= 0 otherwise this function will be skipped
+static int __Pyx__MatchCase_ClassPositional(void *__pyx_refnanny, PyObject *subject, PyTypeObject *type, PyObject *fixed_names[], Py_ssize_t n_fixed, int match_self, PyObject **subjects[], Py_ssize_t n_subjects)
```
scoder, line 822: Probably the same issue as before with the unused refnanny pointer.

Cython/Utility/MatchCase.c
```diff
@@ -731,3 +731,233 @@ static PyObject* __Pyx_MatchCase_DoubleStarCapture{{tag}}(PyObject *mapping, PyO
...
+        } else if (contains == 1) {
+            goto raise_error;
+        }
+    }
+    Py_DECREF(attrs_set);
+    return 0;
+
+    raise_error:
+    {
+        __Pyx_TypeName tp_name = __Pyx_PyType_GetFullyQualifiedName(type);
```
scoder: Needs error handling in Limited API. There's no helper function for the formatting order used here, sadly.

Cython/Utility/MatchCase.c
```diff
@@ -731,3 +731,233 @@ static PyObject* __Pyx_MatchCase_DoubleStarCapture{{tag}}(PyObject *mapping, PyO
...
+        if (!PyTuple_CheckExact(match_args)) {
+            __Pyx_TypeName type_typename = __Pyx_PyType_GetFullyQualifiedName(type);
+            __Pyx_TypeName match_args_type_name = __Pyx_PyType_GetFullyQualifiedName(Py_TYPE(match_args));
+            PyErr_Format(PyExc_TypeError, __Pyx_FMT_TYPENAME ".__match_args__ must be a tuple (got " __Pyx_FMT_TYPENAME ")",
+                type_typename,
+                match_args_type_name
+            );
+            Py_DECREF(match_args);
+            __Pyx_DECREF_TypeName(type_typename);
+            __Pyx_DECREF_TypeName(match_args_type_name);
```
scoder:
```suggestion
            __Pyx_RaiseTypeErrorWithTypes(
                __Pyx_FMT_TYPENAME ".__match_args__ must be a tuple (got " __Pyx_FMT_TYPENAME ")",
                type, Py_TYPE(match_args));
            Py_DECREF(match_args);
```

Cython/Utility/MatchCase.c
```diff
@@ -731,3 +731,233 @@ static PyObject* __Pyx_MatchCase_DoubleStarCapture{{tag}}(PyObject *mapping, PyO
...
+        allowed = __Pyx_PyTuple_GET_SIZE(match_args);
+#if !CYTHON_ASSUME_SAFE_SIZE
+        if (unlikely(allowed < 0)) goto end;
+#endif
+    } else {
+        allowed = 0;
+    }
+    if (allowed < n_subjects) {
+        const char *plural = (allowed == 1) ? "" : "s";
+        __Pyx_TypeName type_name = __Pyx_PyType_GetFullyQualifiedName(type);
```
scoder: Need error handling in Limited API.

Cython/Utility/MatchCase.c
```diff
@@ -731,3 +731,233 @@ static PyObject* __Pyx_MatchCase_DoubleStarCapture{{tag}}(PyObject *mapping, PyO
...
+            result = -1;
+            goto end;
+        } 
+#endif
+        if (!PyUnicode_CheckExact(name)) {
+            __Pyx_TypeName name_type_name = __Pyx_PyType_GetFullyQualifiedName(Py_TYPE(name));    
+            PyErr_Format(PyExc_TypeError,
+                         "__match_args__ elements must be strings "
+                         "(got " __Pyx_FMT_TYPENAME ")", name_type_name);
+            __Pyx_DECREF_TypeName(name_type_name);
```
scoder:
```suggestion
            __Pyx_RaiseTypeErrorWithObjectType(
                     "__match_args__ elements must be strings (got " __Pyx_FMT_TYPENAME ")",
                     name);
```

scoder (commented):
I only found minor issues. Seems fine to go in with a few cleanups and reconsiderations.

Cython/Compiler/MatchCaseNodes.py
da-woods: `memoryview` doesn't set `_Py_TPFLAGS_MATCH_SELF` so shouldn't be in this list. I'll add a test for it because it's easy to see how how it could accidentally end up included.

Cython/Compiler/MatchCaseNodes.py
da-woods: I've renamed this to `EvaluateWithKeysAndSubjectsArrays` which I think explains its role a bit better as a single class.

(I've also refactored it a bit to avoid the awkward `Inner` class)

Cython/Compiler/MatchCaseNodes.py
da-woods: Renamed to something `type_guard` which I think explains its role (checks if the argument is a type, and returns it) a little better

Cython/Utility/MatchCase.c
da-woods: I've provided a helper function/macro which can handle "type as the first argument; any combination of other arguments" to cover the two cases here.

Cython/Utility/MatchCase.c
```diff
@@ -731,3 +731,230 @@ static PyObject* __Pyx_MatchCase_DoubleStarCapture{{tag}}(PyObject *mapping, PyO
...
+        allowed = 1;
+    } else if (match_args) {
+        allowed = __Pyx_PyTuple_GET_SIZE(match_args);
+#if !CYTHON_ASSUME_SAFE_SIZE
+        if (unlikely(allowed < 0)) goto end;
+#endif
+    } else {
+        allowed = 0;
+    }
+    if (allowed < n_subjects) {
```
scoder: Marking this condition as `unlikely()` might help to ensure that the macro code (retrieving the type name and making the varargs call) gets properly moved out of the way.
```suggestion
    if (unlikely(allowed < n_subjects)) {
```

Cython/Utility/MatchCase.c
```diff
@@ -731,3 +731,230 @@ static PyObject* __Pyx_MatchCase_DoubleStarCapture{{tag}}(PyObject *mapping, PyO
...
+        PyExc_TypeError,
+        __Pyx_FMT_TYPENAME "() got multiple sub-patterns for attribute %R",
+        type, attr
+    );
+    bad:
+    Py_DECREF(attrs_set);
+    return -1;
+}
+
+// Adapted from ceval.c "match_class" in CPython
```
scoder, line 816: There were recent changes to that code in https://github.com/python/cpython/issues/138912 that seem relevant. Can still be integrated later, but worth having on the list.

Cython/Utility/ObjectHandling.c
```diff
@@ -3027,18 +3021,124 @@ static CYTHON_INLINE PyObject* __Pyx_{{typeobj}}_Multiply(PyObject *seq, Py_ssiz
...
+{
+    va_list arg_list;
+    va_start(arg_list, message);
+    __Pyx_TypeName tp_name = va_arg(arg_list, __Pyx_TypeName);
+    va_end(arg_list);
+#if CYTHON_COMPILING_IN_LIMITED_API && __PYX_LIMITED_VERSION_HEX < 0x030d0000
+    if (!tp_name) return;
+#else
+    CYTHON_UNUSED_VAR(tp_name);
+#endif
```
scoder, line 3087: Looking behind the macros here, I think we only need to extract the type name in the Limited API, right? Otherwise, we use `tp->tp_name` that we can just pass through. But I'm not sure that it's worth doing something about it, the C compiler would probably just eliminate the unused variable access, in which case we'd add complexity to our code without any gain.

Might be worth a "has been considered" comment in the `#else` case, though.

scoder (approved):

Cython/Utility/MatchCase.c
da-woods: The first suggestion there (that you can't get duplicates in a size-1 array) is a useful point, and also applies to the mapping check too.  I've implemented it.

The second doesn't apply to the Cython implementation.

The third is something I already did.

scoder: Excellent. Thanks for checking. (And for being smart.)

PR #7736, opened 2026-06-04
Document using thread-locals for safety

itamarst:
This is a pattern I used in a recent Scikit-Learn PR to get a significant speed-up, seemed worth documenting.

### DIFF
--- a/docs/src/userguide/freethreading.rst
+++ b/docs/src/userguide/freethreading.rst
@@ -348,6 +348,83 @@ need there is for locking to control their interaction, and the less likely
 they are to slow each other down by invaliding the CPU cache for other
 threads.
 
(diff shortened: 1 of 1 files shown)


scoder:
Thanks.

You can use a verbatim C code docstring in the cdef extern block instead of an external C file.

itamarst:
Oh neat, will update the example.

da-woods:
I definitely don't mind documenting it (in that it's definitely better to show people this kind of thing rather that letting them figuring it out for themselves), but I haven't personally found too many uses for thread-locals.

Just because it seems fairly rare that you need a per-thread global (which is basically what they are). The main use I've found is using to implement a non-thread-safe cache.

da-woods:
Thanks @itamarst

PR #7737, opened 2026-06-08
`infer_sequence_item_type` ignores `may_be_none()` for `T | None`-typed function arguments

shadchin:
`Cython.Compiler.ExprNodes.infer_sequence_item_type` infers a C numeric item type for tuple/list literals composed of function arguments annotated as `T | None` (or `Optional[T]`), even though those references can be `None` at runtime. The inferred C type then propagates into a generator expression's loop variable, producing C code that calls `__Pyx_PyFloat_AsDouble(Py_None)`

## Minimal reproducer

```python
# repro.pyx
def synchronous_gc(
    total_size: float | None = None,
    min_last_access: float | None = None,
    max_object_size: float | None = None,
):
    assert sum(
        1 for x in (total_size, min_last_access, max_object_size) if x is not None
    ) < 2
    return "ok"
```

Compile and call:

```python
>>> import repro
>>> repro.synchronous_gc()
  ...
TypeError: must be real number, not NoneType
```

Generated C (relevant lines, Cython 3.2.5):

```c
double __pyx_v_x;                                  /* <-- loop variable typed as C double */
...
__pyx_t_4 = __Pyx_PyFloat_AsDouble(__pyx_t_3);     /* <-- fails when __pyx_t_3 is Py_None */
if (unlikely((__pyx_t_4 == (double)-1) && PyErr_Occurred())) __PYX_ERR(...);
__pyx_cur_scope->__pyx_v_x = __pyx_t_4;
__pyx_t_3 = PyFloat_FromDouble(__pyx_cur_scope->__pyx_v_x);
```

Expected: `__pyx_v_x` should be `PyObject *` because the tuple items may be `None`.


da-woods:
There's one test failing. To be it looks like a regression (maybe not an ultra-important regression but it'd be nice to keep it working if possible).

shadchin:
Yes, I'm working on it

Cython/Compiler/ExprNodes.py
scoder: The following might work. Rename `has_none` to `allows_none`, and then:
```suggestion
if not allows_none:
    allows_none = any(tp.python_type_constructor_name and tp.allows_none() for tp in item_types)
if not allows_none:
    item_types = [
        # Unpack the Python types to optimise and avoid a mix of inferred C and Python.
        item_type.equivalent_type if item_type.is_pyobject and item_type.equivalent_type else item_type
        for item_type in item_types
    ]
item_type = PyrexTypes.reduce_spanning_types(item_types)
```

scoder (commented):
The condition that you changed removes None values from the items to prevent them from mixing the plain `object` type into the type inference. We need to fix the type inference afterwards, not the None removal, so this is not the right change. Below is a different proposal.

Please create a test case from your reproducer in `tests/run/type_inference.pyx`.

PR #7738, opened 2026-06-10
cpython: Define few function from the ABI 3.14

vallsv:
Hi,

Here is few functions related to `long` (3.14)  and `float` (sounds like it's there for long time)  that i  would like to use at some point.

- https://docs.python.org/3.14/c-api/long.html
- https://docs.python.org/3.11/c-api/float.html

Would you like such contribution?

Tell me if there is more to do.

Thanks a lot.

### DIFF
--- a/Cython/Includes/cpython/float.pxd
+++ b/Cython/Includes/cpython/float.pxd
@@ -36,3 +36,21 @@ cdef extern from "Python.h":
     double PyFloat_AS_DOUBLE(object pyfloat)
     # Return a C double representation of the contents of pyfloat, but
     # without error checking.
+
+    int PyFloat_Pack2(double x, unsigned char *p, int le) except -1
+    # Pack a C double as the IEEE 754 binary16 half-precision format.
+
+    int PyFloat_Pack4(double x, unsigned char *p, int le) except -1
+    # Pack a C double as the IEEE 754 binary32 single precision format.
+
+    int PyFloat_Pack8(double x, unsigned char *p, int le) except -1
+    # Pack a C double as the IEEE 754 binary64 double precision format.
+
+    double PyFloat_Unpack2(const unsigned char *p, int le) except? -1.0
+    # Unpack the IEEE 754 binary16 half-precision format as a C double.
+
+    double PyFloat_Unpack4(const unsigned char *p, int le) except? -1.0
+    # Unpack the IEEE 754 binary32 single precision format as a C double.
+
+    double PyFloat_Unpack8(const unsigned char *p, int le) except? -1.0
+    # Unpack the IEEE 754 binary64 double precision format as a C double.
--- a/Cython/Includes/cpython/long.pxd
+++ b/Cython/Includes/cpython/long.pxd
@@ -1,3 +1,4 @@
+from libc.stdint cimport uint32_t, uint64_t, int32_t, int64_t
 
 cdef extern from "Python.h":
     ctypedef long long PY_LONG_LONG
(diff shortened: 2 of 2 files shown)


Cython/Includes/cpython/float.pxd
```diff
@@ -36,3 +36,21 @@ cdef extern from "Python.h":
...
     # without error checking.
+
+    int PyFloat_Pack2(double x, unsigned char *p, int le) except? -1
+    # Pack a C double as the IEEE 754 binary16 half-precision format.
+
+    int PyFloat_Pack4(double x, unsigned char *p, int le) except? -1
+    # Pack a C double as the IEEE 754 binary32 single precision format.
+
+    int PyFloat_Pack8(double x, unsigned char *p, int le) except? -1
+    # Pack a C double as the IEEE 754 binary64 double precision format.
```
da-woods:
```suggestion
    int PyFloat_Pack2(double x, unsigned char *p, int le) except -1
    # Pack a C double as the IEEE 754 binary16 half-precision format.

    int PyFloat_Pack4(double x, unsigned char *p, int le) except -1
    # Pack a C double as the IEEE 754 binary32 single precision format.

    int PyFloat_Pack8(double x, unsigned char *p, int le) except -1
    # Pack a C double as the IEEE 754 binary64 double precision format.
```

The exception values from these aren't ambiguous - `-1` is always an error.

Cython/Includes/cpython/long.pxd
```diff
@@ -147,3 +156,24 @@ cdef extern from "Python.h":
...
+
+    object PyLong_FromNativeBytes(const void *buffer, size_t n_bytes, int flags)
+    # Create a Python integer from the value contained in the firstn_bytes
+    # of buffer, interpreted as a two’s-complement signed number.
+
+    object PyLong_FromUnsignedNativeBytes(const void *buffer, size_t n_bytes, int flags)
+    # Create a Python integer from the value contained in the first n_bytes of
+    # buffer, interpreted as an unsigned number.
+
+    Py_ssize_t PyLong_AsNativeBytes(object pylong, void *buffer, Py_ssize_t n_bytes, int flags)  except? -1
```
da-woods:
```suggestion
    Py_ssize_t PyLong_AsNativeBytes(object pylong, void *buffer, Py_ssize_t n_bytes, int flags)  except -1
```

Cython/Includes/cpython/long.pxd
```diff
@@ -147,3 +156,24 @@ cdef extern from "Python.h":
...
+    # Copy the Python integer value pylong to a native buffer of size n_bytes.
+    # The flags can be set to -1 to behave similarly to a C cast, or to values
+    # to control the behavior.
+
+    int PyLong_AsInt32(object obj, int32_t *value)
+    int PyLong_AsInt64(object obj, int64_t *value)
+    # Set *value to a signed C int32_t or int64_t representation of obj.
+
+    int PyLong_AsUInt32(object obj, uint32_t *value)
+    int PyLong_AsUInt64(object obj, uint64_t *value)
```
da-woods:
```suggestion
    int PyLong_AsInt32(object obj, int32_t *value) except -1
    int PyLong_AsInt64(object obj, int64_t *value) except -1
    # Set *value to a signed C int32_t or int64_t representation of obj.

    int PyLong_AsUInt32(object obj, uint32_t *value) except -1
    int PyLong_AsUInt64(object obj, uint64_t *value) except -1
```

da-woods:
> Would you like such contribution?

Yes thanks - we try to keep these files up-to-date.

da-woods:
It looks like you might need to `cimport` a few of the sized int types from `libc.stdint`

vallsv:
@da-woods Thanks a lot for the comments. It makes it clear now.
Ill rewrite that.

vallsv:
It should be fine now. Thanks again for the review.

scoder:
Thanks. These are special and rare enough to go in without usage test.

PR #7739, opened 2026-06-11
[3.2.x] Shared module: Use `SharedUtilitySourceDescriptor` instead of temporary directory (#7723)

matusvalo:
This PR uses `SharedUtilitySourceDescriptor` instead of an empty pyx file in the temp dir. This simplifies the logic and also assures that `__pyx_f` is unchanged across builds.

Backport of https://github.com/cython/cython/pull/7723
Alternative to https://github.com/cython/cython/pull/7634

### DIFF
--- a/Cython/Build/SharedModule.py
+++ b/Cython/Build/SharedModule.py
@@ -1,13 +1,10 @@
 import os
-import re
-import shutil
-import tempfile
 
 from Cython.Compiler import (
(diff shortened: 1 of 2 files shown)


matusvalo:
Based on https://github.com/cython/cython/pull/7634 there is issue with reproducible builds hence, I propose this as hot fix.

PR #7740, opened 2026-06-12
HUGE PERFORMANCE IMPROVEMENTS for typed code

jairov4:
# HUGE PERFORMANCE IMPROVEMENTS for typed code

This branch contains a series of heavy compiler optimizations and new features that eliminate
Python-level overhead for typed code all of them already working. 

I am already using these optimizations with large codebases with massive speedups and my goal is just to share what is working for me to the community.
However, I may have missed some edge cases, let me know any bug you find.

Changes are trongly focused in the embedded use case and achieving MAXIMUM PERFORMANCE from 
Python code with zero or minimal changes, however many of them apply to pyrex too.

Try yourself online: https://cytrainer.gercon.com.co/

I have tested it in MacOS and Windows using Python 3.13 and 3.14; 
Generated code compiled using Apple Clang, MSVC and Clang 21 from MacPorts.

I depict the most important changes below:

---

### What it does

Properties and methods declared in `@cclass` are automatically promoted to `cpdef`
(C-callable + Python-callable) when they can be made so without breaking the Python
protocol.  The promoted functions are added to the class vtable and exported, so any
other Cython module that `cimport`s the class can call them without touching the Python
attribute machinery.

`auto_cpdef` already existed, but here we extended it.

### Before (default mode)

```python
# a.pyx — if t is of type B declared in b.pyx, calling b.B.prop from another module
t.prop
```

Generated (default):
```c
__pyx_t_3 = __Pyx_PyObject_GetAttrStr(
    ((PyObject *)__pyx_v_item),
    __pyx_mstate_global->__pyx_n_u_prop);   // full Python attribute lookup
```

### After (`auto_cpdef=True`)

The vtable gains slots for each promoted property and its setter:

```c
// b.c — vtable struct definition
struct __pyx_vtabstruct_1b_B {
  int (*__pyx___init__)(struct __pyx_obj_1b_B *, int __pyx_skip_dispatch);
  PyObject *(*mtd)(struct __pyx_obj_1b_B *, int __pyx_skip_dispatch);
  PyObject *(*class_tag)(PyObject *);
  int (*__pyx_f_1b_1B_4prop___get__)(struct __pyx_obj_1b_B *);   // ← cproperty slot
  int (*__pyx_f_1b_1B_4prop___set__)(struct __pyx_obj_1b_B *, int);
  float (*__pyx_f_1b_1B_11prop_second___get__)(struct __pyx_obj_1b_B *);
};
```

Callers in `a.c` use the vtable slot — no Python object creation, NO dict lookup:

```c
// a.c — vtable dispatch to cproperty getter (returns C int, not PyObject*)
__pyx_t_5 = ((struct __pyx_vtabstruct_1b_B *)__pyx_v_tt->__pyx_vtab)
               ->__pyx_f_1b_1B_4prop___get__(__pyx_v_tt);
```

The property getter itself is a tight C function with no boxing:

```c
// b.c — the generated cproperty getter
int __pyx_f_1b_1B_4prop___get__(struct __pyx_obj_1b_B *__pyx_v_self) {
  int __pyx_r;
  __pyx_r = __pyx_v_self->_B__prop_int;  // direct struct field read
  goto __pyx_L0;
  ...
}
```

### Type inference cascade

Because the getter returns `C int` rather than `PyObject*`, Cython's type inferencer
propagates this through expressions.  The local variable `temp` is inferred as `int`:

```c
// a.c (auto_cpdef / lto mode) — func_in_a locals
int __pyx_v_temp;    // ← inferred C int because t.prop returns int
```

In default mode the same variable is `PyObject *__pyx_v_temp` because `t.prop` returns
a boxed Python integer.

---

### What it does

Allows the compiler to assume the modules being compiled together are going to be linked together.
Very useful in embedded contexts.

With `lto=True` (combined with `auto_cpdef=True`) the compiler emits direct `extern`
declarations for the producer module's C symbols instead of going through imported
function pointers.  This lets the C compiler / linker inline or optimise across module
boundaries.

### Import style comparison

**auto_cpdef only** (function-pointer import, resolved at Python `import` time):
```c
// a.c — dynamic function-pointer slots
static int (*__pyx_f_1b_1B_4prop___get__)(struct __pyx_obj_1b_B *);  // ptr
static PyObject *(*__pyx_f_1b_func_in_b)(int __pyx_skip_dispatch);
```

**lto mode** (direct `extern`, resolved at link time):
```c
// a.c — direct symbol references; linker resolves these
 CYTHON_INLINE int __pyx_f_1b_1B_4prop___get__(struct __pyx_obj_1b_B *); /*proto*/
 CYTHON_INLINE int __pyx_f_1b_1B_4prop___set__(struct __pyx_obj_1b_B *, int);
 CYTHON_INLINE float __pyx_f_1b_1B_11prop_second___get__(struct __pyx_obj_1b_B *);
```

With `CYTHON_INLINE` on the prototypes, the C compiler can inline these one-liner
accessors across translation units when building with `-O2 -flto` or similar.

### super().__init__() becomes a direct C call

```python
# a.pyx — A inherits B; super().__init__() known statically
class A(B):
    def __init__(self) -> None:
        super().__init__()   # direct call to B.__init__
```

Generated in lto mode:
```c
// a.c — A.__init__: super().__init__() → direct cross-module C call
__pyx_f_1b_1B___init__(((struct __pyx_obj_1b_B *)__pyx_v_self), 1);
// no vtable indirection, no Python method lookup
```

---

### What it does

`SomeClass(args)` in typed context (where the static type of the result is known) is
split into `tp_new` + a direct C call to `__init__`, bypassing the Python
`type.__call__` machinery entirely.  This works even for cross-module classes and for
`__init__` with optional typed parameters.

### Generated code for `B()` (same-module, no args)

```c
// Equivalent of b.B() from a.pyx — tp_new then direct C __init__
__pyx_t_3 = __Pyx_tp_new(
    ((PyObject *)__pyx_mstate_global->__pyx_ptype_1b_B),
    __pyx_mstate_global->__pyx_empty_tuple);   // allocate
__pyx_f_1b_1B___init__(((struct __pyx_obj_1b_B *)__pyx_t_1), 1);  // initialise
```

### Cross-module constructor with optional typed args

```python
# b.pyx
from cython import cclass, int, float

@cclass
class BWithOptInit:
    x: int
    y: float
    def __init__(self, x: int, y: float = 1.5) -> None:
        self.x = x; self.y = y
```

The optional arg is packed into an `__pyx_opt_args` struct — the same mechanism
used by `cpdef` functions.

```c
// a.c — BWithOptInit(5) — required-only call
__pyx_t_13 = __Pyx_tp_new(...__pyx_ptype_1b_BWithOptInit...);
__pyx_f_1b_12BWithOptInit___init__(__pyx_v_bopt, 5, 1, NULL);  // NULL = no opt args
```

```c
// a.c — BWithOptInit(5, 2.5) — explicit optional arg
struct __pyx_opt_args_1b_12BWithOptInit___init__ __pyx_t_23;
__pyx_t_23.__pyx_n = 1;
__pyx_t_23.y = 2.5;         // C double, not PyObject*
__pyx_t_3 = __Pyx_tp_new(...__pyx_ptype_1b_BWithOptInit...);
__pyx_f_1b_12BWithOptInit___init__(__pyx_v_bopt2, 5, 1, &__pyx_t_23);
```

No `PyDict_New`, no `PyArg_ParseTuple`, no `PyObject_Call`.
Not even boxing for arguments.

### Vtable slot for `__init__`

```c
// b.c — vtable slot prototype for BWithOptInit.__init__
struct __pyx_vtabstruct_1b_BWithOptInit {
  int (*__pyx___init__)(struct __pyx_obj_1b_BWithOptInit *, int,
                        int __pyx_skip_dispatch,
                        struct __pyx_opt_args_1b_12BWithOptInit___init__ *);
};
```

In `auto_cpdef` mode callers use this vtable slot; in `lto` mode they call
`__pyx_f_1b_12BWithOptInit___init__` directly.

### Works in all expression contexts

The optimisation fires whether the constructor result is:
- assigned to a typed local (`bopt: BWithOptInit = BWithOptInit(5)`)
- used as a function argument (`_take_bopt(BWithOptInit(9))`)
- returned from a function (`return BWithOptInit(x)`)

---

## 4. `cpdef __init__` — vtable-based constructor dispatch

- Init is optimized too, gaining ability to receive C specialized types.
- `super().__init__()` resolves statically (no vtable indirection in `lto` mode,
  vtable in `auto_cpdef` mode)

---

## 5. `cpdef` classmethods

`@classmethod` on a `@cclass` is promoted to a vtable slot and exported.  Cross-module
callers call it as a direct C function pointer rather than going through
`PyObject_GetAttr` + `PyObject_Call`.

```python
# b.pyx
class B:
    @classmethod
    def class_tag(cls) -> object:
        return cls.__name__
```

```c
// b.c — vtable entry
struct __pyx_vtabstruct_1b_B {
  ...
  PyObject *(*class_tag)(PyObject *);   // ← classmethod slot
  ...
};
```

```c
// a.c — cross-module classmethod call via vtable pointer
__pyx_t_13 = __pyx_vtabptr_1b_B->class_tag(
    ((PyObject *)__pyx_mstate_global->__pyx_ptype_1b_B));
```

---

## 6. `python_subclassing=False` — eliminate override-check boilerplate

Every single `cpdef` method normally emits an "if overridden in Python subclass" guard.  With
`python_subclassing=False` this guard is removed; the `__pyx_skip_dispatch` parameter
is kept (vtable ABI compatibility) but marked `CYTHON_UNUSED`.

```c
// b.cpp (python_subclassing=False) — no dispatch block, UNUSED parameter
int __pyx_f_1b_1B___init__(struct __pyx_obj_1b_B *__pyx_v_self,
                            CYTHON_UNUSED int __pyx_skip_dispatch) {
  // "/* Check if overridden in Python */" block is gone
  ...
}
```

The check `grep -c "Check if overridden in Python" b.cpp` returns **0** in this mode.

However, still python_subclassing=True by default to preserve the original behavior.

---

### What it does

A new analysis pass (`NoexceptInference`) inspects function bodies.  Functions that
proved cannot raise (no `raise`, no Python calls, no propagating calls) are
automatically marked `noexcept`.  Call sites to such functions omit the
`if (unlikely(__pyx_r == -1 && PyErr_Occurred()))` check.

Also, a `noexcept` method definition skips the Traceback machinery and saving cycles. 

### Noexcept call site — no check

```c
// a.c — A.__init__ is provably noexcept; call site has no error guard
__pyx_f_1a_1A___init__(__pyx_v_t, 1);
// ← no "if (unlikely(" follows
```

### Raising function — check kept

```c
// a.c — DerivedWithCtorRaise.__init__ raises; check is preserved
__pyx_t_2 = __pyx_f_1a_20DerivedWithCtorRaise___init__(
    ((struct __pyx_obj_1a_DerivedWithCtorRaise *)__pyx_t_1), 1, 1);
if (unlikely(__pyx_t_2 == ((int)-1))) __PYX_ERR(0, 70, __pyx_L41_error)
```

The inference is **per-call-site** for `super().__init__()` and direct constructor
calls: each call site independently decides whether to emit the check based on the
callee's inferred spec.

---

### Closures and generators

Previously, any property or method that referenced a closure variable or `yield`
could not be promoted to `cpdef`.  This branch implements generalized trampolines:

- The C-callable slot contains the optimised C implementation.
- A separate "wrapper" (`__pyx_wrap_N`) holds the Python-protocol entry point that
  captures the closure or drives the generator.
- The two never conflict; the vtable always points to the C slot.

ALMOST EVERYTHING CAN BE PROMOTED TO CPDEF NOW, including properties, dunders, etc with closures and bodies
with `yield` or lambdas. In many cases capture C variables directly without needing to box them as `PyObject*`
in a closure cell, so they can be used in `cpdef` methods and properties without losing performance.

### Dunders with zero/fixed args

`__init__`, `__len__`, `__bool__`, etc. are now included in auto-promotion when their
signatures match the vtable ABI.  Zero-arg dunders (after `self`) get a dedicated
fast path avoiding Python-level call.

```python
len(t)
```

```c
((struct __pyx_vtabstruct_1a_A *)__pyx_v_t->__pyx_base.__pyx_vtab)->__pyx___len__(__pyx_v_t, 0); ...
```
---

### Dunders for binops

Binops like `__add__` that take one argument are promoted to `cpdef` when their signature matches the vtable ABI, even if they have a closure. 
This means that expressions like `a + b` where `a` and `b` are typed Cython classes can be compiled to a direct C call to the `__add__` implementation 
producing beautiful C code with NO Python overhead.

```c
  /* "auto_cpdef_dunders.py":78
 *     a = Vec2(1, 3) + Vec2(2, 4)
 *     print(a.coords())
*/
  __pyx_t_2 = ((PyObject *)__pyx_tp_new_18auto_cpdef_dunders_Vec2(((PyTypeObject *)__pyx_mstate_global->__pyx_ptype_18auto_cpdef_dunders_Vec2), __pyx_mstate_global->__pyx_empty_tuple, NULL)); if (unlikely(!__pyx_t_2)) __PYX_ERR(0, 78, __pyx_L1_error)
  __Pyx_GOTREF((PyObject *)__pyx_t_2);
  __pyx_t_1 = ((PyObject *)__pyx_t_2);
  __pyx_t_2 = 0;
  __pyx_f_18auto_cpdef_dunders_4Vec2___init__(((struct __pyx_obj_18auto_cpdef_dunders_Vec2 *)__pyx_t_1), 1, 3, 1); 
  __pyx_t_3 = ((PyObject *)__pyx_tp_new_18auto_cpdef_dunders_Vec2(((PyTypeObject *)__pyx_mstate_global->__pyx_ptype_18auto_cpdef_dunders_Vec2), __pyx_mstate_global->__pyx_empty_tuple, NULL)); if (unlikely(!__pyx_t_3)) __PYX_ERR(0, 78, __pyx_L1_error)
  __Pyx_GOTREF((PyObject *)__pyx_t_3);
  __pyx_t_2 = ((PyObject *)__pyx_t_3);
  __pyx_t_3 = 0;
  __pyx_f_18auto_cpdef_dunders_4Vec2___init__(((struct __pyx_obj_18auto_cpdef_dunders_Vec2 *)__pyx_t_2), 2, 4, 1); 
  __pyx_t_3 = ((PyObject *)__pyx_f_18auto_cpdef_dunders_4Vec2___add__(((struct __pyx_obj_18auto_cpdef_dunders_Vec2 *)__pyx_t_1), ((struct __pyx_obj_18auto_cpdef_dunders_Vec2 *)__pyx_t_2), 0)); if (unlikely(!__pyx_t_3)) __PYX_ERR(0, 78, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_3);
  __Pyx_DECREF((PyObject *)__pyx_t_1); __pyx_t_1 = 0;
  __Pyx_DECREF((PyObject *)__pyx_t_2); __pyx_t_2 = 0;
  __pyx_v_a = ((struct __pyx_obj_18auto_cpdef_dunders_Vec2 *)__pyx_t_3);
  __pyx_t_3 = 0;
  __pyx_t_4 = __pyx_f_18auto_cpdef_dunders_4Vec2_coords(__pyx_v_a, 0); if (unlikely(PyErr_Occurred())) __PYX_ERR(0, 79, __pyx_L1_error)
  __pyx_t_3 = __pyx_convert__to_py___pyx_ctuple_int__and_int(__pyx_t_4); if (unlikely(!__pyx_t_3)) __PYX_ERR(0, 79, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_3);
  if (__Pyx_PrintOne(0, __pyx_t_3) < 0) __PYX_ERR(0, 79, __pyx_L1_error)
  __Pyx_DECREF(__pyx_t_3); __pyx_t_3 = 0;
```

**Note**: There exist some special cases for mirror binops (`__radd__`, etc.) but still handled.

## 9. Dataclass support

`@dataclass` (both `@cython.dataclasses.dataclass` and `from dataclasses import dataclass`)
works correctly with all the above optimisations:

- `__init__` is cpdef-promoted; optional typed fields become `opt_args` struct members.
- The `opt_args` struct uses C types (`float y`) not `PyObject *`:

```c
// a.c — opt_args struct for DataClass.__init__
struct __pyx_opt_args_1b_9DataClass___init__ {
  int __pyx_n;
  float y;   // ← C float, not PyObject*
};
```

- `from_x` classmethod is promoted to the vtable.
- `frozen=True` dataclasses with `freelist` continue to work.

---

## 10. Parallel compilation (`cython -j N`)

`cython -j 4 module_a.pyx module_b.pyx ...` compiles multiple files in parallel using
`ProcessPoolExecutor`.  Global compiler options (`cimport_from_pyx`, etc.) are captured
before spawning workers and restored inside each worker process to avoid data races.

---

## 11. Enhanced enum support

Pure Python support for enums as C `enum` constants.  Cross-module access
resolves to the integer constant at compile time:

```python
@cclass
class Color(Enum):
    RED = 1
```

```c
// a.c — Color.RED used as C enum value directly
__pyx_e_5Color_RED = 1
```

No `PyObject_GetAttr`, no `int()` conversion. Even in Pure-Python mode the enums are still generated
as C enums and accessed as C integers, so this is a pure win.
When they need to interface to python they get boxed as `PyObject*` as before, so no loss of functionality.

---

## 12. Warning and error deduplication

When compiling multiple files, the same performance hint or warning
can fire once per file that `cimport`s a shared header.  Previously this flooded the
terminal with hundreds of identical lines.

```text
warning: events.pyx:95:0: Cannot reliably handle Cython dataclasses with base types in external modules since it is not possible to tell what fields they have
warning: events.pyx:95:0: Cannot reliably handle Cython dataclasses with base types in external modules since it is not possible to tell what fields they have
warning: events.pyx:95:0: Cannot reliably handle Cython dataclasses with base types in external modules since it is not possible to tell what fields they have
warning: events.pyx:95:0: Cannot reliably handle Cython dataclasses with base types in external modules since it is not possible to tell what fields they have
warning: events.pyx:95:0: Cannot reliably handle Cython dataclasses with base types in external modules since it is not possible to tell what fields they have
...
```

Now, Each unique formatted line is emitted exactly once per cythonize call; subsequent duplicates
are silently dropped.

```
warning: events.pyx:95:0: Cannot reliably handle Cython dataclasses with base types in external modules since it is not possible to tell what fields they have
```

---

## 13. Integer-literal narrowing in binary operations

A plain integer literal like `2` in `x // 2` has no declared C type.
So `cython.int_var // 2` produced a `long` result instead of the natural `int`, and
`(temp // 2, temp)` where `temp: int` produced a `(long, int)` ctuple that
doesn't match `tuple[int, int]`.

```python
# Before: typeof(g) == "long"  (literal 4 widened to c_long)
g = 4 // <int>2

# After: typeof(g) == "int"   (4 fits in c_int; result stays c_int)
g = 4 // <int>2
```

When combined with cproperty type inference, a tuple built from `int`-typed results
infers as a C tuple:

```python
# a.pyx
temp  = t.prop + t.prop   # t.prop returns C int → temp: int
temp3 = (temp // 2, temp) # both operands int → ctuple (int, int), not (long, int)
assert typeof(temp3) in ('tuple object', '(int, int)')
```

```c
// a.c — locals in func_in_a (auto_cpdef/lto mode)
int __pyx_v_temp;
__pyx_ctuple_int__and_int __pyx_v_temp3;  // ← C tuple, zero allocation
```

Without the narrowing fix `temp3` would be `__pyx_ctuple_long__and_int`, a type
mismatch that either caused a compile-time warning or silently widened the slot.

---

## 14. Value-Type Semantics for frozen dataclasses

Introduce a new `@cython.value_type` decorator that applies value-type semantics to a class, making it behave like a C 
struct with no Python overhead.
It is strictly limited to final, frozen dataclasses with only C-typed fields.
This is ideal for small data containers like `Vec2` that benefit from direct C operations and can be safely copied by value.
Fully stack allocated and no gc overhead involved.

This is a pure python code still works without Cython, but when compiled with Cython it generates beautiful C code.

```python
from __future__ import annotations
from dataclasses import dataclass
from cython import final, value_type, cclass, double

@value_type
@final
@cclass
@dataclass(frozen=True)
class Vec2:
    x: double
    y: double = 0.0

    def __add__(self, other: Vec2) -> Vec2:
        return Vec2(self.x + other.x, self.y + other.y)

def hot_make_and_add(a: double, b: double) -> Vec2:
    return Vec2(a, b) + Vec2(1.0, 1.0)
```

Look at the BEAUTIFUL C code generated for a simple `Vec2` class with an `__add__` method:

```c
static __pyx_val_18value_type_methods_Vec2 __pyx_f_18value_type_methods_hot_make_and_add(double __pyx_v_a, double __pyx_v_b, CYTHON_UNUSED int __pyx_skip_dispatch) {
  __pyx_val_18value_type_methods_Vec2 __pyx_r;
  __pyx_val_18value_type_methods_Vec2 __pyx_t_1;
  __pyx_val_18value_type_methods_Vec2 __pyx_t_2;
  __pyx_val_18value_type_methods_Vec2 __pyx_t_3;

/* "value_type_methods.py":83
 * def hot_make_and_add(a: cython.double, b: cython.double) -> Vec2:
 *     return Vec2(a, b) + Vec2(1.0, 1.0)             # <<<<<<<<<<<<<<
*/
  __pyx_t_1.x = __pyx_v_a;
  __pyx_t_1.y = __pyx_v_b;
  __pyx_t_2.x = 1.0;
  __pyx_t_2.y = 1.0;
  __pyx_t_3 = __pyx_f_18value_type_methods_4Vec2___add__((&__pyx_t_1), __pyx_t_2, 0); 
  __pyx_r = __pyx_t_3;
  ...
  return __pyx_r;
```

---

# How to reproduce it

You can interactively try this in: https://cytrainer.gercon.com.co/

The improvements are layered and interact: each one unlocks the next.  
The test file `tests/build/embed_modules_optimize.srctree` covers most of them end-to-end, and the
generated C in `TEST_TMP/build/embed_modules_optimize/` can be used to verify the codegen
claims.

```
python3.13 runtests.py --no-code-style --no-cleanup embed_modules_optimize
```


matusvalo:
Thank you @jairov4 for your PR. But I must say that the diff is huge. There is no chance that this PR is going to be reviewed... It should be split into smaller digestible pieces...

scoder:
> Thank you @jairov4 for your PR. But I must say that the diff is huge. There is no chance that this PR is going to be reviewed... It should be split into smaller digestible pieces...

Yes, definitely too large as a whole. You're already describing the features separately, so it should be possible to submit separate PRs for the related commits.

scoder:
> submit separate PRs for the related commits

… which, as I see from https://github.com/cython/cython/issues?q=is%3Apr%20author%3Ajairov4 , you have already done for a few of these.

I'm sorry that you mostly got short and somewhat negative responses. That is not because we rejected your contributions. But because they look AI coded from a quick glance, I think the reviewers (including me) considered the actual code cheap to produce and concentrated directly on the more general aspects, assuming that you could easily adapt the design. Most of the features that you name in this ticket are interesting for us, in one way or another, and we would appreciate having them included.

IMHO, the least controversial features are 3, 5, 7, 10, 12. Maybe you could concentrate on these first and create separate PRs for them? Once (most of) these are in, we could see how we can get the ball rolling for the rest.

da-woods:
FWIW my biggest current issue with https://github.com/cython/cython/pull/7626 was that the PR description turned out to be completely untrue because auto_cpdef of methods in C classes are at least partly supported. That was the point that I stopped engaging as a reviewer because it didn't seem like either me, you or whatever AI was involved actually knew what had been changed.

Based on the description here (without attempting to see whether the code actually does what is described) direct constructor calls seem like the most generally useful thing.

I'm generally sceptical of "link everything together in a one giant multi-module .so file" as an approach that Cython should be encouraging. My feeling is that this might be better living in your personal fork.

jairov4:
tbh it is difficult to me rollback and sequence again many PRs for each change, actually since first submission so far many new fixes and improvements were flowing in my fork. I know it is a big PR of course, I am aware.
I have created this site to help anyone to try those changes https://cytrainer.gercon.com.co/ as I mention I just wanted to share with the community what is working for me.

PR #7741, opened 2026-06-12
Improve ci-yml file filter

da-woods:
The issue was that if anything matched `negative-docs` then it ended up overriding `any-files` even if there were other files changed.

The solution I've come up with isn't particularly general. It should work for this particular set of path filters but it doesn't solve the problem well.

I quite like the "applied in order" aspect of the regular github actions path filter, but it doesn't combine well with required actions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,12 +31,10 @@ jobs:
       pull-requests: read
     outputs:
       ci-changes: |-
-        ${{
-          steps.positive-github == 'true' || steps.positive-docs == 'true' || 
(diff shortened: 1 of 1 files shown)


da-woods:
I've pushed a test commit on my repo (https://github.com/da-woods/cython/actions/runs/27432012789/job/81084473290) with a mixture of docs and non-docs changes and it seems to trigger correctly so I think this works. But it's github actions so who really knows?

scoder:
> But it's github actions so who really knows?

Whatever we can get to work for us is probably just fine.

PR #7743, opened 2026-06-12
Improve some type annotations in the source code

matusvalo:


scoder:
Fine in general but there's an issue with the string constant declaration. If it's not an easy fix, maybe exclude that and commit the rest for now?

matusvalo:
> there's an issue with the string constant declaration
Yes I did the same mistake already before and the frustration led me to add support for subscripted types :-/

> If it's not an easy fix, maybe exclude that

I renamed the variable. Now should be OK in my opinion even better since the same variable is not used for different data types...

PR #7744, opened 2026-06-12
Fix 32 bit immortalization (by turning it off)

### LINKED ISSUE #7742: [BUG] immortalize_constants is broken on 32bit systems
### Describe the bug

This line of code is passing UINT32_MAX (`_Py_IMMORTAL_REFCNT_LOCAL`) to a function that takes a `Py_ssize_t` as an argument. On 32 bit systems, this overflows to -1, leading to crashes.

https://github.com/cython/cython/blob/2d06b4992eeafbf52171a24b79df761a09727385/Cython/Compiler/Code.py#L2529

NumPy hit this trying to build win32 wheels for our nightly CI: https://github.com/numpy/numpy/pull/31607

One fix is to use the code path in `Py_SET_REFCNT` for overflowing arguments:

https://github.com/python/cpython/blob/f4f102027a9b0edc72a048f17b696aa92d2e6893/Include/refcount.h#L177-L182

This code already only fires on the free-threaded build, so I think it's safe to substitute the code that manually sets PyObject internals.

This happens for all Cython modules on 32bit architectures using the free-threaded build. It triggers an assert in CPython on 3.15t:

```
Assertion failed: refcnt >= 0, file C:\Users\goldbaum\AppData\Local\Python\pythoncore-3.15t-32\Include\refcount.g, line 155
Fatal Python error: Aborted
```

### Expected behaviour

No asserts when I import Cython modules.

### OS

Windows

### Python version

3.15t 32 bit

### Cython version

3.2.5

### Additional context

I used Claude to identify the bug and for the suggested fix.

I'm going to work around this in the NumPy CI by disabling constant immortalization for the NumPy cython tests, with a TODO to revert that once this issue is fixed in Cython in a released version.

da-woods:
Fixes #7742.

Also work around a bug in Python where the overflow check meant we didn't set the thread id to unowned.

Also use `PyUnstable_SetImmortal` on Python 3.15+ instead.

### DIFF
--- a/Cython/Compiler/Code.py
+++ b/Cython/Compiler/Code.py
@@ -2516,19 +2516,26 @@ class GlobalState:
         writer.putln("{")
         writer.putln(f"PyObject **table = {array_cname};")
         writer.putln(f"for (Py_ssize_t i=0; i<{constant_count}; ++i) {{")
-        writer.putln("#if CYTHON_COMPILING_IN_CPYTHON_FREETHREADING")
+        writer.putln("#if PY_VERSION_HEX >= 0x030F0000")
+        writer.putln("PyUnstable_SetImmortal(table[i]);")
+        writer.putln("#elif CYTHON_COMPILING_IN_CPYTHON_FREETHREADING")
         # We don't want to set the refcount on shared constants (e.g. cached integers)
         # because setting the refcount isn't thread-safe. The chances are that most of the constants
         # that this applies to are already immortal though so that isn't a great loss.
+        # Overflow, e.g. on 32 bit systems.
+        writer.putln("if ((PY_SSIZE_T_MAX <= _Py_IMMORTAL_REFCNT_LOCAL)) break;")
         writer.putln("#if PY_VERSION_HEX < 0x030E0000")
         writer.putln("if (_Py_IsOwnedByCurrentThread(table[i]) && Py_REFCNT(table[i]) == 1)")
         writer.putln("#else")
         writer.putln("if (PyUnstable_Object_IsUniquelyReferenced(table[i]))")
         writer.putln("#endif")
         writer.putln("{")
-        writer.putln("Py_SET_REFCNT(table[i], _Py_IMMORTAL_REFCNT_LOCAL);")
+        # Go one higher than we think we need to because of a bug in SET_REFCNT check in CPython
+        writer.putln("Py_SET_REFCNT(table[i], ((Py_ssize_t)_Py_IMMORTAL_REFCNT_LOCAL + 1));")
         writer.putln("}")
         writer.putln("#else")
+        # Overflow, e.g. on 32 bit systems.
+        writer.putln("if ((PY_SSIZE_T_MAX < _Py_IMMORTAL_INITIAL_REFCNT)) break;")
         writer.putln("Py_SET_REFCNT(table[i], _Py_IMMORTAL_INITIAL_REFCNT);")
         writer.putln("#endif")
         writer.putln("}")  # for()


ngoldbaum:
I think this looks good. One tiny wrinkle is `PyUnstable_SetImmortal` will never immortalize string constants. I don't know if that behavior difference matters for Cython's use.

You can add 32 bit free-threaded windows CI if you'd like, see https://github.com/numpy/numpy/pull/31607 for an example.

da-woods:
> One tiny wrinkle is `PyUnstable_SetImmortal` will never immortalize string constants.

We already don't. Presumably `PyUnstable_SetImmortal` has a good reason for not supporting it. 

I was half tempted to use `PyUnstable_SetImmortal` on the string constant array anyway on the basis that we'd get the benefit of it ever changed in future. But I haven't done so here.

Cython/Compiler/Code.py
```diff
@@ -2516,19 +2516,26 @@ def immortalize_constants(array_cname, constant_count, writer):
...
+        writer.putln("if ((PY_SSIZE_T_MAX <= _Py_IMMORTAL_REFCNT_LOCAL)) break;")
         writer.putln("#if PY_VERSION_HEX < 0x030E0000")
         writer.putln("if (_Py_IsOwnedByCurrentThread(table[i]) && Py_REFCNT(table[i]) == 1)")
         writer.putln("#else")
         writer.putln("if (PyUnstable_Object_IsUniquelyReferenced(table[i]))")
         writer.putln("#endif")
         writer.putln("{")
-        writer.putln("Py_SET_REFCNT(table[i], _Py_IMMORTAL_REFCNT_LOCAL);")
+        # Go one higher than we think we need to because of a bug in SET_REFCNT check in CPython
+        writer.putln("Py_SET_REFCNT(table[i], ((Py_ssize_t)_Py_IMMORTAL_REFCNT_LOCAL + 1));")
```
scoder, line 2534: Since we should only reach this code in CPython 3.14, can't we just call `_Py_SetImmortal()` there, like `pythoncapi_compat.h` does?
https://github.com/python/pythoncapi-compat/blob/75a796764d63327268d195e2d5c044e564d0dada/pythoncapi_compat.h#L2692-L2693

da-woods: I think this covers 3.13t as well (although it's possibly debatable what the real value is in supporting that).

It also needs an internal include to access, although obviously that's something we could do.

scoder: My guess is that the traction of the latest Py3.1?t improvements is still high enough for most users to keep upgrading quickly, so Py3.14t is probably the minimum for production currently.

Anyway, I was just throwing in the question. It's clean for Py3.15 now and if this change gets it to work for Py3.13/3.14, that's fine with me.

scoder (approved):

Cython/Compiler/Code.py
da-woods: > My guess is that the traction of the latest Py3.1?t improvements is still high enough for most users to keep upgrading quickly, so Py3.14t is probably the minimum for production currently.

I think a lot of people have taken that view, and have decided to drop support for 3.13t pretty much as soon as 3.14t was released. I'm not sure we want to be quite that strict but it probably has few users and we shouldn't support it for too long.

ngoldbaum: > It also needs an internal include to access, although obviously that's something we could do.

FWIW, the way I think about this is despite the use of internals, the defintions in the pythoncapi_compat are to some extent "officially" blessed as the correct translations of modern idioms to older versions of the C API. So copying what the compat headers do almost completely mitigates use of internals, in my mind.

da-woods:
3.2.x commit c8b236333

PR #7745, opened 2026-06-12
Deprecate generate_pxi

### LINKED ISSUE #5047: Remove `generate_pxi`??

da-woods:
Eventually closes #5047

It's completely untested, and undocumented. I didn't manage to find any obvious users of it by searching github (but maybe I've missed some...).
Deprecating it is a fairly gentle approach but it may help us discover if there are any legitimate uses of it.

I'm made it self-disabling, just on the off-chance that we forget to chase it up. Although I think we will want to remove it properly.

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -353,8 +353,14 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
             h_code_main = globalstate.parts['type_declarations']
             h_code_end = globalstate.parts['end']
             if options.generate_pxi:
-                result.i_file = replace_suffix_encoded(result.c_file, ".pxi")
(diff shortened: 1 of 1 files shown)


da-woods:
I found one not too modern user https://github.com/LeeKamentsky/pyemd (not been updated in 12 years).

PR #7746, opened 2026-06-13
Split the "tp_new" function into allocation and initialisation

scoder:
The intention is to allow different entry points for allocation if we find quicker special cases, whereas the initialisation always needs to happen.

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -1671,12 +1671,138 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
         else:
             format_vectorcall_conditional = lambda vc, non_vc: non_vc
 
+        signature = format_vectorcall_conditional(
+            "{unused_marker}PyObject *const *args, {unused_marker}Py_ssize_t nargs, {unused_marker}PyObject *kwnames",
+            "{unused_marker}PyObject *a, {unused_marker}PyObject *k")
+        call_args = format_vectorcall_conditional(
+            "args, nargs, kwnames",
+            "a, k")
+
+        self._generate_tpnew_initialisation_function(scope, code, cclass_entry, is_vectorcall, signature, call_args)
+
+        code.start_slotfunc(
+            scope, PyrexTypes.py_objptr_type,
+            "tp_new_vectorcall" if is_vectorcall else "tp_new",
+            f"PyTypeObject *t, {signature.format(unused_marker='')}", needs_prototype=True)
+
+        code.putln("PyObject *o;")
+
+        base_type = scope.parent_type.base_type
+        if base_type:
+            self._generate_allocation_from_basetype(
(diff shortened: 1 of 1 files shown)


da-woods:
I think I'd avoid the name `init` just because it's a bit close to `__init__`. Maybe spell it out fully as `initialization`.

It does seem like two different operations and a fairly large function so it probably does make sense to be able to access the component parts individually.

I'm not sure how much it'd change #7698. In that I don't particularly see why we'd want to do the allocation differently (and the general rule seems to be that if you implement vectorcall then it should do the same thing as `tp_new` but hopefully faster).

scoder:
> I don't particularly see why we'd want to do the allocation differently

Right, it's not a big difference. The allocation remains the same. But the internal call to `__cinit__` gets a third and fourth signature. For the case of no-args `__cinit__`, we could even consider removing the arguments from the initialisation function entirely, although it might be simpler to take the used/unused decision only there and not further up. Since it's an internal function, it won't make a difference to the C compiler.

I've refactored it further to get an idea of how far we can reasonably push the split. I wouldn't mind doing it again if you want to merge your PR #7698 first.

scoder:
Updated for https://github.com/cython/cython/pull/7698

While working on this, I noticed that #7698 degraded the optimisation for direct module internal calls to `exttype.__new__()` which now goes through the fallback call. If we find direct calls to `exttype.__new__()`, I think we should always generate a direct entry point for the instantiation of the type (vectorcall or specialised), and then direct the call to that.

This equally applies to closure classes.

da-woods:
> This equally applies to closure classes.

I don't think it's too bad for closures because the empty dict and empty tuple to vectorcall conversion should be fairly good. Although obviously not as good as a direct call.  I agree with the general point though.

My original version of https://github.com/cython/cython/pull/7698 (commits https://github.com/cython/cython/pull/7698/commits/ec03c388361a867275d1b5c38b1d749b655f374e and https://github.com/cython/cython/pull/7698/commits/c1252d0bd518dc1e0182669e33abd6cbd10c6ef0) generated both the vectorcall and varargs function wrappers rather than having `tp_new` implemented in terms of vectorcall.

It was a bit ugly and generated a bit more code. But it did work, and could probably be made neater. Maybe that would still be the better thing to do?

scoder:
> It was a bit ugly and generated a bit more code. But it did work, and could probably be made neater. Maybe that would still be the better thing to do?

Hmm. Yes, I think _something_ like that could work better. If we could generate (up to) three entry points (as far as needed and supported by the signature), one for `tp_new`, `tp_vectorcall` and `tp.__new__(tp)`, and direct them all at the same internal function (which could be the `tp.__new__()` function or something really internal), I think that would serve all cases best and allow us to easily switch them on and off via C compile time flags.

da-woods:
Ignoring the "duplicate all the calls or not" question, the PR here looks good to me.

PR #7747, opened 2026-06-13
Avoid unused warnings about refnanny in mapping

da-woods:
I can't actually get GCC to warn me about it, but I'm relatively sure this is the right fix. The argument is unused when refnanny is off.

(One could also omit the argument. I think I decided it was more readable not to put macros in the function signature. But it would work either way.)


scoder:
LGTM

PR #7748, opened 2026-06-13
Avoid useless incref-decref cycles in the inner listcomp loop

scoder:
Instead, steal the item reference if we own it.


scoder:
There is no visible difference in the benchmarks (and I didn't really expect one). But the microbenchmark `[x+1 for x in it]` shows a 15% speedup on my side, so it clearly does make a difference.

PR #7750, opened 2026-06-14
Implement pattern matching "or" cases

da-woods:
This is the last implementation part of pattern matching (#4029).

I think this bit's fairly simple. Only notable details are:
* cache the result of "sequence or mapping" to share between each of the cases
* Evaluation of the individual sub-cases is essentially transformed into `idx = (1 if case1_test() else 0) or (2 if case2_test() else 0) or ...` when `idx` is used in a if statement to decide what subjects to assign.


Cython/Compiler/MatchCaseNodes.py
```diff
@@ -521,6 +539,14 @@ def is_irrefutable(self):
     def irrefutable_message(self):
         return self.get_first_irrefutable().irrefutable_message()
 
+    def is_sequence_or_mapping(self):
+        # this affects if the caller generates a temp for it. If so the
+        # this node can forward the temp to the relevant alternative
```
```suggestion
        # This affects if the caller generates a temp for it. If so then
        # this node can forward the temp to the relevant alternative.
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -505,7 +516,14 @@ def analyse_pattern_expressions(self, env, sequence_mapping_temp):
...
     """
     alternatives   list of PatternNodes
+
+    generated:
+    which_alternative_temp  - an integer temp node. 0 for failed; 1, 2...
+                              identify the alternative that succeeded
     """
+    which_alternative_temp = None
+    sequence_mapping_temp = None  # used in a similar way to MatchCaseNode,
+                # to avoid recalcutating if we're a sequence or mapping
```
```suggestion
                # to avoid recalculating if we're a sequence or mapping.
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -552,27 +578,52 @@ def is_simple_value_comparison(self):
...
-                operand1=binop,
-                operand2=a.get_simple_comparison_node(subject_node),
-            )
-        return binop
+        checks = []
+        for a in self.alternatives:
+            checks.append(a.get_simple_comparison_node(subject_node))
+        if any(isinstance(ch, ExprNodes.BoolNode) and ch.value for ch in checks):
+            # handle the obvious very simple case
+            return ExprNodes.BoolNode(self.pos, value=True)
```
scoder: What about evaluation errors along the way? If we don't get to look at the boolean value, users should probably still know about the failure.

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -552,27 +578,52 @@ def is_simple_value_comparison(self):
...
-        error(self.pos, "'or' cases aren't fully implemented yet")
-        return ExprNodes.BoolNode(self.pos, value=False)
+        if self.is_really_simple_value_comparison():
+            return self.get_simple_comparison_node(subject_node)
+
+        cond_exprs = []
+        for n, a in enumerate(self.alternatives, start=1):
+            a_test = a.get_comparison_node(subject_node, sequence_mapping_temp)
+            a_value = ExprNodes.IntNode(a.pos, value=str(n))
+            if isinstance(a_test, ExprNodes.BoolNode) and a_test.value:
```
scoder: If the condition value is known at compile time, even when false, we could use it to simplify the expression, couldn't we?

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -584,18 +635,79 @@ def analyse_pattern_expressions(self, env, sequence_mapping_temp):
...
+        if self.sequence_mapping_temp:
+            self.sequence_mapping_temp.allocate(code)
+            code.putln(
+                "%s = 0; /* sequence/mapping test temp */"
+                % self.sequence_mapping_temp.result()
+            )
+            # For things that are a sequence at compile-time it's difficult
+            # to avoid generating the sequence mapping temp. Therefore, silence
+            # an "unused error"
+            code.putln("(void)%s;" % self.sequence_mapping_temp.result())
```
```suggestion
        if self.sequence_mapping_temp:
            self.sequence_mapping_temp.allocate(code)
            tempvar = self.sequence_mapping_temp.result()
            code.putln(f"{tempvar} = 0; /* sequence/mapping test temp */")
            # For things that are a sequence at compile-time it's difficult
            # to avoid generating the sequence mapping temp. Therefore, silence
            # an "unused error".
            code.putln(f"(void){tempvar}")
```

Cython/Compiler/MatchCaseNodes.py
```diff
@@ -584,18 +635,79 @@ def analyse_pattern_expressions(self, env, sequence_mapping_temp):
...
+            sequence_mapping_count = 0
+            for a in self.alternatives:
+                if a.is_sequence_or_mapping():
+                    sequence_mapping_count += 1
+            if sequence_mapping_count >= 2:
+                self.sequence_mapping_temp = AssignableTempNode(
+                    self.pos, PyrexTypes.c_uint_type,
+                    is_addressable=True
+                )
+                sequence_mapping_temp = self.sequence_mapping_temp
```
```suggestion
        if not sequence_mapping_temp:
            sequence_mapping_condition_usage = sum(1 for a in self.alternatives if a.is_sequence_or_mapping())
            if sequence_mapping_condition_usage >= 2:
                self.sequence_mapping_temp = AssignableTempNode(
                    self.pos, PyrexTypes.c_uint_type,
                    is_addressable=True
                )
```

scoder (commented):
A few comments, then this should be good to go in.

Cython/Compiler/MatchCaseNodes.py
da-woods: Yes that's true. I'll drop this and add a test

PR #7752, opened 2026-06-16
Pattern matching benchmark

da-woods:
This is a small set of benchmarks for pattern matching that I wrote at the time and updated to work with the new benchmark runner.

They're clearly not hugely important, but since they already exist I think we may as well use them.

This is the last bit of #4029 - my belief is that it's now perfect and will never need to be changed again.


da-woods:
My plan is here that if no-one objects in ~24hrs then I'll merge it.

da-woods:
It looks like I need to exclude this benchmark for older versions of Cython....

Demos/benchmarks/run_benchmarks.py
```diff
@@ -130,6 +134,13 @@ def copy_benchmarks(bm_dir: pathlib.Path, benchmarks=None, cython_version=None):
             continue
         bm_file = bm_dir / bm_src_file.name
 
+        if cython_version and bm_src_file.name in BENCHMARK_MIN_VERSIONS:
+            _, *cy_version = cython_version
+            if tuple(cy_version) < BENCHMARK_MIN_VERSIONS[bm_src_file.name]:
```
scoder: `cython_version` is a tuple already.
```suggestion
            if cython_version[1:] < BENCHMARK_MIN_VERSIONS[bm_src_file.name]:
```
And I know that you've tried, but ISTM that `benchmark_revision` is a better place to decide about the benchmarks to run for the current revision than this function.

scoder (commented):
Any benchmark will probably do for now. If you want to look into extending it, these articles have several non-trivial examples:
https://benhoyt.com/writings/python-pattern-matching/
https://betterstack.com/community/guides/scaling-python/python-pattern-matching/

Here's a command parser example based on patterns:
https://medium.com/@vv2599/pattern-matching-in-python-a-practical-deep-dive-9197a2e0f96f#7a9a

And here are some examples that were written in the light of the original PEP discussions:
https://github.com/gvanrossum/patma/tree/master/examples
The most interesting seems an expression evaluation engine including tests, which could probably be used as benchmark commands directly.
https://github.com/gvanrossum/patma/blob/master/examples/expr.py
https://github.com/gvanrossum/patma/blob/master/examples/test_expr.py

Demos/benchmarks/run_benchmarks.py
da-woods: I can move it out. I put it there because it seemed like `PROCESSED_BENCHMARKS` was the closest existing thing for a quick fix.

> Any benchmark will probably do for now.

This was mostly a quick packaging of what I'd already done so I'm not inclined to add a lot to it right now.

But I do agree that what I have here is mostly just testing the individual elements in isolation which isn't really the full point of the feature (although practically, most uses of it probably are on the simpler end).

scoder:
I took the liberty to add a test (obviously working) based on your benchmark that mixes yields into the game. I noticed that the resulting C code does a lot of temp juggling, so I thought that it would be good to have an example of how that looks inside of a generator.

https://github.com/scoder/cython/commit/ebbc156b0b29c5e05688a7abdaf106413812b02d

PR #7754, opened 2026-06-17
Tests: Add a pattern matching test for yielding from match case statements (temp handling etc.)

scoder:


da-woods (approved):

tests/run/extra_patma_py.py
```diff
@@ -347,3 +347,81 @@ def test_or_evaluation_failure(x):
...
+        case {"a": "xxx", A.b: x, **extra}:
+            yield x
+        case 5 as y:
+            yield y
+        case set(y):
+            yield y
+        case A(b=1):
+            yield x
+        case C(a, None):
+            yield a
```
da-woods, line 432:
```suggestion
            yield a

             
def test_yield_from_case_if(x):
    """
    >>> g = test_yield_from_case_if([5])
    >>> next(g)
    >>> try:
    ...    g.send(5)
    ... except StopIteration:
    ...    pass
    A 5
    >>> g = test_yield_from_case_if([5])
    >>> next(g)
    >>> try:
    ...    g.send(6)
    ... except StopIteration:
    ...    pass
    B
    """
    match x:
        case [out] if out == (yield):
            print(f"A {out}")
        case _:
            print("B")
```

Another place where it's possible to put a `yield` (and where you probably wouldn't expect one to appear).

da-woods:
I'll try to have a look at the PyPy failure later, but it looks like there's potentially something wrong

da-woods:
The PyPy crash is when running the code directly in PyPy (https://github.com/pypy/pypy/issues/5506). It works fine when compiled in Cython. So I suggest you just disable it if pypy and not cython.compiled

tests/run/extra_patma_py.py
```diff
@@ -347,3 +351,106 @@ def test_or_evaluation_failure(x):
...
+class C:
+    __match_args__ = ("a", "b", "c")
+
+    def __init__(self):
+        self.a = 1
+        self.b = None
+        self.c = "string"
+
+
+@skip_in_pure_pypy  # crash
```
scoder:
```suggestion
@skip_in_pure_pypy  # crash, see https://github.com/pypy/pypy/issues/5506
```

PR #7755, opened 2026-06-17
Remove PyPy pin

da-woods:
I think the fixed version (7.3.23) should now be available


da-woods:
Assuming it works I'll merge this. If it doesn't work then I'll wait

PR #7757, opened 2026-06-20
Remove `from StringIO import StringIO`

da-woods:
and a few other related Py2 imports.


runtests.py
```diff
@@ -39,16 +39,8 @@
...
-try:
-    from StringIO import StringIO
-except ImportError:
-    from io import StringIO  # doesn't accept 'str' in Py2
-
-try:
-    import cPickle as pickle
-except ImportError:
-    import pickle
+from io import open as io_open, StringIO
```
scoder, line 42: `io.open` is just an alias for `open`. That can be cleaned up as well.

scoder (approved):

PR #7763, opened 2026-06-23
Improve handling on non-heaptype bases on Py3.9 Limited API

da-woods:
1. Fixes an issue in `__Pyx_MergeVtables` where accessing the final `object` base fails because it isn't a heap-type.
2. Raise an error in the constructor for a type than cannot be instantiated because we can't reach its base.

This will affect inheritance from external types.


da-woods:
I've trigger the wheel builder action for this on https://github.com/da-woods/cython/actions/runs/28009155756 (since I think that's where we were seeing the issue in practice)

scoder (approved):
Right, seems as good as it gets.

da-woods:
The wheels look happy at least

PR #7764, opened 2026-06-23
Emit compile-time error for 'not None' parameters with None default

### LINKED ISSUE #7762: not None parameter with = None default compiles silently but default is unreachable
## Summary

Cython accepts `def f(T x not None = None)` without any error or warning for extension types T, but the default value is unreachable — calling the function without the argument triggers `TypeError` at runtime because the default violates the `not None` constraint. It would be nice for this contradictory declaration to produce a compile-time error (or at minimum a warning). The `not None` declaration is a Cython addition to normal Python syntax, so even though it is only supported on Python `def` functions, I would still expect it to be subject to Cython's compile-time checks.

Cython does properly catch this when using primitive types that always fail at runtime because variables of those types unequivocally do not support assignment of None.

## Reproduction

```cython
# not_none_default.pyx
cdef class Foo:
    pass

def do_thing(Foo x not None = None):
    print(x)

def with_str(str s not None = None):
    print(s)

def untyped(x not None = None):
    print(x)
```

Compile and call:

```python
>>> import not_none_default
>>> not_none_default.do_thing()
TypeError: Argument 'x' has incorrect type (expected not_none_default.Foo, got NoneType)
>>> not_none_default.do_thing(None)
TypeError: Argument 'x' has incorrect type (expected not_none_default.Foo, got NoneType)
>>> not_none_default.do_thing(not_none_default.Foo())  # valid explicit argument
<not_none_default.Foo object at ...>
>>> not_none_default.with_str()
TypeError: Argument 's' has incorrect type (expected str, got NoneType)
>>> not_none_default.untyped()
TypeError: Argument 'x' must not be None
```

The function compiles, but the parameter is effectively required — the `= None` default can never satisfy the `not None` constraint. The declaration is contradictory and the default value is unreachable.

The int case fails to compile as expected
```cython
# not_none_default_int.pyx
def with_int(int x not None = None):
    print(x)
```

## Expected behavior

Cython should emit a compile-time error when a parameter is declared with both `not None` and a default value of `None`, since the default can never be used without raising `TypeError`.

Proposed error message:

```
Parameter 'x' is declared 'not None' but has a default value of None
```

Alternatively, if there's a use case for this pattern that I'm not seeing (e.g., documentation of intent for subclasses), a warning would be acceptable — but it would be nice for it to not pass silently.

## Environment

- Cython 3.2.5
- Python 3.13
- Linux x86_64

vyasr:
Adds a compile-time check that rejects parameters declared with `not None` that also have a default value of `None`, since the default can never be used without raising `TypeError` at runtime.

Fixes #7762

### DIFF
--- a/Cython/Compiler/Nodes.py
+++ b/Cython/Compiler/Nodes.py
@@ -3419,6 +3419,9 @@ class DefNode(FuncDefNode):
                     arg.accept_none = True
                 elif arg.not_none:
                     arg.accept_none = False
+                    if arg.default and arg.default.constant_result is None:
+                        error(arg.pos,
+                              "Parameter '%s' is declared 'not None' but has a default value of None" % arg.name)
                 elif (arg.type.is_extension_type or arg.type.is_builtin_type
                         or arg.type.is_buffer or arg.type.is_memoryviewslice):
                     if arg.default and arg.default.constant_result is None:
--- a/tests/errors/e_notnone_default.pyx
+++ b/tests/errors/e_notnone_default.pyx
@@ -1,0 +1,35 @@
+# mode: error
+
+cdef class Foo:
+    pass
+
+def ext_type(Foo x not None = None):
+    pass
+
+def builtin_type(str s not None = None):
+    pass
+
+def untyped(x not None = None):
+    pass
+
+def object_type(object o not None = None):
+    pass
+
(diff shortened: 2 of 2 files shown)


scoder:
Thanks

PR #7766, opened 2026-06-25
Fix ambiguous unique_ptr(NULL) comparison in C++ mode

### LINKED ISSUE #7765: [BUG] Cython 3.3.0a1 generates ambiguous unique_ptr(NULL) cast in C++ comparisons, breaks GCC 14
## Summary

Cython 3.3.0a1 (and current master) generates invalid C++ code when a `unique_ptr` attribute is compared to `NULL` in a boolean context (e.g., `if` statement or similar). The generated code casts `NULL` to `unique_ptr`, which is ambiguous with GCC 14.

## Minimal Reproducer

```cython
# repro.pyx
from libcpp.memory cimport unique_ptr

cdef extern from *:
    """
    struct MyStruct {
        int value;
    };
    """
    cdef cppclass MyStruct:
        int value

cdef class Wrapper:
    cdef unique_ptr[MyStruct] c_obj

cpdef bint check_not_null(Wrapper w):
    if w.c_obj != NULL:
        return True
    return False
```

Compile with:
```bash
cython --cplus repro.pyx -o repro.cxx
g++ -std=c++20 -fPIC -shared -I$(python -c "import sysconfig; print(sysconfig.get_path('include'))") repro.cxx -o repro.so
```

## Error

```
repro.cxx: error: call of overloaded 'unique_ptr(NULL)' is ambiguous
  __pyx_t_1 = (__pyx_v_w->c_obj != ((std::unique_ptr<MyStruct> )NULL));
                                                                 ^~~~
/usr/include/c++/14/bits/unique_ptr.h:353:19: note: candidate: 'constexpr std::unique_ptr<_Tp, _Dp>::unique_ptr(std::nullptr_t)'
/usr/include/c++/14/bits/unique_ptr.h:311:9: note: candidate: 'std::unique_ptr<_Tp, _Dp>::unique_ptr(pointer)'
```

## Environment

- **Cython 3.3.0a1 / master** — FAILS
- **Cython 3.2.5** — PASSES (generates `(__pyx_v_w->c_obj != NULL)` directly)
- **GCC 14.3.0**
- **C++ standard**: C++20

### Cython 3.2.5 (correct)
```cpp
__pyx_t_1 = ((__pyx_v_w->c_obj != NULL) != 0);
```
`NULL` is passed directly to `unique_ptr::operator!=`, which works fine.

### Cython 3.3.0a1 (broken)
```cpp
__pyx_t_1 = (__pyx_v_w->c_obj != ((std::unique_ptr<MyStruct> )NULL));
```
`NULL` is explicitly cast to `unique_ptr`, which requires constructing a `unique_ptr` from `NULL`. This is ambiguous because GCC's `NULL` (`__null`) matches both:
- `unique_ptr(pointer)` (pointer = `MyStruct*`)
- `unique_ptr(nullptr_t)`

## Root Cause

The regression was introduced by commit `23581d608` ("Always evaluate comparisons as bint if that is the expected result").

In `coerce_to_boolean`, the method now unconditionally sets `is_temp = True`. With `is_temp = True`, `generate_evaluation_code` takes the `generate_operation_code()` path, which computes a `common_type` and uses `result_as(common_type)`:

```python
common_type = type1  # unique_ptr<MyStruct>
code2 = operand2.result_as(common_type)  # casts NULL to unique_ptr → broken!
```

Since `analyse_cpp_comparison()` already coerces operands to the correct argument types of the resolved operator, the subsequent `result_as(common_type)` in `generate_operation_code` is redundant and harmful for C++ comparisons.

## Workaround

In user code, avoid `ptr != NULL` and use truthiness instead:

```cython
# Instead of:
if w.c_obj != NULL:

# Use:
if w.c_obj:
```

This avoids the comparison codepath entirely.

vyasr:
## Summary

AI disclosure: I provided an example of some code that was previously working under older Cython versions that was failing under Cython 3.3.0a1 to an agent, and it wrote up issue #7765 and came up with the fix in this PR. I verified that its tests are a faithful representation of the issue I observed, and the fix looks reasonable to me, but there may be a more concise fix here.

Fixes #7765.

Cython 3.3.0a1 introduced a regression where C++ comparisons (e.g. `unique_ptr != NULL`) generate ambiguous C++ code that fails to compile with GCC 14.

## Root Cause

Commit 23581d608 ("Always evaluate comparisons as bint if that is the expected result") changed `coerce_to_boolean` to unconditionally set `is_temp = True`. This routes C++ comparison codegen through `generate_operation_code()`, which casts both operands to a `common_type` via `result_as()`. For C++ types like `unique_ptr`, this generates:

```cpp
(__pyx_v_w->c_obj != ((std::unique_ptr<MyStruct> )NULL));
```

GCC 14 reports this as ambiguous because `NULL` (`__null`) matches both the `pointer` and `nullptr_t` constructors of `unique_ptr`.

## Fix

Since `analyse_cpp_comparison()` already coerces operands to the resolved operator's argument types, the subsequent `result_as(common_type)` cast is redundant for C++ comparisons. This PR skips it and uses the operands' `result()` directly when `is_cpp_comparison()` is true.

## Test

Added `tests/run/cpp_unique_ptr_null_comparison.pyx` covering:
- `unique_ptr == NULL` / `!= NULL`
- `unique_ptr == nullptr` / `!= nullptr`
- `unique_ptr` in boolean context (`if ptr:`)

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -14270,6 +14270,30 @@ class CmpNode:
                 operand1.result(),
                 operand2.result()))
 
+        elif self.is_cpp_comparison():
+            # C++ comparisons already had operands coerced in
+            # analyse_cpp_comparison(). Don't re-cast them to a
+            # common_type, as casting NULL to a C++ class type
+            # (e.g. unique_ptr) generates ambiguous code with GCC 14.
+            code1 = operand1.result()
+            code2 = operand2.result()
+            statement = "%s = %s(%s %s %s);" % (
+                result_code,
+                coerce_result,
+                code1,
+                self.c_operator(op),
+                code2)
+            if self.exception_check == '+':
+                translate_cpp_exception(
+                    code,
+                    self.pos,
+                    statement,
+                    result_code if self.type.is_pyobject else None,
+                    self.exception_value,
+                    self.in_nogil_context)
+            else:
+                code.putln(statement)
+
         else:
             type1 = operand1.type
             type2 = operand2.type
@@ -14282,22 +14306,12 @@ class CmpNode:
                 common_type = type1
             code1 = operand1.result_as(common_type)
             code2 = operand2.result_as(common_type)
-            statement = "%s = %s(%s %s %s);" % (
+            code.putln("%s = %s(%s %s %s);" % (
                 result_code,
                 coerce_result,
                 code1,
                 self.c_operator(op),
-                code2)
-            if self.is_cpp_comparison() and self.exception_check == '+':
-                translate_cpp_exception(
-                    code,
-                    self.pos,
-                    statement,
-                    result_code if self.type.is_pyobject else None,
-                    self.exception_value,
-                    self.in_nogil_context)
-            else:
-                code.putln(statement)
+                code2))
 
     def c_operator(self, op):
         if op == 'is':
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -14275,14 +14275,7 @@ class CmpNode:
             # analyse_cpp_comparison(). Don't re-cast them to a
             # common_type, as casting NULL to a C++ class type
             # (e.g. unique_ptr) generates ambiguous code with GCC 14.
-            code1 = operand1.result()
-            code2 = operand2.result()
-            statement = "%s = %s(%s %s %s);" % (
-                result_code,
-                coerce_result,
-                code1,
-                self.c_operator(op),
-                code2)
+            statement = f"{result_code} = {coerce_result}({operand1.result()} {self.c_operator(op)} {operand2.result()});"
             if self.exception_check == '+':
                 translate_cpp_exception(
                     code,
@@ -14306,12 +14299,7 @@ class CmpNode:
                 common_type = type1
             code1 = operand1.result_as(common_type)
             code2 = operand2.result_as(common_type)
-            code.putln("%s = %s(%s %s %s);" % (
-                result_code,
-                coerce_result,
-                code1,
-                self.c_operator(op),
-                code2))
+            code.putln(f"{result_code} = {coerce_result}({code1} {self.c_operator(op)} {code2});")
 
     def c_operator(self, op):
         if op == 'is':
--- a/tests/run/cpp_smart_ptr.pyx
+++ b/tests/run/cpp_smart_ptr.pyx
@@ -114,3 +114,66 @@ def test_dynamic_pointer_cast():
     a = shared_ptr[A](new A())
     b = dynamic_pointer_cast[B, A](a)
     assert b.get() == NULL
+
+
+cdef extern from *:
+    """
+    struct MyStruct {
+        int value;
+    };
+    """
+    cdef cppclass MyStruct:
+        int value
+
+
+cdef class Wrapper:
+    cdef unique_ptr[MyStruct] c_obj
+
+    def set_obj(self):
+        self.c_obj.reset(new MyStruct())
+
+
+def test_unique_ptr_null_comparison():
+    """
+    >>> test_unique_ptr_null_comparison()
+    True
+    False
+    True
+    """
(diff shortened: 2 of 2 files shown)


tests/run/cpp_unique_ptr_null_comparison.pyx
```diff
@@ -0,0 +1,69 @@
+# mode: run
```
da-woods: Unless there's a good reason not to, I think this should be moved to `tests/run/cpp_smart_ptr.pyx` (just because adding a new test file costs notably more CI time than expanding an existing file).

Cython/Compiler/ExprNodes.py
```diff
@@ -14273,15 +14273,26 @@ def generate_operation_code(self, code, result_code,
         else:
```
da-woods: The basic fix looks reasonable to me.

I wonder if it might be cleaner to move `if self.is_cpp_comparsion()` up here to a separate block (along with the `translate_cpp_exception` block below).

Just because the `is_cpp_comparison` block doesn't really look to share that much with rest of the code here.

Cython/Compiler/ExprNodes.py
```diff
@@ -14270,6 +14270,30 @@ def generate_operation_code(self, code, result_code,
...
+            # common_type, as casting NULL to a C++ class type
+            # (e.g. unique_ptr) generates ambiguous code with GCC 14.
+            code1 = operand1.result()
+            code2 = operand2.result()
+            statement = "%s = %s(%s %s %s);" % (
+                result_code,
+                coerce_result,
+                code1,
+                self.c_operator(op),
+                code2)
```
scoder: `code1` and `code2` aren't overly helpful names (which you just inherited here), and their expression is short enough now to go into an f-string.
```suggestion
            statement = f"{result_code} = {coerce_result}({operand1.result()} {self.c_operator(op)} {operand2.result()});"
```

Cython/Compiler/ExprNodes.py
```diff
@@ -14282,22 +14306,12 @@ def generate_operation_code(self, code, result_code,
...
-                translate_cpp_exception(
-                    code,
-                    self.pos,
-                    statement,
-                    result_code if self.type.is_pyobject else None,
-                    self.exception_value,
-                    self.in_nogil_context)
-            else:
-                code.putln(statement)
+                code2))
```
scoder: And since you're already touching this line, let's just update it as well.
```suggestion
            code.putln(f"{result_code} = {coerce_result}({code1} {self.c_operator(op)} {code2});")
```

tests/run/cpp_smart_ptr.pyx
```diff
@@ -114,3 +114,66 @@ def test_dynamic_pointer_cast():
...
+def test_unique_ptr_null_comparison():
+    """
+    >>> test_unique_ptr_null_comparison()
+    True
+    False
+    True
+    """
+    cdef Wrapper w = Wrapper()
+    print(w.c_obj == NULL)
+    print(w.c_obj != NULL)
```
scoder, line 145: Pointer comparisons _can_ use `==` and `!=`, but I would generally prefer `is` and `is not`. We should test both variants here.

scoder (commented):
Looks good to me overall, just a few style improvements.

tests/run/cpp_smart_ptr.pyx
da-woods: I'm not convinced `is` and `is not` should work here, since these comparisons are using an overloaded c++ `operator==`

scoder:
Will merge when CI is happy. Thanks!

tests/run/cpp_unique_ptr_null_comparison.pyx
vyasr: Fixed.

Cython/Compiler/ExprNodes.py
vyasr: Done.

vyasr:
Thanks for applying those fixes!

PR #7768, opened 2026-06-25
Fix cyfunction annotate setter on Python 3.14

### LINKED ISSUE #7767: [BUG] 3.2.6 release broke `@functools.wraps`
### Describe the bug

It seems that gh-7676 broke `functools.wraps`:
```
import_tests/test_import.py:14: in <module>
    @cupy.memoize()
     ^^^^^^^^^^^^^^
cupy/_util.pyx:54: in cupy._util.memoize.decorator
    ???
../../../miniforge3/envs/cupy-dev/lib/python3.14/functools.py:56: in update_wrapper
    setattr(wrapper, attr, value)
E   TypeError: __annotate__() missing 1 required positional argument: 'format'
```

```cython
import functools

def create(f):
    @functools.wraps(f)
    def cyfunc():
        pass

```
Then call `create()` with somethign like:
```
def myfunc(a : Any):  # has annotations
    pass
```

Doing it in pure Cython is a bit hard because Cython attaches an `__annotate__` that *doesn't* need a `format` argument, which is probably why this slipped through.

### Cython version

3.2.6

### Additional context

(Broke CuPy CI, but seems likely to cause more issues to me.)

JelleZijlstra:
Fixes #7767.

There were a few issues:
- Cython assumed it could call `__annotate__` with no arguments, which is not true for the functions CPython generates
- Calling it eagerly is the wrong thing to do anyway: it loses lazy evaluation of annotations. This PR adds lazy evaluation similar to what CPython does.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -255,6 +255,9 @@ Bugs fixed
   ``@functools.wraps`` work in Python 3.14+.
   (Github issue :issue:`7675`)
 
+* Assigning a Python 3.14+ ``.__annotate__`` function to a Cython compiled function no
+  longer evaluates annotations eagerly.
+
 * The ``--embed-positions`` option no longer includes absolute file paths in the C code.
   (Github issue :issue:`6755`)
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -173,6 +173,10 @@ Features added
 Bugs fixed
 ----------
 
+* Assigning a Python 3.14+ ``.__annotate__`` function to a Cython compiled function no
+  longer evaluates annotations eagerly.
+  (Github issue :issue:`7767`)
+
 * Generated Cython language features like properties, auto-pickle or dataclasses
   now use a critical section (on the object itself) as guard for concurrent access.
   (Github issue :issue:`6621`)
@@ -254,9 +258,6 @@ Bugs fixed
 * A minimal implementation of ``.__annotate__`` was added to Cython compiled functions to make
   ``@functools.wraps`` work in Python 3.14+.
   (Github issue :issue:`7675`)
-
-* Assigning a Python 3.14+ ``.__annotate__`` function to a Cython compiled function no
-  longer evaluates annotations eagerly.
 
 * The ``--embed-positions`` option no longer includes absolute file paths in the C code.
   (Github issue :issue:`6755`)
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -549,6 +549,7 @@ __Pyx_CyFunction_get_kwdefaults(PyObject *op, void *context) {
 }
 
 static PyObject *__Pyx_CyFunction_get_annotate_from_dict_if_exists(PyObject *op_in); /*proto*/
+static int __Pyx_CyFunction_set_annotate_in_dict_if_exists(PyObject *op_in, PyObject *value); /*proto*/
 static int __Pyx_CyFunction_set_annotate_in_dict(PyObject *op_in, PyObject *value); /*proto*/
 
 static int
@@ -566,7 +567,7 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
     __Pyx_BEGIN_CRITICAL_SECTION(op_in);
     __Pyx_Py_XDECREF_SET(op->func_annotations, value);
     __Pyx_END_CRITICAL_SECTION();
-    if (unlikely(__Pyx_CyFunction_set_annotate_in_dict(op_in, Py_None) < 0)) return -1;
+    if (unlikely(__Pyx_CyFunction_set_annotate_in_dict_if_exists(op_in, Py_None) < 0)) return -1;
     return 0;
 }
 
@@ -584,6 +585,20 @@ __Pyx_CyFunction_get_annotate_from_dict_if_exists(PyObject *op_in) {
     PyObject *annotate = PyDict_GetItemString(dict, "__annotate__");
     Py_XINCREF(annotate);
     return annotate;
+}
+
+static int
+__Pyx_CyFunction_set_annotate_in_dict_if_exists(PyObject *op_in, PyObject *value) {
+    PyObject *dict;
+#if PY_VERSION_HEX >= 0x030C0000 && !CYTHON_COMPILING_IN_LIMITED_API
+    PyObject **dictptr = _PyObject_GetDictPtr(op_in);
+    if (unlikely(!dictptr)) return 0;
+    dict = *dictptr;
+#else
+    dict = __Pyx_as_CyFunctionObject(op_in)->func_dict;
+#endif
+    if (!dict) return 0;
+    return PyDict_SetItemString(dict, "__annotate__", value);
 }
 
 static int
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -631,39 +631,22 @@ __Pyx_CyFunction_set_annotate_in_dict(PyObject *op_in, PyObject *value) {
 }
 
 static PyObject *
-__Pyx_CyFunction_get_annotations_locked(PyObject *op_in) {
+__Pyx_CyFunction_get_annotations_or_annotate_locked(PyObject *op_in, PyObject **annotate) {
     __pyx_CyFunctionObject *op = __Pyx_as_CyFunctionObject(op_in);
-    PyObject *annotate = NULL;
     PyObject *result = op->func_annotations;
+    *annotate = NULL;
     if (result) {
         Py_INCREF(result);
         return result;
     }
 
-    annotate = __Pyx_CyFunction_get_annotate_from_dict_if_exists(op_in);
-    if (unlikely(!annotate && PyErr_Occurred())) return NULL;
-    if (annotate && annotate != Py_None) {
-        PyObject *format = PyLong_FromLong(1L);  // annotationlib.Format.VALUE
-        if (unlikely(!format)) {
-            Py_DECREF(annotate);
-            return NULL;
-        }
-        result = __Pyx_PyObject_CallOneArg(annotate, format);
-        Py_DECREF(format);
-        Py_DECREF(annotate);
-        if (unlikely(!result)) return NULL;
-        if (unlikely(!PyDict_Check(result))) {
-            PyErr_SetString(PyExc_TypeError, "__annotate__ must return a dict");
-            Py_DECREF(result);
-            return NULL;
-        }
-        if (!op->func_annotations) {
-            Py_INCREF(result);
-            op->func_annotations = result;
-        }
-        return result;
+    *annotate = __Pyx_CyFunction_get_annotate_from_dict_if_exists(op_in);
+    if (unlikely(!*annotate && PyErr_Occurred())) return NULL;
+    if (*annotate && *annotate != Py_None) {
+        return NULL;
     }
-    Py_XDECREF(annotate);
+    Py_XDECREF(*annotate);
+    *annotate = NULL;
 
     result = op->func_annotations;
     if (unlikely(!result)) {
@@ -677,10 +660,35 @@ __Pyx_CyFunction_get_annotations_locked(PyObject *op_in) {
 
 static PyObject *
 __Pyx_CyFunction_get_annotations(PyObject *op_in, void *context) {
+    PyObject *annotate = NULL;
     PyObject *result = NULL;
+    __pyx_CyFunctionObject *op = __Pyx_as_CyFunctionObject(op_in);
     CYTHON_UNUSED_VAR(context);
     __Pyx_BEGIN_CRITICAL_SECTION(op_in);
-    result = __Pyx_CyFunction_get_annotations_locked(op_in);
+    result = __Pyx_CyFunction_get_annotations_or_annotate_locked(op_in, &annotate);
+    __Pyx_END_CRITICAL_SECTION();
+    if (result) return result;
+    if (unlikely(!annotate)) return NULL;
+
+    PyObject *format = PyLong_FromLong(1L);  // annotationlib.Format.VALUE
+    if (unlikely(!format)) {
+        Py_DECREF(annotate);
+        return NULL;
+    }
+    result = __Pyx_PyObject_CallOneArg(annotate, format);
+    Py_DECREF(format);
+    Py_DECREF(annotate);
+    if (unlikely(!result)) return NULL;
+    if (unlikely(!PyDict_Check(result))) {
+        PyErr_SetString(PyExc_TypeError, "__annotate__ must return a dict");
+        Py_DECREF(result);
+        return NULL;
+    }
+    __Pyx_BEGIN_CRITICAL_SECTION(op_in);
+    if (!op->func_annotations) {
+        Py_INCREF(result);
+        op->func_annotations = result;
+    }
     __Pyx_END_CRITICAL_SECTION();
     return result;
 }
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -631,24 +631,8 @@ __Pyx_CyFunction_set_annotate_in_dict(PyObject *op_in, PyObject *value) {
 }
 
 static PyObject *
-__Pyx_CyFunction_get_annotations_or_annotate_locked(PyObject *op_in, PyObject **annotate) {
-    __pyx_CyFunctionObject *op = __Pyx_as_CyFunctionObject(op_in);
-    PyObject *result = op->func_annotations;
-    *annotate = NULL;
-    if (result) {
-        Py_INCREF(result);
-        return result;
-    }
-
-    *annotate = __Pyx_CyFunction_get_annotate_from_dict_if_exists(op_in);
-    if (unlikely(!*annotate && PyErr_Occurred())) return NULL;
-    if (*annotate && *annotate != Py_None) {
-        return NULL;
-    }
-    Py_XDECREF(*annotate);
-    *annotate = NULL;
-
-    result = op->func_annotations;
+__Pyx_CyFunction_get_annotations_locked(__pyx_CyFunctionObject *op) {
+    PyObject* result = op->func_annotations;
     if (unlikely(!result)) {
         result = PyDict_New();
         if (unlikely(!result)) return NULL;
@@ -665,10 +649,21 @@ __Pyx_CyFunction_get_annotations(PyObject *op_in, void *context) {
     __pyx_CyFunctionObject *op = __Pyx_as_CyFunctionObject(op_in);
     CYTHON_UNUSED_VAR(context);
     __Pyx_BEGIN_CRITICAL_SECTION(op_in);
-    result = __Pyx_CyFunction_get_annotations_or_annotate_locked(op_in, &annotate);
+    if (op->func_annotations) {
+        result = __Pyx_CyFunction_get_annotations_locked(op);
+    }
     __Pyx_END_CRITICAL_SECTION();
     if (result) return result;
-    if (unlikely(!annotate)) return NULL;
+
+    annotate = __Pyx_CyFunction_get_annotate_from_dict_if_exists(op_in);
+    if (unlikely(!annotate && PyErr_Occurred())) return NULL;
+    if (!annotate || annotate == Py_None) {
+        Py_XDECREF(annotate);
+        __Pyx_BEGIN_CRITICAL_SECTION(op_in);
+        result = __Pyx_CyFunction_get_annotations_locked(op);
+        __Pyx_END_CRITICAL_SECTION();
+        return result;
+    }
 
     PyObject *format = PyLong_FromLong(1L);  // annotationlib.Format.VALUE
     if (unlikely(!format)) {
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -548,7 +548,7 @@ __Pyx_CyFunction_get_kwdefaults(PyObject *op, void *context) {
     return result;
 }
 
-static PyObject *__Pyx_CyFunction_get_annotate_from_dict(PyObject *op_in); /*proto*/
+static PyObject *__Pyx_CyFunction_get_annotate_from_dict_if_exists(PyObject *op_in); /*proto*/
 static int __Pyx_CyFunction_set_annotate_in_dict(PyObject *op_in, PyObject *value); /*proto*/
 
 static int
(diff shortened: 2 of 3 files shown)


Cython/Utility/CythonFunction.c
```diff
@@ -562,9 +566,29 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
...
     __Pyx_END_CRITICAL_SECTION();
+    if (unlikely(__Pyx_CyFunction_set_annotate_in_dict(op_in, Py_None) < 0)) return -1;
     return 0;
 }
 
+static PyObject *
+__Pyx_CyFunction_get_annotate_from_dict(PyObject *op_in) {
+    PyObject *dict = __Pyx_PyObject_GetAttrStr(op_in, PYIDENT("__dict__"));
+    if (unlikely(!dict)) return NULL;
+    PyObject *annotate = PyDict_GetItemString(dict, "__annotate__");
```
da-woods: Should this be `PyDict_GetItemStringRef` on versions that support it?

tests/run/cyfunction.pyx
da-woods, line 454: I think the most important test is that `wraps` works with a non-cyfunction and a cyfunction (presumably in both directions, although I think only the cyfunction wrapping a non-cyfunction direction was broken).

In that that's the behaviour that this is trying to support and everything else is really implementation details.

Cython/Utility/CythonFunction.c
```diff
@@ -579,11 +603,46 @@ __Pyx_CyFunction_get_annotations_locked(__pyx_CyFunctionObject *op) {
...
 static PyObject *
 __Pyx_CyFunction_get_annotations(PyObject *op_in, void *context) {
-    PyObject *result;
+    PyObject *annotate = NULL;
+    PyObject *result = NULL;
     __pyx_CyFunctionObject *op = __Pyx_as_CyFunctionObject(op_in);
     CYTHON_UNUSED_VAR(context);
     __Pyx_BEGIN_CRITICAL_SECTION(op_in);
-    result = __Pyx_CyFunction_get_annotations_locked(op);
+    if (op->func_annotations) {
```
da-woods: One consequence of this change is that looking up `__annotations__` on a cyfunction without any annotations will now always instantiate `__dict__` as a way of checking  (which most cyfunctions don't have) so it ends up being relatively expensive

Cython/Utility/CythonFunction.c
```diff
@@ -579,11 +603,46 @@ __Pyx_CyFunction_get_annotations_locked(__pyx_CyFunctionObject *op) {
...
+        result = __Pyx_CyFunction_get_annotations_locked(op);
+    }
+    __Pyx_END_CRITICAL_SECTION();
+    if (result) return result;
+
+    annotate = __Pyx_CyFunction_get_annotate_from_dict(op_in);
```
da-woods: Personally I think it'd be clearer to move most of this in `__Pyx_CyFunction_get_annotations_locked` rather than jumping in an out of critical sections. I'm aware that the critical section might end up interrupted but it's less to think about, and I don't think critical sections on cyfunctions are heavily contested (they're only really used for the introspection-type attributes)

da-woods (commented):
What I'm sceptical of here is putting it in `__dict__` because that typically wouldn't be instantiated, and it makes looking up `__annotations__` on something that doesn't have annotations expensive.

I'm wondering if `func_annotations` could double up and store either the annotations dict, or a callable instead (since a function is it's distinguishable from a `dict` anyway... ignoring the unlikely complication of callable `dict` subclasses). In that scheme the `__annotate__` function wouldn't survive after it's been used to generate `__annotations__`. I realize that isn't a perfect replica of how real Python functions work, but I think that's fine? The only objective is to make `wraps` work kind of OK and not to implement real deferred annotations.

JelleZijlstra:
We could maybe also store an (annotate, annotations) tuple in the func_annotations slot?

I'll get back to this later today.

da-woods:
Yeah the tuple could probably also work as a reasonable compromise between "don't take up extra space in the very common no-`__annotate__` case" and "avoid instantiating `__dict__` for this"

da-woods:
Another alternative might be: keep in in `__dict__` for now, as you're doing, but avoid instantiating `__dict__` if it doesn't already exist. I think that's a case of using `_PyObject_GetDictPtr` for version that support managed dict, and `->func_dict` otherwise.

JelleZijlstra:
Do you have a strong preference? I feel the last alternative (only materialize the dict if we need it) is cleanest and simplest, but happy to go with another approach if it's your preference.

da-woods:
> I feel the last alternative (only materialize the dict if we need it) is cleanest and simplest

Yes - I think I agree with that.

Cython/Utility/CythonFunction.c
JelleZijlstra: This code is changing a bit but I think the separate critical sections are cleaner, since various parts of this code may call arbitrary Python code (e.g., calling `__annotate__`), and there might be subtle deadlocks if we do that while holding the critical section.

Cython/Utility/CythonFunction.c
JelleZijlstra: Yeah, good point. Changes I'm about to push will avoid materializing the dict when we don't need to.

tests/run/cyfunction.pyx
JelleZijlstra: Added some tests.

Cython/Utility/CythonFunction.c
JelleZijlstra: Made it call this on recent versions.

Cython/Utility/CythonFunction.c
```diff
@@ -562,9 +567,69 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
...
     __Pyx_Py_XDECREF_SET(op->func_annotations, value);
     __Pyx_END_CRITICAL_SECTION();
+    if (unlikely(__Pyx_CyFunction_set_annotate_in_dict_if_exists(op_in, Py_None) < 0)) return -1;
     return 0;
 }
 
+static PyObject *
+__Pyx_CyFunction_get_annotate_from_dict_if_exists(PyObject *op_in) {
+    PyObject *dict;
+#if PY_VERSION_HEX >= 0x030C0000 && !CYTHON_COMPILING_IN_LIMITED_API
```
scoder: PyPy seems to provide `_PyObject_GetDictPtr()`, not sure about GraalPython but probably not. It would be better to use `PyObject_GenericGetDict()` outside of `CYTHON_COMPILING_IN_CPYTHON`.

Cython/Utility/CythonFunction.c
```diff
@@ -562,9 +567,69 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
...
+    PyObject **dictptr = _PyObject_GetDictPtr(op_in);
+    if (unlikely(!dictptr)) return NULL;
+    dict = *dictptr;
+#else
+    dict = __Pyx_as_CyFunctionObject(op_in)->func_dict;
+#endif
+    if (!dict) return NULL;
+#if (defined(Py_LIMITED_API) && Py_LIMITED_API >= 0x030d0000) || (!defined(Py_LIMITED_API) && PY_VERSION_HEX >= 0x030d0000)
```
scoder:
```suggestion
#if __PYX_LIMITED_VERSION_HEX >= 0x030d0000
```

Cython/Utility/CythonFunction.c
```diff
@@ -562,9 +567,69 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
...
+    PyObject *dict;
+#if PY_VERSION_HEX >= 0x030C0000 && !CYTHON_COMPILING_IN_LIMITED_API
+    PyObject **dictptr = _PyObject_GetDictPtr(op_in);
+    if (unlikely(!dictptr)) return 0;
+    dict = *dictptr;
+#else
+    dict = __Pyx_as_CyFunctionObject(op_in)->func_dict;
+#endif
+    if (!dict) return 0;
+    return PyDict_SetItemString(dict, "__annotate__", value);
```
scoder: I think we can afford to instantiate the string constant upfront. That avoids the need to call the `…String()` C-API variants.
```suggestion
    return PyDict_SetItem(dict, PYIDENT("__annotate__"), value);
```
etc. in all other places.

Cython/Utility/CythonFunction.c
```diff
@@ -562,9 +567,69 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
...
+    PyObject **dictptr = _PyObject_GetDictPtr(op_in);
+    if (unlikely(!dictptr)) return NULL;
+    dict = *dictptr;
+#else
+    dict = __Pyx_as_CyFunctionObject(op_in)->func_dict;
+#endif
+    if (!dict) return NULL;
+#if (defined(Py_LIMITED_API) && Py_LIMITED_API >= 0x030d0000) || (!defined(Py_LIMITED_API) && PY_VERSION_HEX >= 0x030d0000)
+    PyObject *annotate = NULL;
+    if (unlikely(PyDict_GetItemStringRef(dict, "__annotate__", &annotate) < 0)) return NULL;
```
scoder:
```suggestion
    if (unlikely(PyDict_GetItemRef(dict, PYIDENT("__annotate__"), &annotate) < 0)) return NULL;
```

Cython/Utility/CythonFunction.c
scoder: Could you please reverse this condition and swap the implementation blocks? I prefer moving special cases first and the general/normal case last, both because it makes special cases easier to remove later (due to dependency EOL) and because they often end up accompanied with more special cases, and then they chain more cleanly if put first. The same applies to various other places below.

Cython/Utility/CythonFunction.c
```diff
@@ -562,9 +567,69 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
     __Pyx_BEGIN_CRITICAL_SECTION(op_in);
     __Pyx_Py_XDECREF_SET(op->func_annotations, value);
     __Pyx_END_CRITICAL_SECTION();
+    if (unlikely(__Pyx_CyFunction_set_annotate_in_dict_if_exists(op_in, Py_None) < 0)) return -1;
     return 0;
 }
 
+static PyObject *
+__Pyx_CyFunction_get_annotate_from_dict_if_exists(PyObject *op_in) {
```
scoder: It looks like this function could benefit from an error return code to distinguish between "don't have a dict" and "failure while trying to get the dict".

Cython/Utility/CythonFunction.c
```diff
@@ -579,16 +644,51 @@ __Pyx_CyFunction_get_annotations_locked(__pyx_CyFunctionObject *op) {
...
-    PyObject *result;
+    PyObject *annotate = NULL;
+    PyObject *result = NULL;
     __pyx_CyFunctionObject *op = __Pyx_as_CyFunctionObject(op_in);
     CYTHON_UNUSED_VAR(context);
     __Pyx_BEGIN_CRITICAL_SECTION(op_in);
-    result = __Pyx_CyFunction_get_annotations_locked(op);
+    if (op->func_annotations) {
+        result = __Pyx_CyFunction_get_annotations_locked(op);
+    }
```
scoder: This is really just
```suggestion
    result = __Pyx_XNewRef(op->func_annotations);
```

Cython/Utility/CythonFunction.c
```diff
@@ -579,16 +644,51 @@ __Pyx_CyFunction_get_annotations_locked(__pyx_CyFunctionObject *op) {
...
+
+    PyObject *format = PyLong_FromLong(1L);  // annotationlib.Format.VALUE
+    if (unlikely(!format)) {
+        Py_DECREF(annotate);
+        return NULL;
+    }
+    result = __Pyx_PyObject_CallOneArg(annotate, format);
+    Py_DECREF(format);
+    Py_DECREF(annotate);
+    if (unlikely(!result)) return NULL;
```
scoder, line 659: The first failure is so unlikely that I'd rather reduce the code duplication:
```suggestion
    PyObject *format = PyLong_FromLong(1L);  // annotationlib.Format.VALUE
    if (likely(format)) {
        result = __Pyx_PyObject_CallOneArg(annotate, format);
        Py_DECREF(format);
    }
    Py_DECREF(annotate);
    if (unlikely(!result)) return NULL;
```

CHANGES.rst
```diff
@@ -255,6 +255,9 @@ Bugs fixed
   ``@functools.wraps`` work in Python 3.14+.
   (Github issue :issue:`7675`)
 
+* Assigning a Python 3.14+ ``.__annotate__`` function to a Cython compiled function no
+  longer evaluates annotations eagerly.
+
```
scoder: This appears in the wrong spot (in an already released version section). You can merge master and move it up to the future version, or I can do it later.

Cython/Utility/CythonFunction.c
```diff
@@ -579,16 +644,51 @@ __Pyx_CyFunction_get_annotations_locked(__pyx_CyFunctionObject *op) {
...
+    if (unlikely(!PyDict_Check(result))) {
+        PyErr_SetString(PyExc_TypeError, "__annotate__ must return a dict");
+        Py_DECREF(result);
+        return NULL;
+    }
+    __Pyx_BEGIN_CRITICAL_SECTION(op_in);
+    if (!op->func_annotations) {
+        Py_INCREF(result);
+        op->func_annotations = result;
+    }
```
scoder: Can we be sure that `op->func_annotations` holds the correct value if it's set already, or should we overwrite it?

scoder (commented):
Mostly minor issues from my side, thanks for writing this up.

Cython/Utility/CythonFunction.c
da-woods: > PyPy seems to provide `_PyObject_GetDictPtr()`, not sure about GraalPython but probably not. It would be better to use `PyObject_GenericGetDict()` outside of `CYTHON_COMPILING_IN_CPYTHON`.

It looks like we use `_PyObject_GetDictPtr()` elsewhere so I think it's fine.

Cython/Utility/CythonFunction.c
da-woods: I think you could argue that it's an arbitrary timing thing that some race has happened, both values were the right answer at some point, and it doesn't really which you return and which you persist. I don't think there's a single logical order that makes sense.

If I'd have written this, I'd have probably returned the existing value. But I don't have a reason to claim that this is a more correct choice.

Cython/Utility/CythonFunction.c
da-woods: I don't feel particularly strongly about this so happy to leave it as is since we have something that works currently.

I think critical sections should be deadlock-free. So the arbitrary Python code in `__annotate__` could definitely end up using the same critical section and breaking it, but should never deadlock. You couldn't assume that `func_annotate` was unchanged after the call to the arbitrary Python code though.

Cython/Utility/CythonFunction.c
da-woods: Thinking about this more: overwriting the existing value might be inconsistent (because you wouldn't have called `__annotate__` had there been an existing value). I think the choice of what to return really is arbitrary.

So I'd claim this implements (one of the) sensible options and so is fine.

Cython/Utility/CythonFunction.c
scoder: Right, I double checked and it's a historically named, de facto public function. Both PyPy and GraalPy provide it. Fine then.

Cython/Utility/CythonFunction.c
JelleZijlstra: Actually CPython unconditionally overwrites func_annotations in this case, so I think it's more consistent to do that.

Cython/Utility/CythonFunction.c
JelleZijlstra: Apparently there's no way for it to fail while getting a dict. But I put the dict getting in a separate function for clarity.

Cython/Utility/CythonFunction.c
```diff
@@ -579,16 +644,44 @@ __Pyx_CyFunction_get_annotations_locked(__pyx_CyFunctionObject *op) {
...
+    Py_DECREF(annotate);
+    if (unlikely(!result)) return NULL;
+    if (unlikely(!PyDict_Check(result))) {
+        PyErr_SetString(PyExc_TypeError, "__annotate__ must return a dict");
+        Py_DECREF(result);
+        return NULL;
+    }
+    __Pyx_BEGIN_CRITICAL_SECTION(op_in);
+    Py_CLEAR(op->func_annotations);
+    op->func_annotations = __Pyx_NewRef(result);
```
scoder:
```suggestion
    __Pyx_Py_XDECREF_SET(op->func_annotations, __Pyx_NewRef(result));
```

Cython/Utility/CythonFunction.c
```diff
@@ -562,9 +568,68 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
...
+    if (!dict) {
+        dict = PyDict_New();
+        if (unlikely(!dict)) return -1;
+        op->func_dict = dict;
+    }
+    Py_INCREF(dict);
+    result = PyDict_SetItem(dict, PYIDENT("__annotate__"), value);
+    Py_DECREF(dict);
+#else
+    dict = __Pyx_PyObject_GetAttrStr(op_in, PYIDENT("__dict__"));
```
scoder: `PyObject_GenericGetDict()` ?

Cython/Utility/CythonFunction.c
JelleZijlstra: Ah thanks, I was looking for a SETREF equivalent in Cython

Cython/Utility/CythonFunction.c
JelleZijlstra: Done

Cython/Utility/CythonFunction.c
```diff
@@ -562,9 +568,68 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
...
+    }
+    Py_INCREF(dict);
+    result = PyDict_SetItem(dict, PYIDENT("__annotate__"), value);
+    Py_DECREF(dict);
+#else
+    dict = PyObject_GenericGetDict(op_in, NULL);
+    if (unlikely(!dict)) return -1;
+    result = PyDict_SetItem(dict, PYIDENT("__annotate__"), value);
+    Py_DECREF(dict);
+#endif
```
scoder: Together with the Limited API guard, this now matches the code of the `CyFunction.__dict__` implementations.
```suggestion
static int
__Pyx_CyFunction_set_annotate_in_dict(PyObject *op_in, PyObject *value) {
    PyObject *dict;
    int result;
#if CYTHON_COMPILING_IN_LIMITED_API && __PYX_LIMITED_VERSION_HEX < 0x030A0000
    dict = __Pyx_CyFunction_get_dict(__Pyx_as_CyFunctionObject(op_in), NULL);
#else
    dict = PyObject_GenericGetDict(op_in, NULL);
#endif
    if (unlikely(!dict)) return -1;
    result = PyDict_SetItem(dict, PYIDENT("__annotate__"), value);
    Py_DECREF(dict);
```

Cython/Utility/CythonFunction.c
```diff
@@ -547,6 +548,11 @@ __Pyx_CyFunction_get_kwdefaults(PyObject *op, void *context) {
     return result;
 }
 
+static int __Pyx_CyFunction_get_dict_if_exists(PyObject *op_in, PyObject **dict); /*proto*/
+static int __Pyx_CyFunction_get_annotate_from_dict_if_exists(PyObject *op_in, PyObject **annotate); /*proto*/
+static int __Pyx_CyFunction_set_annotate_in_dict_if_exists(PyObject *op_in, PyObject *value); /*proto*/
+static int __Pyx_CyFunction_set_annotate_in_dict(PyObject *op_in, PyObject *value); /*proto*/
```
scoder: I think we only need one of these now, the rest is used in order.
```suggestion
static int __Pyx_CyFunction_set_annotate_in_dict_if_exists(PyObject *op_in, PyObject *value); /*proto*/
```

Cython/Utility/CythonFunction.c
```diff
@@ -562,9 +565,59 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
...
+    if (unlikely(__Pyx_CyFunction_set_annotate_in_dict_if_exists(op_in, Py_None) < 0)) return -1;
     return 0;
 }
 
+static int
+__Pyx_CyFunction_get_dict_if_exists(PyObject *op_in, PyObject **dict) {
+    /* Return 1 if the function dict exists, 0 otherwise.  This cannot fail:
+     * _PyObject_GetDictPtr() may clear errors internally, but never reports them. */
+    *dict = NULL;
+#if CYTHON_COMPILING_IN_LIMITED_API || PY_VERSION_HEX < 0x030C0000
```
da-woods:
```suggestion
#if CYTHON_COMPILING_IN_PYPY
    *dict = PyObject_GenericGetDict(op_in, NULL);
#elif CYTHON_COMPILING_IN_LIMITED_API || PY_VERSION_HEX < 0x030C0000
```

It looks to me like PyPy is a bit of a mess with their various dict getting functions. They don't seem to use `__dictoffset__` and `_PyObject_GetDictPtr` doesn't seem to get the same thing as `PyObject_GenericGetDict`.

Suggestion above will work, but will do an unnecessary instantiation. I haven't managed to find anything better, so I suggest that we live with it for PyPy.

(I'll send PyPy a bug report on this)

Cython/Utility/CythonFunction.c
```diff
@@ -562,9 +565,59 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
...
     __Pyx_END_CRITICAL_SECTION();
+    if (unlikely(__Pyx_CyFunction_set_annotate_in_dict_if_exists(op_in, Py_None) < 0)) return -1;
     return 0;
 }
 
+static int
+__Pyx_CyFunction_get_dict_if_exists(PyObject *op_in, PyObject **dict) {
+    /* Return 1 if the function dict exists, 0 otherwise.  This cannot fail:
+     * _PyObject_GetDictPtr() may clear errors internally, but never reports them. */
+    *dict = NULL;
```
scoder:
```suggestion
```

Cython/Utility/CythonFunction.c
```diff
@@ -562,9 +565,59 @@ __Pyx_CyFunction_set_annotations(PyObject *op_in, PyObject* value, void *context
...
+#if CYTHON_COMPILING_IN_LIMITED_API || PY_VERSION_HEX < 0x030C0000
+    *dict = __Pyx_as_CyFunctionObject(op_in)->func_dict;
+#else
+    PyObject **dictptr = _PyObject_GetDictPtr(op_in);
+    if (unlikely(!dictptr)) return 0;
+    *dict = *dictptr;
+#endif
```
scoder, line 583:
```suggestion
#else
    PyObject **dictptr = _PyObject_GetDictPtr(op_in);
    *dict = likely(dictptr) ? *dictptr : NULL;
#endif
```

da-woods:
For reference the PyPy dict bug is https://github.com/pypy/pypy/issues/5515.  It looks like PyPy always instantiates a dict on class creation anyway, so there's nothing to be gained by trying not to instantiate it (therefore I think the workaround here is fine)

PR #7770, opened 2026-06-26
Fix __dealloc__ refcount guard resurrecting dying objects on free-threading (GH-7769)

### LINKED ISSUE #7769: [BUG] `__dealloc__` refcount guard can resurrect a dying object on free-threaded builds
### Describe the bug

The guard Cython emits around a user `__dealloc__` (`ModuleNode.generate_usr_dealloc_call`) marks the dying object as **live to other threads** on free-threaded builds, so a concurrent `PyUnstable_TryIncRef()` succeeds on an object already being deallocated → use-after-free (aborts with `Py_SET_REFCNT: Assertion 'refcnt >= 0' failed`).

```c
Py_SET_REFCNT(o, Py_REFCNT(o) + 1);   // guard
__pyx_pw_..._1__dealloc__(o);
Py_SET_REFCNT(o, Py_REFCNT(o) - 1);
```

At `tp_dealloc` the object is unowned/merged-zero (`ob_ref_shared == _Py_REF_MERGED`), where `TryIncRef` correctly refuses. But `Py_SET_REFCNT(o, 1)` on an unowned object takes the unowned branch — `ob_ref_shared = _Py_REF_SHARED(1, _Py_REF_MERGED) == 7` — i.e. "merged, **shared count 1**", indistinguishable from a live object, so `TryIncRef` now succeeds for the whole `__dealloc__`. The non-atomic unbump then clobbers the racing thread's increment.

```
                ob_ref_shared   TryIncRef
entry           3 (merged,0)    refused
after  +1 bump  7 (merged,1)    SUCCEEDS   <-- looks alive to other threads
racing TryIncRef 11 (merged,2)
after  -1 bump  3 (merged,0)    (clobbers the +4)  -> freed; other thread holds a dangling ref
```

`TryIncRef` is correct — a merged shared count of 1 is the legitimate encoding of a live unowned object; the bug is the guard writing it onto a dying one. Only `__dealloc__` is affected (`__del__`/`tp_finalize` has no guard).

`cythonize -i bug.pyx` on `python3.14t`, then `python -c "import bug; bug.run()"` → aborts within a few thousand iterations on an unfixed build; `PYTHON_GIL=1` never faults.

```cython
# cython: freethreading_compatible=True
cdef extern from *:
    """
    #include <stdatomic.h>
    static _Atomic(PyObject*) slot;
    static atomic_long caught;
    static void publish(PyObject *o){ atomic_store(&slot, o); }
    static void enable(PyObject *o){ PyUnstable_EnableTryIncRef(o); }
    static void consume(int (*dying)(PyObject*)){
        PyObject *o = atomic_load(&slot);
        if (o && PyUnstable_TryIncRef(o)) {      /* borrowed -> strong upgrade */
            if (dying(o)) atomic_fetch_add(&caught, 1);
            Py_DECREF(o);
        }
    }
    static long caught_count(void){ return atomic_load(&caught); }
    """
    void publish(object)
    void enable(object)
    void consume(int (*)(object) noexcept)
    long caught_count()

from threading import Thread

cdef int is_dying(object o) noexcept:
    return (<D>o).dying

cdef class D:
    cdef public int dying
    def __cinit__(self):
        enable(self); publish(self)          # publish a borrowed pointer
    def __dealloc__(self):
        self.dying = 1                       # mark mid-deallocation
        cdef int i
        for i in range(200): pass

def _consume(long n):
    cdef long i
    for i in range(n): consume(is_dying)

def run(long producer=100000, long consumer=2000000):
    t = Thread(target=_consume, args=(consumer,)); t.start()
    for _ in range(producer):
        d = D(); d = None
    t.join()
    print("TryIncRef hit an in-__dealloc__ object", caught_count(), "times")
```

### Expected behaviour

The guard should keep the object alive without making it acquirable by other threads — as CPython's own `_PyObject_ResurrectStart/End` do, by re-taking ownership and leaving the *shared* refcount at the refuse sentinel:

```c
op->ob_tid = _Py_ThreadId(); op->ob_ref_local = 1; op->ob_ref_shared = 0;  // TryIncRef refused
```

The recursion guard still holds (owner-thread inc/dec moves `ob_ref_local`). These stores need no `Py_BUILD_CORE` — available in the `CYTHON_COMPILING_IN_CPYTHON_FREETHREADING` config. Patch + test to follow as a PR.

### OS

Linux (any free-threaded build)

### Python version

3.14t / 3.13t (`Py_GIL_DISABLED`)

### Cython version

master (3.3.0a1) and 3.2.x

### Additional context

Distinct from #7391 (Limited-API symbol availability on the same lines) and #7115; neither covers this race.

cyx-6:
Fixes #7769.

On free-threaded builds the reference-count guard Cython emits around a user `__dealloc__` (`Py_SET_REFCNT(o, Py_REFCNT(o) + 1)`) marks the dying object as live to other threads: on the unowned, merged object this writes `ob_ref_shared = _Py_REF_SHARED(1, _Py_REF_MERGED)` ("merged, shared count 1"), so a concurrent `PyUnstable_TryIncRef()` succeeds and resurrects an object already in `__dealloc__` (use-after-free; aborts with `Py_SET_REFCNT: Assertion 'refcnt >= 0' failed`).

The fix keeps the object alive owner-locally — re-take ownership and leave the shared refcount at the refuse sentinel (`ob_ref_shared = 0`), mirroring CPython's `_PyObject_ResurrectStart/End` — so the keep-alive is invisible to other threads and `TryIncRef` keeps refusing. The recursion guard still holds via `ob_ref_local`. The GIL build is unchanged.

Only the non-limited free-threaded build is addressed (the only free-threaded build that currently exists). abi3t (PEP 803) cannot express this with the current stable ABI and is left for when CPython provides a stable in-dealloc resurrection primitive.

The added test reproduces the race: a `cdef class` with `__dealloc__` is published to a borrowed slot and another thread upgrades it via `PyUnstable_EnableTryIncRef`/`TryIncRef` while it is mid-`__dealloc__`. It aborts on an unfixed free-threaded build and passes with the fix; on non-free-threaded / pre-3.14 builds the consumer is a no-op so it passes trivially.

### DIFF
--- a/Cython/Utility/ExtensionTypes.c
+++ b/Cython/Utility/ExtensionTypes.c
@@ -295,6 +295,34 @@ static int __Pyx_PyType_Ready(PyTypeObject *t) {
 #define __Pyx_TRASHCAN_END
 #endif
 
+/////////////// DeallocKeepAlive.proto ///////////////
+
+// Keep `o` alive across a call into user __dealloc__ code, to prevent the user
+// code from triggering recursive deallocation, then drop the temporary reference.
+//
+// On the GIL build this is a plain refcount bump.  On free-threading we must NOT
+// use Py_SET_REFCNT(o, 1): the object reaching tp_dealloc is unowned and merged
+// (ob_ref_shared == _Py_REF_MERGED), so Py_SET_REFCNT takes the unowned branch and
+// writes ob_ref_shared = _Py_REF_SHARED(1, _Py_REF_MERGED), i.e. "merged, shared
+// count 1" -- indistinguishable from a live object.  A concurrent
+// PyUnstable_TryIncRef() would then succeed and resurrect the dying object
+// (use-after-free).  Instead we mirror CPython's _PyObject_ResurrectStart/End:
+// re-take ownership and keep the *shared* refcount at the refuse sentinel (0), so
+// the bump is invisible to other threads and TryIncRef keeps refusing.
+
+#if CYTHON_COMPILING_IN_CPYTHON_FREETHREADING
+#define __Pyx_DeallocKeepAliveBegin(o) do {                                  \
+        _Py_atomic_store_uintptr_relaxed(&(o)->ob_tid, _Py_ThreadId());      \
+        _Py_atomic_store_uint32_relaxed(&(o)->ob_ref_local, 1);              \
+        _Py_atomic_store_ssize_relaxed(&(o)->ob_ref_shared, 0);             \
+    } while (0)
+#define __Pyx_DeallocKeepAliveEnd(o) \
+        _Py_atomic_store_uint32_relaxed(&(o)->ob_ref_local, 0)
+#else
+#define __Pyx_DeallocKeepAliveBegin(o) Py_SET_REFCNT(o, Py_REFCNT(o) + 1)
+#define __Pyx_DeallocKeepAliveEnd(o)   Py_SET_REFCNT(o, Py_REFCNT(o) - 1)
+#endif
+
 /////////////// CallNextTpDealloc.proto ///////////////
 
 static void __Pyx_call_next_tp_dealloc(PyObject* obj, destructor current_tp_dealloc);
--- a/tests/run/freethreaded_dealloc_resurrect_T7769.pyx
+++ b/tests/run/freethreaded_dealloc_resurrect_T7769.pyx
@@ -13,45 +13,93 @@
 # has already started.  The fix keeps the object alive owner-locally, leaving the
 # shared refcount at the refuse sentinel so cross-thread TryIncRef keeps failing.
 #
-# This test publishes each instance to a borrowed slot and, from another thread,
-# repeatedly upgrades that borrowed pointer with the documented
-# PyUnstable_EnableTryIncRef/PyUnstable_TryIncRef protocol.  If a TryIncRef ever
-# succeeds on an object that is already in __dealloc__, the run is buggy (and an
-# unfixed free-threaded build also aborts with
-# "Py_SET_REFCNT: Assertion `refcnt >= 0' failed").  The PyUnstable_TryIncRef API
-# exists only on CPython 3.14+, and the race only exists on free-threaded builds,
-# so the consumer is a no-op elsewhere and the test then trivially passes.
+# To detect this deterministically (rather than relying on timing luck), a
+# dedicated probe thread does a single cross-thread PyUnstable_TryIncRef each time
+# an object enters __dealloc__: __dealloc__ hands "self" to the probe thread and
+# blocks until the probe reports back.  A separate thread is required -- a
+# same-thread TryIncRef would take the owner fast path and succeed even with the
+# fix, so it could not tell buggy from fixed.  If the probe ever acquires a strong
+# reference to an object that is mid-__dealloc__, the guard resurrected it.
+#
+# PyUnstable_TryIncRef exists only on CPython 3.14+ and the race only exists on
+# free-threaded builds, so on every other build the helpers are no-ops and the
+# test passes trivially.
 
 cdef extern from *:
     """
     #if PY_VERSION_HEX >= 0x030E0000 && defined(Py_GIL_DISABLED)
     #include <stdatomic.h>
-    static _Atomic(PyObject*) _ka_slot = NULL;
-    static atomic_long _ka_caught = 0;
-    static void _ka_publish(PyObject *o) { atomic_store(&_ka_slot, o); }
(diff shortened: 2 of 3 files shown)


tests/run/freethreaded_dealloc_resurrect_T7769.pyx
```diff
@@ -0,0 +1,96 @@
...
+        _ka_enable(self)
+        _ka_publish(self)
+
+    def __dealloc__(self):
+        # Mark the object as mid-deallocation, then widen the window a little so
+        # the consumer thread has a chance to race the guard.
+        self.dying = 1
+        cdef int i
+        for i in range(200):
+            pass
```
da-woods: Would `thrd_sleep` work (if you're using C11 anyway)? You could add a longer wait which would make the test more robust, and it'd be less likely to be optimized out.

tests/run/freethreaded_dealloc_resurrect_T7769.pyx
```diff
@@ -0,0 +1,96 @@
...
+# unfixed free-threaded build also aborts with
+# "Py_SET_REFCNT: Assertion `refcnt >= 0' failed").  The PyUnstable_TryIncRef API
+# exists only on CPython 3.14+, and the race only exists on free-threaded builds,
+# so the consumer is a no-op elsewhere and the test then trivially passes.
+
+cdef extern from *:
+    """
+    #if PY_VERSION_HEX >= 0x030E0000 && defined(Py_GIL_DISABLED)
+    #include <stdatomic.h>
+    static _Atomic(PyObject*) _ka_slot = NULL;
```
da-woods: The borrowed reference bit of this test feels a bit dubious to me.  I think it's relying fairly strongly on the fact that the GC never ends up called in the specific loops that you have and I don't see why that's really guaranteed (although I do see why it doesn't happen in practice). Because I think if the GC got called then `PyUnstable_TryIncRef` would no longer be a valid thing to do on the borrowed reference.

I'm not quite sure what I'd propose doing about this though. I think it would be possible to write the test with swapping a list element, but it might need more luck with the timing to get it to work reliably?

Cython/Utility/ExtensionTypes.c
```diff
@@ -295,6 +295,34 @@ static int __Pyx_PyType_Ready(PyTypeObject *t) {
...
+// (ob_ref_shared == _Py_REF_MERGED), so Py_SET_REFCNT takes the unowned branch and
+// writes ob_ref_shared = _Py_REF_SHARED(1, _Py_REF_MERGED), i.e. "merged, shared
+// count 1" -- indistinguishable from a live object.  A concurrent
+// PyUnstable_TryIncRef() would then succeed and resurrect the dying object
+// (use-after-free).  Instead we mirror CPython's _PyObject_ResurrectStart/End:
+// re-take ownership and keep the *shared* refcount at the refuse sentinel (0), so
+// the bump is invisible to other threads and TryIncRef keeps refusing.
+
+#if CYTHON_COMPILING_IN_CPYTHON_FREETHREADING
+#define __Pyx_DeallocKeepAliveBegin(o) do {                                  \
```
da-woods, line 314: This bit all looks right to me.

scoder: One issue is that it copies CPython's implementations details rather deeply. That's fine as a fix for now, but there should (eventually, hopefully) be an official way in FT-CPython to do this. We shouldn't have to do it manually here.

tests/run/freethreaded_dealloc_resurrect_T7769.pyx
cyx-6: Yes, these are important concerns. 

I have refactored to a handshake for deterministic reproduction: `__dealloc__` blocks until a probe thread has done one cross-thread `PyUnstable_TryIncRef`. Since `__dealloc__` runs before `tp_free`, the object is still alive for that read, so `PyUnstable_TryIncRef` hits the exact in-dealloc part.

As for the list-swapping, it is subtle that `l[0]` uses `_Py_TryIncrefCompare` (rechecks the slot and backs out on a swap), so it can't ever hand back a dying object. A bare `PyUnstable_TryIncRef` is exactly what we use in our downstream project, and the only interface we know to safely check a dying object state across threads.

tests/run/freethreaded_dealloc_resurrect_T7769.pyx
cyx-6: reworked as there is no such dummy loop now

da-woods (approved):
Looks good to me (assuming the tests pass)

PR #7773, opened 2026-06-27
Docs: Add section about type inference

matusvalo:


da-woods:
I think this needs a sentence to make it clear why users care about type inference. Something like

> If Cython is able to infer a specific type then it may be able to generate faster code based on that type, in the same way that it can generate faster code when you manually tell it the type of an object.

matusvalo:
> I think this needs a sentence to make it clear why users care about type inference. Something like
> 
> > If Cython is able to infer a specific type then it may be able to generate faster code based on that type, in the same way that it can generate faster code when you manually tell it the type of an object.

Thanks! Included your proposal.

docs/src/userguide/language_basics.rst
```diff
@@ -549,6 +549,71 @@ can group them into a :keyword:`cdef` block like this:
...
+
+.. tabs::
+    .. group-tab:: Pure Python
+
+        .. code-block:: python
+
+            @cython.cfunc
+            def add_multiply(i: cython.int, j: cython.int, k: cython.int) -> cython.int:
+                x = i + j           # x is Python object
+                return x * j        # not optimized
```
```suggestion
                x = i + j           # x is a Python 'int' object
                return x * j        # evaluated as object operation
```

docs/src/userguide/language_basics.rst
```diff
@@ -549,6 +549,71 @@ can group them into a :keyword:`cdef` block like this:
...
+
+    .. group-tab:: Cython
+
+        .. code-block:: cython
+
+            cdef int add_multiply(int i, int j, int k):
+                x = i + j           # x is Python object
+                return x * j        # not optimized
```
```suggestion
                x = i + j           # x is a Python 'int' object
                return x * j        # evaluated as object operation
```

scoder (approved):

PR #7774, opened 2026-06-27
Fix use of uninitialized monitoring fields in generator/coroutine objects

GeorgOstrovski:
`__Pyx__Coroutine_NewInit()` does not initialize `monitoring_states` and `monitoring_version` on the generator object. Since `PyObject_GC_New()` does not zero memory, these contain garbage when first read by `PyMonitoring_EnterScope()` via `__Pyx__TraceStartGen()`.

The fix zero-initializes both fields at construction, consistent with how `__Pyx_TraceStartFunc` already handles regular functions.

### DIFF
--- a/Cython/Utility/Coroutine.c
+++ b/Cython/Utility/Coroutine.c
@@ -1676,6 +1676,10 @@ static __pyx_CoroutineObject *__Pyx__Coroutine_NewInit(
     Py_XINCREF(code);
     gen->gi_code = code;
     gen->gi_frame = NULL;
+#if CYTHON_USE_SYS_MONITORING && (CYTHON_PROFILE || CYTHON_TRACE)
+    memset(gen->$monitoring_states_cname, 0, sizeof(gen->$monitoring_states_cname));
+    gen->$monitoring_version_cname = 0;
+#endif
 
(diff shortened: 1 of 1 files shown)


da-woods:
That definitely looks reasonable. I'm slightly puzzled why tests/run/line_profile_test.srctree isn't already broken (since it at least superficially looks to test monitoring of generators and coroutines).

scoder:
Hmm, interesting. The documentation of [`PyObject_GC_New`](https://docs.python.org/3/c-api/gcsupport.html#c.PyObject_GC_New) refers to the documentation of [`PyObject_New()`](https://docs.python.org/3/c-api/allocation.html#c.PyObject_New), which has this note:
```
The returned memory is not guaranteed to have been completely zeroed before it was initialized.
```
It's a bit funny how it says "not completely" when the implementation rather says "not at all". :)

Thanks for the report and the fix.

da-woods:
@GeorgOstrovski there's really no need to merge in master. The overall effect was to cancel the already running tests and block the auto-merge.

GeorgOstrovski:
Apologies - I saw what seemed to be test failures (unsure if related or not) and thought that a merge may resolve these.

PR #7776, opened 2026-06-27
Make AsyncGenObject 8 bytes smaller

scoder:
160->152 bytes (-5%) on Linux AMD64

I considered bit fields because all three attributes are flags, but they don't make a difference due to struct alignment.

I also took the liberty to remove some old comments while at it that simply refer to attributes being implemented differently in CPython and Cython – useless information at this point.

### DIFF
--- a/Cython/Utility/AsyncGen.c
+++ b/Cython/Utility/AsyncGen.c
@@ -45,12 +45,13 @@ if (likely(__pyx_AsyncGen_init($module_cname) == 0)); else
 //@requires: Coroutine.c::Coroutine
 
 #define __Pyx_AsyncGen_USED
+
 typedef struct {
     __pyx_CoroutineObject coro;
     PyObject *ag_finalizer;
-    int ag_hooks_inited;
-    int ag_closed;
-    int ag_running_async;
+    char ag_hooks_inited;
+    char ag_closed;
+    char ag_running_async;
 } __pyx_PyAsyncGenObject;
 
 #define __Pyx_AsyncGen_CheckExact(obj) Py_IS_TYPE(obj, CGLOBAL(__pyx_AsyncGenType))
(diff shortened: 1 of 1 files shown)


Cython/Utility/AsyncGen.c
```diff
@@ -346,22 +347,24 @@ __Pyx_async_gen_self_method(PyObject *g, PyObject *arg) {
...
      PyDoc_STR("qualified name of the async generator"), 0},
-    //REMOVED: {(char*) "ag_await", (getter)coro_get_cr_await, NULL,
-    //REMOVED:  (char*) PyDoc_STR("object being awaited on, or None")},
+    {"ag_running", (getter)__Pyx_async_gen_get_running, NULL, NULL, 0},
     {0, 0, 0, 0, 0} /* Sentinel */
 };
 
 static PyMemberDef __Pyx_async_gen_memberlist[] = {
     //REMOVED: {(char*) "ag_frame",   T_OBJECT, offsetof(__pyx_PyAsyncGenObject, ag_frame),   READONLY},
-    {"ag_running", T_BOOL,   offsetof(__pyx_PyAsyncGenObject, ag_running_async), READONLY, NULL},
```
da-woods, line 362: I think this is still fine. The documentation of `T_BOOL` says that it should point to a `char` anyway. Unless there's good reason to want to implement it as a function?

scoder: Ha, right. I had actually tried bit fields and various field reorderings and this was a left-over. I'll revert that part.

scoder: … we might be fixing a Schrödingbug on big endian machines here then, right?

EDIT: The tests only use the flag attribute once, checking that it's 0 when it's _not_ running, which doesn't preclude the case that it's _always_ 0.

https://github.com/cython/cython/blob/072985c5cf33d0824853602be063e08e9821b203/tests/run/test_asyncgen.py#L531

scoder: I'll use the function implementation as an untested fix for 3.2.x because I noticed that there are a lot of issues in master that need proper 3.3-level fixing following CPython.

3.2.x fix is in https://github.com/cython/cython/commit/86ba58478a62ed8c3ef50ff402a673e86318fa18
PR for 3.3 is https://github.com/cython/cython/pull/7777

da-woods: IBM do provide free Github actions runners for s390x: https://github.com/IBM/actionspz/

I have no idea how much of an effort it is to set up, or how quick it runs, (or anything else). But if we wanted to add a big-endian test then it's probably the easiest way of doing it.  I don't know of any other semi-commonly used big-endian systems.

da-woods: (I'm not really suggesting we should do add s390x tests. Just saying "it's there")

scoder: We've had a few big-endian issues in the past and only became aware of them because someone else noticed. We should notice them ourselves. I've asked for a runner.

https://github.com/IBM/actionspz/issues/121

PR #7777, opened 2026-06-28
Fix AsyncGenerator following CPython

scoder:
* Fix ".close()" of async generators' asend/athrow to properly raise a GeneratorExit.
* Fix ".close()" and ".throw()" to block access when the generator is running.
* Set ".ag_running" during ".close()".
* Copy the related tests from CPython.
* Add a complete test for the ".ag_running" flag.

Follows https://github.com/python/cpython/issues/117714
Follows https://github.com/python/cpython/issues/117881


scoder:
This is really mostly copying (and slightly adapting) code from CPython to fix new tests, so should be fine in general.

Issue #7729, opened 2026-05-30
[BUG] `list[cython.int]` generates working but bad code


da-woods:
### Describe the bug

I assume the idea of `list[cython.int]` is that the result of indexing it is a C int. 

### Code to reproduce the behaviour:

```cython
import cython

def f(l: list[cython.int]):
    l0 = l[0]
    print(l0*2)
```

This generates the C code:

```c
__pyx_t_1 = __Pyx_GetItemInt_List(__pyx_v_l, 0, long, 1, __Pyx_PyLong_From_long, 0, 1, 1, __Pyx_ReferenceSharing_FunctionArgument); if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 4, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_1);
  __pyx_t_2 = __Pyx_PyLong_As_int(__pyx_t_1); if (unlikely((__pyx_t_2 == (int)-1) && PyErr_Occurred())) __PYX_ERR(0, 4, __pyx_L1_error)
  __Pyx_DECREF(__pyx_t_1); __pyx_t_1 = 0;
  __pyx_t_1 = __Pyx_PyLong_From_int(__pyx_t_2); if (unlikely(!__pyx_t_1)) __PYX_ERR(0, 4, __pyx_L1_error)
  __Pyx_GOTREF(__pyx_t_1);

  if (__Pyx_PyInt_FromNumber(&__pyx_t_1, NULL, 0) < (0)) __PYX_ERR(0, 4, __pyx_L1_error)
  __pyx_v_l0 = ((PyObject*)__pyx_t_1);
  __pyx_t_1 = 0;
```

* `l0` appears to be inferred as a python int (probably... it might just be a PyObject).
* The list is index to produce a Python object.
* The Python object is converted to a C int.
* The C int is converted back to to a Python int.
* For some reason `__Pyx_PyInt_FromNumber` is called to do a bit more work on the Python int.

### Expected behaviour

Something efficient, either:
* `l0` inferred as `cython.int`, one conversion
* `l0` inferred as a Python object, double conversion via C int optimized out.

In this case the first version is better, but in other cases the second version might be better.

### OS

Linux

### Python version

3.13

### Cython version

Current master

### Additional context

I'm not completely convinced that `list[cython.int]` is a sensible thing to support because I think it may mislead users about what they're storing. But that's probably separate

scoder:
It's marked as unsafe because you are multiplying by 2.

[...]

PR #7780, opened 2026-06-28
Check the type of the NextNode.

### LINKED ISSUE #7775: [BUG] Segfault passing non-int when annotation is list[int]
### Describe the bug

There is a change in Cython 3.3.0a1 that leads to a segfault with the example code

```cython
def check(N: list[int]):
    return any(n < 0 for n in N)

class A:
    pass

check([A()])
```
Result with Cython 3.3.0a1:
```
>>> import pyximport; pyximport.install()
(None, <pyximport.pyximport.PyxImportMetaFinder object at 0x7238f521d400>)
>>> import check
Segmentation fault (core dumped)
```

### Expected behaviour

Previous Cython versions do not segfault and allow other types to pass through even though the annotation says int. With python-flint's fmpz integer type the code would still work.

Possibly it should fail with an exception if the type is not int?

### OS

Linux x86-64

### Python version

3.13.1

### Cython version

3.3.0a1

### Additional context

This popped up in CI in https://github.com/flintlib/python-flint/pull/405. The original problem was not a segfault but rather test failures like this:
```
___________________ test_docstrings[flint.types.fmpq_mpoly.fmpq_mpoly.deflation] ____________________
flint.types.fmpq_mpoly.fmpq_mpoly.deflation, line: 1033
Failed example:
	q.inflate(N) == f
Exception raised:
      File "/.../uv/python/cpython-3.13.1-linux-x86_64-gnu/lib/python3.13/doctest.py", line 1395, in __run
        exec(compile(example.source, filename, "single",
        ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                     compileflags, True), test.globs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "<doctest flint.types.fmpq_mpoly.fmpq_mpoly.deflation[5]>", line 1, in <module>
        q.inflate(N) == f
        ~~~~~~~~~^^^
      File "flint/types/fmpq_mpoly.pyx", line 997, in flint.types.fmpq_mpoly.fmpq_mpoly.inflate
    ValueError: all inflate strides must be non-negative
```
By the time I got down to a minimal reproducer it was a segfault so I'm not sure if that segfault is actually the same as the original problem I was investigating.

matusvalo:
This ensure correctness of variable used in iteration loop.

Fixes #7775


tests/run/pep526_variable_annotations.py
```diff
@@ -479,6 +479,34 @@ def test_iteration_over_frozendict_with_subscript():
...
+    """
+    >>> test_generator_next_node_coercion([A()])  # doctest:+IGNORE_EXCEPTION_DETAIL
+    Traceback (most recent call last):
+    ...
+    TypeError: '<' not supported between instances of 'A' and 'int'
+    """
+    return any(n < 0 for n in N)
+
+
+def test_iterator_next_node_coercion(N: list[int]):
```
matusvalo: Not sure whether we need to test both cases, since they use the same code path...

tests/run/pep526_variable_annotations.py
```diff
@@ -603,6 +479,34 @@ def test_iteration_over_frozendict_with_subscript():
...
+
+
+def test_generator_next_node_coercion(N: list[int]):
+    """
+    >>> test_generator_next_node_coercion([A()])  # doctest:+IGNORE_EXCEPTION_DETAIL
+    Traceback (most recent call last):
+    ...
+    TypeError: '<' not supported between instances of 'A' and 'int'
+    """
+    return any(n < 0 for n in N)
```
da-woods: I think the exception detail is a bit confusing here and it might be clearer to omit it and just write `TypeError:`.

The message you show is for pure Python, but Cython checks it earlier and gives `Expected int, got A`.  I was just looking at it and thinking "surely that can't be right and Cython does something different"

tests/run/pep526_variable_annotations.py
scoder: It doesn't hurt much to test if it works in both usages. But these are not "PEP-526 variable annotation" tests (e.g. `x: int = 5`) but tests related to for-loop iteration and type inference behaviour. I would drop them into `type_inference.pyx` because they really test how mismatches on a declared item type impact the behaviour of inferred for-loop variables. Without type inference, we'd not have a problem here.

scoder (commented):
Fix looks perfect, test should be moved.

scoder:
Sorry for committing an only partially tested change here. I think the previous assumption (relying on `.equivalent_type`) was actually wrong because there is no real guarantee about the object type read from a Python iterable. That's always a plain `object` as far as compile time is concerned. Any necessary coercions must _always_ be applied to it after reading it, whether the target is a Python object type or not. Let's see what the other tests say.

matusvalo:
No problem. I think it is 99% correct. Only following case was not covered: `for e in [Exception()]:`. I tried to fix it. If CI/CD is green I will migrate the test to different place as suggested in comment above.

Cython/Compiler/ExprNodes.py
```diff
@@ -3545,10 +3545,11 @@ def infer_type(self, env, iterator_type=None):
 
     def analyse_types(self, env):
         item_type = self.infer_type(env, self.iterator.type)
-        if self.iterator.type.is_pyobject and not item_type.is_pyobject:
-            # We definitely read a Python object from the iterable but inferred a C type for it,
-            # probably by anticipating to unpack it.  Do the coercion outside to allow undoing it later.
-            self.type = item_type.equivalent_type or py_object_type
+        if self.iterator.type.is_pyobject and \
+                not (item_type.is_builtin_type and item_type.name == 'type'):
```
scoder: Hmm. You could check `item_type is type_type` here, but I actually wonder why there should be a special case like this. In the concrete failing test, we are iterating over a sequence of exception _instances_. Why should `type_type` as `item_type` be the correct result of the iterator type inference here? Shouldn't it infer a concrete exception type instead?

matusvalo: > You could check item_type is type_type her

I saw this check elsewhere in cython codebase so I use it too. But I have no problem rewriting it.

>  Why should type_type as item_type be the correct result of the iterator type inference here? Shouldn't it infer a concrete exception type instead?

Honestly? No idea. Before adding this condition cython adds `PyType_CheckExact()` condition which clearly fails. Hence, avoiding coercion for pytype fixes generated C code. And clearly now CI/CD is green except Graalpy which I don't know why is failing.

Maybe there is deeper issue, but I cannot tell with my limited knowledge. Is somehow #7783 related to this PR?

scoder: > Is somehow https://github.com/cython/cython/pull/7783 related to this PR?

I'm hoping that it will help to improve the type inference situation for exceptions, but it is not a fix for the problem here.

scoder:
The problem surfaces here:
https://github.com/cython/cython/blob/ef6af38d1c07810a88a8fdf76dc29e5743e0553d/Cython/Compiler/ExprNodes.py#L6237-L6249
The last `return` statement returns the function entry's type, which in this case is `type`. There's something fishy about it in that it can sometimes return `type` and sometimes the concrete builtin's type, for the same builtin, depending on the usage of that builtin. For `range()`, for example, I get `type` when called as iterator in a for-loop but `range object` when called separately. So there's something going wrong in the called function's entry assignment, somewhere earlier. It should always select the entry of the concrete type whenever it's called, not the generic `type` type of the name.

matusvalo:
Hmm I see. Here is explicit example and the output of debug print of `result_type`:

```cython
cdef float main():
    for e in [ValueError(3)]:
        pass

print(main())
```


```
(.venv313) matus@homepc:~/dev/cython/test$ cython a.pyx
(<FileSourceDescriptor:/home/dev/cython/test/a.pyx>, 18, 24) ValueError object Entry(<7f3f30443a10>, name=ValueError, type=ValueError object)
(<FileSourceDescriptor:/home/dev/cython/test/a.pyx>, 18, 24) ValueError object Entry(<7f3f30443a10>, name=ValueError, type=ValueError object)
(<FileSourceDescriptor:/home/dev/cython/test/a.pyx>, 18, 24) ValueError object Entry(<7f3f30443a10>, name=ValueError, type=ValueError object)
(<FileSourceDescriptor:/home/dev/cython/test/a.pyx>, 18, 24) ValueError object Entry(<7f3f30443a10>, name=ValueError, type=ValueError object)
(<FileSourceDescriptor:/home/dev/cython/test/a.pyx>, 18, 24) ValueError object Entry(<7f3f30443a10>, name=ValueError, type=ValueError object)
(<FileSourceDescriptor:/home/dev/cython/test/a.pyx>, 18, 24) ValueError object Entry(<7f3f30443a10>, name=ValueError, type=ValueError object)
(<FileSourceDescriptor:/home/dev/cython/test/a.pyx>, 18, 24) type object Entry(<7f3f30443af0>, name=ValueError, type=type object)
(<FileSourceDescriptor:/home/dev/cython/test/a.pyx>, 18, 24) type object Entry(<7f3f30443af0>, name=ValueError, type=type object)
(<FileSourceDescriptor:/home/dev/cython/test/a.pyx>, 18, 24) type object Entry(<7f3f30443af0>, name=ValueError, type=type object)
```

One observiation: There are two different entries (one `type` object and other one `ValueError object` as mentioned above)

UPDATE: OK I dig out stacktraces where the entries were created:

ValueError object:
```python
  /home/dev/cython/.venv313/bin/cython(10)<module>()
-> sys.exit(setuptools_main())
  /home/dev/cython/Cython/Compiler/Main.py(821)setuptools_main()
-> return main(command_line = 1)
  /home/dev/cython/Cython/Compiler/Main.py(849)main()
-> result = compile(sources, options)
  /home/dev/cython/Cython/Compiler/Main.py(740)compile()
-> return compile_multiple(source, options, cache)
  /home/dev/cython/Cython/Compiler/Main.py(696)compile_multiple()
-> context = Context.from_options(options)
  /home/dev/cython/Cython/Compiler/Main.py(94)from_options()
-> return cls(options.include_path, options.compiler_directives,
  /home/dev/cython/Cython/Compiler/Main.py(71)__init__()
-> from . import Builtin, CythonScope
  <frozen importlib._bootstrap>(1415)_handle_fromlist()
  <frozen importlib._bootstrap>(488)_call_with_frames_removed()
  <frozen importlib._bootstrap>(1360)_find_and_load()
  <frozen importlib._bootstrap>(1331)_find_and_load_unlocked()
  <frozen importlib._bootstrap>(935)_load_unlocked()
  <frozen importlib._bootstrap_external>(1026)exec_module()
  <frozen importlib._bootstrap>(488)_call_with_frames_removed()
  /home/dev/cython/Cython/Compiler/Builtin.py(900)<module>()
-> init_builtins()
  /home/dev/cython/Cython/Compiler/Builtin.py(850)init_builtins()
-> init_builtin_exceptions()
  /home/dev/cython/Cython/Compiler/Builtin.py(833)init_builtin_exceptions()
-> builtin_types[name] = builtin_scope.declare_builtin_type(
  /akdjome/dev/cython/Cython/Compiler/Symtab.py(1337)declare_builtin_type()
-> entry = self.declare_type(name, type, None, visibility='extern')
  /home/dev/cython/Cython/Compiler/Symtab.py(657)declare_type()
-> entry = self.declare(name, cname, type, pos, visibility, shadow,
  /home/dev/cython/Cython/Compiler/Symtab.py(616)declare()
-> entry = Entry(name, cname, type, pos = pos)
  /home/dev/cython/Cython/Compiler/Symtab.py(241)__init__()
```

type object:

```python
  /home/dev/cython/.venv313/bin/cython(10)<module>()
-> sys.exit(setuptools_main())
  /home/dev/cython/Cython/Compiler/Main.py(821)setuptools_main()
-> return main(command_line = 1)
  /home/dev/cython/Cython/Compiler/Main.py(849)main()
-> result = compile(sources, options)
  /home/dev/cython/Cython/Compiler/Main.py(740)compile()
-> return compile_multiple(source, options, cache)
  /home/dev/cython/Cython/Compiler/Main.py(696)compile_multiple()
-> context = Context.from_options(options)
  /home/dev/cython/Cython/Compiler/Main.py(94)from_options()
-> return cls(options.include_path, options.compiler_directives,
  /home/dev/cython/Cython/Compiler/Main.py(71)__init__()
-> from . import Builtin, CythonScope
  <frozen importlib._bootstrap>(1415)_handle_fromlist()
  <frozen importlib._bootstrap>(488)_call_with_frames_removed()
  <frozen importlib._bootstrap>(1360)_find_and_load()
  <frozen importlib._bootstrap>(1331)_find_and_load_unlocked()
  <frozen importlib._bootstrap>(935)_load_unlocked()
  <frozen importlib._bootstrap_external>(1026)exec_module()
  <frozen importlib._bootstrap>(488)_call_with_frames_removed()
  /home/dev/cython/Cython/Compiler/Builtin.py(900)<module>()
-> init_builtins()
  /home/dev/cython/Cython/Compiler/Builtin.py(850)init_builtins()
-> init_builtin_exceptions()
  /home/dev/cython/Cython/Compiler/Builtin.py(833)init_builtin_exceptions()
-> builtin_types[name] = builtin_scope.declare_builtin_type(
  /home/dev/cython/Cython/Compiler/Symtab.py(1344)declare_builtin_type()
-> var_entry = Entry(
> /home/dev/cython/Cython/Compiler/Symtab.py(241)__init__()
```

matusvalo:
OK @scoder maybe this will help you - here is the source of two different entry types:

https://github.com/cython/cython/blob/f71b67cf3678850acd1f9d29545c753e5559302c/Cython/Compiler/Symtab.py#L1334-L1346

On the line 1334, the entry is created with type `ValueError object`, on the line 1341, the entry is created with the type `type object`

matusvalo:
OK and here is answer where `function.entry` with type `ValueError object` is replaced with entry `type object`:

https://github.com/cython/cython/blob/f71b67cf3678850acd1f9d29545c753e5559302c/Cython/Compiler/ExprNodes.py#L2477-L2478

scoder:
Right, that makes sense. Then we're assigning the `var_entry` (i.e. the entry's representation when used in a variable-like context) as the call node's entry at some point where we should be assigning the normal entry.

Edit: The place that you found in ExprNodes.py seems too generic to change, though. I'm puzzled why we never had this problem with the other builtins, only with the exception types now.

tests/run/pep526_variable_annotations.py
matusvalo: Based on your previous comment here: https://github.com/cython/cython/issues/7729#issuecomment-4583410709 I moved the tests to separate file.

matusvalo:
> Right, that makes sense. Then we're assigning the `var_entry` (i.e. the entry's representation when used in a variable-like context) as the call node's entry at some point where we should be assigning the normal entry.
> 
> Edit: The place that you found in ExprNodes.py seems too generic to change, though. I'm puzzled why we never had this problem with the other builtins, only with the exception types now.

OK for now I consider this PR as done. I moved the tests as requested and that's all from my side for now...

scoder:
I think I fixed the type inference bug in https://github.com/cython/cython/pull/7848, specifically https://github.com/cython/cython/pull/7848/commits/0494f2572c17519b2f56f34bf650d4465897eb07

matusvalo:
> I think I fixed the type inference bug in #7848, specifically [0494f25](https://github.com/cython/cython/commit/0494f2572c17519b2f56f34bf650d4465897eb07)

I confirm that with in the PR #7848 following diff:

```diff
diff --git a/Cython/Compiler/ExprNodes.py b/Cython/Compiler/ExprNodes.py
index 28a203cee..a68ded1f7 100644
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -3548,10 +3548,10 @@ class NextNode(AtomicExprNode):

     def analyse_types(self, env):
         item_type = self.infer_type(env, self.iterator.type)
-        if self.iterator.type.is_pyobject and not item_type.is_pyobject:
+        if self.iterator.type.is_pyobject:
             # We definitely read a Python object from the iterable but inferred a C type for it,
             # probably by anticipating to unpack it.  Do the coercion outside to allow undoing it later.
-            self.type = item_type.equivalent_type or py_object_type
+            self.type = py_object_type
             return self.coerce_to(item_type, env)
         else:
             self.type = item_type
```

passes both test added in this PR + test `test_exceptions` which previously was failing.

PR #7782, opened 2026-06-30
Fix cdef function optional argument captured by closures not moved into the closure scope

donlon:
### Minimal reproducible example

``` Python
cdef _queue_scan_reg(
    object length = None, # int | None
):
    print(f'length: {length}') # Segfault caused by null pointer __pyx_cur_scope->__pyx_v_length
    def _on_done():
        # NameError: free variable 'length' referenced before assignment in enclosing scope
        return length
    return _on_done()

def run():
    res = _queue_scan_reg(196883)
    print(res)
```

When a cdef function is declared with optional argument and has nested function inside that captures this optional argument, the generated code didn't move the parsed argument (`__pyx_v_length` in this case) into the closure scope (`__pyx_cur_scope->__pyx_v_length`). Further uses of this variable through `__pyx_cur_scope->__pyx_v_length`  cause errors as it's a null pointer.

This patch simply fixes this bug by removing the `arg.default` condition when generating the assignment code for the captured variables (and now it's similar to how arguments of [non-cdef functions](https://github.com/cython/cython/blob/0d6e04ce50c2a4c412dc9b58539145b64094558b/Cython/Compiler/Nodes.py#L3749) are processed).

### DIFF
--- a/tests/run/cdef_optional_arg_closure.pyx
+++ b/tests/run/cdef_optional_arg_closure.pyx
@@ -15,10 +15,12 @@ cdef test(object register=None, object length=None):
     return length, _on_done()
 
 
-cdef test_multi(object device_index, object register=2, object tdi_data=3):
-    def inner():
-        return (device_index, register, tdi_data)
-    return inner()
+def run_default():
+    """
+    >>> run_default()
+    (None, None)
+    """
+    return test(0)
 
 
 def run_passed():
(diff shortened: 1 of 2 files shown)


donlon:
Hi @da-woods, I just found that the graalpy25 run failed, but I suspect that the problem wasn't introduced by this commit. The CI run on #7780 also failed with the same errors.

The log show that the latest successful run on master branch used an older Graalpy25 (`Python 3.12.8 (Fri Apr 10 10:27:46 UTC 2026)`) while the recent failing runs are using a slightly newer (`Python 3.12.8 (Tue Jun 23 15:26:01 UTC 2026)`). The problem might be caused by the GraalPy update, which is pretty interesting...

da-woods:
GraalPy is occasionally flaky. I'll just rerun it before suspecting an issue

da-woods:
> GraalPy is occasionally flaky. I'll just rerun it before suspecting an issue

No... this is new and different test failures. I suspect you're right and that it's the new GraalPy version.

tests/run/cdef_optional_arg_closure.pyx
```diff
@@ -0,0 +1,61 @@
...
+    def _on_done():
+        return length
+    return length, _on_done()
+
+
+cdef test_multi(object device_index, object register=2, object tdi_data=3):
+    def inner():
+        return (device_index, register, tdi_data)
+    return inner()
+
```
da-woods, line 24: Could you add an example where a non-object variable gets captured, e.g.

```
cdef test_nonobj(object device_index, int register=2):
    ...
```

I suspect it'll be fine, but it's not exactly the same.

da-woods (commented):
Apart from one extra test, this looks sensible to me

donlon:
Done!

da-woods (approved):

scoder:
Thanks

scoder:
3.2.x commit is https://github.com/cython/cython/commit/dd041b83d671e668be1353b7c4c753a37ae42563

PR #7783, opened 2026-06-30
Infer exception base types

scoder:
Teaches Cython about Python's exception class hierarchy and allows it to infer a concrete exception (base) type as spanning type for two given exceptions.

Moves the long list of known builtin type names from `Code.py` (where it always felt awkward) to `PyrexTypes.py`. `Builtin.py` might seem better but introduces import order issues since the list is used in other modules that `Builtin.py` depends on. It now lives right before `BuiltinObjectType` in `PyrexTypes.py`, which seems sufficiently appropriate.

I also moved `uncachable_builtins` with it to help keeping that set in sight of the known builtins.
I wonder if `renamed_py2_builtins_map` should be moved as well. It's not going to change any more.


da-woods:
What can we usefully do with this information? (In that I haven't found too many places where we can usefully use the knowledge that something is an `(Base)Exception` because most attributes of exceptions could be overridden by base classes anyway)

scoder:
We're only dealing with builtin exceptions here, which have well-defined, known behaviour.

da-woods:
I think we currently infer the types of exceptions from `except` statements:

e.g.

```
except RuntimeError as e:
     ...
```

will infer the type of `e` as `RuntimeError` (assuming nothing else assigns to `e`).

That means that we have to treat exceptions as "`RuntimeError` or subclass" (unlike all other builtin types which are exact types). Which generally means we can't do any fast access to the internals (e.g. I looked into using things like `PyException_Set/GetTraceback` for fast attribute access but you have to assume that `__traceback__` could be overridden in base class).  If you did do anything else you'd need to distinguish between "exact exception type" and "inexact exception type"

Currently I think the only thing you could usefully do with an inferred exception type is to shortcut some exception handling where you're able to identify which one matches at compile-time.

scoder:
> I think we currently infer the types of exceptions from `except` statements

Yes, we do:
https://github.com/cython/cython/blob/b8e835a322439ff79dd03108e542576753720047/Cython/Compiler/TypeInference.py#L170-L172

We can still optimise if we find the exact type at runtime. I'll add a bit of that.

Cython/Utility/Builtins.c
```diff
@@ -931,3 +931,58 @@ static CYTHON_INLINE PyObject* __Pyx__PyFrozenDict_New(PyObject* frozendict_type
...
+
+static CYTHON_INLINE PyObject* __Pyx_Py{{NAME}}_GetArgs(PyObject *exception); /*proto*/
+
+/////////////// ExceptionGetArgs ////////////////
+//@requires: ObjectHandling.c::PyObjectGetAttrStr
+//@requires: Synchronization.c::CriticalSections
+
+static CYTHON_INLINE PyObject* __Pyx_Py{{NAME}}_GetArgs(PyObject *exception) {
+#if CYTHON_COMPILING_IN_CPYTHON
+    if (likely(Py_TYPE(exception) == (PyTypeObject*) &PyExc_{{NAME}})) {
```
da-woods: > We can still optimise if we find the exact type at runtime. I'll add a bit of that.

Yeah that does make sense for the concrete exception types. I'd thought it was practically very unlikely to hit for `Exception` but something like `KeyError` is probably almost always a `KeyError` and not a base class

scoder:
The crash in FT-Python will be fixed by https://github.com/cython/cython/pull/7793

PR #7791, opened 2026-07-02
Allow C property setters to explicitly propagate exceptions

scoder:
They previously called `PyErr_Occurred()` after each call to the `void` setter, which seems needlessly inefficient.

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -8150,10 +8150,11 @@ class AttributeNode(ExprNode):
         elif target and self.obj.type.is_builtin_type:
             error(self.pos, "Assignment to an immutable object field")
         elif self.entry and self.entry.is_cproperty:
-            if not target:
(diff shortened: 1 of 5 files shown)


da-woods (commented):
Yes - seems like a good idea to allow the setter to signal exceptions.

One thing I'm not sure of from looking at the PR is whether the following are allowed:

```
@property1.setter
cdef inline void property1(self, val) except *:
    ...

@property2.setter
cdef inline int property2(self, val) noexcept:
    ...

@property3.setter
cdef inline int property3(self, val) except -1?:
    ...

@property4.setter
cdef inline int property3(self, val) except -2:
    ...
```

It probably doesn't do any harm to allow them although it's maybe slightly pointless.

scoder:
> It probably doesn't do any harm to allow them although it's maybe slightly pointless.

Let's be strict for now and only allow `void` and `0/-1` to keep users from writing code in the wild that we won't be able to hunt down later. I would even require `noexcept` together with `void`. That keeps all future options for us and makes the code's behaviour explicit on user side.

PR #7792, opened 2026-07-02
Relax the assignment checks for builtin types to allow typedef-ed 'object'

### LINKED ISSUE #7789: [BUG] ctypedef object alias rejects string/bytes literal that plain object accepts
### Describe the bug

A type declared with `ctypedef object MyObj` is an alias for the generic Python `object` type, but since Cython 3.1.0 it no longer behaves like one: assigning or returning a literal (`str`, `bytes`, or `unicode` constant) to a `ctypedef object` target fails to compile with `Cannot assign type '… object' to 'MyObj'`, whereas the identical code targeting plain `object` compiles. Only literals are affected — a variable of the same type assigns fine. It reproduces under every `language_level` (`3`, `3str`, `2`). This compiled on Cython 3.0.x.

```cython
# cython: language_level=3

ctypedef object MyObj          # alias for the generic Python `object`

cdef object accept():
    return b"x"                # OK

cdef MyObj reject():
    return b"x"                # ERROR: Cannot assign type 'bytes object' to 'MyObj'

# Same failure with a str literal (any language_level: 3 / 3str / 2):
cdef MyObj reject_str():
    return "x"                 # ERROR: Cannot assign type 'str object' to 'MyObj'

# But a variable of the same type is accepted — only *literals* fail:
cdef MyObj ok_var():
    cdef str s = "x"
    return s                   # OK
```

### Expected behaviour

`ctypedef object MyObj` is an alias for `object`. Since a `str`/`bytes`/`unicode` value is an `object`, returning or assigning one — literal or not — to a `MyObj` target should be accepted, exactly as it is for plain `object` (and as it was in Cython ≤ 3.0.x).

### OS

Windows, Linux

### Python version

3.11, 3.14

### Cython version

3.1.0, 3.2.4

### Additional context

Regression range: fails on 3.1.0, 3.1.6, 3.2.4; works on 3.0.12 and earlier 3.0.x. Python-version- and platform-independent

scoder:
Closes https://github.com/cython/cython/issues/7789

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -1946,7 +1946,7 @@ class UnicodeNode(ConstNode):
                   "Unicode literals do not support coercion to C types other "
                   "than Py_UCS4/Py_UNICODE (for characters), Py_UNICODE* "
                   "(for strings) or char*/void* (for auto-encoded strings).")
-        elif dst_type is not py_object_type:
+        elif dst_type.resolve() is not py_object_type:
             self.check_for_coercion_error(dst_type, env, fail=True)
         return self
 
--- a/Cython/Compiler/PyrexTypes.py
+++ b/Cython/Compiler/PyrexTypes.py
@@ -335,23 +335,23 @@ class PyrexType(BaseType):
 
     @property
     def may_be_pyint_type(self) -> bool:
-        return self is py_object_type or self.is_pyint_type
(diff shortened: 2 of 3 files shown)


scoder:
3.2.x commit is https://github.com/cython/cython/commit/b8412f3503b6c8923dbdb86b2c2d7a20c64b4dbd

PR #7795, opened 2026-07-04
Build: Compile string processor modules

scoder:
Adds `StringEncoding.py` and `LZSS.py` to default compiled modules.
They are rather small (stripped ~290kB on Linux AMD64) but computationally expensive.


scoder:
I also thought it might be worth enabling `CYTHON_USE_TYPE_SPECS` for the build since it tends to reduce the module size when extension types are involved. But we don't seem to use all that many. The overall gain is somewhere around 15kB (0.3%) in my test, total across all default compiled Cython modules. But for compile-all, it's about 1.2MB out of 21MB (~5%), which seems more worth it. `Nodes.py` (4MB) and `ExprNodes.py` (6MB) are really large. Maybe we should start using the shared module in that setup…

da-woods:
> Maybe we should start using the shared module in that setup…

I had a look at that a few weeks ago. It isn't worthwhile until we've made the shared module customizable. Right now the shared module is mostly memoryviews (which we don't use ourselves). I think (but not 100% sure) we also turn off `binding` too so we may not end up sharing cyfunctions either.

scoder:
> It isn't worthwhile until we've made the shared module customizable.

Hmm, right. Even with compile-all, it's only another ~2.5% smaller, ~7.5% total. That's not nothing, minus 1.2MB vs. minus 1.67MB is a visible difference, but it probably also comes with a (slight?) reduction in performance, which I don't think is what you want when you opt into compiling all modules. Enabling type specs is clearly the larger win for now.

> I think (but not 100% sure) we also turn off binding too

We turn off several options:
https://github.com/cython/cython/blob/8f61367d8373c272456343f6b6b335750afb3fec/setup.py#L202-L209

scoder:
> We turn off several options:

Turns out, we were _trying_ to switch them off (by changing `default_directives` from `setup.py`) but werent't, at least not in parallel builds that started separate Cython processes. :-/

And we _do_ need `binding=True` because we use `@functools.wraps()` in a couple of places.

EDIT: Fixing that in https://github.com/cython/cython/pull/7801

matusvalo:
Since, we are compiling `StringEncoding.py` can we also make `EncodedString` and `BytesLiteral` as extention types?

Cython/Compiler/StringEncoding.py
```diff
@@ -277,10 +303,10 @@ def split_string_literal(s, limit=2000):
         return '""'.join(chunks)
 
 
-def encode_pyunicode_string(characters):
+def encode_pyunicode_string(string):
     """Create Py_UNICODE[] representation of a given unicode string.
     """
-    characters = list(map(ord, characters))
+    characters = list(map(ord, string))
```
matusvalo: Annotating as list of C `int`s cython optimizes the code slighly:
* numeric operations are native
* tuple construction is completely avoided
```suggestion
    characters: list[cython.int] = list(map(ord, string))
```

matusvalo: Hmm now thinking about the code, `code_point >= 0x10000` is likely to be False. Hence, I think better annotation is:

```suggestion
    characters: list[int] = list(map(ord, string))
```
This optimizes ` __Pyx_PyObject_CompareBoolGe_object_int` to ` __Pyx_PyObject_CompareBoolGe_int_int`

da-woods:
> can we also make `EncodedString` and `BytesLiteral` as extention types?

I don't think so. `str` and `bytes` are variable-sized extension types. That means that we can't generate a C struct for them because we don't know how big the base class is.

e.g. https://github.com/cython/cython/blob/master/tests/errors/builtin_type_inheritance.pyx

I think sometime in the future we could do [PEP-697 opaque inheritance](https://peps.python.org/pep-0697/) but we don't currently support that. But that would let you inherit from `str` and `bytes` (at a performance cost)

------------------------------------------

Plus inheriting an extension type from _any_ builtin type breaks the Limited API (again, fixed by opaque inheritance). But that's not the main problem.

Cython/Compiler/StringEncoding.py
scoder: You're right, we could do better here. I've rewritten the function to streamline it when compiled.

PR #7797, opened 2026-07-05
Optimise bytearray.extend(bytes)

scoder:
I constrained it to the most common case of `bytes` values for now because mutable objects like `bytearray` and `memoryview` require nested critical section guards, which is not trivial.

This PR includes 3.2.x PR https://github.com/cython/cython/pull/7796 which should be merged first.

### DIFF
--- a/Cython/Compiler/Optimize.py
+++ b/Cython/Compiler/Optimize.py
@@ -3324,12 +3324,15 @@ class OptimizeBuiltinCalls(Visitor.NodeRefCleanupMixin,
         if len(args) != 2:
             return node
 
+        func_type = self.PyByteArray_Extend_func_type
+        utility_code = UtilityCode.load_cached("ByteArrayExtendBytes", "StringTools.c")
+
         value = unwrap_coerced_node(args[1])
         arg_type = value.type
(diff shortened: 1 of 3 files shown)


da-woods:
> [...] `memoryview` require nested critical section guards, which is not trivial.

`memoryview` is (kind of) OK I think. Being held by the memoryview prevents it from being resized/reallocated.

It doesn't prevent other threads from modifying the data, but that's a less serious race in that it won't lead to out of bounds errors, and is the Python version you call won't guard against it either.

---------------------------------------------------------------------------------

There's also `Py_BEGIN_CRITICAL_SECTION2`. But I'd be a bit suspicious of that because `PyByteArray_Resize` internally acquires a single critical section. And I'm not completely sure that they're ensured that this nesting path doesn't get broken. So I probably wouldn't rely on it.

scoder:
I wrote the code so that it easily extends to other input types. Let's keep this PR as simple as it is.

scoder:
`bytearray` as input is even more complex because it allows `b.extend(b)`. CPython has a whole extra code path for that (although for us it would be easy to exclude and pass on).

PR #7798, opened 2026-07-05
Introduce support of `tuple[TYP, ...]`

matusvalo:
Allows cython to understand ellipses as tuple subscript. Currently, this PR does not allow `tuple[...]` which should be allowed but usage is limited because it means tuple with one Ellipsis element inside.

This PR also increases level of warnings to show the user when subscript types are ignored.


tests/errors/e_typing_errors.pyx
```diff
@@ -237,6 +238,22 @@ def invalid_type_count():
     dict1: dict[int]
     dict2: dict[int, int, int]
 
+def invalid_ellipsis():
```
matusvalo, line 241: Following declarations are producing warnings but they are not asserted against `_WARNINGS`. Not sure why though...

da-woods:
> Currently, this PR does not allow `tuple[...]` which should be allowed but usage is limited because it means tuple with one Ellipsis element inside.

Both mypy and ty claim that it's invalid rather than having any special meaning.

Cython/Compiler/ExprNodes.py
```diff
@@ -1403,6 +1403,8 @@ class EllipsisNode(PyConstNode):
     def compile_time_value(self, denv):
         return Ellipsis
 
+    def analyse_as_type(self, env):
+        return Builtin.ellipsis_type
```
da-woods: I don't think this is right.

This also also implies that 
1. `...()` produces an ellipsis instance.
2. `def f(x: ...)` is a function accepting an ellipsis instance (I think `None` is the only thing special-cased like that in the typing system)

matusvalo: Yeah I think you are right. `...` is an instance not a type. Hmm hence the handling of annotation must be aware that annotation is not only containing types but also `Node`s. I will try to craft something (hopefully) smart.

scoder:
I think the issue here (as David noted) is that `...` isn't generally a type. It only has a special meaning in _some_ places where types are used. I think we have to represent the meaning in those places, rather than representing `...` as a type.

A generic tuple type should know that it has unknown length when finding `...` in its argument list, build its `repr()` accordingly, and implement appropriate (partial) type inference. That's not a property of `...`, it's a capability of the tuple type.

matusvalo:
> I think the issue here (as David noted) is that `...` isn't generally a type. It only has a special meaning in _some_ places where types are used. I think we have to represent the meaning in those places, rather than representing `...` as a type.
> 
> A generic tuple type should know that it has unknown length when finding `...` in its argument list, build its `repr()` accordingly, and implement appropriate (partial) type inference. That's not a property of `...`, it's a capability of the tuple type.

Yes I agree. Hence,
* I am using `Ellipsis` as a sentinel instead of `EllipsisType`.
* I moved all logic specific to `Ellipsis` to `PythonTupleTypeConstructor` since it is only specific to tuples

I hope now the code is better organized.

Cython/Compiler/PyrexTypes.py
```diff
@@ -5449,14 +5451,32 @@ def get_container_type(self):
...
-        return super().specialize_here(pos, env, template_values)
+        if (len(template_values) == 2 and template_values[0] is not Ellipsis and template_values[1] is Ellipsis):
+            # ellipsis is allowed only as tuple[TYP, ...]
+            set_uniform_element_type = True
+            template_values = template_values[0:1]
+        else:
+            set_uniform_element_type = False
+        typ = super().specialize_here(pos, env, template_values)
+        if set_uniform_element_type:
+            typ.has_uniform_element_type = True
```
scoder: It seems more likely that the second item is something other than `Ellipsis`, so let's test that first.
```suggestion
        if len(template_values) == 2 and template_values[1] is Ellipsis and template_values[0] is not Ellipsis:
            # ellipsis is allowed only as tuple[TYP, ...]
            has_uniform_element_type = True
            template_values = template_values[0:1]
        else:
            has_uniform_element_type = False
        typ = super().specialize_here(pos, env, template_values)
        if has_uniform_element_type:
            typ.has_uniform_element_type = True
```

Cython/Compiler/PyrexTypes.py
```diff
@@ -5449,14 +5451,32 @@ def get_container_type(self):
...
+        if self.subscripted_types and self.has_uniform_element_type:
+            subscripted_types = [self.subscripted_types[0], "..."]
+            return f"{self._full_type_name(self.name, subscripted_types)} object"
+        else:
+            return super().__str__()
+
     def specialize_here(self, pos, env, template_values=None):
         if (template_values and None not in template_values and
+                Ellipsis not in template_values and
                 not any(v.is_pyobject for v in template_values)):
```
scoder: This seems redundant. I think the `None` is just there as a shortcut for the not so unlikely case that not all types are known (to Cython), but `Ellipsis` doesn't seen just as likely.
```suggestion
                not any(v.is_pyobject for v in template_values)):
```

scoder (approved):

Cython/Compiler/PyrexTypes.py
matusvalo: I needed to revert this change because tests are failing... :-(

matusvalo:
It seems that there is an issue with coverage not related to this PR... Hence, I cannot merge this PR :-(

PR #7799, opened 2026-07-05
Fix some issues with UnicodeConcatInPlace

da-woods:
There's a few issues which I think are mostly theoretical but are worth cleaning up:

* The "function arg" version was wrong. In practice this didn't matter because of point 2.
* The sharing was based on the type of the result rather than the type of the operand (and therefore was always treated as a temp).
* self += self could go wrong if it was reallocated (because we don't always hold a separate reference for the rhs). This is a real issue I think, but hard to reproduce reliably.

It's hard to check the threading logic particular reliably because we don't make the read-write access to globals/non-locals intrinsically thread safe. But that's a design choice elsewhere and not something intrinsic to unicode.

### DIFF
--- a/Cython/Utility/ObjectHandling.c
+++ b/Cython/Utility/ObjectHandling.c
@@ -2934,8 +2934,8 @@ static CYTHON_INLINE PyObject *__Pyx_PyUnicode_ConcatInPlaceImpl(PyObject **p_le
     }
     new_len = left_len + right_len;
 
-    if (left != right &&
-            __Pyx_unicode_modifiable(left, unsafe_shared)
+    if (left != right
+            && __Pyx_unicode_modifiable(left, unsafe_shared)
             && PyUnicode_CheckExact(right)
             && PyUnicode_KIND(right) <= PyUnicode_KIND(left)
             // Don't resize for ascii += latin1. Convert ascii to latin1 requires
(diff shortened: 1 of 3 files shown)


Cython/Utility/ObjectHandling.c
```diff
@@ -2969,7 +2934,8 @@ static CYTHON_INLINE PyObject *__Pyx_PyUnicode_ConcatInPlaceImpl(PyObject **p_le
     }
     new_len = left_len + right_len;
 
-    if (__Pyx_unicode_modifiable(left, unsafe_shared)
+    if (left != right &&
+            __Pyx_unicode_modifiable(left, unsafe_shared)
             && PyUnicode_CheckExact(right)
             && PyUnicode_KIND(right) <= PyUnicode_KIND(left)
```
scoder: For consistency, let's put all `&&` left.
```suggestion
    if (left != right
            && __Pyx_unicode_modifiable(left, unsafe_shared)
            && PyUnicode_CheckExact(right)
            && PyUnicode_KIND(right) <= PyUnicode_KIND(left)
```

scoder (approved):

scoder:
3.2.x commit is https://github.com/cython/cython/commit/4f7b56d2460cf8895642d8952ab56fa8059a2335

PR #7800, opened 2026-07-05
Simple feature list for shared module based on utility file names

### LINKED ISSUE #7759: [ENH] Let users select the shared utility module code

We currently decide on a general basis which utility code we extract into the shared module and which not, thus making the shared module larger or the user modules larger for everyone. Some features are large, some are just widely used, some are large and less widely used, which leaves us with a permanent trade-off.

If users could easily select certain features in or out, then they could adapt the shared module to their needs and take the trade-off decision themselves.

I envision a positive list of general features that users want to see included, configured as part of the shared module setup.
This list would then be mapped to a larger list of utility code names that Cython uses internally.

The list wouldn't have to be exhaustive but only list major features like
- `memoryview`
- `import`
- `pickle`
- `extensiontypes`
- `fusedfunction`
- `builtins` (possible more specific, like `builtinstr`)

To allow adding to this list later, Cython should just ignore (and probably warn about) unknown feature names that future Cython version might understand. Something that's unknown now would not end up in the shared module anyway (or users must depend on a later Cython version).

It's probably fine to constrain this configuration to the shared module, at least for the beginning. If users want to use features in their code that they excluded from the shared module, then they'll get a failure at import time at the latest, which seems fine.
Alternatively, we could be nice and automatically put excluded shared code into the user module if that's easily possible, but that doesn't seem a required feature.

We could provide include/exclude configuration options. I think that a single positive list has the advantage of allowing future feature additions that would automatically be excluded by existing builds that don't use them anyway. OTOH, if we decide to add a complete _existing_ feature to the list of shared code, users of this feature would have to adapt their setup in order to use a new Cython version. That seems fine, given the "ignore unknown names" constraint and the presumably low frequency of this happening.

### Additional context

_No response_

scoder:
This PR allows users to select the features in the shared utility module.

I had first envisioned a semantic list of feature names, but while putting it together, it seemed to correspond broadly with the utility file names that we use. Thus, the list of features to chose from is automatically gathered by finding the names of C files in `Cython/Utility/` that export code, as for the utility code collection before. I added a utility code tag `//@feature: NAME` that allows setting a different (more specific) name for a code section. I do that for the `AutoPickle` feature.

Commenting and trying it out is welcome.

### DIFF
--- a/Cython/Build/SharedModule.py
+++ b/Cython/Build/SharedModule.py
@@ -5,6 +5,7 @@ from Cython.Compiler import (
 )
 from Cython.Compiler.StringEncoding import EncodedString
 from Cython.Compiler.Scanning import SharedUtilitySourceDescriptor
+from Cython.Compiler.Errors import warning
 
 
 def list_of_features(included=None, excluded=None, _list_of_features=[]):
@@ -17,12 +18,25 @@ def list_of_features(included=None, excluded=None, _list_of_features=[]):
     return _select_features(feature_names, included, excluded)
 
 
-def _select_features(names, included, excluded):
+def _select_features(all_names, included, excluded):
+    names = all_names
     if included:
         included = set(included)
+        unknown = included.difference(all_names)
+        if unknown:
+            warning(None, f"Unknown shared module features selected: {', '.join(sorted(unknown))}")
+
         names = [name for name in names if name in included]
+
     if excluded:
         excluded = set(excluded)
+        unknown = excluded.difference(all_names)
+        if unknown:
+            warning(None, f"Unknown shared module features excluded: {', '.join(sorted(unknown))}")
+        unused = excluded.difference(names)
+        if unused:
+            warning(None, f"Shared module features excluded that was not included: {', '.join(sorted(unused))}")
+
         names = [name for name in names if name not in excluded]
 
     return names
--- a/Cython/Build/SharedModule.py
+++ b/Cython/Build/SharedModule.py
@@ -7,7 +7,54 @@ from Cython.Compiler.StringEncoding import EncodedString
 from Cython.Compiler.Scanning import SharedUtilitySourceDescriptor
 
 
-def create_shared_library_pipeline(context, scope, options, result):
+def list_of_features(included=None, excluded=None, _list_of_features=[]):
+    if _list_of_features:
+        feature_names = _list_of_features[:]
+    else:
+        feature_names = sorted({feature_name for feature_name, _, _ in _iter_exports()})
+        _list_of_features[:] = feature_names
+
+    return _select_features(feature_names, included, excluded)
+
+
+def _select_features(names, included, excluded):
+    if included:
+        included = set(included)
+        names = [name for name in names if name in included]
+    if excluded:
+        excluded = set(excluded)
+        names = [name for name in names if name not in excluded]
+
+    return names
+
+
+def _iter_exports(selected_features=None):
+    UtilityCode = Code.UtilityCode
+    match_special = UtilityCode.get_special_comment_matcher('/')
+
+    for c_utility_file in os.listdir(Code.get_utility_dir()):
+        if not c_utility_file.endswith('.c'):
+            continue
+
+        file_base_name = os.path.splitext(os.path.basename(c_utility_file))[0]
+        feature_name = file_base_name.partition('_')[0]
+
+        if selected_features is not None and feature_name not in selected_features:
+            continue
+
+        for line in Code.read_utilities_hook(c_utility_file):
+            if '////' not in line:
+                continue
+            if not ((m := match_special(line)) and (name := m.group('name'))):
+                continue
+            if not (section_title := UtilityCode.match_section_title(name)):
+                continue
+            name, section_type = section_title.groups()
+            if section_type == 'export':
+                yield (feature_name, name, c_utility_file)
+
+
+def create_shared_library_pipeline(context, scope, options, result, selected_features=None):
 
     parse = Pipeline.parse_stage_factory(context)
 
@@ -15,11 +62,13 @@ def create_shared_library_pipeline(context, scope, options, result):
         def generate_tree(compsrc):
             tree = parse(compsrc)
 
-            tree.scope.use_utility_code(
-                MemoryView.get_view_utility_code(options.shared_utility_qualified_name))
+            if selected_features is None or 'MemoryView' in selected_features:
+                tree.scope.use_utility_code(
+                    MemoryView.get_view_utility_code(options.shared_utility_qualified_name))
 
-            tree.scope.use_utility_code(MemoryView._get_memviewslice_declare_code())
-            tree.scope.use_utility_code(MemoryView._get_typeinfo_to_format_code())
+                tree.scope.use_utility_code(MemoryView._get_memviewslice_declare_code())
+                tree.scope.use_utility_code(MemoryView._get_typeinfo_to_format_code())
+
             context.include_directories.append(Code.get_utility_dir())
             return tree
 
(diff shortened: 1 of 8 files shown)


matusvalo:
Thank you working on this. This will allow shared module be much useful feature for broader set of projects. Here are my 5 cents:

## Errors

When shared feature is missing we should clearly explain users what happened. E.g. following flow: 

```cython
from array import array

def add_one(int[:] buf):
    for x in range(buf.shape[0]):
        buf[x] += 1


a = array('i', [1, 2, 3, 4])
add_one(a)
print(a.tolist())
```

If user by mistake exclude the memoryview from shared utility generation code:
```
cython --generate-shared shared.c  --shared-exclude MemoryView
cython --shared shared memview.pyx
```

Compilation will succeed but importing module will cause cryptic error:
```
>>> import memview
    import memview
  File "memview.pyx", line 1, in init memview
    from array import array
ImportError: shared does not export expected C function __Pyx_ValidateAndInit_memviewslice
```

Should print error like "Add feature MemoryView to shared utility." Or something like that. **Ideally** we should error out during compilation but not sure if there is robust way to do it.

## CLI

* We need to clarify better when `--shared-exlude` and `--shared-only` needs to be used. If I am not wrong they should be used together with `--generate-shared` argument. They should error out otherwise.

* Related to new arguments, let me bikeshed a little. Have you consider use `feature` e.g. `--feature-exclude`, `--feature-only`.
I have not strong opinion about that. Maybe it fits better for CLI proposed lower (and hence not for this PR) 

* Long term (not this PR of course) I would do overhaul of whole CLI because it's getting complicated. We can:
  - `cython MODULE.pyx` -> `cython compile MODULE.pyx`
  - `cython --shared cyshared MODULE.pyx` -> `cython compile --shared MODULE.pyx`
  - `cython --generate-shared cyshared.c` -> `cython generate-shared cyshared.c`
  - `cython --generage-shared cyshared.c --shared-exclude MemoryView` -> `cython generate-shared --shared-exclude MemoryView cyshared.c`
  - `cythonize -fib MODULE.pyx` -> `cython build -fi MODULE.PYX`

## setuptools

We should have a simple and stupid way how to integrate it to setuptools:

```python

extensions = [
    Extension("*", ["**/*.pyx"]),
    # Providing 'sources' is optional for the shared module.
    # If missing, the module package will be used for the path in 'build_dir'.
    Extension("mypkg.shared._cyutility", sources=["mypkg/shared/_cyutility.c"], shared_exclude='MemoryView'),
]

setup(
  ext_modules = cythonize(extensions, shared_utility_qualified_name = 'mypkg.shared._cyutility')
)
```

da-woods:
I think that ideally excluding something from the shared module should just allow it to be included unshared. On the basis that you might have performance reasons for wanting to exclude a specific feature

I don't necessarily think that has to be in the initial implementation, but I think that should in part of an ideal implementation.

> setuptools

I'm not sure there's a great need to prioritise setuptools any more. Meson and/or scikit-build-core are probably more important (although the support might be implemented outside Cython). It's not that we should completely ignore setuptools, but I don't *personally* feel obliged to do much with it.

--------------------------------------------

This looks basically good to me. I think it will probably be hard to recategorize things as soon as we have a release with them categorized.  The initial classification doesn't look bad though.

I don't think `MemoryView` will appear in the list of features in the help but I think `MemoryView_c` will. (This is based on quickly readying the code, rather than running it)

matusvalo:
> I think that ideally excluding something from the shared module should just allow it to be included unshared. On the basis that you might have performance reasons for wanting to exclude a specific feature

I agree with you. This would be best solution. What I consider to be difficult to get right is user experience. Simplest solution would be to provide the list of excluded features in **both** commands (`cython --shared` and `cython --generate-shared`) which I consider to be pain. Ideally user should provide the list of features **once** and everything will compile correctly. Having this kind of behaviour would require some kind of discovery of shared features during compilation of .pyx file. Not sure how complex this would be. Theoretically, this can be done in `setuptools` logic though (but it does not solve running `cython` commands manually).

> I'm not sure there's a great need to prioritise setuptools any more.

setuptools is still first popular in python ecosystem (even cython uses it). But I understand that from scientific perspective meson is standard though. I suppose it would not be much difficult to implement and cython documentation still prefers setuptools soo my humble oppinion is that we should implement it...

scoder:
I agree that the CLI of `cython.py` is probably complicated enough to merit sub-commands. Lacking that for the moment, I think we should keep `--shared-only` and `--shared-exclude` to make it match up with `--shared`. Maybe `--generate-shared` should have been `--shared-generate`, but if we start making changes in that corner then sub-commands are certainly the cleaner alternative.

scoder:
> Should print error like "Add feature MemoryView to shared utility."

The error is agnostic to the missing feature, and I'd rather not add a textual mapping from symbols back to feature names to all modules. That would defeat the purpose of the shared module. But we could probably add a message like "Does the share utility module contain all required features?" to hint in the right direction. We _could_ also generate a list of features that the current module reads utility code for, and then say something like "This module needs features x, y, z. Please make sure the shared module has them.".

scoder:
> I'm not sure there's a great need to prioritise setuptools any more.

I'm not arguing with that, but … we use it for Cython. :)

scoder:
> excluding something from the shared module should just allow it to be included unshared

I agree. I don't think the current state precludes that for the future. With this PR, users can start to configure what they want and need to take care that they don't exclude something that's used. They'll notice at import time. Eventually, we can add an automatic switch to put excluded features into each module separately, so that users do _not_ have to take care any more.

The advantage of setuptools here is that `cythonize()` sees all user modules together with the shared module configuration and can take steps accordingly.

We could also move everything to C compile time, as we often do, and put exclusive guards around the export sections and the implementation code, so that users can set a C macro like `CYTHON_SHARE_MEMORYVIEWS=1` globally in their build to control the binary modules sizes, at the cost of larger, redundant C files. That would speed up user side experiments a lot.

scoder:
> I don't think `MemoryView` will appear in the list of features in the help but I think `MemoryView_c` will.

I took care of that by looking only at the first of the name before `"_"`.

scoder:
> I think it will probably be hard to recategorize things as soon as we have a release with them categorized.

I think the `//@feature` tag reduces the risk of that. If we want to move code around, we can keep the old feature name. There's also no reason why a code section could only have one tag, we could add sub-feature names when a finer granularity proves relevant.

da-woods:
> > I'm not sure there's a great need to prioritise setuptools any more.
> 
> I'm not arguing with that, but … we use it for Cython. :)

I wasn't really arguing for "no setuptools at all"; more "no more effort into setuptools than needed"

The other thing we should probably be aware of is that the individual utility code parts have dependencies. I'm fairly sure that memoryviews with require extension types just based on how it's implemented.

I think that's all fine though and that the dependencies will be included in the shared utility file even if they don't end up exported.

Cython/Build/SharedModule.py
```diff
@@ -5,40 +5,117 @@
...
+    for c_utility_file in os.listdir(Code.get_utility_dir()):
+        if not c_utility_file.endswith('.c'):
+            continue
+
+        file_base_name = os.path.splitext(os.path.basename(c_utility_file))[0]
+        current_feature_name = file_base_name.partition('_')[0]
+
+        if selected_features is not None and current_feature_name not in selected_features:
+            # Disable whole file feature, including any @features declared therein.
+            continue
```
da-woods: I'm not completely convinced by having "whole file features" and "subfeatures" for a couple of reasons:
* I don't think it's obviously got this structure from the help, and if we do document it then it seems like an extra complication to explain.
* I agree with your conclusion that the file names currently do a decent job of grouping things. But I'm not sure we want the utility code location to be dictated by that. It suddenly becomes a user-facing decision where you put a block of utility code when there may be multiple possible options that make sense.

  (For some recent additions I've chosen to add new utility code files, just because the existing files are big and a hard-to-navigate. I don't think any of that would particularly be wrong if those names were exposed as shared feature, but it wasn't _really_ what I was thinking when putting them there)

  Letting the tag be an override of the feature name rather than a sub-feature avoids that.

da-woods: Just as an example - I think `ObjectHandling` and `Optimize` are probably too generic to be useful features. We don't currently have any of that code in the shared module but if we did I think we'd probably want to override it rather than isn't it as a "subfeature"

scoder: > Just as an example - I think `ObjectHandling` and `Optimize` are probably too generic to be useful features.

I agree – although from prior experience, I do not expect much non-inline code to appear in these files.

I think your arguments against sub-features are sound, though. That would quickly become difficult to explain. I removed them (making `//@feature` _always_ override the file name), and also the ability to assign multiple feature names per section. It's not currently needed and the intended effect of that is unclear. I made the tag repetition an error to prevent accidental misuse on our side.

da-woods:
I'm happy with this as an initial version.  I think long-term we should make it possible to use parts that you're choosing not to take from the shared module but that can be put off to later.

PR #7801, opened 2026-07-06
Build: Fix setting compiler directives.

scoder:
We were trying to switch off some directives in the build (by changing the `default_directives` from `setup.py`) but weren't, at least not for parallel builds that start separate Cython instances. The directives are now configured per `Extension`.

### DIFF
--- a/Cython/Compiler/Code.py
+++ b/Cython/Compiler/Code.py
@@ -2,6 +2,8 @@
 #   Code output module
 #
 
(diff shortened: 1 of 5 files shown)


scoder:
I added explicit `binding=True` and `binding=False` to modules that either need it or shouldn't need it, but might have missed less visible ones. Thus, the global setting is still off.

I'll merge this since it fixes a real issue and enables further experiments.

PR #7803, opened 2026-07-07
Add likely()/unlikely() compiler hints

### LINKED ISSUE #7667: [ENH] Add if-unlikely construct

Apologies if this is already implemented in some way, or there is already an issue or a PR, but I couldn't find it. 

In my code, I would like to mark certain if branches as unlikely.

```cython
if (cython.unlikely(some_condition)):
    # do something
```

I noticed that cython already adds `unlikely` for the branches that raise exceptions. Maybe it wouldn't be to difficult to expose a generic mechanism for user. 

I'm happy to try and do it myself, just want get an opinion first regarding syntax and general feasibility. Does approach with `cython.unlikely` look good? Are there any caveats?

_No response_

### Additional context

_No response_

matusvalo:
Closes #7667


matusvalo:
For now, no tests are added. I will add them later...

Cython/Compiler/ExprNodes.py
```diff
@@ -14716,6 +14716,36 @@ def binop_node(pos, operator, operand1, operand2, inplace=False, **kwargs):
...
+    def analyse_types(self, env):
+        self.arg = self.arg.analyse_types(env)
+        self.arg = self.arg.coerce_to_boolean(env)
+        return self
+
+    def calculate_result_code(self):
+        return self.arg.result()
+
+    def result(self):
+        return f'{self.expect}({super().result()})'
```
matusvalo, line 14743: Hmm here I should use somehow `code.unlikely()` and `code.likely()` (the latter one needs to be created)

matusvalo: I have one question though. In code.pyx there is a method:
https://github.com/cython/cython/blob/fcee6d0a8aaf7ed428c62c98cab355cfc9009755/Cython/Compiler/Code.py#L2965-L2969

But in other places is used "naked" likely/unlikely - e.g. here:

https://github.com/cython/cython/blob/fcee6d0a8aaf7ed428c62c98cab355cfc9009755/Cython/Compiler/Code.py#L3320

Hence, it means that even user provides `gcc_branch_hints`, not all `likely/unlikely` are eliminated. Is it intentional? Or it is a "bug"?

matusvalo:
Hmm I live only once, so I right away used `likely()`/`unlikely()` over cython utility code. But I did it very conservatively, only around conditions leading to errors...

Cython/Compiler/ExprNodes.py
scoder: I think we just got lazy after a while and started using them inflationarily, without taking care of the option flag.

Cython/Utility/MemoryView.pyx
```diff
@@ -168,12 +168,12 @@ cdef class array:
...
 
-        if not self._shape:
+        if cython.unlikely(not self._shape):
             raise MemoryError, "unable to allocate shape and strides."
 
         # cdef Py_ssize_t dim, stride
         for idx, dim in enumerate(shape):
-            if dim <= 0:
+            if cython.unlikely(dim <= 0):
                 raise ValueError, f"Invalid shape in axis {idx}: {dim}."
```
scoder: I think we automatically inject `unlikely()` these days when we detect that the branch only raises an exception. Same for the `FusedFunction.pyx` cases, I don't think these are needed.

Cython/Compiler/ExprNodes.py
```diff
@@ -14716,6 +14716,36 @@ def binop_node(pos, operator, operand1, operand2, inplace=False, **kwargs):
         **kwargs)
 
 
+class LikelyNode(ExprNode):
```
scoder: I think "branch hint" is a more general term than "likely or unlikely".
```suggestion
class BranchHintNode(ExprNode):
```

Cython/Compiler/ExprNodes.py
```diff
@@ -14716,6 +14716,36 @@ def binop_node(pos, operator, operand1, operand2, inplace=False, **kwargs):
...
+    #  arg       ExprNode
+    #  expect    string
+
+    subexprs = ['arg']
+    type =PyrexTypes.c_bint_type
```
scoder: The code checker should really have detected this, but I think we might have disabled it because of … keyword arguments? We're very inconsistent with those.

```suggestion
    type = PyrexTypes.c_bint_type
```

Cython/Compiler/ExprNodes.py
```diff
@@ -14716,6 +14716,36 @@ def binop_node(pos, operator, operand1, operand2, inplace=False, **kwargs):
         **kwargs)
 
 
+class LikelyNode(ExprNode):
+    #  arg       ExprNode
+    #  expect    string
```
```suggestion
    #  expect    string ("likely" or "unlikely")
```

scoder (commented):
As I mentioned in the ticket, we should reject the usage outside of (complete) if-conditions, at least for the beginning, so we may not even need the `LikelyNode` since if-statements have it as a feature.

https://github.com/cython/cython/issues/7667#issuecomment-4393854506

matusvalo:
> As I mentioned in the ticket, we should reject the usage outside of (complete) if-conditions

It should be implemented that way (tests are pending). This is allowed:
```python
if likely(...):
   ...
```

```python
if ...:
   ...
elif likely(...):
   ...
```

Everything else should raise compilation error. Or am I missing something?

Cython/Compiler/ExprNodes.py
matusvalo: Hmm not sure then what is right to do. Should I use `unlikely()` method from Code.py or this is OK. I think that if we are not using globally  that method, then `Options.gcc_branch_hints` is misleading and should be deprecated/removed.

Cython/Utility/MemoryView.pyx
matusvalo: I was not aware about that. Good to know!

matusvalo: Fixed.

Cython/Compiler/ExprNodes.py

matusvalo:
> > As I mentioned in the ticket, we should reject the usage outside of (complete) if-conditions
> 
> It should be implemented that way (tests are pending). This is allowed:
> 
> ```python
> if likely(...):
>    ...
> ```
> 
> ```python
> if ...:
>    ...
> elif likely(...):
>    ...
> ```
> 
> Everything else should raise compilation error. Or am I missing something?

Hmm OK, conditions inside comprehensions are also supported as side effect. But conditional assignment (`a = 1 if cond else 2`) does not support branch hints though.

da-woods:
I'm just noting a few other places that it _could_ go. I agree with @scoder here that initially we should probably reject anything except `if` and `elif` but it's worth being aware of where people might try to use it:

```
while likely(...):  # maybe not high value because the C compiler probably assumes that loops are likely to continue.
```

```
x if likely(...) else y  # I think you already noted that this doesn't currently work.
```

```
match m:
   case ... if likely(...):  # should probably eventually be supported
```

```
match m:
    case likely(...):  # I think this is a syntax nightmare and should probably never be supported.
```

```
try:
    ...
except likely(...):  # probably shouldn't be supported ever?
    ...
```

parts of conditional expressions e.g.:

```
if likely(x) or y:  # maybe?? The probably reduces to C-ifs so might compile OK.
```

Cython/Compiler/ExprNodes.py
da-woods: `Options.gcc_branch_hints` is used in `Nodes.py` to define `likely` and `unlikely` (although in a way that will only work if it's set before `Nodes` is imported). So I think it's fine to always use generate `likely()` or `unlikely()` in the code and rely on the macros.

The other place where it probably should be used is `Code.redef_builtin_expect`

matusvalo:
> I'm just noting a few other places that it could go

I prefer to keep this feature as much simple as possible (only `if` conditions). After implementing this PR I learned that Cython already automatically inserts `likely()` for condition branches only raising exception. 

Hence, this feature I think is limited only for high performance functions... For more complex stuff than bare simple if condition I would prefer those two options:

1. mark utility function raising exception (`__Pyx_Raise` I think) as `cold` this will set likeness automatically in all conditions leading to the exception
2. (my preferred way) is to add first class support of PGO to cython. Using PGO is even recommended way over `likely/unlikely`. We can e.g. add builtin support of PGO into Cython's distutils/setuptools code. (I am aware about last discussion not to add anything to setuptools)

That said, I still thinks that having `likely`/`unlikely` in cython is nice to have but I would implement only basic cases...

da-woods:
> I prefer to keep this feature as much simple as possible (only `if` conditions).

Yes, definitely agree.

>After implementing this PR I learned that Cython already automatically inserts `likely()` for condition branches only raising exception.

I assume this does work if I do `if cython.likely(something): raise RuntimeError()`?

(It certainly seemed to work with the more natural `unlikely` because it looked like you applied that to the utility code).

> first class support of PGO to cython

I assume you mean "compile the .c code with PGO" rather than "try to generate different .c code after profiling"?

matusvalo:
> I assume this does work if I do if cython.likely(something): raise RuntimeError()?

It is not documented, I just learned it from comment by @scoder here: https://github.com/cython/cython/pull/7803#discussion_r3541416550
I **suppose** that if condition branch contains **only** `raise <something>` , cython automatically injects `unlikely()`/`likely()`.

>  it looked like you applied that to the utility code

Yes, but I put it only in places which are more complex - e.g. in MemoryView.pyx the condition branch calls function inside:
https://github.com/cython/cython/blob/2dc07522ec0e90ba8600d58216bbd134e0d744bf/Cython/Utility/MemoryView.pyx#L156-L157

> I assume you mean "compile the .c code with PGO" rather than "try to generate different .c code after profiling"?

I have an idea, or maybe just a dream. I ran into PGO when studying documentation around `__builtin_expect ` where GCC documentation recommended PGO. When digging more in deep I learnt it means just 3 steps:
1. compile C code with PGO profile enabled
2. run code on some data or production load (I suppose more production load, better)
3. compile the same C code again using PGO collected profiles.

My idea is just to stream line the compilation expirience with PGO. Just to extend somehow Cython builtin setuptools to allow following workflow (e.g. can be even used for Cython)
1. `python setup.py build_ext --enable-pgo-profile` - This will run Cython + GCC with PGO profiling enabled - `setup.py` `Extention` definition could somehow specify details (where profile files are stored etc)
2. Run the test suite (e.g. in case Cython test suite)
3. `python setup.py build_ext --use-pgo-profile` - This command will skip Cython compilation just run GCC compilation of already generated C files using collected profile as input.

> [!NOTE]
> I know that bigger projects are using meson etc. But I don't know meson and it's up to meson plugin (not cython) to allow this kind of compilation...

da-woods:
> > I assume this does work if I do if cython.likely(something): raise RuntimeError()?
> 
> It is not documented, I just learned it from comment by @scoder here: [#7803 (comment)](https://github.com/cython/cython/pull/7803#discussion_r3541416550) I **suppose** that if condition branch contains **only** `raise <something>` , cython automatically injects `unlikely()`/`likely()`.

Ah - that wasn't quite my question.

I was really asking: does you new `cython.likely` directive work if it contradicts the automatic unlikely?

matusvalo:
> I was really asking: does you new `cython.likely` directive work if it contradicts the automatic unlikely?

Good question. No idea. Maybe whole universe will explode in supernova :-D

da-woods:
I gave it a quick go. It looks to compile to

```
if (unlikely(likely(__pyx_t_1))) {
```

The C code compiles and runs without destroying the universe though.

matusvalo:
> I gave it a quick go. It looks to compile to
> 
> ```
> if (unlikely(likely(__pyx_t_1))) {
> ```
> 
> The C code compiles and runs without destroying the universe though.

I can try to fix that somehow... to force `likely()` instead of `unlikely()`...

da-woods:
> My idea is just to stream line the compilation expirience with PGO. Just to extend somehow Cython builtin setuptools to allow following workflow (e.g. can be even used for Cython)

I did find https://github.com/esoma/pgo. It's obviously not maintained. It seems like it still works with some very minor fixes (I had to remove `dry_run=self.dry_run` from a couple of places in it).

I tried it with a Cython module and that seemed like it was fine (although its hard to tell from a quick test if the PGO changed anything)

scoder:
I added PGO support to our IPythonMagic a while ago. You can take that as an example.

https://github.com/cython/cython/blob/2dc07522ec0e90ba8600d58216bbd134e0d744bf/Cython/Build/IpythonMagic.py#L264-L284

I second the general recommendation of automatic PGO over manual micro-optimisations like `__builtin_expect()`, but it does require suitable benchmark code for the profile generation to push the tuning in the right direction and to prevent it from accidentally pessimising important code paths that your benchmark simply doesn't exercise.

CPython uses parts of its test suite, although they include error tests (that probably shouldn't be preferred by the C compiler), but the bulk of the executed code tends to go through important C code paths that are worth optimising.

matusvalo:
I am closing this PR in favor of #7812.

PR #7806, opened 2026-07-10
Fix scope tracking transform

scoder:
There were several issues with it.
- It didn't always initialise its attributes, only when starting from a `ModuleNode`.
- Storing away the attributes inefficiently created a new tuple.
- Calls to `.visit_scope()` needlessly went through the vtable.

### DIFF
--- a/Cython/Compiler/Visitor.py
+++ b/Cython/Compiler/Visitor.py
@@ -368,6 +368,11 @@ class ScopeTrackingTransform(CythonTransform):
     #scope_type: can be either of 'module', 'function', 'cclass', 'pyclass', 'struct'
     #scope_node: the node that owns the current scope
 
+    def __init__(self, context):
+        super().__init__(context)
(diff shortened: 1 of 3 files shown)


Cython/Compiler/Visitor.py
```diff
@@ -368,31 +368,34 @@ class ScopeTrackingTransform(CythonTransform):
...
     #scope_node: the node that owns the current scope
 
-    def visit_ModuleNode(self, node):
+    def __init__(self, context):
+        super().__init__(context)
         self.scope_type = 'module'
-        self.scope_node = node
-        self._process_children(node)
-        return node
+        self.scope_node = None
```
da-woods, line 374: I think this is why the fix always felt slightly dodgy to me before in that it won't *generally* work.

However, in practice it looks like `scope_node` is only used in very limited situations so it works well enough to improve the tests.

da-woods (approved):

PR #7807, opened 2026-07-10
[Bug] Fix __Pyx_ArgsSlice_FASTCALL when CYTHON_VECTORCALL is disabled

fgallaire:
When CYTHON_VECTORCALL is 0, the _FASTCALL argument macros alias their _VARARGS counterparts and the args parameter is a tuple, not a C array. The old '|| (CYTHON_COMPILING_IN_CPYTHON && CYTHON_ASSUME_SAFE_MACROS && !CYTHON_AVOID_BORROWED_REFS)' clause dates from when the macro took the address of __Pyx_Arg_FASTCALL(args, start) (a GET_ITEM lvalue inside the tuple); with the current 'args + start' pointer arithmetic it hands the tuple object itself to __Pyx_PyTuple_FromArray, which no longer compiles (incompatible pointer type) - any def with *args built with -DCYTHON_VECTORCALL=0 fails to build. Alias _FASTCALL to _VARARGS like every other fastcall macro.

### DIFF
--- a/Cython/Utility/FunctionArguments.c
+++ b/Cython/Utility/FunctionArguments.c
@@ -963,10 +963,10 @@ static CYTHON_INLINE int __Pyx_MergeKeywords(PyObject *kwdict, PyObject *source_
 
 #define __Pyx_ArgsSlice_VARARGS(args, start, stop) PyTuple_GetSlice(args, start, stop)
 
-#if CYTHON_VECTORCALL || (CYTHON_COMPILING_IN_CPYTHON && CYTHON_ASSUME_SAFE_MACROS && !CYTHON_AVOID_BORROWED_REFS)
+#if CYTHON_VECTORCALL
 #define __Pyx_ArgsSlice_FASTCALL(args, start, stop) __Pyx_PyTuple_FromArray(args + start, stop - start)
 #else
-#define __Pyx_ArgsSlice_FASTCALL(args, start, stop) PyTuple_GetSlice(args, start, stop)
+#define __Pyx_ArgsSlice_FASTCALL __Pyx_ArgsSlice_VARARGS
 #endif
 
 /////////////// fastcall ///////////////


da-woods (approved):
Thanks - that looks right.

Out of interest, why are you trying to compile with vectorcall turned off? (This is just curiosity and doesn't affect the validity of the fix)

fgallaire:
> Out of interest, why are you trying to compile with vectorcall turned off? (This is just curiosity and doesn't affect the validity of the fix)

I'm compiling Cython extensions for an unusual target: a WebAssembly build where the "interpreter" is a bridge that emulates the C API on top of Brython (a Python implementation running in JavaScript in the browser). Python objects live on the JS side as opaque handles, not as real PyObject structs in linear memory, and each C API function is a thin JS shim.

So anything in Cython's generated code that relies on CPython internals rather than the API surface breaks: direct struct field access, borrowed-reference tricks, unicode/long internals, etc. That's why I build with CYTHON_COMPILING_IN_CPYTHON=0, CYTHON_USE_TYPE_SLOTS=0, CYTHON_AVOID_BORROWED_REFS=1 and friends.

That combination (METH_FASTCALL entry points still being generated, but the "compiling in CPython" fast paths disabled) is how I ended up on the portability fallback of __Pyx_ArgsSlice_FASTCALL, which assumes args is a real tuple while FASTCALL hands it a C array.

da-woods:
That all makes sense - we left those options configurable largely for the benefit of people try to use Cython in weird situations that we hadn't thought of. Which definitely covers your use case.

> That's why I build with `CYTHON_COMPILING_IN_CPYTHON=0` [...]

I don't think this is currently something that is user-overrideable. I think it would have worked if you had managed to override it.

--------

Anyway - thanks. I'll merge this and cherry-pick it back to the 3.2 branch.

da-woods:
3.2.x commit https://github.com/cython/cython/commit/299791c57ab3d4ceb85d7ce84d41aa1adf7a9a65

PR #7808, opened 2026-07-10
Migrate Cython CLI to support sub-commands

### LINKED ISSUE #7802: Use sub-commands in `cython` CLI
Migrate to (and, initially, add) providing sub-commands for the different features of the `cython` command:

  - `cython MODULE.pyx` -> `cython compile MODULE.pyx`
  - `cython --shared cyshared MODULE.pyx` -> `cython compile --shared MODULE.pyx`
  - `cython --generate-shared cyshared.c` -> `cython generate-shared cyshared.c`
  - `cython --generage-shared cyshared.c --shared-exclude MemoryView` -> `cython generate-shared --shared-exclude MemoryView cyshared.c`
  - `cythonize -fib MODULE.pyx` -> `cython build -fi MODULE.PYX`

_Originally posted by @matusvalo in https://github.com/cython/cython/issues/7800#issuecomment-4896378636_

raza-khan0108:
## Summary
This pull request adds subcommand support to the `cython` CLI while preserving full backward compatibility with the legacy command structure:
- `cython MODULE.pyx` &rarr; `cython compile MODULE.pyx`
- `cython --shared cyshared MODULE.pyx` &rarr; `cython compile --shared MODULE.pyx` (with `--shared` defaulting to `'cyshared'` if the name is omitted)
- `cython --generate-shared cyshared.c` &rarr; `cython generate-shared cyshared.c`
- `cythonize -fib MODULE.pyx` &rarr; `cython build -fi MODULE.pyx`

## Subcommands Overview

1. **`compile`**: Translates `.pyx` or `.py` files to C/C++. Supports all existing compiler options.
2. **`generate-shared`**: Generates a shared C/C++ module with specified utility features (formerly `--generate-shared`).
3. **`build`**: Compiles and builds extension modules using setuptools/distutils (acting as a clean alias for the `cythonize` CLI).

## Key Implementation Details
- **Backward Compatibility**: A custom argument preprocessing step checks if a subcommand was supplied. If not, it falls back to the legacy parser, ensuring existing build scripts and integrations do not break.
- **`--shared` Defaulting**: Preprocesses `--shared` to default to `'cyshared'` if no value is explicitly supplied.
- **Reusable Parsers**: Modified `Cythonize` argument parser to accept an `add_help` argument. This allows reusing the entire `cythonize` parser definition under the new `build` subcommand without option conflicts.
- **Build Delegation**: If the `build` subcommand is detected, the CLI hands off control directly to `Cython.Build.Cythonize.main` and exits.

## Changes
- **`Cython/Build/Cythonize.py`**: Added `add_help` parameter to `create_args_parser` for subcommand parent parser support.
- **`Cython/Compiler/CmdLine.py`**:
  - Factored compile option additions into `add_compile_arguments`.
  - Implemented `create_cython_subcommand_parser` with subparsers.
  - Implemented `preprocess_shared_args` and subcommand detection logic.
  - Updated `parse_command_line` to route subcommands/legacy configurations.
- **`Cython/Compiler/Tests/TestCmdLine.py`**: Added comprehensive test coverage for `compile`, `generate-shared`, and `build` subcommands, as well as `--shared` defaulting.

## Verification
Ran the command-line parser test suite directly:
```bash
python -c "import sys; sys.path.insert(0, '.'); import unittest; import Cython.Compiler.Tests.TestCmdLine as t; unittest.main(t)"
```
All **56 unit tests passed successfully** (including existing legacy tests and the new subcommand/defaulting tests).

### DIFF
--- a/Cython/Build/Cythonize.py
+++ b/Cython/Build/Cythonize.py
@@ -192,11 +192,12 @@ def benchmark(code, setup_code=None, import_module=None, directives=None):
             f"slower than the best time ({format_time(fastest)}).")
 
 
-def create_args_parser():
+def create_args_parser(add_help=True):
     from argparse import ArgumentParser, RawDescriptionHelpFormatter
     from ..Compiler.CmdLine import ParseDirectivesAction, ParseOptionsAction, ParseCompileTimeEnvAction
 
     parser = ArgumentParser(
+        add_help=add_help,
         formatter_class=RawDescriptionHelpFormatter,
         epilog="""\
 Environment variables:
--- a/Cython/Compiler/CmdLine.py
+++ b/Cython/Compiler/CmdLine.py
@@ -4,6 +4,7 @@
 
 
 import os
+import sys
 from argparse import ArgumentParser, Action, SUPPRESS, RawDescriptionHelpFormatter
 from . import Options
 
@@ -102,16 +103,7 @@ class CythonArgumentParser(ArgumentParser):
         pass
 
 
-def create_cython_argparser():
-    description = "Cython (https://cython.org/) is a compiler for code written in the "\
-                  "Cython language.  Cython is based on Pyrex by Greg Ewing."
-
-    parser = CythonArgumentParser(
-        description=description,
-        argument_default=SUPPRESS,
-        formatter_class=RawDescriptionHelpFormatter,
-    )
-
+def add_compile_arguments(parser, include_sources=True):
     parser.add_argument("-V", "--version", dest='show_version', action='store_const', const=1,
                       help='Display version number of cython compiler')
     parser.add_argument("-l", "--create-listing", dest='use_listing_file', action='store_const', const=1,
@@ -193,19 +185,11 @@ def create_cython_argparser():
     parser.add_argument('-M', '--depfile', action='store_true', help='produce depfiles for the sources')
 
     # Shared module setup.
-
-    parser.add_argument("--generate-shared", dest='shared_c_file_path', action='store', type=str,
-                        help='Generates shared module with specified name.')
     parser.add_argument("--shared", dest='shared_utility_qualified_name', action='store', type=str,
                         help='Imports utility code from shared module specified by fully qualified module name.')
-    parser.add_argument("--shared-only", dest='shared_utility_features_enabled',
-                        action=ParseCommaListAction, type=str, metavar='FEATURES',
-                        help='Comma separate list of features to move to the shared module exclusively.')
-    parser.add_argument("--shared-exclude", dest='shared_utility_features_disabled',
-                        action=ParseCommaListAction, type=str, metavar='FEATURES',
-                        help='Comma separate list of features to exclude from the shared module.')
 
-    parser.add_argument('sources', nargs='*', default=[])
+    if include_sources:
+        parser.add_argument('sources', nargs='*', default=[])
 
     # TODO: add help
     parser.add_argument("-z", "--pre-import", dest='pre_import', action='store', type=str, help=SUPPRESS)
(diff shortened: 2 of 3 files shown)


da-woods:
A very superficial comment is that it can't be 100% compatible since right now `cython compile` attempts to convert a file called `compile` (no extension) into `compile.c`. I don't particularly think that matters though because I don't think anyone will actually be doing this (and we've made bigger breaking changes before).

Could you explain what the benefit of this is? As I see it, subcommands are mainly used by "multi-tool" type utilities to avoid adding too many commands to the global command-line namespace. We wouldn't really get the benefit of this (because we won't actually remove the `cythonize` command). So I guess the main questions are "does it make the shared utility code easier to use?" or "does it make it easier to add new features to the shared utility code?".

One slightly ugly pre-existing difference between `cython` and `cythonize` is that `#distutils: language = c++` directives lines are ignored in `cython` and but used in `cythonize`. That becomes a little uglier if it's a single command (just because the rule behind it is a bit harder to explain). I might try to fix that separately rather than treat it as a problem with this PR.

scoder:
> Could you explain what the benefit of this is?

Note that we asked for it in https://github.com/cython/cython/issues/7802

da-woods:
> > Could you explain what the benefit of this is?
> 
> Note that we asked for it in #7802

I missed that and read this as an unsolicited rewrite. I haven't looked at the comments of the PR at all yet, but I'll treat the justification as justified.

scoder:
> I'll treat the justification as justified.

It's still up for discussion, I think. The ticket was mostly an idea, not a finished design. Now that there's an implementation, I think the next step are really trying it out and seeing if it's an improvement over the status quo.

I'll ask in @rgommers as a vivid user of the `cython` command, whether sub-commands feel like something that is worth having from a user perspective.

raza-khan0108:
Regarding the file named `compile` edge case: users can still compile it by specifying a path like `./compile` or using the explicit `cython compile compile` command.

The main benefits of this subcommand structure are:
1. **Unified CLI**: Consolidates separate tools (`cython`, `cythonize`, `--generate-shared`) into a single command space (`cython compile`, `cython build`, `cython generate-shared`).
2. **Clarity**: Each subcommand gets its own focused help menu (e.g., `cython build -h`), preventing option pollution at the top level.
3. **Extensibility**: Makes it much easier to introduce new utilities under the `cython` command in the future.

matusvalo:
I tend to prefer this PR style. Current cython CLI is a little bit messy:

* `cython file.pyx` - compiles to .c file
* `cythonize file.pyx` - compiles to .c file
* `cythonize -b file.pyx` - compiles to .c file + builds .c file
* When generating shared module the `cython` command expects different set of parameters than normally etc...

da-woods:
I personally prefer the distinction between cython and cythonize. I can't *really* justify why, outside familiarity. I think it's partly that they have different dependencies and slightly different rules (although maybe the different rules should just be fixed). That doesn't mean that `cython build` shouldn't exist, but just that I probably won't use it.

I do agree that the shared module stuff is getting a bit crowded and cleaning that up has to be a good thing.

rgommers:
I think this is useful indeed, beyond the `generated-shared` cleanup I see an improvement opportunity here because it allows splitting the many options (see `cython --help`) _and_ removing the obsolete ones in the new interface. The current PR doesn't do any of that yet, but adding a new CLI that doesn't have backwards compat constraints is a nice one-off chance to do a reset. For example:

- drop `-2`, `-3`,  `--3str`, `-+` (only keep `--cplus`),
- consolidate the near-duplicates (e.g., three separate `--annotate[...]` flags and two `--embed` ones)) into single flags that take a value. 

There are a number of other flags that may either be obsolete, have a suboptimal default, or needs better docs. For example:
- `-Werror`/`-Wextra` have minimal docs in `--help` and nothing in html docs, and at first I thought they were stray commands drifting in from `cythonize` to pass on to a C compiler, while on closer inspection they're for Cython compile warnings/errors.
- `--fast-fail` should be the default I'd think, that's what all package authors always want. I don't know of many compilers that keep going by default.
- is `--force` doing anything useful (just don't pass `-t` instead, if I understand the "(overrides implied -t)" correctly)?
- is `--compile-time-env` a thing you want to keep, given the `DEF` deprecation?

Regarding `cython build`: that's probably too generic a name, I'd lean towards something like `build-module`.

scoder:
Let's take the overall CLI cleanup to Cython 3.4 to give us a bit more time for a good design.

Since we're adding new options for the shared module generation, maybe we could take out that part already and fold only those options into a separate `generate-shared` subcommand for the upcoming 3.3?

For that command, I think the following "compile options" should not be available: `--version`, `--embed*`, `-2/3/3str`, `--lenient`, `--capi-reexport-cincludes`, `-E`.

I think `--version` should generally not be in the sub-commands but only available as a global option.

@raza-khan0108, could you create a separate PR only with that one sub-command? Thank you for working on this.

raza-khan0108:
That makes complete sense—deferring the broader CLI hierarchy (`compile` and `build` subcommands) to Cython 3.4 will definitely give us more time to refine the overall subcommand structure cleanly.

I have just extracted the `generate-shared` subcommand (`cython generate-shared OUTPUT_FILE`) into a separate PR for the upcoming Cython 3.3 release: https://github.com/cython/cython/pull/7842

In that separate PR:
- **Restricted Compile Options**: All the compile options you listed (`--embed*`, `-2`/`-3`/`--3str`, `--lenient`, `--capi-reexport-cincludes`, and `-E`) are strictly excluded from the `generate-shared` subparser.
- **Global `--version`**: `-V` / `--version` is only registered on the global root parser (`cython --version generate-shared ...` works, but `cython generate-shared --version` correctly raises an `argparse` unknown option error).
- **Backward Compatibility**: The legacy `--generate-shared=OUTPUT_FILE` flat option is kept intact on the main parser for backward compatibility.
- **Test Coverage**: Added comprehensive unit tests and error check assertions to `TestCmdLine.py` verifying the subcommand behavior and confirming that all excluded options and extra source files are properly rejected.

I'll keep this PR (`#7808`) open for the broader Cython 3.4 CLI cleanup.

PR #7812, opened 2026-07-11
Add likely()/unlikely() compiler hints - simplified version

### LINKED ISSUE #7667: [ENH] Add if-unlikely construct

matusvalo:
This is alternative to #7803 but without the need of having separate BranchHintNode class.

~The downside of this implementation that cannot be explicitly asserted whether branch hint was generated.~

Fixes #7667


tests/errors/e_cmp.pyx
```diff
@@ -1,20 +1,37 @@
...
-	bool = int1 == ptr2 # error
-	bool = ptr2 == ptr3 # error
+cdef void likely_foo(int j):
+    cdef bint bol
+    cdef int i = 5
 
-	bool = 1 in 2 in 3
+    bol = cython.likely(i == j)
+    bol = cython.unlikely(i == j)
+    a = i if cython.likely(j > 5) else -i
```
scoder: Instead of rejecting this, it should be trivial to add a `branch_hint` attribute to `CondExprNode` in the same way as `IfClauseNode`. Could you do that?

Cython/Compiler/Optimize.py
```diff
@@ -5392,7 +5392,8 @@ def _set_ifclause_branch_hint(self, clause, statements_node, inverse=False):
             if not isinstance(node, non_branch_nodes):
                 if next_node_pos == len(statements) and isinstance(node, (Nodes.RaiseStatNode, Nodes.ReraiseStatNode)):
                     # Anything that unconditionally raises exceptions at the end should be considered unlikely.
-                    clause.branch_hint = 'likely' if inverse else 'unlikely'
+                    if not clause.branch_hint:
```
scoder: This can be done much earlier, at the beginning of the method. There is really nothing to do here if there is an explicit branch hint already.

scoder (commented):
That's much better.

We should mention in the docs that these functions are for rare and very clear use cases, not for guessing.

Cython/Compiler/Optimize.py
scoder: Fixed as part of the branch update.

tests/errors/e_cmp.pyx
matusvalo: I think it is done. Maybe some finetuning/polishing can be done but the logic is there.

matusvalo:
> We should mention in the docs

The docs are pending on my todo list. Any idea where it is best to put it in?

Cython/Compiler/ExprNodes.py
```diff
@@ -13889,7 +13890,10 @@ def generate_evaluation_code(self, code):
         code.mark_pos(self.pos)
         self.allocate_temp_result(code)
         self.test.generate_evaluation_code(code)
-        code.putln("if (%s) {" % self.test.result())
+        condition = self.test.result()
+        if self.branch_hint:
+            condition = '%s(%s)' % (self.branch_hint, condition)
+        code.putln("if (%s) {" % condition)
```
scoder: Let's use f-srings these days.

Cython/Compiler/ParseTreeTransforms.py
```diff
@@ -4239,6 +4240,26 @@ def visit_GeneratorBodyDefNode(self, node):
...
+        self.visitchildren(node)
+        return node
+
+    def visit_CondExprNode(self, node):
+        condition = node.test
+        if isinstance(condition, ExprNodes.SimpleCallNode):
+            function = condition.function.as_cython_attribute()
+            if function in ('likely', 'unlikely'):
+                node.branch_hint = function
+                node.test = condition.args[0]
```
scoder: I wouldn't mind giving both node types the same attribute name here so that these two methods can become one. Doesn't need to be in this PR but could be.

Cython/Compiler/ExprNodes.py

Cython/Compiler/ParseTreeTransforms.py
matusvalo: Done.

matusvalo: Renamed attribute and common logic merged together.

matusvalo:
> > We should mention in the docs
> 
> The docs are pending on my todo list. Any idea where it is best to put it in?

I have an idea where to put documentation, but it is non-trivial change in docs. Hence, In order to not block this PR, I will document this PR in separate PR.

scoder (approved):

PR #7814, opened 2026-07-12
[3.2.x] Update OS version for 3.2.x docs

da-woods:
It looks like the old version is no-longer supported. This matches what the master branch is already set to.


da-woods:
I'll merge this once I've checked that it works and had a look at the generated docs

PR #7815, opened 2026-07-12
Mark some internal functions as @cfunc.

scoder:
Reduce code, enable direct calls and optimisations.

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -4348,7 +4348,7 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
 # cimport/export code for functions and pointers.
 
 @cython.cfunc
-def _deduplicate_inout_signatures(item_tuples) -> tuple[list[str], tuple[str], tuple[str]]:
+def _deduplicate_inout_signatures(item_tuples) -> tuple[list[str], tuple[str, ...], tuple[str, ...]]:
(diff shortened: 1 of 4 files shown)


Cython/Compiler/ModuleNode.py
```diff
@@ -4347,7 +4347,8 @@ def generate_exttype_vtable_init_code(self, entry, code):
 
 # cimport/export code for functions and pointers.
 
-def _deduplicate_inout_signatures(item_tuples):
+@cython.cfunc
+def _deduplicate_inout_signatures(item_tuples) -> tuple[list[str], tuple[str], tuple[str]]:
```
matusvalo: Are the two inner tuples containing just one string? Didn't you mean

```suggestion
def _deduplicate_inout_signatures(item_tuples) -> tuple[list[str], tuple[str, ...], tuple[str, ...]]:
```

Note that there is pending PR to support `tuple[str, ...]`

scoder: Ah, right. They are arbitrarily long tuples of `str`. I forgot that that's spelled differently, thanks. ~I'll just use plain `tuple`.~ I'll use that instead to let it (eventually) know that the last tuple contains `str` object that don't need extra string formatting.

matusvalo:
Keep in mind that PR implementing the syntax used in this PR needs to be merged #7798. The code will be compiled and will work but Cython won't fully infer type of tuple items.

PR #7818, opened 2026-07-12
docs: Create Performance Optimization section and add branch hints docs

matusvalo:
I decided to make `early_binding_for_speed` more generic so we can document there performance optimization related features.

> [!NOTE]
> The documentation related to early bindings was unchanged. 

Related to #7812


scoder:
This will break existing web links to the page that you renamed. We should keep it in place. I'll have to take a look later for a good place to put this, unless you already found alternatives.

matusvalo:
> This will break existing web links to the page that you renamed.

There is a plugin in sphinx to create redirects (https://github.com/documatt/sphinx-reredirects). If the change makes sense we can try to use that to ensure working links...

matusvalo:
OK I implemented redirects which should provide correct redirects for external links...

You can test it using this link: https://cython--7818.org.readthedocs.build/en/7818/src/userguide/early_binding_for_speed.html

scoder:
The "early binding for speed" page was always there to present the basic idea of going from dynamic code to statically optimised code. That is a very important concept in Cython, and far, far more important than the specialized `unlikely()` marker. I'd rather not present both on the same page.

You are right that there isn't a perfect place to present this, but it's really an expert feature that most users don't need to learn about. There is a section on "configurable optimisations" on the "source files and compilation" page that describes the compiler directives and all that. I think we can briefly present the branch hints in there. They aren't really directives, but they activate an optimisation feature, locally for a given condition. It's not too far fetched to find them there.

I also don't think we need to extensively describe the automatic branch hints for failure branches. We don't describe other C code optimisations in that level of detail either. And it's really just that, an optimisation detail.

matusvalo:
Created alternative PR #7820

PR #7820, opened 2026-07-13
Document cython.likely() and cython.unlikely()

matusvalo:
Related to https://github.com/cython/cython/pull/7812

### DIFF
--- a/docs/src/userguide/source_files_and_compilation.rst
+++ b/docs/src/userguide/source_files_and_compilation.rst
@@ -872,8 +872,17 @@ Configurable optimisations
 Branch hints
 ^^^^^^^^^^^^
 
-Branch hints tell the compiler which path of a conditional branch is more commonly taken.
-This information helps the compiler and CPU generate better machine code:
+Cython automatically inserts branch hints in some places where the result of a condition
+is clearly skewed, e.g. when an if-condition can only lead to an exception being raised
+and is therefore very unlikely to be taken and not a performance critical path.
+Branch hints tell the C/C++ compiler which path of a conditional branch is more likely
+and should be translated into a more efficient binary code sequence than the other
+(unlikely) branch, if it can.
+
+In performance critical cases where neither Cython nor the C compiler can decide this
+themselves because the outcome is not obvious from the code (they cannot know the
+data that the code will commonly process), users can manually provide branch hints
+for an if-clause since Cython 3.3:
 
 - ``cython.likely()`` - Marks a condition as expected to be true
 - ``cython.unlikely()`` - Marks a condition as expected to be false
@@ -881,13 +890,13 @@ This information helps the compiler and CPU generate better machine code:
 .. code-block:: python
 
     if cython.likely(x > threshold):
-        # This branch is expected to execute frequently
+        # This branch is expected to execute frequently.
         process_common_case(x)
     else:
         process_rare_case(x)
 
-The compiler uses these hints to optimize instruction layout and branch prediction, potentially
-improving cache hit rates and reducing pipeline flushes.
+The compiler can use these hints to optimize instruction layout and branch prediction,
+potentially improving cache hit rates and reducing pipeline flushes.
 
 .. warning::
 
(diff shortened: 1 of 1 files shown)


matusvalo:
Few notes to the PR:
1. I agree with the reason why to abandon #7818
2. I am not 100% comfortable with the placement of the docs  - it's kind of strange place but currently there is no good place for that
3. As long-term ideal solution I would one of following:
   - Create reference documentation of `cython` "module"
   - Create separate page where would be "Advanced" stuff documented
4. (not full related one) source files and compilation page contains branch hints [compilation option](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#Cython.Compiler.Options.gcc_branch_hints). This option is broken (it does not remove hints just part of them). I would remove it from docs and raise deprecation warning when user tries to set it...

docs/src/userguide/source_files_and_compilation.rst
```diff
@@ -898,6 +869,37 @@ Configurable optimisations
     completely wrong.
     Disabling this option can also reduce the code size.
 
+Branch hints
+^^^^^^^^^^^^
+
+Branch hints tell the compiler which path of a conditional branch is more commonly taken.
+This information helps the compiler and CPU generate better machine code:
```
scoder: I think we can connect this better to the previous section to make it feel less alien here.
```suggestion
Branch hints
^^^^^^^^^^^^

Cython automatically inserts branch hints in some places where the result of a condition
is clearly skewed, e.g. when an if-condition can only lead to an exception being raised
and is therefore very unlikely to be taken and not a performance critical path.
Branch hints tell the C/C++ compiler which path of a conditional branch is more likely
and should be translated into a more efficient binary code sequence than the other
(unlikely) branch, if it can.

In performance critical cases where neither Cython nor the C compiler can decide this
themselves because the outcome is not obvious from the code (they cannot know the
data that the code will commonly process), users can manually provide branch hints
for an if-clause since Cython 3.3:
```

docs/src/userguide/source_files_and_compilation.rst
```diff
@@ -898,6 +869,37 @@ Configurable optimisations
...
+.. code-block:: python
+
+    if cython.likely(x > threshold):
+        # This branch is expected to execute frequently
+        process_common_case(x)
+    else:
+        process_rare_case(x)
+
+The compiler uses these hints to optimize instruction layout and branch prediction, potentially
+improving cache hit rates and reducing pipeline flushes.
```
scoder:
```suggestion
The compiler can use these hints to optimize instruction layout and branch prediction,
potentially improving cache hit rates and reducing pipeline flushes.
```

docs/src/userguide/source_files_and_compilation.rst
```diff
@@ -898,6 +869,37 @@ Configurable optimisations
...
+.. warning::
+
+   Branch hints are for **rare and very clear use cases**, not for guessing. Only use them when:
+
+   - You have concrete profiling data showing the actual branch frequency
+   - You have domain knowledge that makes the prediction obvious
+
+   Incorrect hints can make code slower, not faster. The compiler's own heuristics (e.g.
+   `PGO <https://en.wikipedia.org/wiki/Profile-guided_optimization>`_) are often
+   better than manual guessing.
```
scoder:
```suggestion
   Incorrect hints will make your code slower instead of faster, and can silently become
   misleading over time when data, requirements or surrounding/calling code change.
   The compiler's own heuristics are often better than manual guessing, especially
   when based on real profiling data with
   `PGO <https://en.wikipedia.org/wiki/Profile-guided_optimization>`_.
```

docs/src/userguide/source_files_and_compilation.rst
```diff
@@ -898,6 +869,37 @@ Configurable optimisations
...
+Branch hints tell the compiler which path of a conditional branch is more commonly taken.
+This information helps the compiler and CPU generate better machine code:
+
+- ``cython.likely()`` - Marks a condition as expected to be true
+- ``cython.unlikely()`` - Marks a condition as expected to be false
+
+.. code-block:: python
+
+    if cython.likely(x > threshold):
+        # This branch is expected to execute frequently
```
scoder:
```suggestion
        # This branch is expected to execute frequently.
```

docs/src/userguide/source_files_and_compilation.rst
scoder: I looked for a better link than the generic Wikipedia article (the English version of which mostly mentions MSVC) and couldn't really find one that is likely to stay alive for a longer time. Especially the C compiler's own documentation on PGO is really unimpressive.

For clang, it's a lengthy section in
https://clang.llvm.org/docs/UsersManual.html#profile-guided-optimization

The GCC docs even split it across two pages describing instrumentation and optimisation separately.
https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fprofile-generate
https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-fprofile-use

matusvalo: I chose wikipedia because it gives high level overview and the link is stable (should not change over time). I found C compilers can change overtime... Do you want me to change the link? I think, in current shape, if user wants to use PGO, he needs to do more deep study in any case...

PR #7821, opened 2026-07-14
Improve type annotations in FlowControl.py

matusvalo:


Cython/Compiler/FlowControl.pxd
```diff
@@ -3,19 +3,19 @@ cimport cython
...
-    cdef public object i_input
-    cdef public object i_output
-    cdef public object i_gen
-    cdef public object i_kill
-    cdef public object i_state
+    cdef object i_input
+    cdef object i_output
+    cdef object i_gen
+    cdef object i_kill
+    cdef object i_state
```
scoder, line 18: I haven't investigated the effect on the C code, so I can't say if it's better or worse, but it _might_ be worth typing these as `int` by now.
```suggestion
    cdef int i_input
    cdef int i_output
    cdef int i_gen
    cdef int i_kill
    cdef int i_state
```
EDIT: I "obviously" meant Python `int`, not what I wrote above…

scoder (approved):

PR #7823, opened 2026-07-15
Generate error when special attribute __dict__ is declared as public or readonly

donlon:
Hi,

In extension types, as a special attribute, \_\_dict__ shouldn't be declared as `public` or `readonly` (similar to \_\_weakref__). Currently Cython compiler does not check this restriction, and code such as

```
cdef class C:
    cdef public dict __dict__
```

generates code failed to compile due to redefinition errors.

This commit adds a compiler error for this case, so users can identify the mistake quickly with a clearer diagnostic.


da-woods:
Having an error here seems like a good idea to me and the tests look fine.

I'd personally prefer to cut a tiny bit of duplication e.g.:

```
                if obj_type.is_extension_type and entry.name in ["__dict__", "__weakref__"]:
                    error(self.pos, f"Illegal use of special attribute {entry.name}")
```

(Github interface doesn't seem to let me make it as a code suggestion for some reason)

da-woods:
The failing coverage test definitely isn't to do with this PR

donlon:
Yeah, now it does look more clear.

da-woods:
Thanks - I've set this to merge once the tests pass

PR #7828, opened 2026-07-16
Handle long C strings in MSVC

### LINKED ISSUE #7824: [BUG] str/bytes constants longer than 65535 bytes are silently truncated to len % 2**16 when built with MSVC
### Describe the bug

I hit this in a production system on Windows: a string constant defined in a .pyx file came out truncated at 2**16 characters after building the extension, with a completely clean build log. It affects both `"..."` and `b"..."` literals. The same code works fine on Linux (gcc), which is why it took me a while to track down.

After digging in, the root cause turned out to be MSVC, not the generated code — cl.exe silently emits only `len mod 65536` bytes of a string literal's initializer data when the (concatenated) literal is longer than 65535 bytes. No error, no warning, even at /W3. But since Cython already splits long literals into 2000-char pieces to work around another MSVC limit (`StringEncoding.split_string_literal`), I think Cython is the right place to also work around this one — or at least fail loudly instead of shipping corrupted constants.

I searched the tracker for similar issues; the closest are #4924 (link.exe hangs with many constants) and #5290 (big integer literals fail on MSVC), but neither covers this. I also checked current master:
`split_string_literal` still only does the 2000-char piece splitting and nothing limits the total literal size, so master generates the same affected code as the 3.0.10 I tested with.

```bat
python -c "a='A'*(2**16+3); print(f'a=\"{a}\"\nprint(a)')" > main.pyx
python -c "import main"
```

with this setup.py:

```python

ext = Extension(name="main", sources=["main.pyx"], language="c++")
setup(ext_modules=cythonize(ext, compiler_directives={'language_level': 3}))
```

Actual output:

```
C:\Users\win10\Desktop\cythonbugtest>python -c "import main"
AAAA
```

Exactly 4 characters, because 65539 chars + 1 NUL terminator = 65540 bytes and 65540 mod 65536 = 4. The build log contains no warning or error at all.

### Expected behaviour

`print(a)` prints all 65539 characters (as it does on Linux with gcc, and as plain CPython does when the same file is executed as a .py module). If the target compiler genuinely can't represent the constant, I'd expect the build to fail with an error rather than produce a silently corrupted extension.

### OS

Windows 10 x64

### Python version

3.11.9 (MSC v.1938, 64 bit)

### Cython version

3.0.10

### Additional context

_No response_

da-woods:
Fixes #7824

This handles Python strings, or things that get added to the stringtab. It can't (easily) handle a lot of other places, like docstrings, where the string literal is just passed directly. This seems to account for almost all uses of `as_c_string_literal` so I've just disallowed generating the array definition from that. The other case that it can't easily handle is `BytesNodes.generate_evaluation_code`

```
            # not const => use plain C string literal and cast to mutable type
            literal = self.value.as_c_string_literal()
            # C++ may require a cast
            result = typecast(self.type, PyrexTypes.c_void_ptr_type, literal)
```

I've made it conditional on MSVC. Compiling these array definitions is hugely slow on GCC (I got bored waiting for it to finish so I don't actually know how long...).


scoder:
Looks like you beat me to it. I had come up with this:
https://github.com/cython/cython/compare/master...scoder:cython:msvc_long_strings

I actually had a more efficient escaping version along the way but dropped it because the user C strings come in with a prepared escaped byte string version and I didn't want to change that. Thus the `re` version.

I kept the C string writing part in `Code.py` because it didn't feel like it belonged into `StringEncoding.py`. I also extended the string constant in `msvc_string` test because I noticed that the compressed string is only some 44k long, not 65k+.

da-woods:
It's probably better to use your version. I'm fairly sure I've messed up strings with escaped characters here and it doesn't seem worth fixing if you have a version that works.

The one thing I'd take from this version is the `#ifdef _MSC_VER`. Using the array took GCC compile time from <1s  to >5mins on my laptop (msvc compile times aren't so dramatically affected I think)

PR #7830, opened 2026-07-17
Protect long C strings from silent truncation in MSVC

### LINKED ISSUE #7824: [BUG] str/bytes constants longer than 65535 bytes are silently truncated to len % 2**16 when built with MSVC

scoder:
According to MS's documentation in https://learn.microsoft.com/en-us/cpp/c-language/maximum-string-length?view=msvc-170 MSVC fails to handle C strings longer than "roughly 65535 bytes", which "can be constructed by concatenating strings". We already did the latter for years, but failed to help MSVC over the first limitation.

Borrows good ideas from https://github.com/cython/cython/pull/7828

Closes https://github.com/cython/cython/issues/7824


Cython/Compiler/Code.py
```diff
@@ -1895,11 +1895,7 @@ def generate_string_constants(self):
         c_consts.sort()
         decls_writer = self.parts['string_decls']
         for _, cname, escaped_value in c_consts:
-            cliteral = StringEncoding.split_string_literal(escaped_value)
-            decls_writer.putln(
-                f'static const char {cname}[] = "{cliteral}";',
-                safe=True,  # Braces in user strings are not for indentation.
-            )
+            _write_cstring_const(decls_writer, escaped_value, cname, len(escaped_value))
```
da-woods, line 1901: It probably doesn't really matter (since `length` just picks the cut-off) but should `length` be for the unescaped value?

da-woods (approved):

Cython/Compiler/Code.py
scoder: Preferably yes, but the encoded value is prepared here. Let me see if I can get at the original value safely at this point.

scoder: So, no, we don't store the original byte string value (or its length) in the `StringConst` object, so we can't easily recover it here – short of passing it through the whole machinery.

It's safe to make the cut-off at a shorter string length, though, and it only makes a difference up to the actual 64k character length, from which on we'd use the split formats either way. It only means that we'll store some not so long strings in more C code than strictly necessary.

PR #7831, opened 2026-07-17
Prioritize Limited API check over PyPy and GraalPy check

da-woods:
Both PyPy and GraalPy have expressed an ambition to support the Limited API in future so for now, moving the check makes it easier to test and develop this support. I've added a warning since it's unlikely to work in practice for the moment.

### DIFF
--- a/Cython/Utility/ModuleSetupCode.c
+++ b/Cython/Utility/ModuleSetupCode.c
@@ -62,11 +62,7 @@
 // when doing version checks.
 #define __PYX_LIMITED_VERSION_HEX PY_VERSION_HEX
 
-#ifndef CYTHON_SKIP_NONCPYTHON_DETECTION
-#define CYTHON_SKIP_NONCPYTHON_DETECTION 0
-#endif
-
-#if defined(GRAALVM_PYTHON) && !CYTHON_SKIP_NONCPYTHON_DETECTION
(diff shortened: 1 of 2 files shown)


da-woods:
@mattip This is my proposed macro to let you test it. It's pretty simple and cheap to add so I think it's worth doing even if it's probably a while before it works.

mattip:
Thanks. Makes sense for now.

scoder:
If the intention is to allow use of the Limited API instead, why not just add `&& !defined(Py_LIMITED_API)` to the PyPy and GraalPy sections?

In retrospect, moving the Limited API section to the front would have been more future proof, but that would now impact the git history. Not sure if it's worth that.

da-woods:
My assumption is that right now there's people that definite `Py_LIMITED_API` unconditionally in their build scripts. And that for now the right thing to do for them is to mostly ignore that and take the PyPy path.

In a future where PyPy actually supports it properly, the right thing to do is to detect it first.

scoder:
> definite `Py_LIMITED_API` unconditionally in their build scripts

Hmm. But wouldn't it then be better to stop them early than to make them get used to having it in their build scripts?

da-woods:
I've moved the `Py_LIMITED_API` check first. I do agree that it's the logically consistent thing to do although I suspect it will catch a few people out.

PR #7832, opened 2026-07-17
Support parsing pxd files with parse_from_strings() (#7827)

### LINKED ISSUE #7827: [BUG] `parse_from_strings()` fails to parse .pxd files with default values of *
### Describe the bug

As part of adding a feature to `cython-lint`, I need to start parsing .pxd files, not just .pyx.

However, the `parse_from_strings` usage fails for some .pxd because of of the syntax for default values:

```
...
Cython.Compiler.Errors.CompileError: 
Error compiling Cython file:
...

cdef void simultaneous_sort(
    floating* values,
    intp_t* indices,
    intp_t n,
    bint use_three_way_partition=*,
                                  ^
sklearn/utils/_sorting.pxd:9:34: Expected an identifier or literal
```

```cython
from Cython.Compiler.TreeFragment import parse_from_strings, StringParseContext
from Cython.Compiler.Errors import init_thread

init_thread()
filename = 'sklearn/utils/_sorting.pxd'
with open(filename) as fd:
    code = fd.read()
context = StringParseContext(filename)
context.set_language_level(3)
tree = parse_from_strings(filename, code, context=context)
print(tree)
```

Where the file being parsed is https://github.com/scikit-learn/scikit-learn/blob/e67ae2f1412a4c67ce4c24061ce44af2af3d5bcb/sklearn/utils/_sorting.pxd

### Expected behaviour

Parse the file successfully.

### OS

Linux

### Python version

3.14

### Cython version

3.2.8

### Additional context

_No response_

itamarst:
Fixes #7827

### DIFF
--- a/Cython/Compiler/TreeFragment.py
+++ b/Cython/Compiler/TreeFragment.py
@@ -79,7 +79,7 @@ def parse_from_strings(name, code, pxds=None, level=None, initial_pos=None,
                      scope = scope, context = context, initial_pos = initial_pos)
     ctx = Parsing.Ctx(allow_struct_enum_decorator=allow_struct_enum_decorator)
 
-    if level is None or level == "module_pxd":
+    if level is None or level in ("module", "module_pxd"):
         in_pxd = (level == "module_pxd")
         tree = Parsing.p_module(scanner, in_pxd, module_name, ctx=ctx)
         tree.is_pxd = in_pxd
@@ -238,7 +238,7 @@ class TreeFragment:
             for key, value in pxds.items():
                 fmt_pxds[key] = fmt(value)
             mod = t = parse_from_strings(name, fmt_code, fmt_pxds, level=level, initial_pos=initial_pos)
-            if level is None:
+            if level is None or level in ("module", "module_pxd"):
                 t = t.body  # Make sure a StatListNode is at the top
             if not isinstance(t, StatListNode):
                 t = StatListNode(pos=mod.pos, stats=[t])
--- a/Cython/Compiler/TreeFragment.py
+++ b/Cython/Compiler/TreeFragment.py
@@ -79,13 +79,12 @@ def parse_from_strings(name, code, pxds=None, level=None, initial_pos=None,
                      scope = scope, context = context, initial_pos = initial_pos)
     ctx = Parsing.Ctx(allow_struct_enum_decorator=allow_struct_enum_decorator)
 
-    if level is None or level in ("module", "module_pxd"):
-        scanner.parse_comments = False
+    if level is None or level == "module_pxd":
         in_pxd = (level == "module_pxd")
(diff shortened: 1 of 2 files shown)


Cython/Compiler/TreeFragment.py
```diff
@@ -79,10 +79,12 @@ def parse_from_strings(name, code, pxds=None, level=None, initial_pos=None,
...
-        tree = Parsing.p_module(scanner, 0, module_name, ctx=ctx)
+    if level is None or level in ("module", "module_pxd"):
+        scanner.parse_comments = False
+        in_pxd = (level == "module_pxd")
+        tree = Parsing.p_module(scanner, in_pxd, module_name, ctx=ctx)
         tree.scope = scope
-        tree.is_pxd = False
+        tree.is_pxd = in_pxd
     else:
         tree = Parsing.p_code(scanner, level=level, ctx=ctx)
```
da-woods, line 88: My suggestion about `parse_comments` was for the `else` path since I think it basically never makes sense to have it on there. I think it wants leaving on in the conventional `p_module` path. At very least it's harmless there because it gets turned off inside `p_module` after it's used.

itamarst: Ah, I see.

da-woods:
That looks happier at least.

I assume you've checked that this lets you fix your cython-lint problem?

itamarst:
Not yet, almost there.

itamarst:
OK, `cython-lint`'s usage of `parse_from_strings` can, with this branch of Cython, parse every .pxd in scikit-learn, at least. And all tests pass for `cython-lint`.

itamarst:
Looks like the failed CI job is due to some upload error, so should pass if re-run.

Cython/Compiler/TreeFragment.py
```diff
@@ -79,11 +79,12 @@ def parse_from_strings(name, code, pxds=None, level=None, initial_pos=None,
                      scope = scope, context = context, initial_pos = initial_pos)
     ctx = Parsing.Ctx(allow_struct_enum_decorator=allow_struct_enum_decorator)
 
-    if level is None:
-        tree = Parsing.p_module(scanner, 0, module_name, ctx=ctx)
-        tree.scope = scope
-        tree.is_pxd = False
+    if level is None or level == "module_pxd":
```
da-woods: To me, it feels a little morally dubious to treat "module_pxd" differently from "module".

My proposed fix, which I think makes it a bit more symmetric, is at 86410b4cf2abcd76bbc06ac782d0c6c333840c49

itamarst: OK, I updated with this, thank you.

da-woods (approved):
I'm happy with this I think. I think this is an improvement in "doing what was probably intended" and should help out the cython-linter tool

PR #7835, opened 2026-07-18
Fix "plain" return from prange loop

### LINKED ISSUE #7587: [BUG] Return from prange does not work in simple cases
### Describe the bug

See the example

```cython
# distutils: extra_compile_args = -fopenmp
# distutils: extra_link_args = -fopenmp

from cython.parallel cimport prange

cdef int _f():
    for _ in prange(10, nogil=True):
        return 5

def f():
    return _f()
```

Calling `f()` returns `0`.

To me it doesn't look like `prange` is generating the relevant gotos in `switch(__pyx_parallel_why)`.

If you take a more complicated loop with an `else` block then it does work - this appears to be what we test.

### Expected behaviour

5

### OS

Linux

### Python version

3.13

### Cython version

Current master (but I've gone back a bit and this looks like a long-standing bug)

### Additional context

_No response_

da-woods:
The `else` case (which is what we tested) worked because we checked the loop exit mode to see if we should enter the "else" block. However, without the `else` we didn't check for early exits from the loop, and so didn't follow a "proper" return value and thus overwrote the return value with a default return value.

Fixes #7587

### DIFF
--- a/Cython/Compiler/Nodes.py
+++ b/Cython/Compiler/Nodes.py
@@ -10125,6 +10125,7 @@ class ParallelStatNode(StatNode, ParallelNode):
 
         self.any_label_used = False
         self.breaking_label_used = False
+        self.return_label_used = False
         self.error_label_used = False
 
         self.parallel_private_temps = []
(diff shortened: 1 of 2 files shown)


scoder (approved):
Looks like a fairly minimal fix. I'd have used local variables in the loop rather than perpetually reading and assigning instance attributes, but that's how the code is currently written.

PR #7837, opened 2026-07-19
Allow temps in the PyInit sub-functions

scoder:
Ties a declaration code writer to the `FuncState` that is created for the init sub-functions and uses it to inject the temp variable declarations when ending the function generation.


da-woods:
That all looks reasonable to me.

PR #7838, opened 2026-07-19
Deduplicate constant frozensets

### LINKED ISSUE #2741: Cache constant frozen-sets
Some set constants can be replaced by frozensets, e.g. in cases like `x in {1,2,3}`. This would allow them to become global constants, created once at module init time. There is already a framework for merging tuple and string constants, see `TupleNode`. The `ConstantFolding` transformation could detect these cases and set a flag attribute on the `SetNode` instances that would trigger constant frozenset behaviour.

scoder:
Original patch by Zhenbo Li in https://github.com/cython/cython/pull/4098

Closes https://github.com/cython/cython/issues/2741
Closes https://github.com/cython/cython/pull/4098

### DIFF
--- a/Cython/Utility/Builtins.c
+++ b/Cython/Utility/Builtins.c
@@ -672,12 +672,12 @@ static CYTHON_INLINE PyObject *__Pyx_PyDict_SetDefault(PyObject *d, PyObject *ke
 
 //////////////////// pyfrozenset_new.proto ////////////////////
 
-static CYTHON_INLINE PyObject* __Pyx_PyFrozenSet_New(PyObject* it);
+static PyObject* __Pyx_PyFrozenSet_New(PyObject* it);
 
 //////////////////// pyfrozenset_new ////////////////////
 //@requires: ObjectHandling.c::PyObjectCallNoArg
 
-static CYTHON_INLINE PyObject* __Pyx_PyFrozenSet_New(PyObject* it) {
+static PyObject* __Pyx_PyFrozenSet_New(PyObject* it) {
     if (it) {
         PyObject* result;
 #if CYTHON_COMPILING_IN_PYPY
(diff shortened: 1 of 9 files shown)


da-woods (commented):
Maybe not for this PR, but I wonder if doing `__Pyx_PyFrozenSet_FromArray` (in the same way as we have list and tuple from array functions) would be more efficient than `PyFrozenSet_New`.  Mainly because it avoids an intermediate container, and because `PyFrozenSet_New` doesn't particularly look to have fast-paths for list/tuple iterators.

That probably matters more for the non-literal version (which may be used more than once) than for the module-init version

da-woods:
Other than that thought, it looks basically fine to me.

scoder:
> `__Pyx_PyFrozenSet_FromArray`

Right, `PySet_Add()` also works on frozenset, so this would save code and potentially time. I'll try it.

Cython/Utility/Builtins.c
```diff
@@ -723,6 +723,46 @@ static CYTHON_INLINE PyObject* __Pyx_PyFrozenSet_New(PyObject* it) {
...
 
+//////////////////// pyfrozenset_fromarray.proto ////////////////////
+
+static PyObject* __Pyx_PyFrozenSet_FromArray(PyObject** values, Py_ssize_t length);
+
+//////////////////// pyfrozenset_fromarray ////////////////////
+
+static PyObject* __Pyx_PyFrozenSet_FromArray(PyObject** values, Py_ssize_t length) {
+#if CYTHON_COMPILING_IN_PYPY
+    // PyPy currently lacks PyFrozenSet_New()
```
da-woods: Might be worth double-checking that this is still true (if you haven't already done so): https://github.com/pypy/pypy/commit/32449ff30530abd6f1b7979f954183e9a967ddad

scoder: Thanks, I hadn't verified it yet. According to their release tags, it's been there since PyPy3.5 (v5.9.0), which we don't support any more. I'll remove both special cases.

scoder:
One issue with deduplicating `frozenset()` is that `PySet_Add()` still works on it, so users can do
```cython
fs = frozenset()
PySet_Add(fs, x)
```
and that will change all empty frozensets in the module. It's not only an issue for empty frozensets but IMHO most likely to appear there. Users need to explicitly call `PyFrozenSet_New()` if they want to use `PySet_Add()` on the result.

I think that's an acceptable limitation, but one that is worth documenting somewhere since CPython doesn't currently deduplicate frozensets (and the empty one isn't a singleton there), so this is specific to Cython. I'd say it's only slightly worse than tuples, for which you could also call `PyTuple_SET_ITEM()` on a deduplicated constant.

Another issue is refcounting since we effectively take away the possibility to create thread-local constant frozensets. I'll mark them as immortal to help with that.

scoder:
I also considered excluding frozensets at the module level similar to how we avoid constants and use a slower construction for tuples. But my intuition tells me that they are most likely to end up as user level constants either way, so we'd gain almost nothing. I'll not special case them for now.

da-woods:
> I also considered excluding frozensets at the module level similar to how we avoid constants and use a slower construction for tuples.

My (untested) feeling is that this is less relevant now that we build the tuples from an array instead of using `PyTuple_Pack` (where I think the vararg handling was noticeably slow). Especially if the C compiler can see the `_FromArray` definition and choose to expand the loop or not depending on the local optimization settings.

scoder:
I think this is ready for 3.3. Any more suggestions or objections?

CI failed to run for the PR, apparently due to a GHA hickup, but it validated on the branch.

PR #7839, opened 2026-07-19
Try to fix errors on windows in shared utility tests

da-woods:
This is an attempt to fix the intermittent

```
LINK : fatal error LNK1158: cannot run 'rc.exe'
```

that sometimes happens on Windows. It isn't well-tested (because I haven't managed to reproduce the issue locally). My theory is that it's to do with files not being cleaned up correctly between the steps. Therefore, I removed some Windows-specific file names, and also the entire build directory.

Not at all sure it's going to work, but I think it's worth a try.

### DIFF
--- a/tests/run/shared_utility_module.srctree
+++ b/tests/run/shared_utility_module.srctree
@@ -26,17 +26,20 @@ PYTHON -c "import pkg1.pkg11.feature_extensiontypes as ext; assert hasattr(ext.E
 ######## delete_generated_files.py ########
 
 import os
+import shutil
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


da-woods:
There aren't any errors on the Windows tests, but that doesn't prove anything because it is intermittent. I'll merge this in the fairly near future and I guess we find out if it helps as more things get run.

scoder:
Can't be wrong to clean up better.

da-woods:
Unfortunately doesn't seem to have fixed the intermittent issue: https://github.com/cython/cython/actions/runs/29738386312/job/88344388527. I still think it's probably an improvement in principle though

PR #7842, opened 2026-07-20
Add cython generate-shared subcommand for Cython 3.3

raza-khan0108:
# Add `cython generate-shared` subcommand for Cython 3.3

## Summary

As suggested in https://github.com/cython/cython/pull/7808, this PR extracts only the **`generate-shared`** subcommand (`cython generate-shared OUTPUT_FILE`) from the broader CLI hierarchy refactoring so that it can be included in the upcoming **Cython 3.3** release. The comprehensive CLI cleanup (`compile` and `build` subcommands) is deferred to Cython 3.4 to allow more time for design refinement.

### 1. New `generate-shared` Subcommand
- Added `cython generate-shared OUTPUT_FILE` along with the feature selection options:
  - `--shared-only FEATURES` (`options.shared_utility_features_enabled`)
  - `--shared-exclude FEATURES` (`options.shared_utility_features_disabled`)
- Retained full backward compatibility for the legacy flat option `--generate-shared=OUTPUT_FILE` on the main parser.

### 2. Restricted Option Set for Subcommand (`CmdLine.py`)
- Extracted common option setup into `add_compile_arguments(parser, include_sources=True, for_shared=False)`.
- When `for_shared=True`, the following compile options are strictly excluded as they do not apply to shared module generation:
  - `--embed`, `--embed-modules`, `-p` / `--embed-positions`
  - `-2`, `-3`, `--3str`
  - `--lenient`
  - `--capi-reexport-cincludes`
  - `-E` / `--compile-time-env`
- `--version` (`-V`) is restricted to the global root parser level (`cython --version generate-shared ...` works; `cython generate-shared --version` raises an `argparse` unknown option error).

### 3. Subcommand Routing & Error Prevention
- Updated `parse_command_line` to route `generate-shared` invocations to the focused subparser while enforcing strict option boundaries.
- Added checks in `parse_command_line_raw` to reject forbidden options (`--embed*`) and catch negative option strings (e.g. `-2`, `-3`) from being consumed as positional file paths when passed to the subcommand.

### 4. Unit Testing (`TestCmdLine.py`)
- Added `test_generate_shared_subcommand` verifying positional `OUTPUT_FILE` extraction, feature filtering (`--shared-only`/`--shared-exclude`), and compatibility with valid compile flags (`-I`, `-l`, `-D`, `-a`, `-+`, `-X`, `--fast-fail`, `-Werror`, `--cleanup`).
- Added exhaustive error test cases in `test_errors` ensuring that extra `.pyx` source files and every excluded compile option raise exact `argparse` errors when passed inside `generate-shared`.

## Verification

All command-line and compiler tests pass without error:
- `python -m pytest Cython/Compiler/Tests/TestCmdLine.py -v` (53 passed)
- `python -m unittest discover -v Cython.Compiler.Tests` (137 passed)

### DIFF
--- a/Cython/Compiler/CmdLine.py
+++ b/Cython/Compiler/CmdLine.py
@@ -102,18 +102,10 @@ class CythonArgumentParser(ArgumentParser):
         pass
 
 
-def create_cython_argparser():
-    description = "Cython (https://cython.org/) is a compiler for code written in the "\
-                  "Cython language.  Cython is based on Pyrex by Greg Ewing."
-
-    parser = CythonArgumentParser(
-        description=description,
-        argument_default=SUPPRESS,
-        formatter_class=RawDescriptionHelpFormatter,
-    )
-
-    parser.add_argument("-V", "--version", dest='show_version', action='store_const', const=1,
-                      help='Display version number of cython compiler')
+def add_compile_arguments(parser, include_sources=True, for_shared=False):
+    if not for_shared:
+        parser.add_argument("-V", "--version", dest='show_version', action='store_const', const=1,
+                          help='Display version number of cython compiler')
     parser.add_argument("-l", "--create-listing", dest='use_listing_file', action='store_const', const=1,
                       help='Write error messages to a listing file')
     parser.add_argument("-I", "--include-dir", dest='include_path', action='append',
@@ -127,9 +119,10 @@ def create_cython_argparser():
                       help='Compile all source files (overrides implied -t)')
     parser.add_argument("-v", "--verbose", dest='verbose', action='count',
                       help='Be verbose, print file names on multiple compilation')
-    parser.add_argument("-p", "--embed-positions", dest='embed_pos_in_docstring', action='store_const', const=1,
-                      help='If specified, the positions in Cython files of each '
-                           'function definition is embedded in its docstring.')
+    if not for_shared:
+        parser.add_argument("-p", "--embed-positions", dest='embed_pos_in_docstring', action='store_const', const=1,
+                          help='If specified, the positions in Cython files of each '
+                               'function definition is embedded in its docstring.')
     parser.add_argument("--cleanup", dest='generate_cleanup_code', action='store', type=int,
                       help='Release interned objects on python exit, for memory debugging. '
                            'Level indicates aggressiveness, default 0 releases nothing.')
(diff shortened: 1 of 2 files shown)


Cython/Compiler/CmdLine.py
```diff
@@ -221,11 +208,68 @@ def create_cython_argparser():
...
+    add_compile_arguments(parser, include_sources=True, for_shared=False)
+
+    parser.add_argument("--generate-shared", dest='shared_c_file_path', action='store', type=str,
+                        help='Generates shared module with specified name.')
+    parser.add_argument("--shared-only", dest='shared_utility_features_enabled',
+                        action=ParseCommaListAction, type=str, metavar='FEATURES',
+                        help='Comma separate list of features to move to the shared module exclusively.')
+    parser.add_argument("--shared-exclude", dest='shared_utility_features_disabled',
+                        action=ParseCommaListAction, type=str, metavar='FEATURES',
+                        help='Comma separate list of features to exclude from the shared module.')
```
scoder: These are new options in Cython 3.3. We don't need to add them to the existing `--generate-shared` option since the intention is to promote the new sub-command.
```suggestion
```

Cython/Compiler/CmdLine.py
```diff
@@ -221,11 +208,68 @@ def create_cython_argparser():
...
+        formatter_class=RawDescriptionHelpFormatter,
+    )
+    generate_parser.add_argument('shared_c_file_path', metavar='OUTPUT_FILE', type=str,
+                                 help='Path to the generated shared C/C++ file.')
+    generate_parser.add_argument("--shared-only", dest='shared_utility_features_enabled',
+                                 action=ParseCommaListAction, type=str, metavar='FEATURES',
+                                 help='Comma separate list of features to move to the shared module exclusively.')
+    generate_parser.add_argument("--shared-exclude", dest='shared_utility_features_disabled',
+                                 action=ParseCommaListAction, type=str, metavar='FEATURES',
+                                 help='Comma separate list of features to exclude from the shared module.')
```
scoder: In the context of the sub-command, we can shorten the option names.
```suggestion
    generate_parser.add_argument("--only", dest='shared_utility_features_enabled',
                                 action=ParseCommaListAction, type=str, metavar='FEATURES',
                                 help='Comma separate list of features to move to the shared module exclusively.')
    generate_parser.add_argument("--exclude", dest='shared_utility_features_disabled',
                                 action=ParseCommaListAction, type=str, metavar='FEATURES',
                                 help='Comma separate list of features to exclude from the shared module.')
```

Cython/Compiler/CmdLine.py
```diff
@@ -263,9 +317,50 @@ def filter_out_embed_options(args):
...
+        '-E', '--compile-time-env', '--module-name', '--generate-shared', '--shared',
+        '--shared-only', '--shared-exclude', '-z', '--pre-import'
+    }
+
+    is_subcommand = False
+    i = 0
+    while i < len(args):
+        arg = args[i]
+        if arg in ('generate-shared',):
+            is_subcommand = True
```
scoder: Rather than applying the complex heuristic below, why not just require the sub-command to be the first command line argument? It's the first time we add a sub-command, so we cannot break existing usages of it.

Cython/Compiler/CmdLine.py
```diff
@@ -221,11 +208,68 @@ def create_cython_argparser():
             option_name = name.replace('_', '-')
             parser.add_argument("--" + option_name, action='store_true', help=SUPPRESS)
 
+
+def create_cython_argparser():
+    description = "Cython (https://cython.org/) is a compiler for code written in the "\
+                  "Cython language.  Cython is based on Pyrex by Greg Ewing."
```
scoder: This is repeated and certainly worth moving into a common variable, maybe just a global constant.

scoder:
Thanks

raza-khan0108:
Welcome :)

PR #7844, opened 2026-07-20
Allow "cython.py_int" etc. for Python builtins 

scoder:
These were already available in Shadow.py but not in the compiled `cython` namespace.
There isn't really a good way otherwise to use them in a C type context because they are shadowed by C `int`, `float`, etc.

With the recent optimisations for the Python `int` and `float` data types, there is increased interest in using them, although probably mostly in situation where users control the type (and not as input argument).

### DIFF
--- a/Cython/Compiler/CythonScope.py
+++ b/Cython/Compiler/CythonScope.py
@@ -4,9 +4,20 @@ from .UtilityCode import CythonUtilityCode
 from .Errors import error
 from .Scanning import StringSourceDescriptor
 from . import MemoryView
+from . import Builtin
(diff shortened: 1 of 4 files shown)


da-woods:
It does seem useful to expose them. 

I think we should either remove `py_long` or expose it consistent with the others. I'm not sure what use it is being in Shadow.py but not being understood as a type. Given that it's legacy, removing it probably makes most sense?

scoder:
> I'm not sure what use it is being in Shadow.py but not being understood as a type.

There is some surrounding logic associated with it, so I'll not remove it right now. But it's probably dead code.

PR #7848, opened 2026-07-21
Fix type inference for calling builtin types

scoder:
Type inference needs to do the same as type analysis.

This fixes a bug in preparation for https://github.com/cython/cython/pull/7780

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -6290,14 +6290,14 @@ class CallNode(ExprNode):
             if func_type is Builtin.type_type and (
                     entry and entry.is_builtin and entry.name in Builtin.types_that_construct_their_instance):
                 # calling a builtin type that returns a specific object type
-                self.result_ctype = py_object_type
                 self.may_return_none = False
                 if entry.name == 'float':
-                    # the following will come true later on in a transform
+                    # "float(...) -> double" will come true later on in a transform.
                     self.type = PyrexTypes.c_double_type
                     self.result_ctype = PyrexTypes.c_double_type
                 else:
                     self.type = Builtin.builtin_types[entry.name]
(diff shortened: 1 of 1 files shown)


Cython/Compiler/Builtin.py
```diff
@@ -833,17 +827,22 @@ def init_builtin_exceptions():
...
-                name, f"((PyTypeObject*)PyExc_{name})", utility_code=utility_code)
-        else:
-            exc_type = type_entry.type
+        objstruct_cname = None
+        if name == 'BaseException':
+            objstruct_cname = "PyBaseExceptionObject"
+        elif name == 'Exception':
+            objstruct_cname = "PyBaseExceptionObject"
+
+        assert builtin_scope.lookup_here(name) is None  # should no be declared above
```
da-woods:
```suggestion
        assert builtin_scope.lookup_here(name) is None  # should not be declared above
```

?

Cython/Compiler/ExprNodes.py
```diff
@@ -6249,7 +6249,7 @@ def infer_type(self, env):
                     elif func_name == 'bool':
                         return PyrexTypes.c_bint_type
                     elif func_name in Builtin.types_that_construct_their_instance:
-                        return result_type
+                        return Builtin.builtin_types[function.entry.name]
```
da-woods: If I understand right, this essentially falls back to the plain builtin type rather than the specialized one (e.g. `list` instead of `list[int]`

Cython/Compiler/ExprNodes.py
scoder, line 6299: Later in `.analyse_types()`, we do this. It should probably line up with `infer_type()`. It seems worth a look if we can't unify the two more, but that's beyond the scope of this PR..

PR #7849, opened 2026-07-21
Document Python builtin types

matusvalo:
Related to #7844


matusvalo:
I suppose this PR is now not needed anymore since introduction of 777ac6c

docs/src/userguide/language_basics.rst
```diff
@@ -386,7 +346,7 @@ Note that Cython uses array access for pointer dereferencing, as ``*x`` is not v
 whereas ``x[0]`` is.
 
 Also, the Python types ``list``, ``dict``, ``tuple``, etc. may be used for
-static typing, as well as any user defined :ref:`extension-types`.
+static typing, their specialized versions using subscripts and as well as any user defined :ref:`extension-types`.
```
```suggestion
static typing (including their specialized versions with subscripted item types),
as well as any user defined :ref:`extension-types`.
```

docs/src/userguide/language_basics.rst
```diff
@@ -397,12 +357,14 @@ For example
...
+                bar: list[str] = ['Monty', 'Python']
 
     .. group-tab:: Cython
 
         .. code-block:: cython
 
             cdef list foo = []
+            cdef list[str] bar = ['Monty', 'Python']
 
 This requires an *exact* match of the class, it does not allow subclasses.
```
```suggestion
These declarations require an *exact* match of the class, they do not allow subclasses.
```

docs/src/userguide/language_basics.rst
```diff
@@ -315,57 +315,17 @@ all the standard C types, namely ``char``, ``short``, ``int``, ``long``,
...
 
-.. list-table:: Numeric Types
-   :widths: 25 25
-   :header-rows: 1
+        .. csv-table:: C Numeric types
+            :file: numeric_types.csv
+            :header-rows: 1
+            :class: longtable
+
+and builtin python types:
```
```suggestion
and builtin Python types:
```

docs/src/userguide/python_types.csv
```diff
@@ -0,0 +1,16 @@
...
+``bytes``,``bytes``
+``bytearray``,``bytearray``
+``str``,``str``
+``memoryview``,``memoryview``
+``list``,``list``
+``tuple``,``tuple``
+``set``,``set``
+``frozenset``,``frozenset``
+``dict``,``dict``
+``frozendict``,``frozendict``
```
scoder: I'm not sure these are really worth a table. And it's also slightly misleading. The `cython.py_*` names can equally be used in both, and it doesn't explain when or why users should write one or the other. I hope that's better explained by the paragraph that I added in https://github.com/cython/cython/commit/777ac6c2dbed5c4cbb3ab01b1a8b738dca045bc9

matusvalo: Your paragraph definitely helps! I agree that it misleads that `cython.py_*` is not mentioned that can be used in pure python but should be there any need to do that?

For me (personally) more confusing is that one single type (e.g. `int`) has different meaning (cython as C int, pure python `int` python type etc...) and I can imagine that user can get easily lost (this is not critique it just a fact which is caused by the fact that Cython is not young project and has some legacy burden...).

I leave to you and @da-woods whether you think it is good change or not. If you feel that the table is too excessive and obvious feel free to close the PR.

docs/src/userguide/language_basics.rst
```diff
@@ -315,57 +315,17 @@ all the standard C types, namely ``char``, ``short``, ``int``, ``long``,
...
 
-.. list-table:: Numeric Types
-   :widths: 25 25
-   :header-rows: 1
+        .. csv-table:: C Numeric types
+            :file: numeric_types.csv
+            :header-rows: 1
+            :class: longtable
+
+and builtin Python types:
```
```suggestion
```

docs/src/userguide/language_basics.rst
```diff
@@ -315,57 +315,17 @@ all the standard C types, namely ``char``, ``short``, ``int``, ``long``,
...
-   * - ``Py_ssize_t``
-     - ``cython.Py_ssize_t``
-   * - ``Py_hash_t``
-     - ``cython.Py_hash_t``
-   * - ``Py_UCS4``
-     - ``cython.Py_UCS4``
+        .. csv-table:: Python builtin types
+            :file: python_types.csv
+            :header-rows: 1
+            :class: longtable
```
```suggestion
```

scoder (approved):

scoder:
I find the CSV table much more readable, so that's a good idea.

I've removed the Python types table because it carries little information and can be misleading.

PR #7850, opened 2026-07-22
Improve annotation in `Parsing.py`

matusvalo:
This PR introduces:
* `tuple[...]` subscripted annotations
* `unicode` -> `str` annotation update.

Renaming `unicode` to `str` was implemented because `unicode` cannot be used in .py files (importing fails). Hence, I decided to modernize it under assumption that `unicode` == `str`.

### DIFF
--- a/Cython/Compiler/Parsing.py
+++ b/Cython/Compiler/Parsing.py
@@ -3056,9 +3056,9 @@ struct_enum_union = cython.declare(frozenset, frozenset((
 
 
 @cython.cfunc
-def p_sign_and_longness(s: PyrexScanner) -> tuple[cython.py_int, cython.py_int]:
-    signed = 1
-    longness = 0
+def p_sign_and_longness(s: PyrexScanner) -> tuple[cython.int, cython.long]:
+    signed: cython.int = 1
+    longness: cython.long = 0
     while s.sy == 'IDENT' and s.systring in sign_and_longness_words:
         if s.systring == 'unsigned':
             signed = 0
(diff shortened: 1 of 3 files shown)


Cython/Compiler/Parsing.py
```diff
@@ -1317,7 +1317,7 @@ def p_ft_string_literal(s: PyrexScanner) -> tuple:
 
 
 @cython.cfunc
-def _append_escape_sequence(kind, builder, escape_sequence: str, s: PyrexScanner):
+def _append_escape_sequence(kind: str, builder, escape_sequence: str, s: PyrexScanner):
```
matusvalo, line 1320: `kind` is annotated as `str` instead of `Py_UCS4` because it can contain `''` which is not allowed for `Py_UCS4`.

Cython/Compiler/Parsing.py
```diff
@@ -3056,7 +3056,7 @@ def looking_at_dotted_name(s: PyrexScanner) -> cython.bint:
 
 
 @cython.cfunc
-def p_sign_and_longness(s: PyrexScanner) -> tuple:
+def p_sign_and_longness(s: PyrexScanner) -> tuple[cython.py_int, cython.py_int]:
     signed = 1
     longness = 0
```
```suggestion
def p_sign_and_longness(s: PyrexScanner) -> tuple[cython.int, cython.long]:
    signed: cython.int = 1
    longness: cython.long = 0
```

scoder (approved):

Cython/Compiler/Parsing.py
da-woods: In principle we could probably use `'\0'` instead of `''` to represent an empty kind string (which would let it be `Py_UCS4`). Although that's a bigger change that would involve changing it in a lot of places and might just make things a little less clear

matusvalo: The comment is just to explain why `str` is used instead of `Py_UCS4`. There is no intention to change the underlying logic to migrate to `Py_UCS4`...

PR #7852, opened 2026-07-23
Allow `cyfunction.__annotate__` to be pickled

### LINKED ISSUE #7846: [BUG] `cyfunction.__annotate__` instances can't be unpickled
### Describe the bug

[scikit-learn report this](https://github.com/scikit-learn/scikit-learn/issues/34525).

If you wrap a cyfunction in a wrapper function, then the wrapper function `__annotate__` is set to the result of `wrapped.__annotate__`, and this can't be reloaded.

It gives the slightly misleading error message `AttributeError: 'dict' object has no attribute '__annotate__'`.

Essentially the gist is that we create an method with `self` being the annotations dict named `__annotate__`. `__reduce__` for a method object is defined as returning `(__builtins.getattr, (self.__self__, self.__name__))` so when it unpickles it tries to call `getattr(dict, "__annotate__")`

See https://github.com/lesteve/scikit-learn-34525-cython-reproducer for a full example

### Expected behaviour

I think this does need to unpickle cleanly, since wrapped cyfunctions did before (within the ususal limits of cyfunction pickleability).

### OS

_No response_

### Python version

_No response_

### Cython version

master, 3.2.6

### Additional context

I think the challenge here is getting something that doesn't add too much extra code. Getting something that works should be possible, but it'd be nice not to have to add extra C classes for this.

da-woods:
Essentially to allow things that wrap cyfunctions to be pickled, since they will have a copy of `cyfunction.__annotate__`.

For a C method, the pickling mechanism looks up the method name on the type of the `self` argument.
This doesn't work when we use a dict as the `self` argument.

Instead I've used a `cyfunction` as the `self` argument, and created a function called `__annotate_cython__` that can be looked up.

The main notable consequence of this change is that it delays the evaluation of the annotations to when `__annotate__` is called rather than when the attribute is looked up. I think this is fine (especially since it's a cut-down implementation to make some basic functionality work). But it is a small change. It also makes it possible to set up some infinite
recursion, so I've added a little protection to that.

The alternative way of doing it would be to add an extra type for `cyfunction.__annotate__` some some custom pickling.

Fixes #7846

### DIFF
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -594,13 +594,26 @@ __Pyx_CyFunction_get_annotate_from_dict_if_exists(PyObject *op_in, PyObject **an
     return __Pyx_PyDict_GetItemRef(dict, PYIDENT("__annotate__"), annotate);
 }
 
+// if value==NULL, removes annotate from dict.
 static int
 __Pyx_CyFunction_set_annotate_in_dict_if_exists(PyObject *op_in, PyObject *value) {
     PyObject *dict;
     int dict_found;
     dict_found = __Pyx_CyFunction_get_dict_if_exists(op_in, &dict);
     if (!dict_found) return 0;
-    return PyDict_SetItem(dict, PYIDENT("__annotate__"), value);
+    if (value) {
+        return PyDict_SetItem(dict, PYIDENT("__annotate__"), value);
+    } else {
+#if !CYTHON_COMPILING_IN_LIMITED_API && PY_VERSION_HEX >= 0x030d0000
+        return PyDict_Pop(dict, PYIDENT("__annotate__"), NULL);
+#else
+        int result = PyDict_Contains(dict, PYIDENT("__annotate__"));
+        if (result == 1) {
+            result = PyDict_DelItem(dict, PYIDENT("__annotate__"));
+        }
+        return result;
+#endif
+    }
 }
 
 static int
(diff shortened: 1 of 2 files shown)


Cython/Utility/CythonFunction.c
```diff
@@ -668,34 +684,27 @@ __Pyx_CyFunction_get_annotations(PyObject *op_in, void *context) {
...
-#else
-    METH_FASTCALL,
-#endif
+static const PyMethodDef __Pyx_CyFunction_annotate_method = {
+    "__annotate_cython__",
+    __Pyx_CyFunction_annotate_impl,
+    METH_O,
     "Placeholder __annotate__ function to allow 'functools.wraps' to work "
     "on Cython functions."
 };
```
scoder: Is the docstring still appropriate?

scoder:
> The main notable consequence of this change is that it delays the evaluation of the annotations to when `__annotate__` is called rather than when the attribute is looked up.

The PEP says:
"The new attribute is called `__annotate__`, and is a reference to a function which computes and returns that object’s annotations dict."

So this change should make our implementation comply more and not less.

scoder (approved):

Cython/Utility/CythonFunction.c
da-woods: You you mean the name `__annotate__` in the docstring or something else?

I think that's still close enough to accurate - from a user point of view you access this function using the `__annotate__` attribute. Being able to look it up with `__annotate_cython__` is mostly an implementation detail that's just there to make `pickle_load` work.

Cython/Utility/CythonFunction.c
```diff
@@ -958,6 +981,7 @@ __Pyx_CyFunction_reduce(PyObject *m_in, PyObject *args)
 
 static PyMethodDef __pyx_CyFunction_methods[] = {
     {"__reduce__", (PyCFunction)__Pyx_CyFunction_reduce, METH_NOARGS, 0},
+    __Pyx_CyFunction_annotate_method,  // having __annotate_cython__ as an attribute makes it pickleable
```
scoder: My guess is that this is the line that MSVC is unhappy with.

da-woods:
I've picked this back to 3.2.x because it also affects that branch and is a regression since we added `__annotate__`. Commit: dfdb09b3f

PR #7853, opened 2026-07-25
Improve __Pyx_PyObject_CallMethod without vectorcall

da-woods:
i.e. mainly in old Limited API (but also if vectorcall is manually off).

Outside the Limited API there are a couple of functions which do this exactly. We should use them on the basis that they'll hopefully pick the fastest path.

Inside the Limited API we should prefer CallMethodObjArgs on the basis that it avoids vectorcall, or instantiating the bound method, and so is better than doing it outselves manually.

### DIFF
--- a/Cython/Utility/ObjectHandling.c
+++ b/Cython/Utility/ObjectHandling.c
@@ -2398,30 +2398,15 @@ static PyObject *__Pyx_Object_VectorcallMethodKwds(PyObject *name, PyObject *con
 static PyObject* __Pyx_PyObject_CallMethod0(PyObject* obj, PyObject* method_name); /*proto*/
 
 /////////////// PyObjectCallMethod0 ///////////////
-//@requires: PyObjectGetMethod
-//@requires: PyObjectCallOneArg
-//@requires: PyObjectCallNoArg
 
 static PyObject* __Pyx_PyObject_CallMethod0(PyObject* obj, PyObject* method_name) {
 #if CYTHON_VECTORCALL && (__PYX_LIMITED_VERSION_HEX >= 0x030C0000 || !CYTHON_COMPILING_IN_LIMITED_API)
     PyObject *args[1] = {obj};
-    // avoid unused functions
-    (void) __Pyx_PyObject_CallOneArg;
-    (void) __Pyx_PyObject_CallNoArg;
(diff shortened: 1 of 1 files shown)


Cython/Utility/ObjectHandling.c
```diff
@@ -2398,30 +2398,15 @@ static PyObject *__Pyx_Object_VectorcallMethodKwds(PyObject *name, PyObject *con
 static PyObject* __Pyx_PyObject_CallMethod0(PyObject* obj, PyObject* method_name); /*proto*/
 
 /////////////// PyObjectCallMethod0 ///////////////
-//@requires: PyObjectGetMethod
-//@requires: PyObjectCallOneArg
-//@requires: PyObjectCallNoArg
 
 static PyObject* __Pyx_PyObject_CallMethod0(PyObject* obj, PyObject* method_name) {
 #if CYTHON_VECTORCALL && (__PYX_LIMITED_VERSION_HEX >= 0x030C0000 || !CYTHON_COMPILING_IN_LIMITED_API)
```
da-woods, line 2403: I did wonder if it was worth keeping this path and not just using PyObject_CallMethodNoArgs.

I think it probably is (just to inline what Python does internally). But it's likely borderline.

scoder:
Looks good and saves a lot of our own code. If we're lucky, it might show up in the benchmarks for Py3.10, but I'd merge it either way.

Cython/Utility/ObjectHandling.c
scoder: At least in CPython, `PyObject_CallMethodNoArgs()` is an inline function that does pretty much this. It only pulls the trick of using `&obj` directly instead of creating a new 1-arg C array.
https://github.com/python/cpython/blob/b86a41cbf631c959d274ad6180cf0a0ac6f6e180/Include/cpython/abstract.h#L61-L66
But since it's not available in the Limited API, we'd still need the direct vectorcall either way.

scoder:
Hmm. I don't see the benchmarks being conclusive. Several got slower and some faster but none that I would have expected to be different. (I know, some micro-benchmarks are really flaky and there is generally a lot of noise in them…)
It's probably fine to just ignore the benchmarks.

PR #7854, opened 2026-07-25
Improve error handling in __Pyx_ExceptionGroupMatch

da-woods:
In principle a user class could override `BaseExceptionGroup.split` and therefore we shouldn't trust the format of the output of it (i.e. a proper check is better than an assert).

Exception messages are taken from the CPython implementation.

Also simplify call by using existing utility code.

### DIFF
--- a/Cython/Utility/Exceptions.c
+++ b/Cython/Utility/Exceptions.c
@@ -1035,6 +1035,9 @@ static void __Pyx_RaisePreppedException(PyObject *exc); /* proto */
 #endif
 
 /////////////////// ExceptStar ///////////////////////////////
+//@requires: ObjectHandling.c::PyObjectCallMethod1
+//@requires: ObjectHandling.c::RaiseErrorWithObjectTypes
+//@requires: ObjectHandling.c::RaiseErrorWithTypeAndVarargs
 
 static int __Pyx_ValidateStarCatchPatternElement(PyObject *pattern) {
     int is_subclass;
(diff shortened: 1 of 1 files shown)


Cython/Utility/Exceptions.c
```diff
@@ -1134,22 +1137,32 @@ static int __Pyx_ExceptionGroupMatch(PyObject *match_type, PyObject **current_ex
...
 
-        assert(PyTuple_CheckExact(pair));
-        #if CYTHON_ASSUME_SAFE_MACROS
-        // Just skip the assert without safe macros - it's a sanity check rather than important
-        assert(PyTuple_GET_SIZE(pair) == 2);
-        #endif
+        if (unlikely(!PyTuple_CheckExact(pair))) {
+            __Pyx_RaiseErrorWithObjectTypes1(PyExc_TypeError,
+                "%.1s" __Pyx_FMT_TYPENAME ".split must return a tuple, not " __Pyx_FMT_TYPENAME,
+                "", *current_exception, pair);
```
da-woods, line 1148: I've been a bit lazy here to avoid writing another variation of this function without a string argument. So I've just passed an empty string as the first argument.

scoder: Fine with me. Although you could have added a 1-line macro `__Pyx_RaiseErrorWithObjectTypes()` (without the "1" argument suffix) to the `RaiseErrorWithObjectTypes.proto` section to do the same. Using `"%.1s"` is actually a good idea, I should have done that for the `__Pyx_RaiseTypeErrorWithTypes()` macro there.

scoder: BTW, according to the Unicode formatting code in CPython, `"%.0s"` should also work, although it might actually end up doing a tiny bit more work than less. And I can't say how other implementations handle it. Probably better to stick with `"%.1s"` instead.

PR #7861, opened 2026-07-27
Report an error instead of crashing on cast to an unresolved typeof() type

### LINKED ISSUE #7683: [BUG] Cython.cast(cython.typeof(...), value) crashes when typeof() resolves to UnspecifiedType
### Describe the bug

`cython.typeof()` is currently tolerated in annotation contexts even when the type cannot be inferred, but the same unresolved type causes a compiler crash when used inside `cython.cast()`.

Minimal reproducible example:

```python
import cython

def f():
    s = ""
    y: cython.typeof(s)
    b = cython.cast(cython.typeof(y), s)
```

This crashes during compilation with:

```text
  File "....", line 11622, in analyse_types
    if self.type.create_from_py_utility_code(env):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'UnspecifiedType' object has no attribute 'create_from_py_utility_code'
```

However, the following version works correctly because the type can be resolved:

```python
import cython

def f():
    s: str = ""
    y: cython.typeof(s)
    b = cython.cast(cython.typeof(y), s)
```

### Current behavior

If the type cannot be inferred:

```python
s = ""
```

then Cython still allows:

```python
y: cython.typeof(s)
```

but crashes on:

```python
cython.cast(cython.typeof(y), s)
```

instead of gracefully ignoring the unresolved cast or emitting a regular compilation error.

```cython
import cython

def f():
    s: str = ""
    y: cython.typeof(s)
    b = cython.cast(cython.typeof(y), s)
```

### Expected behaviour

If `cython.typeof(...)` resolves to `UnspecifiedType`, `cython.cast()` should not crash the compiler.

A reasonable behavior would be to ignore the cast and return the operand unchanged, similarly to how unresolved `cython.typeof(...)` annotations are currently tolerated.

### OS

_No response_

### Python version

3.14

### Cython version

current main: 401b195debacf5e790c7d081d11e85025442eb16

### Possible fix

In `TransformBuiltinMethods.visit_SimpleCallNode`, the `cast` handling could special-case `UnspecifiedType` before creating the `TypecastNode`.

Something along these lines:

```python
class TransformBuiltinMethods(EnvTransform):
    ...

    def visit_SimpleCallNode(self, node):
        ...

        elif function == 'cast':
            if len(node.args) != 2:
                error(node.function.pos,
                      "cast() takes exactly two arguments and an optional typecheck keyword")
            else:
                type = node.args[0].analyse_as_type(self.current_env())
                if type:
                    #########################
                    if type.is_unspecified:
                        # Ignore unresolved cast instead of crashing
                        return node.args[1]
                    #########################
                    node = ExprNodes.TypecastNode(
                        node.function.pos,
                        type=type,
                        operand=node.args[1],
                        typecheck=False)
                else:
                    error(node.args[0].pos, "Not a type")
```

vsaraikin:
Closes #7683.

`cython.cast(cython.typeof(x), ...)` crashes the compiler when `typeof()` can't be resolved to a concrete type:

```python
import cython

def f():
    s = ""
    y: cython.typeof(s)
    b = cython.cast(cython.typeof(y), s)
```

```
AttributeError: 'UnspecifiedType' object has no attribute 'create_from_py_utility_code'
```

`TypecastNode.analyse_types()` rejects a few invalid cast targets (function types and so on) but not an unspecified one, so it falls through to the Python conversion path and calls `create_from_py_utility_code()` on `UnspecifiedType`, which doesn't have it.

The fix rejects an unspecified cast target with a normal compile error, recovering via `error_type` like the neighbouring checks do. That follows da-woods' point on the issue that `cast` is an explicit request and should fail rather than be silently ignored, while annotations stay tolerant of types Cython doesn't understand.

Added `tests/errors/cast_unspecified_type.pyx` plus a CHANGES entry. The input crashes on master and gives the error with this change.

### DIFF
--- a/tests/errors/cast_unspecified_type.pyx
+++ b/tests/errors/cast_unspecified_type.pyx
@@ -1,0 +1,15 @@
+# mode: error
+# ticket: 7683
+
+import cython
+
+def f():
+    s = ""
+    y: cython.typeof(s)
+    b = cython.cast(cython.typeof(y), s)
+
+
+_ERRORS = u"""
+9:14: Unable to determine the type to cast to
+9:14: Cannot assign type '<error>' to '<unspecified>'
+"""
--- a/tests/errors/cast_unspecified_type.pyx
+++ b/tests/errors/cast_unspecified_type.pyx
@@ -11,5 +11,4 @@ def f():
 
 _ERRORS = u"""
 9:14: Unable to determine the type to cast to
-9:14: Cannot assign type '<error>' to '<unspecified>'
 """
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -98,6 +98,10 @@ Bugs fixed
   It now uses a dedicated enum implementation class to allow this.
   (Github issue :issue:`7185`)
 
+* Casting to an unresolved ``typeof()`` type (e.g. ``cython.cast(cython.typeof(x), ...)``
+  where the type could not be inferred) crashed the compiler instead of reporting an error.
+  (Github issue :issue:`7683`)
+
 * ``cpdef fused`` functions generated redundant code.
   (Github issue :issue:`7778`)
 
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -12044,6 +12044,12 @@ class TypecastNode(ExprNode):
             error(self.pos,
                 "Cannot cast to a function type")
             self.type = PyrexTypes.error_type
+        elif self.type.is_unspecified:
+            # e.g. cython.cast(cython.typeof(x), ...) where typeof() could not
+            # be resolved to a concrete type; report an error instead of crashing.
+            error(self.pos,
+                "Unable to determine the type to cast to")
+            self.type = PyrexTypes.error_type
         self.operand = self.operand.analyse_types(env)
         if self.type is PyrexTypes.c_bint_type:
             # short circuit this to a coercion
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -1152,6 +1152,8 @@ class ExprNode(Node):
         return src
 
     def fail_assignment(self, dst_type):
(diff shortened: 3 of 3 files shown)


tests/errors/cast_unspecified_type.pyx
```diff
@@ -0,0 +1,15 @@
...
+
+def f():
+    s = ""
+    y: cython.typeof(s)
+    b = cython.cast(cython.typeof(y), s)
+
+
+_ERRORS = u"""
+9:14: Unable to determine the type to cast to
+9:14: Cannot assign type '<error>' to '<unspecified>'
```
scoder: We can probably avoid the second (unhelpful) error line with this change:
```patch
index 71023cb09..2df2a5248 100644
@@ -1152,6 +1152,8 @@ class ExprNode(Node):
         return src
 
     def fail_assignment(self, dst_type):
+        if self.type.is_error or dst_type.is_error:
+            return  # Reported elsewhere.
         src_name = self.entry.name if hasattr(self, "entry") else None
         src_resolved = f" (alias of '{self.type.resolve()}')" if self.type.is_typedef else ""
         dst_resolved = f" (alias of '{dst_type.resolve()}')" if dst_type.is_typedef else ""
```

scoder (approved):
Fine overall, with one suggestion for improvement.

scoder:
The test failures are unrelated, we're trying to fix them on master currently.

tests/errors/cast_unspecified_type.pyx
scoder:
```suggestion
```

scoder:
Thanks

scoder:
We usually add the name of external contributors to the changelog entry. Since you added an entry yourself without your name in it, may I ask if you want your name to be excluded?

vsaraikin:
It would be nice to have my name in there. I really like the project and look forward to contributing in the future

PR #7863, opened 2026-07-27
Try to fix Github actions trigger on labels

da-woods:
It looks like "label created" happens when someone creates a new label for the project as a whole. And we want to trigger on "pull request labeled" for when some adds a label to a PR.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -16,14 +16,6 @@ on:
     # * benchmark (explicit)
     # * performance (implicit)
     types: [opened, synchronize, reopened, labeled]
-    paths:
-      # Only run on PRs likely to affect benchmark code generation
-      - ".github/workflows/benchmarks.yml"
-      - "Demos/benchmarks/**"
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -12,6 +12,10 @@ on:
       - "Cython/Compiler/ExprNodes.py"
       - "Cython/Compiler/Optimize.py"
   pull_request:
+    # Accepted labels (in the 'if' condition below) are:
+    # * benchmark (explicit)
+    # * performance (implicit)
+    types: [opened, synchronize, reopened, labeled]
```
da-woods, line 18: `opened, synchronize, reopened` are the default for PRs if you don't specify anything (if we trust the github actions documentation...)

scoder:
whatever works …

da-woods:
At least the "sanitizers" label now seems to trigger it correctly. I'm sure there are other issues (because I certainly don't really understand GitHub actions) but I think it's closer to working as intended.

PR #7864, opened 2026-07-28
Fix `except` clause with fused-type return value 

### LINKED ISSUE #7819: [BUG] Explicit fused function exception specifications fail with "cannot coerce to type that isn't specialized"
### Describe the bug

```
cdef IntOrFloat f(IntOrFloat x) except -1:
                                        ^
fusedreturn.pyx:5:40: Cannot coerce to a type that is not specialized
```

This should work because `-1` would be a valid exception value for both `int` or `float`

```cython
cdef fused IntOrFloat:
    int
    float

cdef IntOrFloat f(IntOrFloat x) except -1:
    return x
```

### Expected behaviour

_No response_

### OS

Linux

### Python version

3.13

### Cython version

Current master

### Additional context

Possibly worth checking if https://github.com/cython/cython/pull/5737 solves it/helps... It certainly contains a test that looks similar

imann128:
Fixes #7819

## Summary
Using an exception check like `except -1` (or `except? -1`) on a function with a fused type failed to compile. 
Cython gave the error: *"Cannot coerce to a type that is not specialized"*.

This happened because Cython tried to convert the exception value (like `-1`) to a specific type before it actually knew which specific type (like `int` or `double`) the fused function was going to use. Fused-type specialization happens later in `FusedNode.py`.

## Files Changed:
- **`Nodes.py`**: If the return type is fused, stop trying to force a type conversion early. Instead, just check that the exception value works for every possible type the fused type could become.
- **`PyrexTypes.py` (`CFuncType.specialize`)**: When Cython generates the real C code for each specific type, update the exception value to use that specific return type. Without this, the code compiled, but generated wrong C checks (like improper type casts or broken `NaN` comparisons for floats).

## Tests:
- Added `tests/run/fused_except_value_GH7819.pyx` to test both `int` and `double` functions. It covers normal return values, the `except?` case where `-1` is a valid return value, and actual error handling.

### DIFF
--- a/Cython/Compiler/Nodes.py
+++ b/Cython/Compiler/Nodes.py
@@ -832,6 +832,19 @@ class CFuncDeclaratorNode(CDeclaratorNode):
                         error(self.exception_value.pos,
                               "Exception value must be a Python exception, or C++ function with no arguments, or *.")
                     exc_val = self.exception_value
+                elif return_type.is_fused:
+                    # Return type is not specialized yet. Since there is no exact return type, the enclosing function
+                    # is copied per fused-type permutation in FusedNode.py(), so it can't convert the exception value
+                    # to a specific C type here. Check the exception as-is against all potential types.
+                    # Also, CFuncType.specialize() re-targets the resulting ExceptionValue
+                    # at the concrete return type of each specialization.
+                    self.exception_value = self.exception_value.analyse_types(env).analyse_const_expression(env)
+                    exc_val = self.exception_value.as_exception_value(env)
+                    permutations = PyrexTypes.get_all_specialized_permutations(return_type.get_fused_types())
+                    if not all(return_type.specialize(mapping).assignable_from(self.exception_value.type)
+                                    for _, mapping in permutations):
+                        error(self.exception_value.pos,"Exception value incompatible with function return type")
+
                 else:
                     self.exception_value = self.exception_value.analyse_types(env).coerce_to(
                         return_type, env).analyse_const_expression(env)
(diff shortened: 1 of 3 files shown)


da-woods:
What's the reason to prefer PR over the existing https://github.com/cython/cython/pull/5737 (which I believe also fixes the issue)?

imann128:
#5737 covers a lot more in breadth. It reworks how exception_value is represented (strings → typed const nodes) to fix #5586/#5709, plus the fused-type fix on top of that. 
It's also been open since Sep 2023, still depends on the unmerged #5710, and hasn't had a commit since Apr 2024.

In the thread, it was also mentioned that the approach felt "hackish" and that it would be simpler to defer specializing the exception value until `FusedNode` runs. That's the approach I took here, without needing the `exception_value` refactor at all.

In analyse (Nodes.py), I added an `elif return_type.is_fused` branch before the coercion path. Instead of `.coerce_to(return_type, ...),` it just checks that the exception value is assignable against every permutation from `get_all_specialized_permutations`, so nothing gets forced into a concrete type yet. Then in `CFuncType.specialize (PyrexTypes.py)`, once the return type is concrete, 
I re-create the exception value against it before building the specialized `CFuncType.` 
That's what generates the right C check for each specialization.

Two small changes, nothing else touched. Non-fused exceptions work exactly like before. Test covers int and double, both except -1 and except? -1, normal returns, and actual errors getting raised.

#5737 is also changing runtime behavior for existing fused functions (defaults switching from 0 to -1 in some cases), which is why it needed the extern/pxd carve-out. This PR doesn't touch any working code path, so there's nothing to regress.

I understand #5737 is solving this more generally, and the `exception_value` cleanup is worth doing on its own, but it's a much bigger refactor to land, and it's been open for a while. This fixes #7819 without depending on any of that.


But that's at your discretion. I am open to discussion

PR #7867, opened 2026-07-29
Add support of PGO builds

matusvalo:
To compile cython with PGO follow next steps:
1. Patch setup.py:
   ```diff
   diff --git a/setup.py b/setup.py
   index dc8c28746..91b625ab6 100755
   --- a/setup.py
   +++ b/setup.py
   @@ -205,6 +205,7 @@ def compile_cython_modules(profile=False, coverage=False, compile_minimal=False,
   
        for ext in extensions:
            ext.cython_directives = cython_directives
   +        ext.pgo_dir = 'pgo'
   
        # Make 'build_ext' use Cython.
        from Cython.Distutils.build_ext import build_ext as cy_build_ext
   ```
2. Build cython with enabled collecting of profile data: `CFLAGS='-g0 -O2' python setup.py build_ext -i --cython-gen-pgo`
3. Run some load to collect profiling data - e.g. test suite can be run: 
   `python runtests.py --cython-only pep memoryview import gil method class fuse func`
   After this step, `pgo` directory should created and filled with pgo profile data
4. Final build of cython using profile data: `CFLAGS='-g0 -O2' python setup.py build_ext --cython-use-pgo`

> [!NOTE]
> `CFLAGS` are used only to ensure optimization and removal of debug symbols, they do not have any effect on PGO and can be avoided.

This PR is a draft for early feedback. PGO for IPythonMagic is broken and does not work (in master too). I am not sure whether it make sense to fix it...

### DIFF
--- a/Cython/Build/Cythonize.py
+++ b/Cython/Build/Cythonize.py
@@ -134,7 +134,7 @@ def _build(ext_modules, parallel):
 
 def run_distutils(args):
     try:
-        from distutils.core import setup
+        from setuptools import setup
     except ImportError:
         try:
             from setuptools import setup
--- a/Cython/Build/IpythonMagic.py
+++ b/Cython/Build/IpythonMagic.py
@@ -444,16 +444,17 @@ class CythonMagics(Magics):
         compiler_type = build_extension.compiler.compiler_type
         if compiler_type == 'unix':
             compiler_cmd = build_extension.compiler.compiler_so
-            # TODO: we could try to call "[cmd] --version" for better insights
             if not compiler_cmd:
                 pass
             elif 'clang' in compiler_cmd or 'clang' in compiler_cmd[0]:
-                compiler_type = 'clang'
+                compiler_type = 'gcc'
             elif 'icc' in compiler_cmd or 'icc' in compiler_cmd[0]:
                 compiler_type = 'icc'
             elif 'gcc' in compiler_cmd or 'gcc' in compiler_cmd[0]:
                 compiler_type = 'gcc'
             elif 'g++' in compiler_cmd or 'g++' in compiler_cmd[0]:
+                compiler_type = 'gcc'
+            else:
                 compiler_type = 'gcc'
         config = PGO_CONFIG.get(compiler_type)
         orig_flags = []
--- a/Cython/Distutils/build_ext.py
+++ b/Cython/Distutils/build_ext.py
@@ -101,7 +101,7 @@ class build_ext(_build_ext):
     def get_extension_attr(self, extension, option_name, default=False):
         return getattr(self, option_name) or getattr(extension, option_name, default)
 
-    def _add_pgo_flags(self, step_name, temp_dir):
+    def _add_pgo_flags(self, ext, step_name, temp_dir):
         compiler_type = self.compiler.compiler_type
         if compiler_type == 'unix':
             compiler_cmd = self.compiler.compiler_so
@@ -121,10 +121,9 @@ class build_ext(_build_ext):
         orig_flags = []
         if config and step_name in config:
             flags = [f.format(TEMPDIR=temp_dir) for f in config[step_name]]
-            for extension in self.extensions:
-                orig_flags.append((extension.extra_compile_args, extension.extra_link_args))
-                extension.extra_compile_args = extension.extra_compile_args + flags
-                extension.extra_link_args = extension.extra_link_args + flags
+            orig_flags.append((ext.extra_compile_args, ext.extra_link_args))
+            ext.extra_compile_args = ext.extra_compile_args + flags
+            ext.extra_link_args = ext.extra_link_args + flags
         else:
             print("No PGO %s configuration known for C compiler type '%s'" % (step_name, compiler_type),
                   file=sys.stderr)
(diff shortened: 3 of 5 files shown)


Cython/Build/IpythonMagic.py
```diff
@@ -444,17 +444,18 @@ def _add_pgo_flags(self, build_extension, step_name, temp_dir):
...
-                compiler_type = 'clang'
+                compiler_type = 'gcc'
             elif 'icc' in compiler_cmd or 'icc' in compiler_cmd[0]:
                 compiler_type = 'icc'
             elif 'gcc' in compiler_cmd or 'gcc' in compiler_cmd[0]:
                 compiler_type = 'gcc'
             elif 'g++' in compiler_cmd or 'g++' in compiler_cmd[0]:
                 compiler_type = 'gcc'
+            else:
+                compiler_type = 'gcc'
```
matusvalo, line 458: At least on debian, one cannot see easily what compiler is installed. `compiler_cmd` is `cc` and `cc --version` does not display `GCC`. (although based on license one can guess the compiler but in any case there is no straightforward way...)

Cython/Build/Cythonize.py
```diff
@@ -134,7 +134,7 @@ def _build(ext_modules, parallel):
 
 def run_distutils(args):
     try:
-        from distutils.core import setup
+        from setuptools import setup
```
matusvalo, line 137: `distutil` is not showing compiler command but `setuptools` does. We should think whether we do not want to use `setuptools` everywhere instead of `distutils` (even more since `distutils` is not core part of python anymore)

Cython/Distutils/build_ext.py
```diff
@@ -82,6 +101,35 @@ def finalize_options(self):
...
+            elif 'g++' in compiler_cmd or 'g++' in compiler_cmd[0]:
+                compiler_type = 'gcc'
+            else:
+                compiler_type = 'gcc'
+        config = PGO_CONFIG.get(compiler_type)
+        orig_flags = []
+        if config and step_name in config:
+            flags = [f.format(TEMPDIR=temp_dir) for f in config[step_name]]
+            for extension in self.extensions:
+                orig_flags.append((extension.extra_compile_args, extension.extra_link_args))
```
matusvalo: this logic has bug. It accumulates the same flags over extensions but for sake of PoC it just works.

Cython/Distutils/build_ext.py
```diff
@@ -113,6 +161,25 @@ def build_extension(self, ext):
...
+        if self.cython_gen_pgo:
+            self._add_pgo_flags('gen', pgo_dir)
+        if self.cython_use_pgo:
+            self.force = 1
+            self._add_pgo_flags('use', pgo_dir)
+            np_pythran = getattr(ext, 'np_pythran', False)
+            c_sources = []
+            for source in ext.sources:
+                base, ex = os.path.splitext(source)
+                if ex in ('.pyx', '.py'):
```
matusvalo, line 174: Here, I want to force recompilation of C files without compilation .pyx to .c.

scoder:
> PGO for IPythonMagic is broken and does not work

Are you sure? There _is_ a test for it in `TestIPythonMagic.py`.

matusvalo:
> > PGO for IPythonMagic is broken and does not work
> 
> Are you sure? There _is_ a test for it in `TestIPythonMagic.py`.

Hmm at least on debian when I followed the documentation i ran into two problems causing compiler crash:
1. debian has gcc disguised as `cc`

   ```
   (cython) dev@homepc:~/dev/cython/Cython/Compiler$ cc
   cc: fatal error: no input files
   compilation terminated.

   (cython) dev@homepc:~/dev/cython/Cython/Compiler$ cc --version
   cc (Debian 14.2.0-19) 14.2.0
   Copyright (C) 2024 Free Software Foundation, Inc.
   This is free software; see the source for copying conditions.  There is NO
   warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
   ```

   which causes this exception: https://github.com/cython/cython/blob/1165d5d1cc5e0aaf329f58957846fcb617310ef1/Cython/Build/IpythonMagic.py#L467. This PR contains fix (ref. https://github.com/cython/cython/pull/7867#discussion_r3678530049) for this (not ideal though but there is not much that can be done...)

2. But even after fixing point 1. I am getting following error:
   ```
   (cython) dev@homepc:~/dev/cython$ ipython
   Python 3.14.2 (main, Jan 27 2026, 23:59:57) [Clang 21.1.4 ]
   Type 'copyright', 'credits' or 'license' for more information
   IPython 9.15.0 -- An enhanced Interactive Python. Type '?' for help.
   Tip: Use the IPython.lib.demo.Demo class to load any Python script as an interactive demo.
   
   In [1]: %load_ext cython
   
   
   In [2]:
   
   In [2]: %%cython --pgo
      ...: def f(x):
      ...:     return x * x
      ...:
      ...: f(2)
      ...:
      ...:
   
   Content of stderr:
   /home/dev/.cache/ipython/cython/_cython_magic_223583eaaba4ece19e1e88fd697c3021c74d69e976533bc5759a42e183b8ba8b.c: In function ‘__Pyx_PyCode_New’:
   /home/dev/.cache/ipython/cython/_cython_magic_223583eaaba4ece19e1e88fd697c3021c74d69e976533bc5759a42e183b8ba8b.c:7739:1: warning: ‘/home/dev/.cache/ipython/cython//home/dev/.cache/ipython/cython/home/dev/.cache/ipython/cython/_cython_magic_223583eaaba4ece19e1e88fd697c3021
   c74d69e976533bc5759a42e183b8ba8b.gcda’ profile count data file not found [-Wmissing-profile]
   ```
   It seems that `gcda` profile file is created **after** main process (ipython) is finished. Hence, as sideeffect, when second time ipython is run and code before is pasted, it runs OK...

Cython/Distutils/build_ext.py
matusvalo: fixed.

PR #7868, opened 2026-07-29
Remove commented out test in test_grammer.py

scopreon:
Remove comments around a test that was previously fixed in #4254

One of the many low hanging fruit in #3638


scoder:
Thanks

PR #7870, opened 2026-07-30
Fix some syntax errors in cygdb script

da-woods:
Fixup to 6dcfa5a66885076994718eb40afd91acc4f57d72. I don't think it solves the Python 3.16 sanitizer tests problem but it definitely solves some invalid syntax.

### DIFF
--- a/Cython/Debugger/Cygdb.py
+++ b/Cython/Debugger/Cygdb.py
@@ -61,8 +61,7 @@ def make_command_file(path_to_debug_info, prefix_code='',
             sitepackages = [ p for p in sitepackages if pathlib.Path(p).is_relative_to(virtualenv) ]
             activate_virtualenv = (
                 f'import sys; sys.path[:0] = {sitepackages!r}; '
-                f'import site; {"; ".join(f"site.addsitedir({p!r})" for p in sitepackages)}'
-                f'{";" if sitepackages else ""}'
+                f'import site; {"".join(f"site.addsitedir({p!r}); " for p in sitepackages)}'
                 f'print("gdb command file: Activating virtualenv: {virtualenv}")'
             )
             platform_machine = platform.machine()
--- a/Cython/Debugger/Cygdb.py
+++ b/Cython/Debugger/Cygdb.py
@@ -43,7 +43,7 @@ def make_command_file(path_to_debug_info, prefix_code='',
         activate_virtualenv = report_virtualenv_error = 'pass'
         check_virtualenv_version = 'False'
         virtualenv = os.getenv('VIRTUAL_ENV')
-        has_gil = getattr(sys, "_is_gil_enabled", True)
+        has_gil = hasattr(sys, "_is_gil_enabled")
         platform_machine = ""
 
         if virtualenv:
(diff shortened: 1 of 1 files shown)


Cython/Debugger/Cygdb.py
```diff
@@ -43,7 +43,7 @@ def make_command_file(path_to_debug_info, prefix_code='',
         activate_virtualenv = report_virtualenv_error = 'pass'
         check_virtualenv_version = 'False'
         virtualenv = os.getenv('VIRTUAL_ENV')
-        has_gil = getattr(sys, "_is_gil_enabled", True)
+        has_gil = hasattr(sys, "_is_gil_enabled")
```
scoder: Why is it enough to know if the attribute exists, rather than knowing whether it's true or false?

EDIT: Specifically, why is this information (attribute exists or not) not provided by the Python version check already?

Cython/Debugger/Cygdb.py
```diff
@@ -61,13 +61,14 @@ def make_command_file(path_to_debug_info, prefix_code='',
             sitepackages = [ p for p in sitepackages if pathlib.Path(p).is_relative_to(virtualenv) ]
             activate_virtualenv = (
                 f'import sys; sys.path[:0] = {sitepackages!r}; '
-                f'import site; {"; ".join(f"site.addsitedir({p!r})" for p in sitepackages)}; '
+                f'import site; {"; ".join(f"site.addsitedir({p!r})" for p in sitepackages)}'
+                f'{";" if sitepackages else ""}'
```
scoder:
```suggestion
                f'import site; {"".join(f"site.addsitedir({p!r}); " for p in sitepackages)}'
```

Cython/Debugger/Cygdb.py
da-woods: `_is_gil_enabled` only exists for freethreaded builds (and tells you if it's switched into its fallback GIL-emulation mode). So the attribute doesn't exist on the Python 3.14 regular GIL build.

`has_gil` is misnamed though - it should be `is_freethreaded_build`

scoder (approved):

scoder:
Sorry for the mis-corrections then. You're right, changing the variable name makes the intention much more obvious.

Cython/Debugger/Cygdb.py
```diff
@@ -61,18 +61,18 @@ def make_command_file(path_to_debug_info, prefix_code='',
...
             check_virtualenv_version = (
                 f'sys.version_info[:2] == {sys.version_info[:2] !r} and '
-                f'getattr(sys, "_is_gil_enabled", True) == {has_gil !r} and '
+                f'hasattr(sys, "_is_gil_enabled") == {is_freethreaded_build !r} and '
                 f'platform.machine() == {platform_machine !r}'
             )
             report_virtualenv_error = (
                 'print("Not activating virtual environment for mismatched Python version %d.%d%s (%s)" % ('
-                f'{sys.version_info[0]}, {sys.version_info[1]}, {"" if has_gil else "t" !r}, {platform_machine !r}))'
+                f'{sys.version_info[0]}, {sys.version_info[1]}, {"t" if is_freethreaded_build else "" !r}, {platform_machine !r}))'
```
scoder: I just noticed that the double formatting makes no sense here. We can just format the values directly into the output string and just print it from within gdb.

Cython/Debugger/Cygdb.py
da-woods: I double-checked this and I was wrong about it. I've changed it to `sysconfig.get_config_var("Py_GIL_DISABLED")` which is what we use elsewhere. (`sys._is_gil_enabled()` definitely does change if the GIL gets enabled at runtime, so isn't good for identifying the build anyway)

scoder:
Sorry for the back and forth. After staring at the code for a while, I got convinced that the bulk of generated checking code really isn't worth it and is best replaced with a straight forward function. This also keeps the user visible gdb namespace free of imported modules and avoids useless startup overhead when we already see from the Python version that it cannot match.

PR #7871, opened 2026-07-30
Format C floats in f-strings without a Python float object

### LINKED ISSUE #2195: Speed up f-string float formatting
There is [dedicated code](https://github.com/cython/cython/blob/e00f9c271592c3d1326f968acd0710a556033bfb/Cython/Utility/TypeConversion.c#L653) for formatting integer literals in f-strings, which avoids converting C integers to Python integer objects just for formatting them as strings. It is used by the `FormattedValueNode` in `ExprNodes.py`.

The same should be done for C float/double values, as Python provides [PyOS_double_to_string()](https://docs.python.org/3/c-api/conversion.html#c.PyOS_double_to_string). See CPython's own float formatting in [unicodeobject.c](https://github.com/python/cpython/blob/master/Objects/unicodeobject.c) and [formatter_unicode.c](https://github.com/python/cpython/blob/master/Python/formatter_unicode.c).

- implement relevant tests in `tests/run/fstring.pyx` and/or check if `tests/run/test_fstring.pyx` (copied from CPython's test suite) covers this sufficiently
- extract the relevant parts of CPython's formatting implementation (potentially limiting it to easy to support formats)
- create a new C utility function in `TypeConversion.c` that formats double values
- implement `CFloatType.can_coerce_to_pystring()` in `PyrexTypes.py` to make it return `True` when supported
- implement `CFloatType.convert_to_pystring()` to call the formatting C function.

This can be done incrementally for selected formatting modifiers, e.g. a format like `f"{x: .2f}"` should be fairly easy to support and covers a large part of the usages.

Hint: this ticket requires good C skills.

vsaraikin:
Closes #2195

Formats C floating point values in f-strings with `PyOS_double_to_string()` instead of building a Python float and calling `PyObject_Format()` on it.

Only two shapes are handled in C: an empty format spec, which `str()` reproduces exactly via the `'r'` format character, and a plain precision such as `.2f`, `.3e` or `.4g`. Everything else (alignment, padding, sign, `%`, thousands separators) still goes through `PyObject_Format()`.

Timings on CPython 3.14, arm64, 200k iterations, best of 15:

```
f"{x}"                  161 -> 157 ns
f"{x:.2f}"               63 ->  37 ns
f"... {i} ... {x:.3e}"  127 ->  92 ns
```

The empty spec is not faster since CPython already routes it straight to `repr`, it just avoids the intermediate float object.

Checked the output against Python's own `format()` over 12023 values (random bit patterns, inf, nan, +-0.0, subnormals) times 6 format specs with no mismatches


Cython/Compiler/PyrexTypes.py
```diff
@@ -2904,6 +2905,29 @@ def __init__(self, rank, math_h_modifier = ''):
...
 
+    @staticmethod
+    def _parse_format(format_spec):
+        # We currently only support str() formatting ('r') and a plain precision, e.g. '.2f'.
+        if not format_spec:
+            return ('r', 0)
+
+        if len(format_spec) > 2 and format_spec[0] == '.' and format_spec[-1] in 'efg':
+            precision = format_spec[1:-1]
+            if precision.isdigit():
```
da-woods: `isdigit` might not be strict enough here. Try `f"{v:.²f}"` which would be a runtime error in Python and current Cython, but I think the `int(precision)` would fail at compile-time here.

Arguably failing at compile-time would be perfectly reasonable, but ideally it should be a "nice" failure.

The simplest fix is probably `precision.isascii() and precision.isdigit()`

da-woods:
Is `{v:g}` (I.e. plain format but no precision) deliberately unsupported? It's fine if it is but it isn't immediately obvious to me if there was a reason.

scoder:
Merging master will fix some of the build failures. Sorry for that.

Cython/Compiler/PyrexTypes.py
vsaraikin: Fixed with isascii(). I went with falling back to PyObject_Format rather than a compile-time error, so `f"{v:.²f}"` now raises the same ValueError Python does instead of a traceback out of the compiler

The same pattern already crashes master for integers: `f"{i:²}"` reaches int(prefix) in CIntLike._parse_format. Fixed that line too since it is the same two words, happy to split it out if you would rather keep this PR to floats

vsaraikin:
> Is `{v:g}` (I.e. plain format but no precision) deliberately unsupported? It's fine if it is but it isn't immediately obvious to me if there was a reason.

No reason, I just kept the first version narrow. Added it: a bare e/E/f/F/g/G now uses Python's default precision of 6. The uppercase variants work for the precision form too

Checked against format() over 4021 values times 28 specs, including the ones that must still fall back like `'.2', '.f', 'n'` and `'%'`, with no mismatches

scoder:
> now uses Python's default precision of 6

Could you please add corresponding plain Python doctests so that we can compare Python's own output for the specific format tests and make sure we notice if it changes in the future?
Like:
```python
    >>> d = 1.5
    >>> f"{d}", f"{d:.2f}", f"{d:.3e}", f"{d:.2g}", f"{d:.2}", f"{d:>10.2f}", f"{d:+.2f}"
    ('1.5', '1.50', '1.500e+00', '1.5', '1.5', '      1.50', '+1.50')
    >>> format_c_doubles(1.5)
    ('1.5', '1.50', '1.500e+00', '1.5', '1.5', '      1.50', '+1.50')
```
EDIT: Rather keep the input value visible in the Cython function call to help debugging test failures.

Cython/Compiler/PyrexTypes.py
```diff
@@ -2625,7 +2625,8 @@ def _parse_format(format_spec):
         if prefix and prefix[0] == '0':
             padding = '0'
             prefix = prefix.lstrip('0')
-        if prefix.isdigit():
+        # Note that isdigit() also accepts non-ASCII digits that int() rejects.
+        if prefix.isascii() and prefix.isdigit():
```
scoder: I think you want to use `.isdecimal()` here.
```python
>>> '٠١٢٣٤٥٦٧٨٩'.isdecimal()
True
>>> int('٠١٢٣٤٥٦٧٨٩')
123456789
>>> float('٠١٢٣٤٥٦٧٨٩')
123456789.0
>>> f'{1.5:٠٩.4}'
'      1.5'
```
https://docs.python.org/3/library/stdtypes.html#str.isdecimal

scoder (commented):
You could add test cases from
https://github.com/python/cpython/blob/10a84540c2c2533e31f5f0649cc4e4afdd991ba2/Lib/test/mathdata/formatfloat_testcases.txt
They use %-formatting, but that should be adaptable programmatically, as in
https://github.com/python/cpython/blob/10a84540c2c2533e31f5f0649cc4e4afdd991ba2/Lib/test/test_float.py#L815-L835

tests/run/fstring_float_testcases.pyx
```diff
@@ -0,0 +1,431 @@
...
+    >>> test_formatfloat_testcases()
+    292 cases
+    """
+    count = 0
+    for line in TESTCASES.splitlines():
+        line = line.strip()
+        if not line or line.startswith('--'):
+            continue
+        lhs, expected = [part.strip() for part in line.split('->')]
+        fmt, arg = lhs.split()
```
scoder: There's no need to fire up a listcomp here if you immediately call `.split()` afterwards (which ignores whitespace).
```suggestion
        if line.startswith('--'):
            continue
        lhs, expected = line.split(' -> ')
        expected = expected.strip()
        fmt, arg = lhs.split()
```

tests/run/fstring_float_testcases.pyx
```diff
@@ -0,0 +1,431 @@
...
+# The '%' format strings used there become f-string format specs by dropping the
+# leading '%', and '%r' becomes an empty spec.
+
+# cython: test_assert_c_code_has = __Pyx_PyUnicode_FromDouble\(
+
+
+def format_double(double x):
+    # Every spec has to stay a literal in the source for Cython to format it in C,
+    # so the test cases are dispatched through this mapping rather than through a
+    # format spec built at runtime.
```
scoder: It seems a bit wasteful to format all variants, then collect them in a dict, and then pick just the one that's relevant for the test. But it's simple and (shallowly) tests a bit more. Why not.

But if we do it this way, then there's no need for cutting off the '%' before the lookup below. Just pass it as is, then it's clear that we are mapping '%' formats to f-strings here. You can also include the `%r` in that case, because why not. And you can add a test that the actual %-formatting (i.e. `fmt % value`), which gets opaquely evaluated by Python's `str` type, gives the same result as Cython's f-string.

tests/run/fstring_float_testcases.pyx
```diff
@@ -0,0 +1,434 @@
...
+
+
+def test_formatfloat_testcases():
+    """
+    >>> test_formatfloat_testcases()
+    292 cases
+    """
+    count = 0
+    for line in TESTCASES.splitlines():
+        line = line.strip()
```
scoder: I don't think this line if of any use. `.splitlines()` removes the line endings and all comments in the test case lines (which we control ourselves) start immediately with `--`, no preceding whitespace. And not stripping empty lines would help us detect trailing whitespace in these lines, which is a good thing. :)

```suggestion
```

tests/run/fstring_float_testcases.pyx
```diff
@@ -0,0 +1,434 @@
...
+        value = float(arg)
+        for x, result in ((value, expected), (-value, '-' + expected)):
+            assert format_double(x)[fmt] == result, (fmt, x)
+            # Python's own %-formatting has to agree with the f-string.
+            assert fmt % x == result, (fmt, x)
+        count += 1
+    print(f"{count} cases")
+
+
+TESTCASES = u"""\
```
scoder: We don't infer the type of (arbitrarily reassignable) global variables, so a type hint will (eventually) reduce the code overhead on use.
```suggestion
TESTCASES: str = u"""\
```

tests/run/fstring_float_testcases.pyx
```diff
@@ -0,0 +1,434 @@
...
+        '%f': f"{x:f}",
+    }
+
+
+def test_formatfloat_testcases():
+    """
+    >>> test_formatfloat_testcases()
+    292 cases
+    """
+    count = 0
```
scoder:
```suggestion
    cdef int count = 0
```

scoder (commented):
Mostly good now, just a few style improvements.

scoder:
Thanks!

PR #7875, opened 2026-07-31
Fix redeclaration error of global annotated variable.

### LINKED ISSUE #7874: [BUG] global variable with annotation and `cython.declare()` causes `redeclared` error
### Describe the bug

Global variable having annotation and also `cython.declare()` causes failure with `redeclared` error.

```cython

import cython
from typing import Tuple, Any

_Empty_Tuple: Tuple[Any, ...] = cython.declare(tuple, ())
```

### Expected behaviour

I suppose annotation should be ignored since it is ignored in following case:
```python
import cython
from typing import Tuple, Any

_Empty_Tuple: Tuple[Any, ...]
```

But it is up to discussion.

### OS

_No response_

### Python version

_No response_

### Cython version

master

### Additional context

This code already exist here: https://github.com/sqlalchemy/sqlalchemy/blob/d9b44cb731bd27e6e82c99a3c0fb598214e8e7ff/lib/sqlalchemy/engine/_util_cy.py#L45

matusvalo:
Currently cython ignores annotations of global variables. Hence, when
global variable is declared `cython.declare` we should not fail with
redeclared error. This PR allows following code snippet:

```python
var: tuple[typing.Any] = cython.declare(tuple, ())
```

Fixes #7874

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -2359,7 +2359,10 @@ class NameNode(AtomicExprNode):
         elif (self.entry and self.entry.is_inherited and
                 self.annotation and env.is_c_dataclass_scope):
             error(self.pos, "Cannot redeclare inherited fields in Cython dataclasses")
-        elif self.entry and self.annotation and env.directives['annotation_typing']:
+        elif self.entry and self.annotation and env.directives['annotation_typing'] and not self.entry.scope.is_module_scope:
+            # Type annotations of global variables (module scope) are ignored by cython.
+            # Hence, we support following declaration:
+            # _Empty_Tuple: tuple[typing.Any] = cython.declare(tuple, ())
             error(self.pos, f"'{self.name}' redeclared")
         if not self.entry:
             if env.directives['warn.undeclared']:
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -2363,9 +2363,9 @@ class NameNode(AtomicExprNode):
             if not self.entry.scope.is_module_scope:
                 error(self.pos, f"'{self.name}' redeclared")
             else:
-            # Type annotations of global variables (module scope) are ignored by cython.
-            # Hence, we support somewhat contradictory declarations like:
-            # _Empty_Tuple: tuple[typing.Any] = cython.declare(tuple, ())
+                # Type annotations of global variables (module scope) are ignored by cython.
+                # Hence, we support somewhat contradictory declarations like:
+                # _Empty_Tuple: tuple[typing.Any] = cython.declare(tuple, ())
                 annotation_type = self.annotation.analyse_as_type(env)
                 entry_type = self.entry.type
                 if not (annotation_type.assignable_from(entry_type) or entry_type.assignable_from(annotation_type)):
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -2361,7 +2361,7 @@ class NameNode(AtomicExprNode):
             error(self.pos, "Cannot redeclare inherited fields in Cython dataclasses")
         elif self.entry and self.annotation and env.directives['annotation_typing'] and not self.entry.scope.is_module_scope:
             # Type annotations of global variables (module scope) are ignored by cython.
-            # Hence, we support following declaration:
+            # Hence, we support somewhat contradictory declarations like:
             # _Empty_Tuple: tuple[typing.Any] = cython.declare(tuple, ())
             error(self.pos, f"'{self.name}' redeclared")
         if not self.entry:
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -2359,11 +2359,19 @@ class NameNode(AtomicExprNode):
         elif (self.entry and self.entry.is_inherited and
                 self.annotation and env.is_c_dataclass_scope):
             error(self.pos, "Cannot redeclare inherited fields in Cython dataclasses")
-        elif self.entry and self.annotation and env.directives['annotation_typing'] and not self.entry.scope.is_module_scope:
+        elif self.entry and self.annotation and env.directives['annotation_typing']:
+            if not self.entry.scope.is_module_scope:
+                error(self.pos, f"'{self.name}' redeclared")
+            else:
             # Type annotations of global variables (module scope) are ignored by cython.
             # Hence, we support somewhat contradictory declarations like:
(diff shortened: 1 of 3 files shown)


tests/run/pep526_variable_annotations.py
```diff
@@ -23,6 +23,7 @@
 
 pyobj_var = 1  # type: annotation
 var: cython.int = 2
+cvar: cython.int = cython.declare(cython.int, 5)
```
scoder, line 26: I think we should also have a test where the two types are not identical, as in the example that initiated this change. They don't really contradict each other but one is more specific than the other.

Cython/Compiler/ExprNodes.py
```diff
@@ -2359,7 +2359,10 @@ def _analyse_target_declaration(self, env, is_assignment_expression):
         elif (self.entry and self.entry.is_inherited and
                 self.annotation and env.is_c_dataclass_scope):
             error(self.pos, "Cannot redeclare inherited fields in Cython dataclasses")
-        elif self.entry and self.annotation and env.directives['annotation_typing']:
+        elif self.entry and self.annotation and env.directives['annotation_typing'] and not self.entry.scope.is_module_scope:
+            # Type annotations of global variables (module scope) are ignored by cython.
+            # Hence, we support following declaration:
```
```suggestion
            # Hence, we support somewhat contradictory declarations like:
```

scoder (approved):

tests/run/pep526_variable_annotations.py
matusvalo: Hmm i suppose that contradictory annotation will be accepted or we do want to do the check?

scoder: It might be difficult to determine when exactly it's contradictory and when it's just diverging for (whatever good or bad) reason, so rejecting would be a harsh response. We might issue a warning when it's not assignable. Maybe test it both ways, if one way works, it's probably ok-ish and the user might know what she's doing; otherwise it's likely to be a mistake.

tests/errors/pure_warnings.py
```diff
@@ -51,6 +53,7 @@ def bar4(a: cython.foo[:]):  # error
 21:14: Unknown type declaration in annotation, ignoring
 35:14: Unknown type declaration 'Bar' in annotation, ignoring
 39:20: Unknown type declaration 'stdint.bar' in annotation, ignoring
+45:0: Annotation type 'list object' is not compatible with declaration type 'dict object'.
```
matusvalo, line 56: I am not sure why warnings are not asserted - this line can be removed and test still is successful. When I compile manually file, the warning is showed but test does not assert the warning against `_WARNINGS`.

Cython/Compiler/ExprNodes.py
```diff
@@ -2360,7 +2360,18 @@ def _analyse_target_declaration(self, env, is_assignment_expression):
...
-            error(self.pos, f"'{self.name}' redeclared")
+            if not self.entry.scope.is_module_scope:
+                error(self.pos, f"'{self.name}' redeclared")
+            else:
+            # Type annotations of global variables (module scope) are ignored by cython.
+            # Hence, we support somewhat contradictory declarations like:
+            # _Empty_Tuple: tuple[typing.Any] = cython.declare(tuple, ())
+                annotation_type = self.annotation.analyse_as_type(env)
+                entry_type = self.entry.type
+                if not (annotation_type.assignable_from(entry_type) or entry_type.assignable_from(annotation_type)):
```
matusvalo, line 2371: This is very conservative condition. It was pretty difficult to find two types not compatible in both ways. I am thinking won't be better:

* annotation type must **not** be assignable to declaration type
* declaration type must be assignable to annotation type

hence:

```python
a: str = cython.declare(int, 5)    # OK
b: int = cython.declare(str, 'hello')   # warning
```

**edit**: funny `str` is not assignable from `int` and `int` is not assignable from `str` 🤷

tests/run/pep526_variable_annotations.py
matusvalo: Fixed.

Cython/Compiler/ExprNodes.py
```diff
@@ -2360,7 +2360,18 @@ def _analyse_target_declaration(self, env, is_assignment_expression):
                 self.annotation and env.is_c_dataclass_scope):
             error(self.pos, "Cannot redeclare inherited fields in Cython dataclasses")
         elif self.entry and self.annotation and env.directives['annotation_typing']:
-            error(self.pos, f"'{self.name}' redeclared")
+            if not self.entry.scope.is_module_scope:
+                error(self.pos, f"'{self.name}' redeclared")
+            else:
+            # Type annotations of global variables (module scope) are ignored by cython.
+            # Hence, we support somewhat contradictory declarations like:
+            # _Empty_Tuple: tuple[typing.Any] = cython.declare(tuple, ())
```
```suggestion
                # Type annotations of global variables (module scope) are ignored by cython.
                # Hence, we support somewhat contradictory declarations like:
                # _Empty_Tuple: tuple[typing.Any] = cython.declare(tuple, ())
```

Cython/Compiler/ExprNodes.py
scoder: Not sure I understand what you're saying, but both assignments look like they deserve a warning.

tests/errors/pure_warnings.py
scoder: The problem is here:
https://github.com/cython/cython/blob/db6bb4ef95a8b1b480f0644accda2a9b6542cfc7/runtests.py#L1531-L1537
If we look for both errors and warnings, then we ignore the warnings.
That's easy to fix, but it might uncover broken warnings in tests, so I'll prepare a PR for it.

EDIT: See https://github.com/cython/cython/pull/7876

Cython/Compiler/ExprNodes.py
scoder: Maybe you meant to say that you _do_ want a warning if the declaration type is not assignable to the annotation type, but you do _not_ want a warning if the annotation type is not assignable to the declaration type? So, basically, the annotation type must be able to accommodate the declared type, but not necessarily the other way round.

I would generally expect the annotated type to be wider in this case (I'd expect declared Cython types to be the most specific, e.g. regarding integer ranges vs a Python `int` annotation), but in the initial tuple example case, it's actually the annotation that is more specific. So we can't rely on that. It's still assignable, though.

I actually meant the "either way" check to reduce the chance of false positives, given our lack of knowledge about use cases. If both ways contradict, it's definitely a reason to warn. If only one of them contradicts and the other is fine, it's likely ok and might not be a good reason to annoy users.

I think we can keep it like this for the time being. It's still going to call out the obvious mistakes.

PR #7879, opened 2026-08-01
Add support for type aliases

matusvalo:
This PR introduces support for following syntax:
```python
type ListOfInts = list[int]

A: ListOfInts = [1,2,3]
```

## Currently works:
* parser (not sure if done right, in python it is soft keyword [1]
* Simple usage with tests
* Using aliases inside type subscriptions
* Composite usage (type alias having type aliases inside)
* C tuple support
* some tests

## Pending:
* Code is written in a dirty way. I suppose at least `TypeAliasType` needs to be implemented in different way, not sure how is the best way, definitely not how it is currently implemented. Also not as Builtin type too.
* warnings when alias is used that does not exist: `type new_alias = list[BadAlias]`
* using and assigning type alias - e.g. `print(TypeAlias)` crashes cython
* More tests of different combinations - alias of struct, etc...

## Python compatibility
* `type` keyword was introduced in python 3.12.
* Older mechanisms are:
  - simple assignment: `TypeAlias = list[int]`. I would prefer to avoid this.
  - type annotated assignment introduced in python 3.10:
    ```
    from typing import TypeAlias
    Vector: TypeAlias = list[float]
    ```
    I would consider support of this syntax in next PR...

> [!NOTE]
> Not intended for Cython 3.3 unless a miracle happens. 

[1] https://docs.python.org/3/reference/lexical_analysis.html#soft-keywords


da-woods:
A little bit of this was in https://github.com/cython/cython/pull/7540. Obviously your PR goes quite a bit further and actually tries to use it. But possibly the parser bit might be worth comparing if you're unsure.

From a quick look I think it's quite similar. I cover an extra bit if syntax with

```
type MyTypeVar[T] = list[T]
```

I suspect you don't actually want to support this syntax on a first go though.

PR #7887, opened 2026-08-05
Fix dict infer indexed type

matusvalo:
The PR fixes issue with type inference of dict `get()` method.

### DIFF
--- a/Cython/Compiler/PyrexTypes.py
+++ b/Cython/Compiler/PyrexTypes.py
@@ -5476,11 +5476,11 @@ class BuiltinTypeConstructorObjectType(BuiltinObjectType, PythonTypeConstructorM
     def infer_indexed_type(self, at_index=None):
         container_type = self.get_container_type()
         if container_type.is_pytuple_type:
-            if at_index is None:
-                return self.get_common_item_type()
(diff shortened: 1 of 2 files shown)


Cython/Compiler/PyrexTypes.py
```diff
@@ -5475,13 +5475,14 @@ def assignable_from(self, src_type):
...
-            # tuple[TYP, ...]
-            return self.get_subscripted_type(0)
-        if container_type.is_pytuple_type and isinstance(at_index, int):
-            return self.get_subscripted_type(at_index)
+        if container_type.is_pytuple_type:
+            if at_index is None:
+                return self.get_common_item_type()
+            if self.has_uniform_element_type:
+                # tuple[TYP, ...]
+                return self.get_subscripted_type(0)
```
scoder, line 5481: This can probably be moved first. It seems quicker than `self.get_common_item_type()`.

scoder (approved):

Cython/Compiler/PyrexTypes.py
matusvalo: 🫡

PR #7888, opened 2026-08-05
Improve inference of builtin methods

matusvalo:
This PR introduces return type inference of the following methods:
* [x]  `dict.popitem()`
* [x] `dict.keys()`
* [x] `dict.values()`
* [x] `dict.items()`
* [x] Avoid registering type to builtin scope
* [ ] Test the return type parsing in `TestBuiltin.TestBuiltinReturnTypes` and adapt the return type tests

> [!NOTE]
> based on #7887


Cython/Compiler/Builtin.py
```diff
@@ -358,6 +358,9 @@ def _generate_divmod_function(scope, argument_types):
 
     ("int",     "&PyLong_Type",     []),
     ("float",   "&PyFloat_Type",   []),
+    ("dict_keys", "&PyDictKeys_Type", []),
+    ("dict_values", "&PyDictValues_Type", []),
+    ("dict_items", "&PyDictItems_Type", []),
```
scoder, line 370: This defines the three types as user visible builtins. They should not be user visible because there is no such type in Python's type system. We can define them as types (internally), but we shouldn't register them in the builtins scope (or any scope at all, I guess).

Cython/Compiler/Builtin.py
```diff
@@ -675,8 +689,14 @@ def find_return_type_of_builtin_method(pos, env, builtin_type, method_name):
...
             if return_type_name == 'I':
                 return builtin_type.infer_indexed_type() or PyrexTypes.py_object_type
+            if return_type_name == 'K' and container_type.is_pyanydict_type:
+                return builtin_type.infer_iterator_type() or PyrexTypes.py_object_type
             if 'T' in subscripted_type_names:
                 subscripted_type_names = subscripted_type_names.replace('T', builtin_type.name)
+            if 'I' in subscripted_type_names:
+                subscripted_type_names = subscripted_type_names.replace('I', builtin_type.infer_indexed_type().name)
+            if 'K' in subscripted_type_names and container_type.is_pyanydict_type:
+                subscripted_type_names = subscripted_type_names.replace('K', builtin_type.infer_iterator_type().name)
```
scoder: We should do these replacements later, after splitting the string below, when mapping type names to types. String replacements seem inherently unsafe here.

matusvalo: I refactored the logic to be more robust and avoid replacing types...

Cython/Compiler/Builtin.py
```diff
@@ -661,33 +675,47 @@ def _generate_divmod_function(scope, argument_types):
...
+    if sig == 'I':
+        return builtin_type.infer_indexed_type() or PyrexTypes.py_object_type
+    if sig == 'K' and builtin_type.is_pyanydict_type:
+        return builtin_type.infer_iterator_type() or PyrexTypes.py_object_type
+    if sig == 'bint':
+        return PyrexTypes.c_bint_type
+    if sig == 'Py_ssize_t':
+        return PyrexTypes.c_py_ssize_t_type
+    else:
+        return builtin_scope.lookup(sig).type
```
matusvalo: This search signature in `builtin_scope`, `dict_*` types among them.

Cython/Compiler/Builtin.py
matusvalo: > This defines the three types as user visible builtins

What does it exactly mean "visible"? The user see it:

```python
d = {"a": 1}
k = d.keys()

type(k)
# <class 'dict_keys'>
```

And can be used in `isinstance()`:
```python
isinstance(k, type(d.keys()))  # True
```

but **cannot be created manually** only via `keys()` (and friends).

Not sure if it counts as "part of Python's type system or not". This type is kind of mentioned also in [python docs](https://docs.python.org/3/library/stdtypes.html?utm_source=chatgpt.com#dictionary-view-objects).

In any case, current code needs to search the type in builtin scope (at least this PR)- e.g.: here:

https://github.com/cython/cython/pull/7888/files#r3740733070

Hence we have two possibilities:
* keep new `dict_*` types in `builtin_scope` and the rest of code can search the type dynamically from it`
* create the types manually in `PyrexTypes` and hard code them across code base (at least in https://github.com/cython/cython/pull/7888/files#r3740733070)

Not sure what is better approach. I don't know what are all implications of having the type in builtin scope so I cannot tell what is better... :-(

scoder: The way you declared them, users can write `k = dict_keys()` and `isinstance(k, dict_keys)` etc. That's probably not intended and definitely not good. The type names should not be user visible.

I think we need to construct the types manually like the other (C-) types in `PyrexTypes.py`, and not add them to a scope at all. At least, that's what I'd try.

scoder: I wrongly wrote "no such type in Python's type system" above, but what I actually meant was "not in Python's builtins". I'm only referring to the names of the types, not the types themselves. We clearly need the types for the return type declarations, so they need to exist, but they should not be part of the builtin scope (or any other scope) and it should not be possible for _users_ to use them as types. It should not be possible to look them up by name. I hope it's clearer now.

Cython/Compiler/Builtin.py
```diff
@@ -496,7 +502,12 @@ def _generate_divmod_function(scope, argument_types):
 # When updating this mapping, also update "unsafe_compile_time_methods" below
 # if methods are added that are not safe to evaluate at compile time.
 # 'T' - type identical to type in dictionary key
-# 'I' - type from subscript - e.g. for list[int], I is `int`
+# 'I' - type of index type:
+#     - for no-dict it is type from subscript - e.g. for list[int], I is `int`
+#     - for dict it is second subscripted type - e.g. for dict[int, str],
+#       I is `str`
+# 'K' - key value type of dictionary - second subscripted type
+#     - e.g. for dict[int, str], v is `str`
```
```suggestion
# 'I' - type of index type:
#     - value type from subscript - e.g. for list[int], I is `int`
#     - for mappings this is the second subscripted type - e.g. for dict[int, str], I is `str`
# 'K' - key type of dictionary - e.g. for dict[int, str], K is `int`
```

scoder:
Sorry for the trial commits. I pushed an untested proposal that moves the three types out of the builtin scope by simply declaring them in a throw-away scope. I think that's a reasonable thing to do here. Adding options for "declaring a type without actually declaring the type" seems too much effort for this simple one-time use case.

scoder:
This seems to work, mostly, but now the `TestBuiltin` test needs to be adapted as well. It also parses the signatures, which is now incomplete. And it should probably also test the signature parsing separately now (and can then use it in the tests for `find_return_type_of_builtin_method()`).

scoder (commented):
I've rewritten a part of the `TestBuiltin.py` test (in master, merged here) in order to include the generics resolution and to make it align better with the changes here.

I'd be happy to get this into 3.3 if you could get the test adapted.

matusvalo:
I'd be happy to get this into 3.3 if you could get the test adapted.

Last few days I had no time but I will try to finish this PR this evening...

matusvalo:
Or you were faster than me...

scoder:
Lovely error in GraalPy:
```
ERROR: test_oldargs0_0_ext (test_call.CFunctionCalls.test_oldargs0_0_ext)
  File "tests/run/test_call.py", line 58, in test_call.CFunctionCalls.test_oldargs0_0_ext
    {}.keys(*())
TypeError: Expected dict_keys, got dict_keys
```
This is because we call `__Pyx_PyDict_keys_CheckExact()` on the result, which is probably a subtype. However, `dict_keys` cannot be subtyped, at least in CPython:
```python
>>> t = type({}.keys())
>>> t
<class 'dict_keys'>
>>> class T(t): pass
  File "<python-input-3>", line 1, in <module>
    class T(t): pass
TypeError: type 'dict_keys' is not an acceptable base type
```
So this is unexpected. I'll report it to GraalPython and exclude the test there.

matusvalo:
Thank you @scoder for finishing up the PR.

PR #7889, opened 2026-08-06
Fix setitem on declared "list[cy.int]"

scoder:
If the item type of a container is declared as a C type, we must still coerce a value of the expected C type to a Python object on index assignments. The type of the item in the container is always a plain Python object, regardless of the declared item type.

Fixes https://github.com/cython/cython/pull/7689

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -4557,11 +4557,10 @@ class IndexNode(_IndexingBaseNode):
 
         self.wrap_in_nonecheck_node(env, getting)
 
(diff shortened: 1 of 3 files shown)


da-woods:
It's not hugely obvious from the PR description what this is fixing, but I checked and agree that it didn't seem to be generating valid code before. Otherwise looks fine.

intellimath:
May be it would be better to consider `list[cy.int]` as wrong declaration, because python list can not contain any of C simple type?

scoder:
> May be it would be better to consider `list[cy.int]` as wrong declaration, because python list can not contain any of C simple type?

See https://github.com/cython/cython/pull/7288#issuecomment-3535132454

PR #7892, opened 2026-08-07
Fold "CmpNode.is_c_string_contains()" into "find_special_bool_compare_function()".

### LINKED ISSUE #3888: [ENH] optimize u"a" in [unicode] (where literal has len == 1)
```cython
cdef bint contains_a_literal(unicode s) except? -1:
    return "a" in s

cdef bint contains_a_ucs4(unicode s) except? -1:
    return <Py_UCS4>"a" in s
```

The former turns into:

```c
  __pyx_t_1 = (__Pyx_PyUnicode_ContainsTF(__pyx_n_u_a, __pyx_v_s, Py_EQ)); if (unlikely(__pyx_t_1 < 0)) __PYX_ERR(0, 2, __pyx_L1_error)
```

Whereas the latter turns into:

```c
  __pyx_r = (__Pyx_UnicodeContainsUCS4(__pyx_v_s, ((Py_UCS4)97)));
```

The latter allows the C compiler much more scope to optimize the code, as it turns into a simple loop rather than invoking `PyUnicode_Contains`.

scoder:
Along the way, optimises `'c' in str`, `b'c' in bytes`, `char in bytearray`.

Closes https://github.com/cython/cython/issues/3888

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -14387,11 +14387,20 @@ class CmpNode:
             or (self.cascade and self.cascade.is_python_result()))
 
     def is_c_string_contains(self):
-        return self.operator in ('in', 'not_in') and \
-               ((self.operand1.type.is_int
-                 and (self.operand2.type.is_string or self.operand2.type.is_pybytes_type)) or
-                (self.operand1.type.is_unicode_char
-                 and self.operand2.type.is_pystr_type))
+        if self.operator not in ('in', 'not_in'):
+            return False
+        op1_type = self.operand1.type
+        op2_type = self.operand2.type
+        if op1_type.is_unicode_char:
+            return op2_type.is_pystr_type
+        if self.operand1.is_string_literal and self.operand1.can_coerce_to_char_literal():
+            if op1_type.is_pystr_type:
+                return op2_type.is_pystr_type
+            if op1_type.is_pybytes_type:
+                return op2_type.is_pybytes_type or op2_type.is_pybytearray_type
+        if op1_type.is_int:
+            return op2_type.is_string or op2_type.is_pybytes_type or op2_type.is_pybytearray_type
+        return False
 
     def is_ptr_contains(self):
         if self.operator in ('in', 'not_in'):
@@ -14660,15 +14669,20 @@ class PrimaryCmpNode(ExprNode, CmpNode):
                 common_type = None
                 if self.cascade:
                     error(self.pos, "Cascading comparison not yet supported for 'int_val in string'.")
+                    self.type = error_type
                     return self
-                if self.operand2.type.is_pystr_type:
+                if type2.is_pystr_type:
+                    self.operand1 = self.operand1.coerce_to(PyrexTypes.c_py_ucs4_type, env)
                     env.use_utility_code(UtilityCode.load_cached("PyUCS4InUnicode", "StringTools.c"))
                 else:
-                    if self.operand1.type is PyrexTypes.c_uchar_type:
+                    if type1 is PyrexTypes.c_uchar_type or type1.is_pybytes_type:
                         self.operand1 = self.operand1.coerce_to(PyrexTypes.c_char_type, env)
-                    if not self.operand2.type.is_pybytes_type:
-                        self.operand2 = self.operand2.coerce_to(bytes_type, env)
-                    env.use_utility_code(UtilityCode.load_cached("BytesContains", "StringTools.c"))
+                    if type2.is_pybytearray_type:
+                        env.use_utility_code(UtilityCode.load_cached("ByteArrayContains", "StringTools.c"))
+                    else:
+                        if not type2.is_pybytes_type:
+                            self.operand2 = self.operand2.coerce_to(bytes_type, env)
+                        env.use_utility_code(UtilityCode.load_cached("BytesContains", "StringTools.c"))
                 self.operand2 = self.operand2.as_none_safe_node(
                     "argument of type 'NoneType' is not iterable")
             elif self.is_ptr_contains():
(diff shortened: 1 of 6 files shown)


Cython/Compiler/Optimize.py
```diff
@@ -1310,8 +1310,7 @@ def extract_conditions(self, cond, allow_not_in):
         if isinstance(cond, ExprNodes.PrimaryCmpNode):
             if cond.cascade is not None:
                 return self.NO_MATCH
-            elif cond.is_c_string_contains() and \
-                   isinstance(cond.operand2, (ExprNodes.UnicodeNode, ExprNodes.BytesNode)):
+            elif cond.operator in ('in', 'not_in') and cond.operand1.type.is_int and cond.operand2.is_string_literal:
```
scoder, line 1313: This condition simplifies a lot, because it only needs to handle what we now get from type analysis.

Cython/Compiler/ExprNodes.py
```diff
@@ -14388,30 +14384,51 @@ def find_special_bool_compare_function(self, env, operand1, result_is_bool=False
...
-            else:
-                if not self.operand2.type.is_pyobject:
-                    self.operand2 = self.operand2.coerce_to_pyobject(env)
-                self.special_bool_cmp_utility_code = UtilityCode.load_cached("PySequenceContains", "ObjectHandling.c")
-                self.special_bool_cmp_function = "__Pyx_PySequence_ContainsTF"
-                return True
-        return False
+                return True, operand1.coerce_to_pyobject(env)
+            elif type2.is_pystr_type:
+                self.operand2 = self.operand2.as_none_safe_node("argument of type 'NoneType' is not iterable")
```
scoder, line 14448: It turns out that this is the error message that CPython uses, so we got it wrong all the time. Even Py2.6 already had it this way. (I also checked Py2.4 and it only says "iterable argument required".) I changed it for the string types that I touched here but left it for the other builtin types above, to avoid even more churn.

Cython/Compiler/ExprNodes.py
```diff
@@ -14608,41 +14625,20 @@ def analyse_types(self, env):
...
-                self.operand1 = self.operand1.coerce_to_pyobject(env)
-            common_type = None  # if coercion needed, the method call above has already done it
+        if self.is_ptr_contains():
+            if self.cascade:
+                error(self.pos, "Cascading comparison not supported for 'val in sliced pointer'.")
+            self.type = PyrexTypes.c_bint_type
+            # Will be transformed by IterationTransform
+            return self
+        found, operand1 = self.find_special_bool_compare_function(env, self.operand1)
+        if found:
```
scoder, line 14682: It is a bit clumsy to return a flag and a node. But what I really like about this is that it moves all decisions into `find_special_bool_compare_function()` and coerces both operands as needed, in one place.

I considered returning only the coerced operand node or None (to say that no special case matched), but I failed to find a good name for such a function. It seems more explicit to return both.

Cython/Compiler/ExprNodes.py
```diff
@@ -14608,41 +14625,20 @@ def analyse_types(self, env):
         if self.cascade:
             self.cascade = self.cascade.analyse_types(env)
 
-        if self.operator in ('in', 'not_in'):
-            if self.is_c_string_contains():
-                common_type = None
-                if self.cascade:
-                    error(self.pos, "Cascading comparison not yet supported for 'int_val in string'.")
-                    return self
```
scoder, line 14663: This limitation is gone. I found it difficult to distribute the implementation over `CmpNode` and `PrimaryCmpNode` and ran into issues with `CascadedCmpNode`, so I fixed up the implementation until it no longer complained. But this is a new feature as well that is not thoroughly tested. Good thing is that it was previously forbidden, so existing code cannot use it. And code with cascaded comparisons is probably rare outside of numeric ranges, so not very likely to appear for strings in combination with `in` tests.

scoder (commented):
A few comments to help with the reviews.

Cython/Compiler/ExprNodes.py
da-woods: One variantion that I don't think you've tested is a single character literal as the middle argument, e.g.

```
c in "f" in some_str
```

Just because the single character literal is special-cased when on the left-hand side but not when on the right-hand side so I can see there's potential to go wrong there.

(I haven't tested this myself either so it's a suggestion of what could go wrong rather than something I believe does go wrong)

scoder: Good call, that was missing. But luckily, it works as it is.

scoder: Ah, beautiful code:
```c
 *     cdef int result = a not in 'f' not in cascade             # <<<<<<<<<<<<<<
 *     return result
 * 
*/
  /* __pyx_t_1 allocated (bint) */
  __pyx_t_1 = (__Pyx_PyUnicode_ContainsTF(__pyx_v_a, __pyx_mstate_global->__pyx_n_u_f, Py_NE)); if (unlikely((__pyx_t_1 < 0))) __PYX_ERR(0, 306, __pyx_L1_error)
  if (__pyx_t_1) {
    if (unlikely(__pyx_v_cascade == Py_None)) {
      PyErr_SetString(PyExc_TypeError, "argument of type \047NoneType\047 is not iterable");
      __PYX_ERR(0, 306, __pyx_L1_error)
    }
    __pyx_t_1 = (__Pyx_UnicodeContainsUCS4(0x66, __pyx_v_cascade, Py_NE)); if (unlikely((__pyx_t_1 < 0))) __PYX_ERR(0, 306, __pyx_L1_error)
  }
```
Ok, for some value of beautiful, that is… but it does nicely separate the two conditions and evaluate them both efficiently.

PR #7895, opened 2026-08-10
Report an error instead of crashing on a C++ constructor call with keyword arguments

### LINKED ISSUE #7206: [BUG] Compiler crash attempting construction with named initialisation
### Describe the bug

Probably shouldn't compile, but probably shouldn't crash too.

```cython
# distutils: language=c++
cdef extern from *:
    """
    class C {
        public:
            int i;
            C() : i(0) {}
            C(int j) : i(j) {}
    };
    """

    cdef cppclass C:
        int i
        C()
        C(int j)

cdef C c = C(i=0)
```

```
Error compiling Cython file:
...
    """

    cdef cppclass C:
        int i

cdef C c = C(i=0)
            ^

wtf.pyx:13:12: Compiler crash in AnalyseExpressionsTransform

File 'Nodes.py', line 409, in analyse_expressions: StatListNode(wtf.pyx:2:0)
File 'Nodes.py', line 6150, in analyse_expressions: SingleAssignmentNode(wtf.pyx:13:0)
File 'Nodes.py', line 6283, in analyse_types: SingleAssignmentNode(wtf.pyx:13:0)
File 'ExprNodes.py', line 7180, in analyse_types: GeneralCallNode(wtf.pyx:13:12,
    result_is_used = True,
    use_managed_ref = True)
File 'ExprNodes.py', line 6068, in analyse_as_type_constructor: GeneralCallNode(wtf.pyx:13:12,
    result_is_used = True,
    use_managed_ref = True)

  File "/usr/local/lib/python3.13/dist-packages/Cython/Compiler/ExprNodes.py", line 6068, in analyse_as_type_constructor
    self.args = [ arg.analyse_types(env) for arg in self.args ]
                                                    ^^^^^^^^^
AttributeError: 'GeneralCallNode' object has no attribute 'args'
```

### Expected behaviour

An error describing the problem (e.g. "C++ constructor got unexpected keyword argument"), rather than a compiler crash

### OS

Linux

### Python version

3.13.3

### Cython version

3.0.11 and 3.1.4

vsaraikin:
Closes #7206

`C(i=0)` on a C++ class crashes the compiler:

```
AttributeError: 'GeneralCallNode' object has no attribute 'args'
```

`analyse_as_type_constructor` sits on `CallNode`, so both call node types reach it. The struct branch reads its arguments through `explicit_args_kwds()`. The C++ branch reads `self.args` and only `SimpleCallNode` has that. Keyword arguments make the node a `GeneralCallNode`.

So the C++ branch now steps aside when there are keyword arguments. Normal call analysis already rejects them with "Non-trivial keyword arguments and starred arguments not allowed in cdef functions."

The second hunk stops that error cascading. It reported and fell through, so the node ended up typed as a Python object and two more messages followed. `C(1, 2, 3)` gets one message, so this should too. Same `error_type` guard as in #7861.

```
C(i=0) // one error, was a crash
C(*a) // one error, was a crash
C(0) // ok
C() // ok
```

Three cpp_stl_algo tests fail here, but they fail on master too. Recent libc++ wants a comparator reference where the generated code passes a function pointer. It also warns that the platform is no longer supported

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -6355,7 +6355,13 @@ class CallNode(ExprNode):
             node = node.analyse_types(env).coerce_to(type, env)
             return node
         elif type and type.is_cpp_class:
-            self.args = [ arg.analyse_types(env) for arg in self.args ]
+            args, kwds = self.explicit_args_kwds()
+            if kwds:
+                # Only SimpleCallNode has '.args'. Leave a call with keyword
+                # arguments to the normal call analysis, which rejects it, rather
+                # than crashing on the missing attribute below.
+                return None
+            self.args = [ arg.analyse_types(env) for arg in args ]
             constructor = type.scope.lookup("<init>")
             if not constructor:
                 error(self.function.pos, "no constructor found for C++  type '%s'" % self.function.name)
@@ -7516,6 +7522,8 @@ class GeneralCallNode(CallNode):
                     error(self.pos,
                           "Non-trivial keyword arguments and starred "
                           "arguments not allowed in cdef functions.")
+                    self.type = error_type
+                    return self
                 else:
                     # error was already reported
                     pass
--- a/tests/errors/cpp_class_kwargs.pyx
+++ b/tests/errors/cpp_class_kwargs.pyx
@@ -1,0 +1,27 @@
+# mode: error
+# tag: cpp
+# ticket: 7206
+
+cdef extern from *:
+    """
+    class C {
+        public:
+            int i;
+            C() : i(0) {}
+            C(int j) : i(j) {}
+    };
+    """
+
+    cdef cppclass C:
+        int i
+        C()
+        C(int j)
+
+
+def f():
+    cdef C c = C(i=0)
+
+
+_ERRORS = u"""
+22:16: Non-trivial keyword arguments and starred arguments not allowed in cdef functions.
+"""
--- a/tests/errors/cpp_class_kwargs.pyx
+++ b/tests/errors/cpp_class_kwargs.pyx
@@ -19,9 +19,11 @@ cdef extern from *:
 
 
 def f():
-    cdef C c = C(i=0)
+    cdef C wrong_keyword_name = C(i=0)
+    cdef C ok = C(j=0)
 
 
 _ERRORS = u"""
-22:16: Non-trivial keyword arguments and starred arguments not allowed in cdef functions.
+22:33: Non-trivial keyword arguments and starred arguments not allowed in cdef functions.
+23:17: Non-trivial keyword arguments and starred arguments not allowed in cdef functions.
 """


scoder:
Thanks.

It's not quite true that keyword arguments are always rejected for C/C++ calls. If the function declaration has named arguments (and not just a list of plain types, as C/C++ allows), we try to map keyword arguments to the declared argument names at compile time. I'm not sure where and when exactly this happens (it's certainly been more than a decade since I implemented this), but it's definitely intended to allow keyword arguments for C/C++ function calls.

tests/errors/cpp_class_kwargs.pyx
```diff
@@ -0,0 +1,27 @@
...
+    """
+
+    cdef cppclass C:
+        int i
+        C()
+        C(int j)
+
+
+def f():
+    cdef C c = C(i=0)
```
scoder: Let's add a (supposedly) valid call case to make it clearer what the problem is.
```suggestion
    cdef C wrong_keyword_name = C(i=0)
    cdef C ok = C(j=0)
```

Cython/Compiler/ExprNodes.py
```diff
@@ -7518,6 +7522,8 @@ def analyse_types(self, env):
                     error(self.pos,
                           "Non-trivial keyword arguments and starred "
                           "arguments not allowed in cdef functions.")
```
scoder, line 7524: I'm surprised that you only get this error message. There are a number of errors that `e_cdef_keywords_T241.pyx` tests for – but not this one. I'm not sure what the difference is in the case of calling a C++ class compared to a normal C function. That seems worth looking into. The other errors seem much more appropriate.

We definitely need a separate test file (as you created) for C++ calls due to the different compile mode.

vsaraikin:
Looked into it. The C++ path never reaches the keyword mapping at all.

`map_to_simple_call_node` bails at its third check. Instrumenting it for `C(j=0)`:

```
function: NameNode | entry.type: C | is_cfunction: 0
```

For a constructor call the function node still refers to the class type rather than to a cfunction, so it returns `self` before any name matching happens. That is also why the message is the generic one: "C function got unexpected keyword argument" lives inside that same function, so the C++ path cannot produce it.

Making `C(j=0)` work needs more than routing it there. `<init>` carries the overloads:

```
ctor.type          void (void)            <- only the first one
all_alternatives   []   and   [('j', 'int')]
```

`map_to_simple_call_node` matches names against a single signature, so it would have to try each alternative and pick the one whose names fit. That is overload resolution by keyword name, which reads like a feature rather than part of this fix.

Right now both `C(i=0)` and `C(j=0)` get the same error, which the updated test shows. Happy to do the mapping here if you want it in this PR, or to leave this as the crash fix and open a separate issue for keyword support

scoder:
Thanks for looking into this. It's good that it didn't work before, then we can't break anything and it's really best left as a new feature. Not in this PR.

We'd have to add a number of new rules to the existing signature matching, such as pruning signatures that lack named keywords used in the call or where a named keyword would overlap with a positional argument. Doesn't sound all trivial.

scoder:
Thanks

scoder:
3.2.x commit is https://github.com/cython/cython/commit/8503201dbae1a226c49c04a55b23b6e067feec7d

PR #7896, opened 2026-08-11
Include subtypes of builtin types in some optimisations.

scoder:
In some cases, we generate inlined special casing code that matches a type slot and target it at certain known types, e.g. list/tuple iteration. We currently exclude subtypes because they can overwrite the base type's behaviour in arbitrary ways.

But it's easy and quick (in CPython) to compare the slot function pointer to that of the base type. If it has not been overwritten by the subclass, we can apply the same optimisation because we know that the behaviour has not changed.

It's not entirely clear that this is worth it and in which cases. Subtypes are rare, and having more guard conditions in place can slow down the really important (plain type) cases.

It should be worth doing when the initial overhead is small compared to the operation, e.g. when deciding whether to optimise a for-loop. For very short-running operations, it's probably less helpful.

### DIFF
--- a/Cython/Utility/ModuleSetupCode.c
+++ b/Cython/Utility/ModuleSetupCode.c
@@ -984,6 +984,24 @@ static CYTHON_INLINE void *__Pyx__PyModule_GetState(PyObject *op)
   #define __Pyx_PyType_TryGetSubSlot(obj, sub, name, func_ctype) __Pyx_PyType_TryGetSlot(obj, name, func_ctype)
 #endif
 
+#if CYTHON_USE_TYPE_SLOTS
+  #define __Pyx_SlotIsInherited(obj, base_type, slot_name)  \
+    (__Pyx_PyObject_GetSlot(obj, slot_name, void*) == __Pyx_PyType_GetSlot(base_type, slot_name, void*))
+  #define __Pyx_SubSlotIsInherited(obj, base_type, sub, slot_name)  \
+    (__Pyx_PyObject_GetSubSlot(obj, sub, slot_name, void*) == __Pyx_PyType_GetSubSlot(base_type, sub, slot_name, void*))
+#else
+  #define __Pyx_SlotIsInherited(obj, base_type, slot_name)  (0)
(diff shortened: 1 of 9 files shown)


Cython/Utility/ModuleSetupCode.c
```diff
@@ -984,6 +984,29 @@ static CYTHON_INLINE void *__Pyx__PyModule_GetState(PyObject *op)
   #define __Pyx_PyType_TryGetSubSlot(obj, sub, name, func_ctype) __Pyx_PyType_TryGetSlot(obj, name, func_ctype)
 #endif
 
+#if CYTHON_COMPILING_IN_CPYTHON && CYTHON_USE_TYPE_SLOTS
```
da-woods, line 987: I think it'll fail for the 3.9 Limited API (where `PyType_GetSlot` only supports heap-types, which does not include most builtin types and just reads some arbitrary memory where the type slots table should be).

The options are:
* Exclude on 3.9 Limited API.
* Use `PyType_TryGetSlot` (plus some extra logic about `NULL` slots since the functions will now return `NULL`)

I'd probably exclude personally, but either would work

da-woods:
I've had some interest in something like this for relaxing the "annotations typing must be exact types" rule (where it contradicts the general Python annotation typing rules). With that said, doing it there would need to be a lot more comprehensive than this (where it's just an optimization for a few places) so it definitely isn't worth getting distracted by here. (There also doesn't seem to be a standard way of annotating an exact type, which would be nice to be able to do)

Cython/Utility/ModuleSetupCode.c
da-woods: Actually, you wouldn't want `TryGetSlot` because that adds an extra "is heap type" check that would always fail. It'd be better to do a runtime version check only if you were going to go for the more complex option on 3.9.

I'd still just exclude it though personally

da-woods: Actually... looking more, this is excluded from the Limited API entirely so the whole thing is a non-issue...

Issue #7894, opened 2026-08-10
[BUG] Compiler crash in ConstantFolding: starred unpacking in a comprehension target on the RHS of an augmented assignment


ThiefTMBadBad:
### Describe the bug

Cython crashes in `ConstantFolding` when a comprehension whose loop target contains a starred name (`*_`) appears as the right-hand side of an **augmented** assignment (`+=`, `|=`).

`Cython/Compiler/Optimize.py:visit_SequenceNode` assumes every arg has a `.target` attribute, but a starred target yields a `NameNode`, which does not.

### Code to reproduce the behaviour

```python
# a.py — valid, working Python; f([('a', 1, 'x'), ('b', 2)]) == [1, 2]
def f(xs):
    out = []
    out += [c for k, c, *_ in xs]
    return out
```

```console
$ cythonize -3 a.py
```

What matters is the **augmented** assignment; a plain `+` compiles fine:

| variant | result |
|---|---|
| `out += [c for k, c, *_ in xs]` | **crash** |
| `out += [x for x, *_ in xs]` | **crash** |
| `out \|= {x for x, *_ in xs}` (set comp) | **crash** |
| `out = [c for k, c in p] + [c for k, c, *_ in t]` (plain `+`) | ok |
| `out = [c for k, c, *_ in xs]` (no augmented assignment) | ok |
| `out += list(c for k, c, *_ in xs)` (genexp through `list()`) | ok |

### Expected behaviour

Compiles, like the plain-`+` and plain-assignment forms do.

### OS

macOS 15 (Darwin 25.4.0) and Debian (`python:3.12` Docker image) — both reproduce.

### Python version

3.12.13, 3.14

### Cython version

**Every version tested crashes**, spanning three major releases:

| Cython | Result |
|---|---|
| 0.29.37 | crash |
| 3.0.0, 3.0.12 | crash |
| 3.1.0, 3.1.4, 3.1.8 | crash |
| 3.2.0, 3.2.4, 3.2.8, 3.2.9 | crash |
| 3.3.0a1 | crash |

### Additional context

Verbatim output on 3.2.9 / CPython 3.12.13:

```
Error compiling Cython file:
------------------------------------------------------------
...
def f(xs):
    out = []
    out += [c for k, c, *_ in xs]
                  ^
------------------------------------------------------------
a.py:3:18: Compiler crash in ConstantFolding

AddNode.operand2 = ComprehensionNode(a.py:3:11,
    has_local_scope = True,
    is_temp = True,
[...]

PR #7897, opened 2026-08-12
Fix ConstantFolding crash on starred names in assignment targets

AbhinavMir:
#7894


Cython/Compiler/Optimize.py
scoder: I think we can simplify this code to:
```suggestion
            if isinstance(arg, ExprNodes.StarredUnpackingNode) and (
                    arg.target.is_sequence_constructor and not arg.target.mult_factor):
                args.extend(arg.target.args)
            else:
                args.append(arg)
```

tests/run/starred_comprehension_target_T7894.py
```diff
@@ -0,0 +1,22 @@
+# mode: run
+# ticket: 7894
```
scoder: The tests can go into `tests/run/extended_unpacking_T235.pyx`.
We try to avoid creating a new test file for each different case because compilation runs are quite costly.

scoder (commented):
Thanks. Looks good, but can be further simplified.

AbhinavMir:
Thanks @scoder ! Reviewed and addressed ! Both make sense to me.

scoder (approved):

scoder:
Thanks

scoder:
3.2.x commit is https://github.com/cython/cython/commit/94f657fbccc7107a99b76a3cdd062f68542f6b02

Issue #7898, opened 2026-08-12
[ENH] Implement PEP 798: Unpacking in Comprehensions


scoder:
### Is your feature request related to a problem? Please describe.

[PEP 798](https://peps.python.org/pep-0798/) in Py3.15 adds unpacking in comprehensions:
```cython
result = [*many for many in values]
```

### Describe the solution you'd like.

The list comprehension case essentially replaces `PyList_Append()` with `PyList_Extend()` (or an optimised variant). Set and dict comprehensions should be equally easy to do using `Py*_Update()`.

Generator expressions are quite a different thing. But we might be able to do this entirely at the node level, injecting another nested for-loop level. That might be a separate PR to the C code changes for comprehensions.

Note the reasons why `yield from` semantics were explicitly not chosen for generator expressions:
https://peps.python.org/pep-0798/#appendix-semantics-of-generator-delegation

The comprehensions seem a good first issue target, assuming a bit of C knowledge.
Look for the `ListCompAppend` utility code and its usages in `ExprNodes.py` to get started.

fzlzjerry:
I'd like to work on the list, set, and dict comprehension part of this.

I confirmed that current `master` rejects the new syntax and that the existing `listcomp`, `setcomp`, and `dictcomp` tests pass on both C and C++ backends. My plan is to represent top-level `*`/`**` comprehension expressions as update operations, using list extend, set update, and dict update while preserving the existing comprehension evaluation order. I'll add run tests for all three forms, including filtering, nested loops, duplicate dict keys, and invalid inputs.

I'll leave generator-expression semantics to a separate follow-up, as suggested in the issue.

scoder:
Great. Note that the dict comprehension part is also a bit more involved because we currently assume key-value pairs in `DictComprehensionAppendNode`. It might be worth adding a new node class for the star unpacking case, similar to the existing `DictComprehensionAppendNode` but with a single result expression in `child_attrs`.

PR #7899, opened 2026-08-12
Support PEP 798 unpacking in comprehensions

fzlzjerry:
This implements the list, set, and dictionary comprehension portion of PEP 798 from #7898.

List and set comprehensions keep using `ComprehensionAppendNode`. During analysis it records whether the expression is starred, and code generation then extends the result list or updates the result set instead of appending one value. Dictionary unpacking uses a dedicated `DictComprehensionUpdateNode` to update from each produced mapping, preserving later-key-wins behavior and the mapping-specific `TypeError`.

The parser accepts top-level `*` and `**` in the corresponding comprehensions, including async comprehensions. Generator expressions are intentionally left for a separate follow-up because their lowering is different.

The new run test covers filtering and nested loops, evaluation order, assignment-expression scope, all three collection types, duplicate dictionary keys, the mapping protocol, invalid inputs, async comprehensions, C and C++ backends, and the Limited API.

Validation:

- `python runtests.py -v --backends=c,cpp listcomp setcomp dictcomp comprehension_unpacking_T7898` (94 passed)
- `python runtests.py -v --no-cpp --limited-api=3.11 comprehension_unpacking_T7898` (12 passed)
- related compiler unit tests (18 passed)
- code-style check

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -9702,6 +9702,40 @@ class ComprehensionAppendNode(Node):
     def annotate(self, code):
         self.expr.annotate(code)
 
+
+class ComprehensionUpdateNode(ComprehensionAppendNode):
+    """Extend or update a comprehension result with each expression value."""
+
+    def generate_execution_code(self, code):
+        self.expr.generate_evaluation_code(code)
+
+        if self.target.type.is_pylist_type:
+            code.globalstate.use_utility_code(
+                UtilityCode.load_cached("ListExtend", "Optimize.c"))
+            code.put_error_if_neg(self.pos, "__Pyx_PyList_Extend(%s, %s)" % (
+                self.target.result(), self.expr.py_result()))
+        elif self.target.type.is_pyset_type:
+            code.globalstate.use_utility_code(
+                UtilityCode.load_cached("PySet_Update", "Builtins.c"))
+            code.put_error_if_neg(self.pos, "__Pyx_PySet_Update(%s, %s)" % (
+                self.target.result(), self.expr.py_result()))
+        elif self.target.type.is_pydict_type:
+            code.globalstate.use_utility_code(
+                UtilityCode.load_cached("RaiseMappingExpected", "FunctionArguments.c"))
+            code.putln("if (unlikely(PyDict_Update(%s, %s) < 0)) {" % (
+                self.target.result(), self.expr.py_result()))
+            code.putln("if (PyErr_ExceptionMatches(PyExc_AttributeError)) "
+                       "__Pyx_RaiseMappingExpectedError(%s);" % self.expr.py_result())
+            code.putln(code.error_goto(self.pos))
+            code.putln("}")
+        else:
+            raise InternalError(
+                "Invalid type for comprehension node: %s" % self.target.type)
+
+        self.expr.generate_disposal_code(code)
+        self.expr.free_temps(code)
+
+
 class DictComprehensionAppendNode(ComprehensionAppendNode):
     child_attrs = ['key_expr', 'value_expr']
 
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -9678,14 +9678,18 @@ class ComprehensionAppendNode(Node):
         steal_temp = False
         if self.target.type.is_pylist_type:
             if self.is_starred:
-                code.globalstate.use_utility_code(
-                    UtilityCode.load_cached("ListExtend", "Optimize.c"))
+                utility_code_name = "ListExtend"
                 function = "__Pyx_PyList_Extend"
+            elif self.expr.is_temp:
+                steal_temp = True
+                utility_code_name = "ListCompAppendAndDecref"
+                function = "__Pyx_ListComp_AppendAndDecref"
             else:
-                steal_temp = self.expr.is_temp
-                code.globalstate.use_utility_code(
-                    UtilityCode.load_cached("ListCompAppendAndDecref" if steal_temp else "ListCompAppend", "Optimize.c"))
-                function = "__Pyx_ListComp_AppendAndDecref" if steal_temp else "__Pyx_ListComp_Append"
+                utility_code_name = "ListCompAppend"
+                function = "__Pyx_ListComp_Append"
+
+            code.globalstate.use_utility_code(
+                UtilityCode.load_cached(utility_code_name, "Optimize.c"))
         elif self.target.type.is_pyset_type:
             if self.is_starred:
                 code.globalstate.use_utility_code(
@@ -9723,12 +9727,10 @@ class DictComprehensionUpdateNode(ComprehensionAppendNode):
     def generate_execution_code(self, code):
         self.expr.generate_evaluation_code(code)
 
-        if not self.target.type.is_pydict_type:
-            raise InternalError(
-                "Invalid type for dict comprehension node: %s" % self.target.type)
+        assert self.target.type.is_pydict_type, self.target.type
 
-        # PyDict_Update raises AttributeError for non-mappings, whereas **
-        # unpacking raises TypeError. Match Python's unpacking error here.
+        # PyDict_Update raises AttributeError for non-mappings, whereas "**dict" unpacking
+        # raises TypeError. Match Python's unpacking error here.
         code.globalstate.use_utility_code(
             UtilityCode.load_cached("RaiseMappingExpected", "FunctionArguments.c"))
         code.putln("if (unlikely(PyDict_Update(%s, %s) < 0)) {" % (
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -9675,35 +9675,24 @@ class ComprehensionAppendNode(Node):
         return self
 
     def generate_execution_code(self, code):
-        if self.is_starred:
-            self.expr.generate_evaluation_code(code)
-
-            if self.target.type.is_pylist_type:
+        steal_temp = False
+        if self.target.type.is_pylist_type:
+            if self.is_starred:
                 code.globalstate.use_utility_code(
                     UtilityCode.load_cached("ListExtend", "Optimize.c"))
                 function = "__Pyx_PyList_Extend"
-            elif self.target.type.is_pyset_type:
+            else:
+                steal_temp = self.expr.is_temp
+                code.globalstate.use_utility_code(
+                    UtilityCode.load_cached("ListCompAppendAndDecref" if steal_temp else "ListCompAppend", "Optimize.c"))
+                function = "__Pyx_ListComp_AppendAndDecref" if steal_temp else "__Pyx_ListComp_Append"
+        elif self.target.type.is_pyset_type:
+            if self.is_starred:
                 code.globalstate.use_utility_code(
                     UtilityCode.load_cached("PySet_Update", "Builtins.c"))
                 function = "__Pyx_PySet_Update"
             else:
-                raise InternalError(
-                    "Invalid type for starred comprehension node: %s" % self.target.type)
-
-            code.put_error_if_neg(self.pos, "%s(%s, %s)" % (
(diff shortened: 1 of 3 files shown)


scoder:
Note that list/set comprehensions share an append node but dictcomps have their own.
I don't think we need a new append node for list/set comps, they can just detect at code generation time that the target is starred.
A new append node for dict comps seems reasonable, though.

da-woods:
This user gives the impression that they're spraying unreviewed AI PRs widely across github in the hope that something sticks. I don't think we should encourage that generally.

scoder:
~Yep, a PR within seven minutes and no response to feedback is telling. I'll reuse the generated tests and implement the rest myself.~

fzlzjerry:
Updated in `f1ca96781` along those lines. List and set comprehensions now keep using `ComprehensionAppendNode`; it records the starred expression during analysis and selects extend/update during code generation. Only `**` dict comprehensions use the dedicated `DictComprehensionUpdateNode`.

I reran the focused list/set/dict comprehension suite on both C and C++ backends (94 tests), the new test under the Python 3.11 limited API (12 tests), the related compiler unit tests (18 tests), and the code-style check.

Cython/Compiler/ExprNodes.py
```diff
@@ -9661,16 +9661,41 @@ class ComprehensionAppendNode(Node):
...
+            self.expr.generate_evaluation_code(code)
+
+            if self.target.type.is_pylist_type:
+                code.globalstate.use_utility_code(
+                    UtilityCode.load_cached("ListExtend", "Optimize.c"))
+                function = "__Pyx_PyList_Extend"
+            elif self.target.type.is_pyset_type:
+                code.globalstate.use_utility_code(
+                    UtilityCode.load_cached("PySet_Update", "Builtins.c"))
+                function = "__Pyx_PySet_Update"
```
scoder: These can be folded into the two single value conditions below. We don't need to duplicate the entire implementation.

Cython/Compiler/ExprNodes.py
```diff
@@ -9702,6 +9727,30 @@ def generate_function_definitions(self, env, code):
...
+            raise InternalError(
+                "Invalid type for dict comprehension node: %s" % self.target.type)
+
+        code.globalstate.use_utility_code(
+            UtilityCode.load_cached("RaiseMappingExpected", "FunctionArguments.c"))
+        code.putln("if (unlikely(PyDict_Update(%s, %s) < 0)) {" % (
+            self.target.result(), self.expr.py_result()))
+        code.putln("if (PyErr_ExceptionMatches(PyExc_AttributeError)) "
+                   "__Pyx_RaiseMappingExpectedError(%s);" % self.expr.py_result())
+        code.putln(code.error_goto(self.pos))
```
scoder, line 9740: I don't see why we'd need special error handling here. Why not just use `code.error_goto_if_neg()` ? Is it to change the error message?

Cython/Compiler/Parsing.py
scoder: We can simplify the comprehension conditions below by moving the first up:
```suggestion
    if s.sy in ('for', 'async') and len(parts) == 1:
        # dict/set comprehension
        item = parts[0]
        if isinstance(item, list) and len(item) == 1:
            item = item[0]
```
The parse error will be the same in the end.

And I don't think the two cases append/extend are special enough to merit a completely separate code branch. The decision can equally be based on `.is_starred` – if the `StarredUnpackingNode` was created as expected.

Cython/Compiler/Parsing.py
```diff
@@ -1507,18 +1505,24 @@ def p_dict_or_set_maker(s: PyrexScanner):
...
-            return ExprNodes.ComprehensionNode(pos, loop=loop, append=append, type=comprehension_type)
-        else:
-            # syntax error, try to find a good error message
-            if len(parts) == 1 and not isinstance(parts[0], list):
-                s.error("iterable unpacking cannot be used in comprehension")
+        elif len(parts) == 1 and not isinstance(parts[0], list):
+            item = parts[0]
+            if target_type == 1:
+                comprehension_type = Builtin.set_type
+                item = ExprNodes.StarredUnpackingNode(item.pos, target=item)
```
scoder: Isn't the wrapping done in `p_starred_expr()` ? It should be.

Cython/Compiler/Parsing.py
fzlzjerry: Simplified in `f816b94f7`: the single-part condition is handled first, the one-item list is unwrapped once, and list/set generation now shares `ComprehensionAppendNode`; its existing `is_starred` flag selects append versus extend/update during code generation.

Cython/Compiler/Parsing.py
fzlzjerry: Yes. `f816b94f7` now calls `p_starred_expr()` while the scanner is still on `*`, so it creates the `StarredUnpackingNode` directly. The wrapper is retained for comprehensions and unwrapped again for ordinary merged set literals. I also reran the existing PEP 448 unpacking suite to cover that adjacent path.

Cython/Compiler/ExprNodes.py
fzlzjerry: Yes, the special handling preserves `**` unpacking exception semantics. `PyDict_Update()` reports `AttributeError` when the object has no mapping protocol, whereas `**` unpacking requires a `TypeError` saying that the object is not a mapping. This follows the existing `MergedDictNode` update path, and `invalid_dict_unpacking` covers the `TypeError` and message. I added a comment in `f816b94f7` to make the reason explicit.

Cython/Compiler/ExprNodes.py
fzlzjerry: Folded in `f816b94f7`. `generate_execution_code()` now selects append/extend or add/update inside the two target-type branches, then shares expression evaluation, error handling, and cleanup.

fzlzjerry:
Addressed the four inline comments in `f816b94f7`.

Validation on the updated branch:

- list/set/dict comprehension suite on C and C++ backends: 94 tests passed
- new unpacking-comprehension test on the Python 3.11 Limited API: 12 tests passed
- existing PEP 448 unpacking suite on C/C++ and language levels 2/3: 116 tests passed
- related compiler unit tests: 18 tests passed
- code-style check passed

Cython/Compiler/ExprNodes.py
```diff
@@ -9673,12 +9677,22 @@ def analyse_expressions(self, env):
...
-            function = "__Pyx_ListComp_AppendAndDecref" if steal_temp else "__Pyx_ListComp_Append"
+            if self.is_starred:
+                code.globalstate.use_utility_code(
+                    UtilityCode.load_cached("ListExtend", "Optimize.c"))
+                function = "__Pyx_PyList_Extend"
+            else:
+                steal_temp = self.expr.is_temp
+                code.globalstate.use_utility_code(
+                    UtilityCode.load_cached("ListCompAppendAndDecref" if steal_temp else "ListCompAppend", "Optimize.c"))
+                function = "__Pyx_ListComp_AppendAndDecref" if steal_temp else "__Pyx_ListComp_Append"
```
scoder: We really have three different cases here, let's make that clearer.
```suggestion
            if self.is_starred:
                utility_code_name = "ListExtend"
                function = "__Pyx_PyList_Extend"
            elif self.expr.is_temp:
                steal_temp = True
                utility_code_name = "ListCompAppendAndDecref"
                function = "__Pyx_ListComp_AppendAndDecref"
            else:
                utility_code_name = "ListCompAppend"
                function = "__Pyx_ListComp_Append"

            code.globalstate.use_utility_code(
                UtilityCode.load_cached(utility_code_name, "Optimize.c"))
```

Cython/Compiler/ExprNodes.py
```diff
@@ -9702,6 +9716,32 @@ def generate_function_definitions(self, env, code):
...
+
+class DictComprehensionUpdateNode(ComprehensionAppendNode):
+    """Update a dict comprehension result with each expression value."""
+
+    def generate_execution_code(self, code):
+        self.expr.generate_evaluation_code(code)
+
+        if not self.target.type.is_pydict_type:
+            raise InternalError(
+                "Invalid type for dict comprehension node: %s" % self.target.type)
```
scoder: An internal assertion is certainly enough here.
```suggestion
        assert self.target.type.is_pydict_type, self.target.type
```

Cython/Compiler/ExprNodes.py
```diff
@@ -9702,6 +9716,32 @@ def generate_function_definitions(self, env, code):
...
+
+        # PyDict_Update raises AttributeError for non-mappings, whereas **
+        # unpacking raises TypeError. Match Python's unpacking error here.
```
scoder: The comment reads badly when split like this.
```suggestion
        # PyDict_Update raises AttributeError for non-mappings, whereas "**dict" unpacking
        # raises TypeError. Match Python's unpacking error here.
```

Cython/Compiler/Parsing.py
```diff
@@ -1495,29 +1494,32 @@ def p_dict_or_set_maker(s: PyrexScanner):
...
-                s.error("iterable unpacking cannot be used in comprehension")
-            else:
-                # e.g. "{1,2,3 for ..."
-                s.expect('}')
-            return ExprNodes.DictNode(pos, key_value_pairs=[])
+
+    if s.sy in ('for', 'async'):
+        # syntax error, e.g. "{1, 2, 3 for ...}"
+        s.expect('}')
+        return ExprNodes.DictNode(pos, key_value_pairs=[])
```
scoder: Why is this needed in addition? Isn't the `s.expect('}')` call below enough ? That's what I intended the `and len(parts) == 1` condition above to make it run into.

I actually wonder if it wouldn't be better to refactor the parsing part inside of the braces into an initial part of the first expression and then two different continuations, depending on whether we find `,` or `for` / `async`. Splitting this into three functions might simplify and clarify this a lot.

Cython/Compiler/Parsing.py
```suggestion
            # merged set/dict literal or unpacking comprehension
```

Cython/Compiler/Parsing.py
```diff
@@ -1531,7 +1533,7 @@ def p_dict_or_set_maker(s: PyrexScanner):
                 if set_items:
                     items.append(ExprNodes.SetNode(set_items[0].pos, args=set_items))
                     set_items = []
-                items.append(part)
+                items.append(part.target if part.is_starred else part)
```
scoder, line 1531: Where does this come from? Doesn't it lose the information whether we are unpacking or not?

tests/run/comprehension_unpacking_T7898.pyx
```diff
@@ -0,0 +1,141 @@
+# mode: run
+# tag: comprehension
```
scoder: You could rename this file to a plain `.py` file and let it run in regular Py3.15 for comparison.
```suggestion
# mode: run
# tag: comprehension, pure3.15
```

fzlzjerry:
Addressed both parser comments in 473f4b37a. I split the first-item parse from the comprehension and literal continuations. The literal path now carries an explicit unpacking flag and only unwraps `*item` when building a `MergedSequenceNode`; `**item` is passed as the mapping to `MergedDictNode`. This also removes the extra `for`/`async` branch, so the final `expect("}")` reports multi-item comprehensions.

I rebuilt Cython in place and reran the comprehension, PEP 448 literal, Limited API, parser unit, and style checks. All pass, and Read the Docs is green as well.

scoder:
I'm sorry, @fzlzjerry, I ended up rewriting the parser part myself. I had a rather clear idea of how to do it, and it was less work to do it than to explain what to do. Rewriting and cleaning up code is a critical part of my work as a maintainer, and when I marked the ticket as "good first issue", I didn't know yet that it would be necessary.

I pushed a PR to https://github.com/cython/cython/pull/7900

Thank you for your contribution and your help in investigating the feature implementation. I'll make sure to give due credit.

fzlzjerry:
Thanks for carrying this forward and for preserving the test contribution and credit. #7900 now covers the full list, set, and dict comprehension scope, so I'm closing this one to avoid duplicate review.

PR #7909, opened 2026-08-22
Use decref-after-update for C attribute setting.

### LINKED ISSUE #7907: [BUG] Race condition writing to exttype attributes
### Describe the bug

Writing to object attributes of extension types uses decref before assignment, leaving the attribute value in an invalid state for a split second:
```c
  __Pyx_GOTREF((PyObject *)__pyx_v_obj->attr);
  __Pyx_DECREF((PyObject *)__pyx_v_obj->attr);  // <- dead reference !
  __pyx_v_obj->attr = __pyx_v_value;
```

```cython
cdef class Value:
    pass

cdef class Ext:
    cdef Value attr

def update():
    x = Ext()
    value = Value()
    x.attr = value  # <-
```

### Expected behaviour

We should use `ExprNode.generate_decref_set()` to generate "read-update-decref" ordering. That's still not completely safe (race condition between read and update) but avoids leaving the potentially dead reference in the attribute before updating it. Thus, concurrent updates would stay unsafe, but an update with only concurrent reads would become safe, which brings it to a reasonable and cheap level of safety that is user controllable.

### OS

_No response_

### Python version

_No response_

### Cython version

master

### Additional context

_No response_

scoder:
Avoids a race condition in which the attribute still holds an invalidated object reference.

Closes https://github.com/cython/cython/issues/7907

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -8504,13 +8504,12 @@ class AttributeNode(ExprNode):
                 rhs.make_owned_reference(code)
                 rhs.generate_giveref(code)
                 code.put_gotref(select_code, self.type)
-                code.put_decref(select_code, self.ctype())
+                self.generate_decref_set(code, rhs.result_as(self.ctype()))
             elif self.type.is_memoryviewslice:
                 from . import MemoryView
                 MemoryView.put_assign_to_memviewslice(
                         select_code, rhs, rhs.result(), self.type, code)
-
-            if not self.type.is_memoryviewslice:
+            else:
                 code.putln(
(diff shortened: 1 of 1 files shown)


scoder:
One difference is that we no longer use `std::move()` for object references. Is that an issue?
It wouldn't be difficult to implement (I guess), I'm just not sure if it's worth the complication.

da-woods:
> One difference is that we no longer use `std::move()` for object references. Is that an issue?
> It wouldn't be difficult to implement (I guess), I'm just not sure if it's worth the complication.

`std::move` does nothing for a pointer. It's harmless but completely unnecessary

PR #7911, opened 2026-08-23
Avoid MSVC C4551 on unused-helper "touch" statements

aytey:
## Problem

Cython marks conditionally-unused static helper functions as "used" (to silence `-Wunused-function`) by referencing them in a discarded expression. Several sites do this with a **bare function name**, e.g. in the `__Pyx_PyType_Ready` helper that every extension type pulls in:

```c
(void)__Pyx_PyObject_CallMethod0;
```

MSVC reads a bare function name as a call that is missing its argument list and emits **warning C4551** (`function call missing argument list`). Because `__Pyx_PyType_Ready` is generated for every `cdef class`, this shows up in essentially any MSVC build.

## Fix

Take the function's **address** instead:

```c
(void)&__Pyx_PyObject_CallMethod0;
```

`&func` is unambiguous — it cannot be parsed as a call — so C4551 does not fire. It is valid C and C++ on every compiler, still marks the symbol as used, and matches the `(void)&func` idiom already used elsewhere in `CythonFunction.c`.

All bare-function `(void)func;` "touch" sites were converted:

| File | Helper(s) |
|---|---|
| `ExtensionTypes.c` | `__Pyx_PyObject_CallMethod0` (×2), `__Pyx_validate_bases_tuple` |
| `MatchCase.c` | `__Pyx_MatchCase_OtherSequenceSliceToList`, `__Pyx_MatchCase_TupleSliceToList` |
| `ObjectHandling.c` | `__Pyx_GetBuiltinName` |
| `Optimize.c` | `__Pyx_PyObject_CallOneArg` |
| `TString.c` | `__Pyx_GetObjectFromTemplateLib` |

Each identifier was checked to be addressable (plain functions, or the one object-like macro aliasing a function); none is a function-like macro, so `&` is safe at every site.

## Test

Adds `tests/run/unused_utility_func_no_c4551.pyx` (`mode: compile`). A `cdef class` pulls in `__Pyx_PyType_Ready` → `__Pyx_PyObject_CallMethod0`, and the test uses the existing generated-C directives to assert the safe form is emitted and the bare form never appears:

```
# cython: test_assert_c_code_has = \(void\)&__Pyx_PyObject_CallMethod0;
# cython: test_fail_if_c_code_has = \(void\)__Pyx_PyObject_CallMethod0;
```

## Verification

- Regenerated C for a `cdef class`: 2× `(void)&__Pyx_PyObject_CallMethod0;`, 0 bare occurrences.
- Compiled the generated C with `clang -Wall -Wextra -Wunused-function`: exit 0, zero warnings — confirming the address-of form is valid C and still suppresses the unused-function warning it was there for.

### DIFF
--- a/Cython/Utility/ExtensionTypes.c
+++ b/Cython/Utility/ExtensionTypes.c
@@ -137,9 +137,10 @@ static int __Pyx_PyType_Ready(PyTypeObject *t) {
 
 #if CYTHON_USE_TYPE_SPECS || !CYTHON_COMPILING_IN_CPYTHON
     // avoid C warning about unused helper function
-    (void)__Pyx_PyObject_CallMethod0;
+    // (take the address so MSVC does not warn C4551 about a missing call)
+    (void)&__Pyx_PyObject_CallMethod0;
 #if CYTHON_USE_TYPE_SPECS
-    (void)__Pyx_validate_bases_tuple;
+    (void)&__Pyx_validate_bases_tuple;
 #endif
 
     return PyType_Ready(t);
@@ -163,7 +164,7 @@ static int __Pyx_PyType_Ready(PyTypeObject *t) {
     #if PY_VERSION_HEX >= 0x030A00b1
         // finally added in Py3.10 :)
         gc_was_enabled = PyGC_Disable();
-        (void)__Pyx_PyObject_CallMethod0;
+        (void)&__Pyx_PyObject_CallMethod0;
 
     #else
         // Call gc.disable() as a backwards compatible fallback, but only if needed.
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -2048,7 +2048,7 @@ static PyType_Spec __pyx_FusedFunctionType_spec = {
 
 static int __pyx_FusedFunction_init(PyObject *module) {
     // Silence unused warning when not in shared module.
-    (void)__Pyx_Get_FusedFunction_Type;
+    (void)&__Pyx_Get_FusedFunction_Type;
 
     $modulestatetype_cname *mstate = __Pyx_PyModule_GetState(module);
     PyObject *bases = PyTuple_Pack(1, mstate->__pyx_CyFunctionType);
--- a/Cython/Utility/FunctionArguments.c
+++ b/Cython/Utility/FunctionArguments.c
@@ -525,7 +525,7 @@ static void __Pyx_RejectUnknownKeyword(
     #if CYTHON_AVOID_BORROWED_REFS
     while (__Pyx_PyDict_NextRef(kwds, &pos, &key, NULL))
     #else
-    (void) __Pyx_PyDict_NextRef;
+    (void) &__Pyx_PyDict_NextRef;
     while (PyDict_Next(kwds, &pos, &key, NULL))
     #endif
     {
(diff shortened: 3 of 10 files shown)


da-woods:
I'm not really convinced by the test (in that we don't _really_ care what the code looks like here, just that it avoids MSVC warnings). Given that it's failing in the shared-utility build I'd propose deleting it rather than fixing it.

The change itself looks fine (personally I think this is a common pattern that MSVC should accept, but we also can't change MSVC so...)

da-woods:
> I'm not really convinced by the test (in that we don't _really_ care what the code looks like here, just that it avoids MSVC warnings). Given that it's failing in the shared-utility build I'd propose deleting it rather than fixing it.

Or if we did want a test, we'd probably want a C lint of some sort to check that we don't add this pattern elsewhere rather than a check for the specific case. But that probably isn't worthwhile given we don't have anything similar right now.

da-woods:
I can see:

```
Cython/Utility/ModuleSetupCode.c:    (void)__pyx__find_code_object;
Cython/Utility/ModuleSetupCode.c:    (void)__pyx__insert_code_object;
```

which you've missed

aytey:
Thanks for the review!

- Dropped the test
- I've fixed `__pyx__find_code_object` / `__pyx__insert_code_object` and:
  - `FunctionArguments.c`: `__Pyx_PyDict_NextRef` (×2)
  - `StringTools.c`: `__Pyx_PyObject_CallMethod1`
  - `Profile.c`: the `__Pyx_TraceReturnCValue` macro, whose `convert_function` argument is always a bare `to_py_function` name

  Left the genuine unused-*variable* touches (`(void)ptr;`, `(void)wraparound;`, `(void)__pyx_exception_already_reported;`, etc.) alone, since C4551 only applies to functions.

Force-pushed with all of the above squashed into the one commit.

da-woods:
Thanks - I'll merge this.

I think it'll just put it in the next 3.3.1 release. It'd be fine to backport, but also doesn't seem like it's hugely important to.

aytey:
> I think it'll just put it in the next 3.3.1 release.

3.3.1 is fine; do you know when that's due?

da-woods:
Depends how broken 3.3.0 turns out to be :)? If we get lots of important-looking bug reports then it could be in the next week. If we don't it might be ~1 months.

aytey:
> Depends how broken 3.3.0 turns out to be :)? If we get lots of important-looking bug reports then it could be in the next week. If we don't it might be ~1 months.

Thanks, all good

Cython/Utility/ExtensionTypes.c
```diff
@@ -137,9 +137,10 @@ static int __Pyx_PyType_Ready(PyTypeObject *t) {
 
 #if CYTHON_USE_TYPE_SPECS || !CYTHON_COMPILING_IN_CPYTHON
     // avoid C warning about unused helper function
-    (void)__Pyx_PyObject_CallMethod0;
+    // (take the address so MSVC does not warn C4551 about a missing call)
```
scoder, line 140: These comments are in some places but not all, so they serve little purpose but make the `void` lines look more important than they are. Either readers already know the pattern, or they wouldn't even notice the difference, or they wouldn't care, or if they do care and don't know it (which is the only case in which the comment would help), then they'd probably still have to look it up because they are more likely to hit a spot without comment than one with comment. I'll remove them.

aytey: Do you want me to do it, or do you want to do it?

scoder: Feel free to send a PR.

aytey: Done: https://github.com/cython/cython/pull/7913

PR #7912, opened 2026-08-23
Disable annotation for shared module

### LINKED ISSUE #7910: [BUG] cython >= 3.2.6 shared utility setuptools annotate option build regression
### Describe the bug

My project had a build system that worked fine on cython 3.2.5 and earlier, based on [Compiling shared module using setuptools](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiling-shared-module-using-setuptools).

However following 3.2.6 (possibly #7723) the build fails **on the first try** with the following error and then **succeeds on the second**:

```
    with Utils.open_source_file(source_filename) as f:
         ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "Cython/Utils.py", line 418, in Cython.Utils.open_source_file
FileNotFoundError: build/build/_cyutility.pyx
```

To get to "first time" again, I just rm all files in build directory.

```bash
```

Build system **setup.py** looks roughly like this, though tried changing the order and presense and location of pretty much everything without finding a workaround (except for the obvious one: run the build twice).

```python
Options.annotate = True
sharedCythonModule = True

#...

if sharedCythonModule:
    extensions = [Extension("_cyutility",
                            extra_link_args=extra_link_args,
                            sources=["_cyutility.c"])]
else:
    extensions = []

for directoryx in ["*"]:
    extensions.append(
        Extension(directoryx, [directoryx + ".pyx"],
                  extra_compile_args=extra_compile_args,
                  extra_link_args=extra_link_args,
                  libraries=libraries,
                  include_dirs=include_dirs,
                  runtime_library_dirs=library_dirs,
                  library_dirs=library_dirs,
                  define_macros=[('CYTHON_TRACE', str(profile))],
              )
        )

if __name__ == '__main__':
    args = {'nthreads': (cpu_count()+1)//2,
            'build_dir': 'build',
            'language_level': 3}
    if sharedCythonModule:
        args['shared_utility_qualified_name'] = '_cyutility'

    setup(
        ext_modules = cythonize(extensions, **args)
    )
```

### Expected behaviour

Should build without error first time like on 3.2.5

### OS

Linux

### Python version

3.13

### Cython version

3.3.0

### Additional context

_No response_

CypherGrue:
Since Cython 3.2.6 #7723, shared module build uses a virtual pyx file. Generating annotations for this file caused builds to fail because the annotation code requires the physical file to exist.

This change disables annotations during the shared module build, mirroring the existing option set and restore pattern already used for cimport_from_pyx. Fixes #7910.

### DIFF
--- a/Cython/Build/SharedModule.py
+++ b/Cython/Build/SharedModule.py
@@ -119,7 +119,7 @@ def create_shared_library_pipeline(context, scope, options, result, selected_fea
             Options.cimport_from_pyx = cimport_from_pyx
             return node
         return inner
-        
+
     orig_annotate = Options.annotate
 
     def set_annotate(annotate):
@@ -127,7 +127,7 @@ def create_shared_library_pipeline(context, scope, options, result, selected_fea
             Options.annotate = annotate
             return node
         return inner
-        
+
     return [
         # "cimport_from_pyx=True" to force generating __Pyx_ExportFunction
         set_cimport_from_pyx(True),
--- a/Cython/Build/SharedModule.py
+++ b/Cython/Build/SharedModule.py
@@ -119,10 +119,19 @@ def create_shared_library_pipeline(context, scope, options, result, selected_fea
             Options.cimport_from_pyx = cimport_from_pyx
             return node
         return inner
+        
+    orig_annotate = Options.annotate
 
+    def set_annotate(annotate):
+        def inner(node):
+            Options.annotate = annotate
+            return node
+        return inner
+        
     return [
         # "cimport_from_pyx=True" to force generating __Pyx_ExportFunction
         set_cimport_from_pyx(True),
+        set_annotate(False),
         generate_tree_factory(context),
         *Pipeline.create_pipeline(context, 'pyx', exclude_classes=()),
         generate_c_utilities,
@@ -132,6 +141,7 @@ def create_shared_library_pipeline(context, scope, options, result, selected_fea
         Pipeline.abort_on_errors,
         Pipeline.generate_pyx_code_stage_factory(options, result),
         set_cimport_from_pyx(orig_cimport_from_pyx),
+        set_annotate(orig_annotate),
     ]
 
 


scoder:
Thanks for figuring this out. This change might be ok for a quick-fix, but looking at the larger picture, the way the options are currently set for the shared module is a hack. There are several places in the code base that say `if Options.annotate or options.annotate`, which is a clear indication that there shouldn't be two ways to do it. We should look at the global config module's `Options.annotate` only once to set the run specific `options.annotate`, and then only use the latter. That would make it easy for the shared module to overwrite the run specific option.

CypherGrue:
I don't have enough kilowatts for the larger picture :man_shrugging:. A separate question is how did the intended-virtual pyx file somehow get touched on the filesystem for the second build pass to succeed. There is still a file name virtual-to-physical leak somewhere even after my changes.

matusvalo:
Hmm should not better place for this kind of test is here?

https://github.com/cython/cython/blob/3292aa6703019891897b8d979582447b5e21ecee/Cython/Compiler/ModuleNode.py#L677

The line can be like:

```python
if (Options.annotate or options.annotate) and not self.scope.context.shared_c_file_path: 
```

Also,
1. I suppose we will need test proving the fix too (see https://github.com/cython/cython/blob/master/tests/run/shared_utility_module.srctree for an example).
2. we could add property `in_shared_utility_module` (maybe also `using_shared_utility_module` too in one refactor scoop) to `Cython.Main.Context` and use it instead and use it also in other places e.g. here: 
   https://github.com/cython/cython/blob/3292aa6703019891897b8d979582447b5e21ecee/Cython/Compiler/ModuleNode.py#L132-L133

But wait for @scoder feedback of my comments before any changes.

Related to comment about duplicated `options... or Options...` I think it needs separated issue and fix.

scoder:
> Hmm should not better place for this kind of test is here?

Oh, right, that looks like a better place to fix this.

> 1. I suppose we will need test proving the fix too (see https://github.com/cython/cython/blob/master/tests/run/shared_utility_module.srctree for an example).

This can be tested as part of that test file by replacing `CYTHON ` with `CYTHON -a ` in the commands near the top of the file.

> needs separated issue and fix

Would be better, yes. Let's keep this fix PR short.

PR #7913, opened 2026-08-23
Remove inconsistent C4551 explanatory comments

aytey:
Follow-up to #7911, actioning @scoder's review comment ([r3838819988](https://github.com/cython/cython/pull/7911#discussion_r3838819988)):

> These comments are in some places but not all, so they serve little purpose but make the `void` lines look more important than they are. Either readers already know the pattern, or they wouldn't even notice the difference, or they wouldn't care, or if they do care and don't know it (which is the only case in which the comment would help), then they'd probably still have to look it up because they are more likely to hit a spot without comment than one with comment. I'll remove them.

Removes the C4551 explanatory comments from the three sites where they were added (`ExtensionTypes.c`, `MatchCase.c`, `ObjectHandling.c`). The `(void)&func;` fix itself is unchanged.


scoder:
Thanks

PR #7916, opened 2026-08-23
Use Limited API version in __PYX_ABI_MODULE_NAME

### LINKED ISSUE #7914: [BUG] 3.3.0 Limited API: shared `cython_function_or_method` type conflicts between modules targeting abi3 < 3.11 and 3.11
### Describe the bug

**note**: The issue report is generated by Claude and reviewed with limited understanding from my side.

Importing two Cython-3.3.0-compiled Limited API modules into one process fails when one targets `Py_LIMITED_API` < `0x030B0000` and the other targets `0x030B0000` (3.11):

```
TypeError: Shared Cython type cython_function_or_method has the wrong size, try recompiling
```

Regression from 3.2.9 (works fine there). Any process mixing abi3 wheels with different targets is affected; first seen in numpy's CI (`test_limited_api`, which builds abi3 3.9–3.15 variants — exactly the 3.11 variants fail on Linux/macOS/Windows).

### Cause

3.3.0 made the `__pyx_CyFunctionObject` layout depend on the Limited API target (`CythonFunction.c`, presumably via the #7519 layout rework):

```c
#if __PYX_LIMITED_VERSION_HEX < 0x030B0000
    // Backwards compatible way of implementing PyGen_GetCode? ...
    PyObject *func_is_coroutine;
#endif
```

while `func_vectorcall` exists only for targets ≥ 3.12 (`CYTHON_VECTORCALL`).

The shared-type registry is the dict of the module named `__PYX_ABI_MODULE_NAME` = `"_cython_" CYTHON_ABI`, and `CYTHON_ABI`'s `__PYX_LIMITED_ABI_SUFFIX` distinguishes vectorcall builds but nothing distinguishes the `< 0x030B0000` layout. So targets 3.9/3.10 and 3.11 share one registry entry with different `basicsize`, and `__Pyx_VerifyCachedType` rejects whichever size registers second. Observed (CPython 3.14.2, Linux, gcc):

| Py_LIMITED_API target | ABI module | CyFunction basicsize |
|---|---|---|
| 0x030A0000 | `_cython_3_3_0limitednofinalize` | 160 |
| 0x030B0000 | `_cython_3_3_0limitednofinalize` | **152** |
| 0x030C0000 | `_cython_3_3_0limited_vectorcallnofinalize` | 160 (no conflict) |

Possible fixes: add an ABI suffix for the `< 0x030B0000` layout split, or make the layout uniform across Limited API targets.

@da-woods 

```bash
printf 'def f():\n    return 1\n' > mod_a.pyx
cp mod_a.pyx mod_b.pyx
INC=$(python3 -c "import sysconfig; print(sysconfig.get_path('include'))")
cython -3 mod_a.pyx && gcc -shared -fPIC -DCYTHON_LIMITED_API=1 -DPy_LIMITED_API=0x030A0000 -I$INC mod_a.c -o mod_a.so
cython -3 mod_b.pyx && gcc -shared -fPIC -DCYTHON_LIMITED_API=1 -DPy_LIMITED_API=0x030B0000 -I$INC mod_b.c -o mod_b.so
python3 -c "import mod_a, mod_b"
```

Output with 3.3.0:

```
TypeError: Shared Cython type cython_function_or_method has the wrong size, try recompiling
```

With 3.2.9 both import fine (basicsize 152 for both targets).

### Expected behaviour

Limited API modules built with the same Cython version but different `Py_LIMITED_API` targets can be imported into the same process (as in ≤ 3.2.9), or are segregated into different shared ABI modules like the vectorcall builds already are.

### OS

Linux

### Python version

3.12

### Cython version

3.3

### Additional context

_No response_

da-woods:
since there are subtle differences and it's difficult to keep track of them all and remember to update it when things change.

Fixes #7914. The cost is that Cython won't identiy cyfunctions compiled with a different Limited API versions as cyfunctions. Which mostly means it won't do any cyfunc-specific optimizations.

### DIFF
--- a/Cython/Utility/ModuleSetupCode.c
+++ b/Cython/Utility/ModuleSetupCode.c
@@ -1478,12 +1478,33 @@ static int __Pyx_init_tpflags_variables(void) {
     // The limited API makes some significant changes to data structures, so we don't
     // want to share the implementations compiled with and without the limited API.
     #if CYTHON_VECTORCALL
-        #define __PYX_VECTORCALL_ABI_SUFFIX  "_vectorcall"
+        #define __PYX_VECTORCALL_ABI_SUFFIX  "vectorcall"
     #else
         #define __PYX_VECTORCALL_ABI_SUFFIX
     #endif
 
-    #define __PYX_LIMITED_ABI_SUFFIX "limited" __PYX_VECTORCALL_ABI_SUFFIX __PYX_AM_SEND_ABI_SUFFIX
+    #if __PYX_LIMITED_VERSION_HEX >= 0x03100000
+        #define __PYX_LIMITED_ABI_V_SUFFIX "16"
+    #elif __PYX_LIMITED_VERSION_HEX >= 0x030F0000
+        #define __PYX_LIMITED_ABI_V_SUFFIX "15"
+    #elif __PYX_LIMITED_VERSION_HEX >= 0x030E0000
+        #define __PYX_LIMITED_ABI_V_SUFFIX "14"
+    #elif __PYX_LIMITED_VERSION_HEX >= 0x030d0000
+        #define __PYX_LIMITED_ABI_V_SUFFIX "13"
+    #elif __PYX_LIMITED_VERSION_HEX >= 0x030C0000
+        #define __PYX_LIMITED_ABI_V_SUFFIX "12"
+    #elif __PYX_LIMITED_VERSION_HEX >= 0x030B0000
+        #define __PYX_LIMITED_ABI_V_SUFFIX "11"
+    #elif __PYX_LIMITED_VERSION_HEX >= 0x030A0000
+        #define __PYX_LIMITED_ABI_V_SUFFIX "10"
+    #elif __PYX_LIMITED_VERSION_HEX >= 0x03090000
+        #define __PYX_LIMITED_ABI_V_SUFFIX "9"
+    #else
+        // 8 and below are unsupported anyway
+        #define __PYX_LIMITED_ABI_V_SUFFIX "8"
+    #endif
+
+    #define __PYX_LIMITED_ABI_SUFFIX "limited" __PYX_LIMITED_ABI_V_SUFFIX "_" __PYX_VECTORCALL_ABI_SUFFIX __PYX_AM_SEND_ABI_SUFFIX
 #else
     #define __PYX_LIMITED_ABI_SUFFIX
 #endif
--- a/tests/run/isolated_limited_api_tests.srctree
+++ b/tests/run/isolated_limited_api_tests.srctree
@@ -57,7 +57,6 @@ for version in range(9, sys.version_info[1]+1):
 
     all_extensions.extend(
         cythonize(extensions, build_dir=build_dir))
-    all_extensions.extend(extensions)
 
 setup(
     ext_modules = all_extensions,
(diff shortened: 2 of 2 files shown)


tests/run/isolated_limited_api_tests.srctree
```diff
@@ -41,20 +42,22 @@ all_extensions = []
 
 for version in range(9, sys.version_info[1]+1):
     version_hex = f"0x03{version:02x}0000"
+    build_dir = cwd / f"py3{version}"
     extensions = []
 
     for module in modules:
+        shutil.copyfile(f"{module}.pyx", f"{module}_{version}.pyx")
```
da-woods, line 49: It looks like `cythonize` is spotting the same file name and caching the `define_macros` somewhere (so they were all being run with `Py_LIMITED_API` set to `0x03090000` which isn't the intent of the test). I'll pick this bit back to 3.2 to fix the test there.

Cython/Utility/ModuleSetupCode.c
```diff
@@ -1478,12 +1478,33 @@ static int __Pyx_init_tpflags_variables(void) {
...
+    #elif __PYX_LIMITED_VERSION_HEX >= 0x030B0000
+        #define __PYX_LIMITED_ABI_V_SUFFIX "11"
+    #elif __PYX_LIMITED_VERSION_HEX >= 0x030A0000
+        #define __PYX_LIMITED_ABI_V_SUFFIX "10"
+    #elif __PYX_LIMITED_VERSION_HEX >= 0x03090000
+        #define __PYX_LIMITED_ABI_V_SUFFIX "9"
+    #else
+        // 8 and below are unsupported anyway
+        #define __PYX_LIMITED_ABI_V_SUFFIX "8"
+    #endif
```
scoder, line 1504: I'm sure there is a level of indirection that could solve this, but it seems a truly hard problem to make the C preprocessor do any form of integer arithmetic. There is a Boost implementation of basic math operations that takes up several header files, and that's probably what it takes.

This is fine for what it does, and IIUC you added a test that intentionally fails in Py3.17 if (or rather when) we forget to update this.

scoder (approved):

scoder:
> cost is that Cython won't identiy cyfunctions compiled with a different Limited API versions as cyfunctions

… and async functions etc. But performance is the price that users have to pay for targeting the Stable ABI. And since the usual goal of that is to reduce the number of Python versions to build for, I don't expect that many different API versions to be targeted (3.9/11/12/15 seem reasonable), so the problem will be mitigated by practice.

da-woods:
> and IIUC you added a test that intentionally fails in Py3.17 if (or rather when) we forget to update this.

Yes, that's a good part of the intent of the test.

>  I don't expect that many different API versions to be targeted (3.9/11/12/15 seem reasonable), so the problem will be mitigated by practice.

I think you'd mainly hit it if you were mixing Cython code from different independent libraries. Which is probably relatively rare.

Will fix the Windows issue when I can. Slightly unfortunately, it seems to work locally for me (at least with a different Python version)

ngoldbaum:
I'm going to merge https://github.com/numpy/numpy/pull/32393 which pins NumPy's CI to require `Cython<3.3`. I'll remove that once there's a bugfix release out. If one of you could send in a PR to remove the pin when the release happens, I'd appreciate that.

ngoldbaum:
Actually never mind, I think I can do it with a test skip instead: https://github.com/numpy/numpy/pull/32407

Upper version pins never lead to fun times...

PR #7917, opened 2026-08-23
Ensure isolated limited test used correct Py_LIMITED_API

da-woods:
Cython was spotting the same file name, and caching the `define_macros` so it wasn't actually compiling with different limited API versions as intended.


da-woods:
PR is into 3.2.x

PR #7918, opened 2026-08-24
Move PyMethodDef structs into module state

scoder:
Allows moving docstrings and function names into the compressed string block and reusing them as unicode strings.

Maintains a bit of memory redundancy because the wrapper function of special methods still requires the UTF-8 encoded C string, and the docstring must be defined on the original function to correctly inherit it in subtypes.

Only applies to normal functions. Excludes extension type methods because they would require moving the whole type definition as well. It could be applied to our heap type declarations by making a part of the module struct conditional on that feature.


Cython/Utility/CythonFunction.c
```diff
@@ -1030,7 +1030,7 @@ static PyObject *__Pyx_CyFunction_Init(PyObject *op_in,
     op->func_name = NULL;
     Py_INCREF(qualname);
     op->func_qualname = qualname;
-    op->func_doc = NULL;
+    op->func_doc = __Pyx_XNewRef(docstring);
```
da-woods, line 1033: I think that lets you simplify the to `__doc__` attribute? I.e. it can probably just be an object memberdef ?

da-woods (commented):
I was wondering if module state was actually the best place for this? On the basis that the data wants to be initialized dynamically, but there's no real need to have a separate copy for every sub-interpreter for example since it's all constant data once initialized.
With that said, if you're initializing the docstrings of cyfunctions that may end up forcing the issue since those are PyObjects and so probably have to live in the module state.

PR #7923, opened 2026-08-24
Unpack coercion nodes in ExprStatNode.

### LINKED ISSUE #7922: [BUG] 3.3.0 regression: discarded `dict[str, int].pop(key, default)` generates invalid C (`&((PyObject *)tmp)`)
### Describe the bug

Since 3.3.0b1, calling `.pop(key, default)` on a dict annotated with `int` values and **discarding the result** generates C that no compiler accepts:

```c
if (__Pyx_PyInt_FromNumber(&((PyObject *)__pyx_t_4), NULL, 1) < (0)) __PYX_ERR(0, 3, __pyx_L1_error)
```

The address-of operator is applied to a cast expression, which is not an lvalue.

```python
# repro.py
def f(counts: dict[str, int], keys: list[str]) -> None:
    for k in keys:
        counts.pop(k, None)
```

```console
$ pip install cython==3.3.0 setuptools
$ cythonize -3 -i repro.py
repro.c: In function '__pyx_pf_5repro_f':
repro.c:2741:32: error: lvalue required as unary '&' operand
error: Command '['gcc', ..., '-c', 'repro.c', ...]' returned non-zero exit status 1.
```

No extension module is produced; `cythonize` exits 1.

### Expected behaviour

Valid C, as produced by 3.2.9 for the same input.

### Environment

| | |
|---|---|
| OS | Debian 13 (`python:3.12-slim` container) and macOS 15 (arm64) |
| Python | 3.12 |
| Compilers | gcc 14.2.0 and clang 17 — both reject the generated C |
| Cython | **3.2.9 — OK**, **3.3.0b1 — fails**, **3.3.0 — fails** |

Control run in the same container: same file, same command, `cython==3.2.9` → exit 0, extension module produced.

### Notes that may help

The failure is narrow — all three conditions are required:

1. the dict is annotated with `int` values (`dict[str, str]` is fine),
2. `pop` is called with a default (`pop(k, None)`),
3. **the result is discarded** — assigning it (`x = counts.pop(k, None)`) produces valid C.

Dropping the annotation also produces valid C, which suggests the int-typed optimisation path is involved. The generated helper is `__Pyx_PyInt_FromNumber`, which appears to come from the "Skip Py -> C -> Py Coercion" work (#7735) — possibly an edge case where the destination is a temporary rather than a variable.

We hit this in a production build: the compile failure was non-fatal in our Dockerfile, so the image shipped without the module. Not your problem, but it is why the report is a bit detailed — the failure is easy to miss in CI.

scoder:
Their result is never actually used for anything.

Closes https://github.com/cython/cython/issues/7922

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -1033,6 +1033,9 @@ class ExprNode(Node):
 
         if self.check_for_coercion_error(dst_type, env):
             return self
+        if not self.result_is_used:
+            # No need to coerce a result that will never be looked at.
+            return self
 
         used_as_reference = dst_type.is_reference
         if used_as_reference and not src_type.is_reference:
@@ -6308,9 +6311,6 @@ class CallNode(ExprNode):
     def coerce_to_result_type(self, env, function, func_type=None):
         # Default to 'object' and then try to find a better type.
         self.type = py_object_type
-        if not self.result_is_used:
-            # Any coercion of unused values would be useless.
-            return self
         if func_type is None:
             func_type = function.type
         if function.is_name:
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -6308,6 +6308,9 @@ class CallNode(ExprNode):
     def coerce_to_result_type(self, env, function, func_type=None):
         # Default to 'object' and then try to find a better type.
         self.type = py_object_type
+        if not self.result_is_used:
+            # Any coercion of unused values would be useless.
+            return self
         if func_type is None:
             func_type = function.type
         if function.is_name:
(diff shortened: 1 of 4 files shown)


matusvalo (approved):
LGTM

scoder:
I was still a bit unhappy with this change because it feels like it solves the issue only by side-effect.

The underlying problem is that we replace `dict.pop()` (which returns a Python object) with a special function that just returns a C `int` for success/failure, but we don't look at what's done with that value afterwards. By the time it's replaced, we already added the type check to it, which remains there after the replacement. If we don't need the result, we shouldn't add the type check in the first case, otherwise, the replacement later wouldn't be safe.

I pushed a somewhat redundant second change that avoids the type coercion where it's added, when wrapping the call node. I think both changes are reasonable in their own right.

scoder:
I went with only avoiding the coercion. Better than trying to clean it up after the fact and risking to miss a few places.

PR #7926, opened 2026-08-24
Handle destruction of opaque cyfuncs safely during cleanup

da-woods:
With module state, it's possible that the utility code types traverse/clear/dealloc may be called during destructor shutdown when the module state is no longer findable. This is especially an issue for the single-phase init mode, but I think is a general problem.

The usual approach we've used in similar situations for heap-types is to bail out and not complete the destruction. For cyfunctions I think we understand the type hierarchy well enough make it work but this may not generally be possible for all types.

### DIFF
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -23,12 +23,6 @@ if (likely(__pyx_CyFunction_init($module_cname) == 0)); else
 
 #define __Pyx_CyFunction_USED
 
-#if CYTHON_OPAQUE_SHARED_TYPES
-#define __Pyx_as_CyFunctionObject(o) ((__pyx_CyFunctionObject *)PyObject_GetTypeData((o), CGLOBAL(__pyx_CyFunctionType)))
-#else
-#define __Pyx_as_CyFunctionObject(o) ((__pyx_CyFunctionObject *)o)
-#endif
-
 #define __Pyx_CYFUNCTION_STATICMETHOD  0x01
 #define __Pyx_CYFUNCTION_CLASSMETHOD   0x02
 #define __Pyx_CYFUNCTION_CCLASS        0x04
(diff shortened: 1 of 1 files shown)


da-woods:
This is causing the sanitizers crash in https://github.com/cython/cython/pull/7916 (but think it's worth dealing with separately)

Cython/Utility/CythonFunction.c
```diff
@@ -134,6 +141,20 @@ static int __pyx_CyFunction_init(PyObject *module);
...
+    while (1) {
+        PyTypeObject *base = (PyTypeObject *)PyType_GetSlot(cyfunctype, Py_tp_base);
+        if (base == &PyBaseObject_Type || base == NULL) {
+            break;
+        }
+        cyfunctype = base;
+    }
+    return (__pyx_CyFunctionObject *)PyObject_GetTypeData((o), cyfunctype);
+}
+#endif
```
scoder, line 156: This is really an internal function/macro that doesn't need to be declared globally. By declaring it here, you can roll back the macro move/change above.
```suggestion
#else
#define __Pyx_as_CyFunctionObject_safe __Pyx_as_CyFunctionObject
#endif
```

Cython/Utility/CythonFunction.c
```diff
@@ -134,6 +141,20 @@ static int __pyx_CyFunction_init(PyObject *module);
 //////////////////// CythonFunctionPerModule /////////////////////
 //@requires: ObjectHandling.c::CachedMethodType
 
+#if CYTHON_OPAQUE_SHARED_TYPES && CYTHON_USE_MODULE_STATE
+static __pyx_CyFunctionObject *__Pyx_as_CyFunctionObject_safe(PyObject *o) {
```
scoder, line 145:
```suggestion
// Slower version of "__Pyx_as_CyFunctionObject" that's safe to use in destructors
// (where the current module may have been lost).
static __pyx_CyFunctionObject *__Pyx_as_CyFunctionObject_safe(PyObject *o) {
```

scoder (commented):
Fine with me otherwise.

PR #7928, opened 2026-08-24
Remove opaque objects define in the 3.15 Limited API

da-woods:
This was written before the freethreading stable ABI was finalized. There's now a different macro that defines that stable ABI so it isn't needed for plain Limited API support.

We'll add the proper switch later when it's actually supported.


da-woods:
Related to the crash in https://github.com/cython/cython/pull/7916 (because it was using opaque objects in the 3.15 stable ABI and there was a crash there). Since it doesn't actually need to use them, it's better to turn them off.

PR #7938, opened 2026-08-26
Fix inherited tp_vectorcall and `__init__`

### LINKED ISSUE #7936: [BUG] `tp_new_vectorcall` calls non-existing base class `__pyx_tp_new_vectorcall_...`, producing module with undefined symbol
### Describe the bug

An extension type that inherits from another extension type in the same module can be given the vectorcall form of `tp_new` while its base class is generated with only the plain form, without the vectorcall. In this case the base class only has `__pyx_tp_new_<module>_<Base>` defined, but the derived type's generated `tp_new_vectorcall` calls `__pyx_tp_new_vectorcall_<module>_<Base>`.

The C compiler warns it as an implicit declaration, then warns about an `int`-to-pointer assignment, but the link of the shared object succeeds, and the failure surfaces at import:

```
ImportError: .../repro.cpython-314-x86_64-linux-gnu.so: undefined symbol: __pyx_tp_new_vectorcall_5repro_Base
```

**`repro.pyx`**:

```cython
cdef class Base:
    cdef object payload

    def __init__(self, *args, **kwargs):
        raise RuntimeError("This class cannot be instantiated directly.")


cdef class Leaf(Base):
    cdef object extra

    def __init__(self, x):
        self.extra = x
```

``` bash
... $ cythonize -i repro.pyx
Compiling .../repro.pyx because it changed.
[1/1] Cythonizing .../repro.pyx
.../repro.c: In function ‘__pyx_tp_new_vectorcall_5repro_Leaf’:
.../repro.c:4834:7: warning: implicit declaration of function ‘__pyx_tp_new_vectorcall_5repro_Base’; did you mean ‘__pyx_tp_new_vectorcall_5repro_Leaf’? [-Wimplicit-function-declaration]
 4834 |   o = __pyx_tp_new_vectorcall_5repro_Base(t,
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |       __pyx_tp_new_vectorcall_5repro_Leaf
.../repro.c:4834:5: warning: assignment to ‘PyObject *’ {aka ‘struct _object *’} from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
 4834 |   o = __pyx_tp_new_vectorcall_5repro_Base(t,
      |     ^
... $ ls *.so
repro.cpython-314-x86_64-linux-gnu.so
... $ python -c "import repro"
    import repro
ImportError: .../repro.cpython-314-x86_64-linux-gnu.so: undefined symbol: __pyx_tp_new_vectorcall_5repro_Base
... $ nm -u repro*.so | grep repro
                 U __pyx_tp_new_vectorcall_5repro_Base
```

Removing the attributes (`payload`, `extra` above) causes the call to an undefined function to occur in a `__pyx_tp_vectorcall_...` function rather than a `__pyx_tp_new_vectorcall_...` function.

### Expected behaviour

The derived class references a function the module defines, and import of the module succeeds.

### OS

Linux x86_64, gcc 13

### Python version

3.14

### Cython version

3.3.0

### Additional context

As a workaround, adding a `__cinit__` that can't be a vectorcall to the deriving class resolves this: 

``` cython
cdef class Leaf(Base):
    cdef object extra

    def __cinit__(self, *args, **kwargs):
        pass

    def __init__(self, x):
        self.extra = x
```

This compiles without warning and can be imported as normal, and `nm -u repro*.so | grep repro` reports nothing.

da-woods:
Fixes #7936

My original reasoning was to only look for the first `__init__` in the type hierarchy in that this will be the one that we're actually calling. However that fails because we won't have generated a vectorcall method for the base type if any of the previous `__init__` functions aren't suitable for `tp_vectorcall`.

### DIFF
--- a/Cython/Compiler/TypeSlots.py
+++ b/Cython/Compiler/TypeSlots.py
@@ -547,7 +547,10 @@ class TpVectorcallSlot(ConstructorSlot):
                 return "0"  # Can't reason about __init__
             # If we don't have the second overloaded alternative for __cinit__ or __init__
             # this means the signature was wrong and so we haven't generated code for the
-            # vectorcall wrappers in DefNode.analyse_declarations
+            # vectorcall wrappers in DefNode.analyse_declarations.
+            # This applies to all base classes (since we need to be able to call their
+            # vectorcall-signature __cinit__, and we won't have generated that function
+            # unless the base class met all the conditions itself).
             cinit_entry = tp.scope.lookup_here("__cinit__")
             if cinit_entry and not (cinit_entry.is_special and cinit_entry.tp_new_can_be_vectorcall):
                 return "0"
--- a/Cython/Compiler/TypeSlots.py
+++ b/Cython/Compiler/TypeSlots.py
@@ -535,7 +535,6 @@ class TpVectorcallSlot(ConstructorSlot):
 
     def slot_code(self, scope):
         tp = scope.parent_type
-        seen_init = False
         while tp:
             if not tp.is_extension_type:
                 return "0"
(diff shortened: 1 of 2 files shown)


scoder:
> we won't have generated a vectorcall method for the base type if any of the previous `__init__` functions aren't suitable for `tp_vectorcall`.

Something like this would make a good code comment to explain why we go through the hierarchy at all and don't stop at the first hit.

scoder:
Other than that, looks fine.

PR #7941, opened 2026-08-27
fix: add `CYTHON_UNUSED` to `__Pyx_pretend_to_initialize`

lucascolley:
While attempting to build SciPy with clang-23, I observed:

```
scipy/special/_ufuncs_cxx.cpython-312-aarch64-linux-gnu.so.p/_ufuncs_cxx.cpp:1488:13: error: unused function template '__Pyx_pretend_to_initialize' [-Werror,-Wunused-template]
 1488 | static void __Pyx_pretend_to_initialize(T* ptr) {
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~
```

After investigating a little with claude code, this seems like the solution.

Ref. https://github.com/scipy/scipy/pull/26041

### DIFF
--- a/Cython/Utility/ModuleSetupCode.c
+++ b/Cython/Utility/ModuleSetupCode.c
@@ -2478,7 +2478,7 @@ static void __Pyx_FastGilFuncInit(void) {
 #endif
 
 template <typename T>
-static void __Pyx_pretend_to_initialize(T* ptr) {
+CYTHON_UNUSED static void __Pyx_pretend_to_initialize(T* ptr) {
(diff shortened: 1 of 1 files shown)


scoder:
Thanks. I think we can entirely exclude the function if it's unused, though.

lucascolley:
thanks for the quick fix!

scoder:
3.2.x commit is https://github.com/cython/cython/commit/da31db56e54ca528c5bc5ee7f00b0fad636e7564

PR #7943, opened 2026-08-27
Improve source file encoding detection

### LINKED ISSUE #7935: False encoding-cookie detection allows comment-hidden code to execute in Cython binaries
## Summary

Cython searches the raw first two source lines for `coding: <word>` without requiring a PEP 263 comment. An ordinary Python string can therefore select a codec for the complete source file. With a stateful codec such as UTF-7, byte sequences that CPython treats as part of a comment can decode into a newline before Cython parses the source. The resulting Cython binary executes code that is inactive under CPython and visually appears inside a comment in the original file.

This is a source-to-binary semantic-injection primitive. It may enable malicious behavior to pass source review or CPython-based tests and become active only in a Cython build. The demonstration below is intentionally non-destructive: it imports a standard-library module, evaluates an arithmetic expression, and runs the local `whoami` command.

## Affected versions tested

- Cython 3.2.9
- Current Cython `master`, version 3.3.0, commit `3292aa6703019891897b8d979582447b5e21ecee`
- Cython 3.3.0
- Python 3.12.13 and Python 3.14.6
- macOS 26.6, ARM64

## Harmless proof of concept

Save the following raw ASCII/UTF-8 bytes as `repro.py`:

```python
"""This is ordinary text; coding: utf-7"""
# +AAo-import subprocess; print("This is only commnet in any python");print("sum =", sum((5, 5))); subprocess.run(["whoami"], check=True)
print("Normal behaviour")
```

### CPython behavior

```sh
python3.12 repro.py
```

```text
Normal behaviour
```

CPython correctly treats the first line as a module docstring and the second line as a comment.

### Cython behavior

```sh
cython --embed -3 -o repro.c repro.py
cc repro.c -o repro $(python3.12-config --includes) $(python3.12-config --ldflags --embed)
./repro
```

```text
<local username>
This is only commnet in any python
sum = 10
Normal behaviour
```

The exact ordering of the username relative to Python's printed lines can vary when standard output is redirected because the child process and embedded Python runtime use different buffering. This does not affect the semantic result: all three injected statements execute only in the Cython artifact.

Cython falsely selects UTF-7 from the first-line string. Under UTF-7, `+AAo-` decodes to a newline, so Cython parses the second raw line as:

```python
#
import subprocess; print("This is only commnet in any python");print("sum =", sum((5, 5))); subprocess.run(["whoami"], check=True)
```

The generated native binary therefore imports a module, evaluates an expression, and launches a local process from text that remains entirely commented out under CPython's correct UTF-8 interpretation.

## Expected behavior

Only a valid PEP 263 comment on the first or second line should select the source encoding. Text inside a string literal must not affect decoding. Cython and CPython should parse the same source semantics and both print only `Normal behaviour`.

## Root cause

`Cython/Utils.py` currently defines an unanchored search:

```python
_match_file_encoding = re.compile(br"(\w*coding)[:=]\s*([-\w.]+)").search
```

`detect_opened_file_encoding()` applies it directly to each of the first two raw lines. The expression does not require optional whitespace followed by `#`, as PEP 263 does, and therefore accepts matches inside arbitrary strings and other source text.

## Suggested remediation

Require the encoding declaration to occur in a first- or second-line comment and match from the start of the line. One minimal direction that preserves Cython's current capture groups is:

```diff
-_match_file_encoding = re.compile(br"(\w*coding)[:=]\s*([-\w.]+)").search
+_match_file_encoding = re.compile(
+    br"^[ \t\f]*#.*?(\w*coding)[:=]\s*([-\w.]+)"
+).match
```

Using or aligning with `tokenize.detect_encoding()` would reduce future divergence from CPython. The maintainers should determine the canonical implementation and ensure that Cython-specific directives remain supported without widening PEP 263 matching to non-comments.

## Suggested regression tests

1. First-line and second-line strings containing `coding: unknown` must remain normal strings.
2. First-line and second-line strings containing `coding: utf-7` must not change source decoding.
3. The harmless proof of concept above must print only `Normal behaviour` after Cython compilation.
4. Valid `# coding: utf-8` and `# -*- coding: latin-1 -*-` comments must continue to work.
5. Cython-specific `c_string_encoding` directives must retain their intended behavior.

_No response_

## Summary

Cython can interpret `coding: <encoding>` text inside an ordinary string on
the first or second physical source line as a PEP 263 source-encoding cookie.
PEP 263 and CPython only recognize such a declaration in a comment.

By placing `coding: utf-7` in a first-line string, an attacker can make Cython
decode the complete file as UTF-7. The UTF-7 sequence `+AAo-` then decodes to a
line-feed character. If that sequence appears after `#` on the next raw line,
the decoded newline terminates the comment for Cython even though CPython sees
the entire physical line as a comment. Consequently, code that is inert under
CPython and visually appears commented out can execute when the Cython-built
artifact is run.

This creates a source-review and test-bypass primitive: a malicious source
change can behave benignly when reviewed or tested with CPython but gain
additional behavior in a Cython build.

## Harmless proof of concept

Save the following ASCII bytes as `repro.py`:

```python
"""This is ordinary text; coding: utf-7"""
# +AAo-print("CYTHON_ONLY")
print("COMMON")
```

### Direct CPython execution

```sh
python3.12 repro.py
```

Output:

```text
COMMON
```

CPython correctly treats the first line as a module docstring and the entire
second physical line as a comment.

### Cython compilation and execution

```sh
cython --embed -3 -o repro.c repro.py
cc repro.c -o repro $(python3.12-config --embed --cflags --ldflags)
./repro
```

Output:

```text
CYTHON_ONLY
COMMON
```

Cython falsely selects UTF-7 based on text in the first-line string. Under
UTF-7, the second raw line is decoded approximately as:

```python
#
print("CYTHON_ONLY")
```

The second statement is therefore included as executable code in the generated
artifact. The hidden statement is not executed by the compilation command
itself; it executes when the resulting binary is run.

## Expected behavior

Cython should recognize a source-encoding declaration only where CPython would
recognize one under PEP 263. Text inside a string literal must not select the
source codec. Both direct execution and the Cython-built artifact should print
only:

```text
COMMON
```

## Security impact

A malicious contributor or compromised source dependency could use this parser
differential to conceal behavior from:

- human review and syntax highlighting, where the payload appears in a comment;
- tests or validation performed by running the file with CPython; and
- source-analysis tooling that follows CPython's encoding-cookie rules.

The additional behavior becomes active when the Cython-generated program or
extension is loaded or executed. Exploitation therefore requires the crafted
source to enter a Cython build and the resulting artifact to be run, but it can
break the common assumption that Cython preserves the executable meaning of
valid Python source.

## Root cause

`Cython/Utils.py` defines an unanchored regular-expression search:

```python
_match_file_encoding = re.compile(br"(\w*coding)[:=]\s*([-\w.]+)").search
```

`detect_opened_file_encoding()` applies this search to the first two raw lines.
Because the expression is not anchored to a comment at the beginning of the
line, it accepts `coding:` matches inside strings and other arbitrary source
text. Cython then opens and decodes the entire source file using the falsely
selected codec before parsing it.

There is also a second-line semantic detail worth preserving in the fix: under
PEP 263, a cookie on line two is valid only when line one contains no executable
statement. Merely requiring `#` on each candidate line would not fully reproduce
CPython's detection rules.

## Suggested remediation

Align source-encoding detection with CPython's `tokenize.detect_encoding()`
behavior, including:

1. requiring an encoding cookie to be in a first- or second-line comment;
2. applying CPython's restriction on when the second line may contain a cookie;
3. preserving BOM/cookie consistency checks; and
4. handling Cython-specific directives separately so they cannot be mistaken
   for a source codec.

If reusing `tokenize.detect_encoding()` directly is not suitable, equivalent
matching and line-selection logic should be implemented rather than only
anchoring the current regular expression.

## Suggested regression tests

1. A first-line string containing `coding: unknown` remains a normal string and
   does not raise `LookupError`.
2. First- and second-line strings containing `coding: utf-7` do not affect
   decoding.
3. The proof of concept above prints only `COMMON` after Cython compilation.
4. Valid first-line and second-line PEP 263 cookies continue to work.
5. A second-line cookie is ignored or rejected in the same cases as CPython
   when the first line contains executable source.
6. UTF-8 BOM handling and BOM/cookie conflicts match CPython.
7. Cython-specific `c_string_encoding` directives retain their intended
   behavior without being treated as source-encoding cookies.

## Confirmed environments

Fresh local reproductions:

- Cython 3.2.9 with Python 3.12.14
- Cython 3.3.0 with Python 3.14.6
- macOS 26.6 (Build 25G72), ARM64

The issue was also checked during the investigation against Cython repository
commit `3292aa6703019891897b8d979582447b5e21ecee`, which reported version 3.3.0.

## Additional non-security symptom

The same root cause can cause valid Python files to fail compilation. For
example:

```python
"""Hello, this is an ordinary Python string; coding: spending time wisely."""
print("ok")
```

CPython prints `ok`, while affected Cython versions raise:

```text
LookupError: unknown encoding: spending
```

This simpler failure originally exposed the overly broad cookie detection; the
UTF-7 proof above demonstrates its security-relevant semantic consequence.

## Disclosure and validation notes

- No destructive payload, persistence, network access, privilege change, or
  external target was used in the proof of concept.
- A search performed during the investigation found no matching public Cython
  report or changelog entry for this behavior.

### OS

_No response_

### Python version

_No response_

### Cython version

_No response_

### Additional context

_No response_

da-woods:
Fixes #7935

A few improvements:
* Non-comment lines are detected, and correctly cause the search to ignore them (and anything after them).
* The c_string_encoding exception-to-the-rule is detected on both the first and the second line instead of just the first.
* The c_string_encoding exception-to-the-rule generates a warning on the basis that it is different from Python.

### DIFF
--- a/Cython/Utils.py
+++ b/Cython/Utils.py
@@ -357,15 +357,16 @@ def detect_opened_file_encoding(f, default='UTF-8'):
         if m:
             return default
         m = _match_file_encoding(line)
-        if m:
-            if m.group(1) != b'c_string_encoding':
-                return m.group(2).decode('iso8859-1')
-            else:
-                from .Compiler.Errors import warning
-                warning(
-                    None,
-                    "c_string_encoding in first two lines of a file is interpreted as a directive "
-                    "in Cython and a source file-encoding in Python", 2)
+        if not m:
+            continue
+        if m.group(1) == b'c_string_encoding':
+            from .Compiler.Errors import warning
+            warning(
+                None,
+                "c_string_encoding in the first two lines of a file is interpreted as a directive "
+                "in Cython and a source file-encoding in Python", 2)
+            continue
+        return m.group(2).decode('iso8859-1')
     return default
 
 
--- a/Cython/Utils.py
+++ b/Cython/Utils.py
@@ -12,6 +12,7 @@ cython.declare(
     os=object, sys=object, re=object, io=object, glob=object, shutil=object, tempfile=object,
     update_wrapper=object, partial=object, wraps=object, cython_version=object,
     _cache_function=object, _function_caches=list, _parse_file_version=object, _match_file_encoding=object,
+    _match_non_comment_line=object
 )
 
 import os
@@ -334,6 +335,7 @@ def decode_filename(filename):
 
 # support for source file encoding detection
 
+_match_non_comment_line = re.compile(br"\s*[^\s#]").match
 _match_file_encoding = re.compile(br"(\w*coding)[:=]\s*([-\w.]+)").search
 
 
@@ -350,13 +352,20 @@ def detect_opened_file_encoding(f, default='UTF-8'):
         if not data:
             break
 
-    m = _match_file_encoding(lines[0])
-    if m and m.group(1) != b'c_string_encoding':
-        return m.group(2).decode('iso8859-1')
-    elif len(lines) > 1:
-        m = _match_file_encoding(lines[1])
+    for line in lines[:2]:
+        m = _match_non_comment_line(line)
         if m:
-            return m.group(2).decode('iso8859-1')
+            return default
+        m = _match_file_encoding(line)
+        if m:
+            if m.group(1) != b'c_string_encoding':
+                return m.group(2).decode('iso8859-1')
+            else:
+                from .Compiler.Errors import warning
+                warning(
+                    None,
+                    "c_string_encoding in first two lines of a file is interpreted as a directive "
+                    "in Cython and a source file-encoding in Python", 2)
     return default
 
 
--- a/Cython/Tests/TestCythonUtils.py
+++ b/Cython/Tests/TestCythonUtils.py
@@ -1,10 +1,10 @@
 import sys
-from io import StringIO
+from io import StringIO, BytesIO
 
 from Cython.Utils import (
     _CACHE_NAME_PATTERN, _build_cache_name, _find_cache_attributes,
     build_hex_version, cached_method, clear_method_caches, try_finally_contextmanager,
-    print_version, normalise_float_repr,
+    print_version, normalise_float_repr, detect_opened_file_encoding
 )
 from Cython.TestUtils import TimedTest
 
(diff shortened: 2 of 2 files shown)


Cython/Utils.py
```diff
@@ -350,13 +352,20 @@ def detect_opened_file_encoding(f, default='UTF-8'):
...
+        m = _match_file_encoding(line)
+        if m:
+            if m.group(1) != b'c_string_encoding':
+                return m.group(2).decode('iso8859-1')
+            else:
+                from .Compiler.Errors import warning
+                warning(
+                    None,
+                    "c_string_encoding in first two lines of a file is interpreted as a directive "
+                    "in Cython and a source file-encoding in Python", 2)
```
scoder:
```suggestion
        if not m:
            continue
        if m.group(1) == b'c_string_encoding':
            from .Compiler.Errors import warning
            warning(
                None,
                "c_string_encoding in the first two lines of a file is interpreted as a directive "
                "in Cython and a source file-encoding in Python", 2)
            continue
        return m.group(2).decode('iso8859-1')
```

scoder (commented):
[PEP-263](https://peps.python.org/pep-0263/) allows `coding=…` in addition to `coding: …`. Tests for that are missing. You could run the test cases a second time with `: ` replaced by `=` to keep it simple.

PR #7944, opened 2026-08-27
Add test for inherited weakref

da-woods:
See comment https://github.com/cython/cython/issues/7399#issuecomment-5384918484. This is a bug specifically in my WIP freethreading build (it works in master).
However, we don't have a test for it so I'm just adding the test in the master branch.

### DIFF
--- a/Cython/Compiler/TypeSlots.py
+++ b/Cython/Compiler/TypeSlots.py
@@ -487,6 +487,7 @@ class ConstructorSlot(InternalMethodSlot):
                 and not scope.has_pyobject_attrs
                 and not scope.has_memoryview_attrs
                 and not scope.has_explicitly_constructable_attrs
+                and not (self.slot_name == 'tp_dealloc' and scope.lookup_here("__weakref__"))
                 and not (self.slot_name in ('tp_new', 'tp_new_vectorcall') and scope.parent_type.vtabslot_cname)):
             entry = scope.lookup_here(self.method) if self.method else None
             if not (entry and entry.is_special):
--- a/tests/run/weakfail.pyx
+++ b/tests/run/weakfail.pyx
@@ -21,3 +21,25 @@ def test_weakref(key):
     foo_dict[key] = obj
     return obj
 
+
+cdef class Inherited1(Foo):
+    pass
+
+cdef class Inherited2(Foo):
+    cdef int something_else
+
+cdef class Inherited3(Foo):
+    cdef object __weakref__  # It's allowed for classes to reimplement this
+
+def test_inherited_weakref(tp):
+    """
+    >>> test_inherited_weakref(Inherited1)
+    >>> test_inherited_weakref(Inherited2)
+    >>> test_inherited_weakref(Inherited3) 
+    """
+    inst = Inherited1()
+    weakinst = weakref.ref(inst)
+    assert weakinst() is not None
+    del inst
+    gc.collect()
+    assert weakinst() is None
--- a/tests/run/weakfail.pyx
+++ b/tests/run/weakfail.pyx
@@ -31,11 +31,18 @@ cdef class Inherited2(Foo):
 cdef class Inherited3(Foo):
     cdef object __weakref__  # It's allowed for classes to reimplement this
 
+cdef class NoWeakref:
+    pass
+
+cdef class Inherited4(NoWeakref):
+    cdef object __weakref__
+
 def test_inherited_weakref(tp):
     """
     >>> test_inherited_weakref(Inherited1)
     >>> test_inherited_weakref(Inherited2)
-    >>> test_inherited_weakref(Inherited3) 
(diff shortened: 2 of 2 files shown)


da-woods:
Assuming it passes I'll just merge this - I don't see adding a bit more test coverage being contraversal

tests/run/weakfail.pyx
```diff
@@ -21,3 +21,25 @@ def test_weakref(key):
...
+cdef class Inherited3(Foo):
+    cdef object __weakref__  # It's allowed for classes to reimplement this
+
+def test_inherited_weakref(tp):
+    """
+    >>> test_inherited_weakref(Inherited1)
+    >>> test_inherited_weakref(Inherited2)
+    >>> test_inherited_weakref(Inherited3) 
+    """
+    inst = Inherited1()
```
dalcinl: Did you mean the following?

```suggestion
    inst = tp()
```

da-woods: Yes thanks. If got refactored late but clearly not thoroughly enough

Cython/Compiler/TypeSlots.py
```diff
@@ -487,6 +487,7 @@ def _needs_own(self, scope):
                 and not scope.has_pyobject_attrs
                 and not scope.has_memoryview_attrs
                 and not scope.has_explicitly_constructable_attrs
+                and not (self.slot_name == 'tp_dealloc' and scope.lookup_here("__weakref__"))
```
da-woods, line 490: This was causing

```
cdef class NoWeakref:
    pass

cdef class Inherited4(NoWeakref):
    cdef object __weakref__
```

to fail (among other issues) because it never cleared but the weaklist because it didn't generate `tp_dealloc`

dalcinl: Just to be sure... `tp_dealloc` will be generated in the `Inherited4` subclass , not in the `NoWeakref` base class, right?

dalcinl (approved):

Cython/Compiler/TypeSlots.py
da-woods: It's generated for both.

The `NoWeakref` base class hits the "doesn't have a base class" rule so always gets `tp_dealloc`.
`Inherited4` gets its own `tp_dealloc` because of the `__weakref__` (which is the important bit for this test)

(The version here does unnecessarily generate `tp_dealloc` for anything that inherits from something that has `__weakref__`. Those are harmless, but in principle could be cleaned up. It's probably not worth it though. It'd need to duplicate some logic from `generate_dealloc_function`)

dalcinl: > The `NoWeakref` base class hits the "doesn't have a base class" rule so always gets `tp_dealloc`.

OK, of course, I did not state my question properly.

> Inherited4 gets its own tp_dealloc because of the __weakref__ 

OK, that's what I meant to ask. Thanks.

dalcinl:
@da-woods Now that you are at it... 

I think you know that a subclass and define a new __weakref__ different than the one on the base class, that is 
```cython
class A:
    cdef object __weakref__
class B(A):
    cdef object __weakref__
```
What I have not thought of is how to make a test for it. 
In any case, are Cython tests covering this situation?

EDIT: Again, I did not expressed myself properly. What I meant is that we do not have a test that asserts that these classes have different weakref lists.

da-woods:
> I think you know that a subclass and define a new **weakref** different than the one on the base class, that is

`Inherited3` does that (and was causing a crash before I added the rules to generate `tp_dealloc`). It doesn't test anything except that `__weakref__` works, so it doesn't really care which weakref is being use.

dalcinl:
> It doesn't test anything except that `__weakref__` works,

OK, I guess that is enough.

da-woods:
> EDIT: Again, I did not expressed myself properly. What I meant is that we do not have a test that asserts that these classes have different weakref lists.

The test would be `assert Inherited3.__weakrefoffset__ != Foo.__weakrefoffset__`

However - this looks to be something where the Cython with heap types (i.e. limited API) and static extension types behave a little differently. For static extension types it doesn't end up using the new `__weakref__` while for heap-types it does.

My view is that it doesn't matter too much as long as something works and so I won't get too worried about fixing it unless someone gives me a real case where it matters.

dalcinl:
> My view is that it doesn't matter too much as long as something works

Definitely. Some day we should consider using a decorator `@cython.weakref` to be used on cdef classes. That would remove the need for that nasty `__weakref__` attribute that is just a ugly CPython implementation detail.

da-woods:
> Some day we should consider using a decorator `@cython.weakref` to be used on cdef classes. That would remove the need for that nasty `__weakref__` attribute that is just a ugly CPython implementation detail.

It'd also allow the `Py_TPFLAGS_MANAGED_WEAKREF` type flag (at least outside the Limited API) which is now modern Python's preferred object layout.

PR #7947, opened 2026-08-30
switch default CYTHON_USE_TYPE_SPECS on pypy to 1

mattip:
In PR #5030, an error was added to [`__Pyx_ApplySequenceOrMappingFlag`](https://github.com/cython/cython/blob/0fca25b1862663f33e23b7a4bbc9a72fabaa9eee/Cython/Utility/ExtensionTypes.c#L1016) for pypy when `CYTHON_USE_TYPE_SPECS` is not active. This breaks using PyPy and NumPy when using the default. PyPy has supported CYTHON_USE_TYPE_SPECS=1 in its releases for quite a while. I have been running Cython tests with that flag for as long as we have supported python3.8 (at least May 2021), so there should be no users of pypy + cython where it will break.


scoder:
We're still fixing up 3.3.0 to make it usable for everyone, so let's sneak this in while we're at it.

PR #7948, opened 2026-08-30
Implement boolean coercion for the C return type.

### LINKED ISSUE #7946: [BUG] cython 3.3.0 mis-optimizes set.add
### Describe the bug

AI generated from a true bug:

Cython 3.3.0 miscompiles `not (x in <set> or add(x))` when add is a local bound to a set's `.add` method (`add = s.add`).
It optimises the call to `PySet_Add()`, then coerces the `(None)` return value to bool with the `PyLong` truthiness fast-path `__Pyx_PyLong_IsNonZero()` instead of `__Pyx_PyObject_IsTrue()`.
That reads `Py_SIZE(Py_None)` — UB, effectively "truthy" — so `x in s or add(x)` is always truthy and not (...) is always False.

```cython
def dedupe_via_local(seq):
    seen = set()
    seen_add = seen.add
    return [x for x in seq if not (x in seen or seen_add(x))]

# dedupe_via_local(["a", "a", "b"])
#   Cython 3.2.9 -> ['a', 'b']   (correct)
#   Cython 3.3.0 -> []           (wrong)
```

`seen.add(x)` inline compiles correctly; only the method-local form breaks.
Also breaks outside comprehensions (plain for loop). Reproduces at `-O0`, default directives, `CPython 3.14` / `x86-64` Linux.

Generated-C smoking gun (3.3.0)
```
__pyx_t_7 = PySet_Add(__pyx_v_seen, __pyx_v_x);          /* int status */
__pyx_t_4 = __Pyx_Owned_Py_None(__pyx_t_7);              /* -> Py_None  */
Py_ssize_t __pyx_temp = __Pyx_PyLong_IsNonZero(__pyx_t_4);   /* BUG: PyLong path on None */
```
vs. the correct dedupe_inline path: `__pyx_t_7 = (PySet_Add(...) != 0);`

The optimised call result node seems to retain an int/long result type from the `PySet_Add` status code, which then steers `CoerceToBooleanNode` into the `PyLong` fast path.

### OS
Any: Linux

### Python version
3.14.x

### Cython version
3.3.0

### Additional context

We've applied this workaround to our cython code for now:
https://github.com/Xpra-org/xpra/commit/316578aa3a233353a4d0d735474d64bfabcab522

scoder:
We represent the result as 'None' in Python objects, so a coercion to boolean must return constant false but still evaluate the expression.

Closes https://github.com/cython/cython/issues/7946

### DIFF
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -15304,6 +15304,8 @@ class CoerceToPyTypeNode(CoercionNode):
             elif arg.type.equivalent_type is not None and arg.type.equivalent_type.is_pyobject:
                 # Includes bint.
                 self.type = arg.type.equivalent_type
+            elif arg.type.is_returncode:
+                self.type = py_object_type  # constant None
             elif arg.type.is_int:
                 self.type = Builtin.int_type
             elif arg.type.is_float:
@@ -15323,6 +15325,9 @@ class CoerceToPyTypeNode(CoercionNode):
     gil_message = "Converting to Python object"
 
     def may_be_none(self):
+        if self.arg.type.is_returncode:
+            # We do not set '.constant_result = None' to prevent accidental node elimination.
+            return True
         # FIXME: is this always safe?
         return False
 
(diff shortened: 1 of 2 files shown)


da-woods:
This makes sense I think

PR #7949, opened 2026-08-30
Make dict iteration work with less trivial dict expressions.

### LINKED ISSUE #7931: [BUG] Optional typed dictionary iteration crashes Cython 3.3.0
### Describe the bug

Cython 3.3.0 crashes while it infers the type of `.items()` on an optional typed dictionary. Cython 3.2.9 compiles the same file. Master at `ab9899598718f3d949e86aff7d34bdee622ebe72` also crashes.

```cython
# cython: language_level=3, annotation_typing=False

def iterate_optional_dict(values: dict[int, float] | None = None):
    for key, value in (values or {}).items():
        pass
```

```console
python -m cython repro.py -o repro.c
```

The compiler raises `TypeError: NoneType object is not iterable` in `SimpleCallNode.infer_type`.

### Expected behaviour

Cython creates the C file, as Cython 3.2.9 does.

### Environment

- OS: Ubuntu on WSL
- Python: 3.12
- Cython: 3.3.0 and master

scoder:
Used to crash Cython 3.3.0.
Closes https://github.com/cython/cython/issues/7931

### DIFF
--- a/tests/run/iterdict.pyx
+++ b/tests/run/iterdict.pyx
@@ -679,12 +679,12 @@ cdef class NotADict:
 
 
 @cython.annotation_typing(False)
-def iterate_optional_dict(values: dict[int, float] | None = None):
+def iterate_optional_dict_untyped(values: dict[int, float] | None = None):
     """
-    >>> iterate_optional_dict()
-    >>> iterate_optional_dict(None)
-    >>> iterate_optional_dict({})
-    >>> iterate_optional_dict({'a': 'b'})
+    >>> iterate_optional_dict_untyped()
+    >>> iterate_optional_dict_untyped(None)
+    >>> iterate_optional_dict_untyped({})
+    >>> iterate_optional_dict_untyped({'a': 'b'})
     a b
     """
     # See https://github.com/cython/cython/issues/7931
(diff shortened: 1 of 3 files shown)


tests/run/iterdict.pyx
```diff
@@ -1,3 +1,10 @@
+# mode: run
+# cython: language_level=2
+
+# Language level 2 to keep ".items()" etc. as unsafe methods.
+# See "cython3.pyx" for corresponding "language_level=3" tests.
```
da-woods, line 5: What are we getting from doing this? In my mind language level 2 is mostly about supporting old Python 2 syntax (so that people don't have to update their files). But this just looks like it selects a method of optimization and doesn't actually change any behaviour. Possibly I'm missing something here though?

With that said, a bugfix PR may not be the place to change it.

tests/run/iterdict.pyx
```diff
@@ -669,3 +676,31 @@ cdef class NotADict:
...
+    >>> iterate_optional_dict({'a': 'b'})
+    a b
+    """
+    # See https://github.com/cython/cython/issues/7931
+    for key, value in (values or {}).items():
+        print(key, value)
+
+
+@cython.annotation_typing(True)
+def iterate_optional_dict(values: dict[int, float] | None = None):
```
da-woods: This need to be given different names in order to usefully run it.

tests/run/iterdict.pyx
scoder: I added the comment to describe the status quo. The test depends on `language_level=2`, which we currently use for all tests unless stated explicitly otherwise. To express that dependency, I added option and comment.

scoder:
I don't consider this the optimal fix as it's just working around symptoms, but it's still reasonable and adds the relevant tests, so the underlying issues (apparently different ways of representing similar user code and once again the different call node interfaces) can be revisited and cleaned up more safely at some point.

PR #7950, opened 2026-08-31
Document how "cpp_locals=True" behaves for extension type attributes

scoder:


docs/src/userguide/wrapping_CPlusPlus.rst
```diff
@@ -826,26 +826,22 @@ utility can be used to generate a C++ ``.cpp`` file, and then compile it
...
-            return c.some_cpp_function()  # OK
-            
+C++ variables behave like normal Python object variables.
+With this directive, they are only initialized at their first assignment, and thus
+they no longer require a nullary constructor to be allocated on the call stack
+or in extension type structs.
+Trying to access an uninitialized C++ variable will generate an ``UnboundLocalError``
+(or similar) in the same way as a Python variable would.  For example:
+
+.. literalinclude:: ../../examples/userguide/wrapping_CPlusPlus/cpp_locals.pyx
```
da-woods, line 843: Did you forget to commit this?

scoder: Obviously :)

tests/run/cpp_locals_directive.pyx
```diff
@@ -270,3 +270,24 @@ def read_global_var_nogil():
     with nogil:
         val = global_var.getX()
     print(val)
+
+
+cdef class PyC:
```
da-woods: I think

```
cdef class HoldsC:
```

above already tests this. Although I don't mind an extra test

docs/examples/userguide/wrapping_CPlusPlus/cpp_locals.pyx
```diff
@@ -0,0 +1,24 @@
...
+# cython: cpp_locals = True
+
+from Rectangle cimport Rectangle
+
+
+cdef class RectOwner:
+    cdef Rectangle c_rect  # not initialized
+
+    def __cinit__(self, x, y, w, h):
+        self.c_rect = Rectangle(x, y, w, h)  # initialized here
```
da-woods:
```suggestion
    def __cinit__(self):
        pass
        
    def setRect(self, x, y, w, h):
        self.c_rect = Rectangle(x, y, w, h)  # initialized here
```

I wonder if this makes the "might be uninitialized" aspect of it more explicit?

tests/run/cpp_locals_directive.pyx
scoder: Ah, right. My scan of the test module must have been too superficial then. I'll remove it.

docs/examples/userguide/wrapping_CPlusPlus/cpp_locals.pyx
scoder: It's documentation, so explicit is always better than implicit.

da-woods: ... but definitely in favour of adding an extension type example.

scoder:
I initially chose `__cinit__` in the example because it seemed the safest option to show in the docs. But there doesn't seem to be much guidance for writing wrapper constructors in the "wrapping C++" docs. That might be worth adding (not in this PR).

scoder:
The "shared-utility" build failures are due to a general problem with parallel builds of the shared module because it enables `cimport_from_pyx`. I'll look into that.

clang seems to be unhappy about a missing `std::optional`.

da-woods:
> clang seems to be unhappy about a missing `std::optional`.

I think think this might just need `-std=c++17` as a compile argument. Which I think most compilers do automatically now, but not clang on OSX apparently.

PR #7952, opened 2026-08-31
Escape depfile targets and remaining make-special characters

r3wretrhy:
Follow-up to #7423.

`write_depfile()` escaped spaces in dependency paths but not in the target path, so a `.c` file under a directory with spaces still produced a depfile GNU make / ninja cannot parse. `#` (make comments) and `$` (make variables) in paths were also left unescaped.

Paths in both the target and the dependency list now use gcc `-M` escaping: space and `#` get a backslash, `$` is doubled. Backslashes are still left unchanged, as in #7423.

Tests:

- `python runtests.py Cython.Tests.TestCythonUtils --no-file`
- `python runtests.py depfile -x numpy -x package --no-cpp`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@ Cython Changelog
 Bugs fixed
 ----------
 
+* Depfiles now escape the target path, as well as ``#`` and ``$`` in paths,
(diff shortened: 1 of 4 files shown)


da-woods:
I'm going to close this because you're clearly bulk generating PRs and I consider this a DDOS of open source.

PR #7954, opened 2026-09-01
Remove dict special cases for language_level=2

da-woods:
This seems to be about handling the "new" `items`, `keys`, `values` methods. If we have an exact `dict` then we know that these exist (because Cython only supports Python 3) so we can always optimize them.


da-woods:
I don't think this should be in the 3.3.1 bugfix release. It's also possible that I've missed some test that it breaks.

But it seems like an unnecessary complication to me so I think it's worth removing eventually.

scoder:
Ah, right. The runtime code has a fallback for the case that it doesn't find a dict, so this really is useless left-over code.

Since we've just touched that code, I wouldn't mind merging it for 3.3.1. It really seems a useless degradation in Py3.

PR #7959, opened 2026-09-02
Fix dangling __Pyx_FakeReference for conditional expr assigned into C++ container element (GH-7927)

### LINKED ISSUE #7927: [BUG] C++: assigning a conditional expression into a container element generates a dangling `__Pyx_FakeReference` temporary
### Describe the bug

In C++ mode, assigning a Python conditional expression directly into a C++ container
element:

```cython
self.logs[0][i] = 1 if cond else 2      # logs: unordered_map[int, vector[int]]
```

generates the ternary's temporary as `__Pyx_FakeReference<int>` instead of `int`, split
across separate C statements. Because `__Pyx_FakeReference` defines no assignment
operator, each branch assignment materializes a *temporary* `int`, binds the
FakeReference's internal pointer to it via the `const T&` converting constructor, and
default-copy-assigns the (immediately dangling) pointer. The temporary dies at the end
of that statement; the subsequent store statement then reads `*ptr` from the dead stack
slot — undefined behavior.

The emitted code shape is identical in at least Cython **3.0.12** and **3.3.0**. Whether
the UB is observable depends on the surrounding codegen/optimizer: binaries we built
with 3.0.12 happened to read the stale value back intact for years; binaries of the
identical `.pyx` built with 3.3.0 stored garbage (stack remnants — counters, packed
values, pointer halves) into the container in production.

- Cython 3.0.12 and 3.3.0 (generated-C shape verified identical for this construct in both)
- Python 3.10.11
- C++ mode (`# distutils: language = c++`), observed with clang 14 (macOS arm64) and
  gcc (manylinux x86_64); directive-independent (reproduces with bare defaults at every
  `language_level`)

`repro.pyx`:

```cython
# distutils: language = c++
from libcpp.unordered_map cimport unordered_map

cdef class Holder:
    cdef unordered_map[int, vector[int]] logs

    def __init__(self, int n):
        cdef vector[int] blanks
        blanks.assign(n, 0)
        self.logs[0] = blanks

    cpdef void write(self, int i, bint cond):
        # BUG: conditional expression assigned directly into a C++ container element
        self.logs[0][i] = 1 if cond else 2

    cpdef int read(self, int i):
        return self.logs[0][i]
```

```
cython --cplus repro.pyx -o repro.cpp
grep -n 'FakeReference<int>' repro.cpp
```

```c
__Pyx_FakeReference<int> __pyx_t_8;
...
if (__pyx_v_cond) {
  __pyx_t_8 = 1;
} else {
  __pyx_t_8 = 2;
}
((__pyx_v_self->logs[0])[__pyx_v_i]) = __pyx_t_8;
```

with the utility class:

```c
template<typename T>
class __Pyx_FakeReference {
  public:
    __Pyx_FakeReference() : ptr(NULL) { }
    __Pyx_FakeReference(const T& ref) : ptr(const_cast<T*>(&ref)) { }
    T *operator->() { return ptr; }
    T *operator&() { return ptr; }
    operator T&() { return *ptr; }
    ...
  private:
    T *ptr;
};
```

`__Pyx_FakeReference` has **no `operator=`**. So `__pyx_t_8 = 1;` resolves as:

1. materialize a temporary `int` with value `1`;
2. convert it to `__Pyx_FakeReference<int>` via `__Pyx_FakeReference(const T& ref)`,
   binding `ptr` to the temporary's address;
3. implicit copy-assignment copies that pointer into `__pyx_t_8`;
4. the temporary `int` is destroyed at the end of the full-expression.

The next statement, `container = __pyx_t_8;`, invokes `operator T&()` and dereferences
the now-dangling pointer. Whatever the reused stack slot holds is written into the
container. (When the ternary's branches are *lvalues* — struct members, vector
elements — the pointer binds live storage and the pattern is harmless; the bug is
specific to prvalue branches combined with the statement-level split that a
container-element assignment target forces.)

### Expected behaviour

The temporary for the conditional expression should be a plain `int` (as is emitted
when the assignment target is a plain local variable), e.g.:

```c
int __pyx_t_8;
if (__pyx_v_cond) { __pyx_t_8 = 1; } else { __pyx_t_8 = 2; }
((__pyx_v_self->logs[0])[__pyx_v_i]) = __pyx_t_8;
```

Rewriting the source to `cdef int code; if cond: code = 1; else: code = 2;
self.logs[0][i] = code` produces exactly that and is the workaround we shipped.

### OS

Linux

### Python version

3.10.11

### Cython version

3.3.0 and 3.0.12

### Additional context

_No response_

jayClean:
Fixes https://github.com/cython/cython/issues/7927

## Root cause

For `container[i] = a if cond else b` in C++ mode, the assignment target
(`operator[]` returning `T&`) makes `SingleAssignmentNode` call
`rhs.coerce_to(T&)` on the `CondExprNode`. `CondExprNode.coerce_to` overrides
the base `ExprNode.coerce_to` to push the target type onto each branch, but
did not strip the reference first the way the base method does. Each prvalue
branch was stamped with type `T&`, the spanning type became `T&`, and
`analyse_result_type` converted it to `__Pyx_FakeReference<T>`. Since
`__Pyx_FakeReference` has no `operator=`, `__pyx_t = a;` binds its internal
pointer to a temporary that dies at the end of the full-expression; the
following store reads the dangling pointer.

## Fix

In `CondExprNode.coerce_to`, when the target is a reference but the
conditional's own type is not (branches are prvalues), coerce to the
referenced base type instead — mirroring `ExprNode.coerce_to`. The
both-branches-are-lvalue-references case is unchanged.

## Test

`tests/run/cpp_conditional_container_assignment.pyx` uses
`test_fail_if_c_code_has = __Pyx_FakeReference<int>` (deterministic: 3
occurrences before the fix, 0 after) plus runtime value checks.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,11 @@ Bugs fixed
 * A dangling pointer was fixed when assigning to attributes of extension types.
   (Github issue :issue:`7907`)
 
+* In C++ mode, assigning a conditional expression directly into a container
+  element (e.g. ``d[k] = a if cond else b``) generated a dangling
+  ``__Pyx_FakeReference`` temporary and could store garbage.
+  (Github issue :issue:`7927`)
+
 * A crash was resolved when mixing Cython modules compiled with different Limited API versions.
   (Github issue :issue:`7914`)
 
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -1602,6 +1602,9 @@ class IntNode(ConstNode):
         return suitable_type
 
     def coerce_to(self, dst_type, env):
+        if dst_type.is_reference:
+            # A literal is a prvalue, so drop the reference (GH-7927).
+            dst_type = dst_type.ref_base_type
         if self.type is dst_type:
             return self
         elif dst_type.is_float or dst_type.is_pyfloat_type:
@@ -1714,6 +1717,9 @@ class FloatNode(ConstNode):
         return float_value
 
     def coerce_to(self, dst_type, env):
+        if dst_type.is_reference:
+            # A literal is a prvalue, so drop the reference (GH-7927).
+            dst_type = dst_type.ref_base_type
         if dst_type.is_pyobject and self.type.is_float:
             return FloatNode(
                 self.pos, value=self.value,
--- a/tests/run/cpp_conditional_container_assignment.pyx
+++ b/tests/run/cpp_conditional_container_assignment.pyx
@@ -1,0 +1,68 @@
+# mode: run
+# tag: cpp, werror, no-cpp-locals
+# cython: test_fail_if_c_code_has = __Pyx_FakeReference<int>
+# cython: test_fail_if_c_code_has = __Pyx_FakeReference<double>
+
+# Assigning a prvalue conditional expression directly into a C++ container
+# element used to make the ternary temp a dangling __Pyx_FakeReference<T>
+# instead of a plain T (GH-7927).
+
+from libcpp.unordered_map cimport unordered_map
+from libcpp.vector cimport vector
+
+
+cdef class Holder:
+    cdef unordered_map[int, vector[int]] logs
+
+    def __init__(self, int n):
+        cdef vector[int] blanks
+        blanks.resize(n, 0)
+        self.logs[0] = blanks
+
+    def store(self, int i, bint cond):
+        self.logs[0][i] = 1 if cond else 2
+
+    def load(self, int i):
+        cdef int result = self.logs[0][i]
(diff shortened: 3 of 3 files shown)


da-woods:
This made sense to me. I did wonder if the problem was really that the `if dst_type.is_reference and not self.type.is_reference` bit was missing from `IntNode` and and `DoubleNode` `coerce_to`, or if it should be applied universally in almost all `coerce_to`'s.

With that said, I failed to find anything else that could produce this behaviour so a local fix is probably fine. `or` would be an obvious candidate although ends up not producing the same behaviour for other reasons.

The comment is a little verbose.

PR #7960, opened 2026-09-02
Fix sorted() optimiser crash on typeless argument

### LINKED ISSUE #7957: [BUG] 3.3.0 regression: compiler crash in EarlyReplaceBuiltinCalls on sorted(x.attr or []) — arg.type is None in _handle_simple_function_sorted
### Describe the bug

Cython 3.3.0 crashes compiling `sorted()` over an `or`-expression whose left side is an attribute (or method call) — `sorted(x.attr or [])`. `_handle_simple_function_sorted` dereferences `arg.type.is_pylist_type` while `arg.type` is `None`:

```
File "Cython/Compiler/Optimize.py", line 2039, in _handle_simple_function_sorted
    if arg.result_in_temp() and (arg.type is PyrexTypes.py_object_type or arg.type.is_pylist_type)
AttributeError: 'NoneType' object has no attribute 'is_pylist_type'
```

Regression: **3.2.9 and 3.1.4 compile the same code fine.**

Two lines, pure Python mode:

```python
def f(route):
    return ",".join(sorted(route.methods or []))
```

```
$ cythonize -3 t.py
t.py:2:26: Compiler crash in EarlyReplaceBuiltinCalls
```

Also crashes with the argument spelled as a method call, and inside `async def`:

```python
async def g(current_user: dict):
    return sorted(current_user.get("perm_roles") or [])
```

(The `sorted(<py_object> or <empty-literal>)` idiom is common in real code — we hit five separate occurrences across one codebase when trying 3.3.0.)

### Expected behaviour

Compiles (as on ≤3.2.9); presumably the optimizer should treat a `None` type as "unknown" and skip the list fast path.

### OS

Linux (Debian 13, docker `python:3.14-slim`)

### Python version

3.14.7

### Cython version

3.3.0 (crash) — 3.2.9, 3.1.4 OK

### Additional context

Full crash traceback (node stack + frames) available on request; the failing condition is visible at `Optimize.py:2039` — the second operand of the `or` is evaluated on `arg.type` which is `None` for this node at `EarlyReplaceBuiltinCalls` time.

mokashang:
`_handle_simple_function_sorted()` dereferences `arg.type` when
choosing between `__Pyx_PySequence_ListKeepNew` and `PySequence_List`
for a `sorted()` call. `arg.type` can validly be `None` for nodes such
as a top-level `or` expression at `EarlyReplaceBuiltinCalls` time, and
cythonizing e.g. `sorted(route.methods or [])` currently aborts with

```
File "Cython/Compiler/Optimize.py", line 2039, in _handle_simple_function_sorted
    if arg.result_in_temp() and (arg.type is PyrexTypes.py_object_type or arg.type.is_pylist_type)
AttributeError: 'NoneType' object has no attribute 'is_pylist_type'
```

This regressed in the builtin-type refactor from #7543, which replaced
the earlier `arg.type in (PyrexTypes.py_object_type, Builtin.list_type)`
membership check (silently `None`-safe) with
`arg.type.is_pylist_type` on the same branch. Cython 3.2.9 and 3.1.4
compile the reported code fine.

The fix extracts the check into a `keep_new` local and guards it with
`arg_type is not None`, so a typeless argument falls through to the
existing `PySequence_List` path — the same conservative choice the old
code made.

A regression test was added to `tests/run/builtin_sorted.pyx` covering
both shapes from the issue (attribute access, and `dict.get()` inside
an `async def`).

Closes #7957

## Testing

- `python runtests.py builtin_sorted` — 22 tests pass. Confirmed the
  new cases in the file fail on `main` (`AttributeError: 'NoneType'
  object has no attribute 'is_pylist_type'`) and pass with the fix.
- `python runtests.py -j4 Optimize builtin_sorted builtin` — 372 tests
  pass with no regressions.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,10 @@ Cython Changelog
 Bugs fixed
 ----------
 
+* A compiler crash in ``EarlyReplaceBuiltinCalls`` on ``sorted(x.attr or [])``
+  and similar patterns was fixed.
+  (Github issue :issue:`7957`)
+
 * A dangling pointer was fixed when assigning to attributes of extension types.
   (Github issue :issue:`7907`)
 
--- a/Cython/Compiler/Optimize.py
+++ b/Cython/Compiler/Optimize.py
@@ -2033,11 +2033,15 @@ class EarlyReplaceBuiltinCalls(Visitor.EnvTransform):
         else:
             # Interestingly, PySequence_List works on a lot of non-sequence
             # things as well.
+            arg_type = arg.type
+            keep_new = (
+                arg.result_in_temp()
+                and arg_type is not None
+                and (arg_type is PyrexTypes.py_object_type or arg_type.is_pylist_type)
+            )
             list_node = ExprNodes.PythonCapiCallNode(
                 node.pos,
-                "__Pyx_PySequence_ListKeepNew"
-                    if arg.result_in_temp() and (arg.type is PyrexTypes.py_object_type or arg.type.is_pylist_type)
-                    else "PySequence_List",
+                "__Pyx_PySequence_ListKeepNew" if keep_new else "PySequence_List",
                 self.PySequence_List_func_type,
                 args=pos_args, is_temp=True)
 
--- a/tests/run/builtin_sorted.pyx
+++ b/tests/run/builtin_sorted.pyx
@@ -111,6 +111,38 @@ def sorted_tuple_literal():
     return sorted((1, 3, 2) * 2)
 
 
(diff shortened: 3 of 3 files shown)


scoder:
Thanks. I think we can go one step further and use type inference instead of accepting that the type might not be known. See https://github.com/scoder/cython/commit/5bc93c8ecf219aacae5348d2fd4a44e69f430ae5

scoder:
Superseded by https://github.com/cython/cython/pull/7961

PR #7964, opened 2026-09-04
Use new GraalPy APIs for array

msimacek:
- Use new Cython-specific GraalPy APIs that I added to make the array module work
- Add a few remaning changes that avoid direct struct access to objects that are handles in GraalPy
- Remove many passing tests from the graalpy bug list
- Test spearately 25.0 (LTS, 3.12) and 25.3 (3.13)

CC @timfel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -398,7 +398,8 @@ jobs:
           - pypy-3.9
           - pypy-3.10
           - pypy-3.11
-          - graalpy25
+          - graalpy25.0  # 3.12
+          - graalpy25.3  # 3.13
         os: [ubuntu-24.04]
         backend: [c]
 
--- a/Cython/Includes/cpython/array.pxd
+++ b/Cython/Includes/cpython/array.pxd
@@ -100,7 +100,10 @@ cdef extern from *:  # Hard-coded utility code hack.
 
         cdef:
             Py_ssize_t ob_size
-            arraydescr* ob_descr    # struct arraydescr *ob_descr;
+
+        @property
+        cdef inline arraydescr* ob_descr(self) noexcept nogil:
+            return __Pyx_PyArray_Descr(self)
 
         @property
         cdef inline __data_union data(self) noexcept nogil:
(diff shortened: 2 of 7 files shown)


da-woods:
I've only had a quick skim on my phone and it broadly looks reasonable. I'll have a proper look later.

For the Cyfunction changes I wonder if it makes sense to use the Limited API structure (where it doesn't "inherit" from `PyCFunction` but instead has it's own methoddef and module):

https://github.com/cython/cython/blob/c9b44365eb309c4ee490a7a314f77d84aa743451/Cython/Utility/CythonFunction.c#L58-L64

That would avoid the need for the special Graal-only accessors I think so be a little bit cleaner.

(If it isn't completely clear what I mean then I'm happy to make that change myself rather than asking you to do it)

msimacek:
Would that have any compatibility consequences?

da-woods:
> Would that have any compatibility consequences?

I don't think so. All of CyFunction works perfectly on the Limited API (at least from 3.10+).

In non-limited API CPython we use the fact that the structure is compatible with `PyCFunction` for performance reasons. But we won't be doing that optimization on Graal anyway. 

So it really is exactly the same data, just arranged in a way that doesn't need access to Python internals.

da-woods:
My proposed replacement for the CythonFunction parts is in #7971. I think that should be simpler for both Cython and GraalPy. But tell me if you disagree.

.github/workflows/ci.yml
```diff
@@ -398,7 +398,8 @@ jobs:
           - pypy-3.9
           - pypy-3.10
           - pypy-3.11
-          - graalpy25
+          - graalpy25.0  # 3.12
```
da-woods, line 401: Is there a reason to test this version and not `25.2`? (which looks like it's also Python 3.12).

(I'm happy to believe that there is a reason - you know better than me what you're trying to do longer-term support for - but thought it was worth checking).

da-woods:
Array API and testing changes look fine to me.

msimacek:
> My proposed replacement for the CythonFunction parts is in https://github.com/cython/cython/pull/7971. I think that should be simpler for both Cython and GraalPy. But tell me if you disagree.

Looks good. I removed my cyfunction changes from this PR.

.github/workflows/ci.yml
msimacek: 25.0 is a LTS release, it's still supported and we're still somewhat able to fix major bugs. 25.2 was one of the monthly feature releases, it's already superseded by 25.3.

da-woods:
Looks like it all still works so I'll set this to merge. Thanks

PR #7966, opened 2026-09-04
Simplification of Cyfunction Limited API raising

da-woods:
We can avoid creating a Python object and reference counting the name, since we now have access to the C string directly. This lets more code be shared with the regular version

### DIFF
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -833,34 +833,24 @@ __Pyx_CyFunction_get_is_coroutine(PyObject *op_in, void *context) {
 
 static void __Pyx_CyFunction_raise_argument_count_error(PyObject *func, const char* message, Py_ssize_t size) {
 #if CYTHON_COMPILING_IN_LIMITED_API
-    PyObject *py_name = __Pyx_CyFunction_get_name(func, NULL);
-    if (!py_name) return;
-    PyErr_Format(PyExc_TypeError,
-        "%.200S() %s (%" CYTHON_FORMAT_SSIZE_T "d given)",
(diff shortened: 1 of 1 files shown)


Cython/Utility/CythonFunction.c
```diff
@@ -833,34 +833,24 @@ __Pyx_CyFunction_get_is_coroutine(PyObject *op_in, void *context) {
...
 #if CYTHON_COMPILING_IN_LIMITED_API
-    PyObject *py_name = __Pyx_CyFunction_get_name(func, NULL);
-    if (!py_name) return;
-    PyErr_Format(PyExc_TypeError,
-        "%.200S() %s (%" CYTHON_FORMAT_SSIZE_T "d given)",
-        py_name, message, size);
-    Py_DECREF(py_name);
+    const char *name = __Pyx_as_CyFunctionObject(func)->func_methoddef->ml_name;
 #else
     const char* name = ((PyCFunctionObject*)func)->m_ml->ml_name;
```
scoder, line 838: Looks like there isn't a macro yet that hides the LAPI/non-LAPI distinction for accessing the CyFunction's `PyMethodDef` pointer. Might be worth having, we need the struct in a couple of places.

da-woods: Yes I agree with that.

This was something I spotted while looking at my alternative suggestion for https://github.com/cython/cython/pull/7964. That PR did add a macro, although only in a Graal/non-Graal way so it probably could have gone further.

So I was trying to keep this minimal in the hope of not deliberately adding conflicts to that PR, and then to propose an improvement to that one with the macros.

PR #7968, opened 2026-09-04
Fix memview for nested structs

haighcam:
Currently buffmt checking is broken for some structs which have a field, that is not the first field, which is a struct that contains a struct as it's first element.

This is due a gap where `__Pyx_BufFmt_Init` will advance down nested structs on the first field, and `__Pyx_BufFmt_ProcessTypeChunk` will advance down on nested structs that are not the first field. This leads to structs that have a non-first field that has a struct with a nested struct as its first field falling through and trying to compare the nested struct type to the type of its first field.

To illustrate consider creating memory views of the following structs:
```python
cdef struct Foo:
    float a;
    float b;
    float c;

cdef struct Bar:
    Foo a;
    float b;

cdef struct Baz:
    float a;
    Foo b;

cdef struct A:
    Bar a;
    float b;

cdef struct B:
    float a;
    Bar b;

cdef struct C:
    Baz a;
    float b;

cdef struct D:
    float a;
    Baz b;
```

Memory views can be created for all except for `B` which fails with the following error:
```
ValueError: Buffer dtype mismatch, expected 'Foo' but got 'float' in 'Bar.a'
```
Which is caused by `__Pyx_BufFmt_ProcessTypeChunk` checking the field 'a' of Bar when it should be checking field 'a' of Foo (the nested struct).

To resolve this the special case in `__Pyx_BufFmt_ProcessTypeChunk` that treats complex numbers as a struct and walks into them, was extended to include normal structs as well.

Additionally a test was added to the buffmt tests that checks this specific case.

### DIFF
--- a/Cython/Utility/Buffer.c
+++ b/Cython/Utility/Buffer.c
@@ -203,12 +203,6 @@ static void __Pyx_BufFmt_Init(__Pyx_BufFmt_Context* ctx,
   ctx->is_complex = 0;
   ctx->is_valid_array = 0;
   ctx->struct_alignment = 0;
-  while (type->typegroup == 'S') {
-    ++ctx->head;
-    ctx->head->field = type->fields;
-    ctx->head->parent_offset = 0;
-    type = type->fields->type;
-  }
 }
 
 static int __Pyx_BufFmt_ParseNumber(const char** ts) {
@@ -470,7 +464,7 @@ static int __Pyx_BufFmt_ProcessTypeChunk(__Pyx_BufFmt_Context* ctx) {
     }
 
     if (type->size != size || type->typegroup != group) {
-      if (type->typegroup == 'C' && type->fields != NULL) {
+      if ((type->typegroup == 'C' || type->typegroup == 'S') && type->fields != NULL) {
         /* special case -- treat as struct rather than complex number */
         size_t parent_offset = ctx->head->parent_offset + field->offset;
         ++ctx->head;
--- a/Cython/Utility/Buffer.c
+++ b/Cython/Utility/Buffer.c
@@ -203,6 +203,12 @@ static void __Pyx_BufFmt_Init(__Pyx_BufFmt_Context* ctx,
   ctx->is_complex = 0;
   ctx->is_valid_array = 0;
   ctx->struct_alignment = 0;
+  while (type->typegroup == 'S') {
+    ++ctx->head;
+    ctx->head->field = type->fields;
+    ctx->head->parent_offset = 0;
+    type = type->fields->type;
+  }
 }
 
 static int __Pyx_BufFmt_ParseNumber(const char** ts) {
--- a/tests/buffers/buffmt.pyx
+++ b/tests/buffers/buffmt.pyx
@@ -440,6 +440,26 @@ def packed_struct_with_ndarrays(fmt):
     cdef object[PackedStructWithNDArrays] buf = MockBuffer(
         fmt, sizeof(PackedStructWithNDArrays))
 
+ctypedef struct Float2:
+    float a
+    float b
+
+ctypedef struct NestedStructFirst:
+    Float2 a
+    float b
+
+ctypedef struct NestedNestedStructSecond:
+    float a
+    NestedStructFirst b
+
+def nested_structs(fmt):
+    """
+    >>> nested_structs("T{f:a:T{T{f:a:f:b:}:a:f:b:}:b:}")
+    """
+
+    cdef object[NestedNestedStructSecond] buf = MockBuffer(
+        fmt, sizeof(NestedNestedStructSecond))
+
 
 # TODO: empty struct
 # TODO: Incomplete structs
--- a/tests/buffers/buffmt.pyx
+++ b/tests/buffers/buffmt.pyx
@@ -237,12 +237,12 @@ def alignment_string(fmt, exc=None):
     except ValueError, e:
         msg = unicode(e).replace("Big", "X").replace("Little", "X").replace("big", "X").replace("little", "X")
         if msg != exc:
-            print msg
-            print "  is not equal to"
-            print exc
+            print(msg)
+            print("  is not equal to")
+            print(exc)
         return
     if exc:
-        print "fail"
+        print("fail")
 
 
 def int_and_long_are_same():
@@ -440,12 +440,13 @@ def packed_struct_with_ndarrays(fmt):
     cdef object[PackedStructWithNDArrays] buf = MockBuffer(
         fmt, sizeof(PackedStructWithNDArrays))
 
-ctypedef struct Float2:
+ctypedef struct Float3:
     float a
     float b
+    float c
 
 ctypedef struct NestedStructFirst:
-    Float2 a
+    Float3 a
     float b
 
 ctypedef struct NestedNestedStructSecond:
@@ -454,7 +455,7 @@ ctypedef struct NestedNestedStructSecond:
 
 def nested_structs(fmt):
     """
-    >>> nested_structs("T{f:a:T{T{f:a:f:b:}:a:f:b:}:b:}")
+    >>> nested_structs("T{f:a:T{T{f:a:f:b:f:c:}:a:f:b:}:b:}")
     """
 
     cdef object[NestedNestedStructSecond] buf = MockBuffer(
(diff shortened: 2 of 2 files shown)


Cython/Utility/Buffer.c
da-woods, line 527: The other fix that seems to work for me is to change this to

```
    } else if (field->type->typegroup == 'S') {
        while (field->type->typegroup == 'S') {
          size_t parent_offset = ctx->head->parent_offset + field->offset;
          if (field->type->fields->type == NULL) continue; /* empty struct */
          field = field->type->fields;
          ++ctx->head;
          ctx->head->field = field;
          ctx->head->parent_offset = parent_offset;
        }
        break;
      } else {
```

I don't necessarily understand all this code hugely well, but that feels like the main "struct handling" code, so it might be a better place for it?

tests/buffers/buffmt.pyx
```diff
@@ -440,6 +440,27 @@ def packed_struct_with_ndarrays(fmt):
...
+    float c
+
+ctypedef struct NestedStructFirst:
+    Float3 a
+    float b
+
+ctypedef struct NestedNestedStructSecond:
+    float a
+    NestedStructFirst b
+
```
da-woods, line 455: I wonder if it's worth trying a few more combinations just to make it a little more comprehensive. Off the top of my head:

```
ctypedef struct NestedNestedStructMiddle:
    float a
    NestedStructFirst b
    float c

ctypedef struct MultipleNestedNestedStructs1:
    NestedStructFirst a
    NestedStructFirst b
    float c


ctypedef struct MultipleNestedNestedStructs2:
    float a
    NestedStructFirst b
    NestedStructFirst c


ctypedef struct MixedNestedStructs1:
   NestedStructFirst a
   Float3 b

ctypedef struct MixedNestedStructs1:
   Float3 a
   NestedStructFirst b
```

Obviously the existing poor testing isn't entirely your fault but this feels like it might give us a bit more confidence here.

da-woods (commented):
That's - this definitely looks like a reasonable bug that's worth fixing.

tests/buffers/buffmt.pyx
scoder: Might be worth writing a code generator for all possible combinations, like at the end of
https://raw.githubusercontent.com/cython/cython/46e961eeed1c4c3206dce21377edccfbb57577eb/tests/run/pycmp_typed.pyx

Cython/Utility/Buffer.c
scoder: I concur that the "special case for complex" probably isn't the right place to fix this. But yes, it's really opaque code. The yet-another-while-loop here makes me wonder whether directly nested structs are really the only issue here, or whether this change only passes the bug on to a different place that we haven't identified yet.

I think I'd feel more comfortable if we could avoid the new while-loop, remove the `break` instead to continue the existing while-loop, and then see under which conditions we need to exit it. I'm not sure that this will lead to a better (or even viable) solution, but it seems more in the spirit of the existing implementation.

haighcam: It seems reasonable to avoid re-using the complex special case, it was just the easiest way to get something working.

Adding something to the existing while loop is probably better since the comment states that it will push or pop the struct stack. I'm not convinced that avoiding the a while loop would be cleaner, since the existing while loop is focused on advancing the field pointer. You could get away with adding an extra state variable that will skip the advancing of the field pointer right after walking into a struct. Or, setting `ctx->head->field = --field` in the struct case (although this would briefly leave that pointer aimed at something outside of the field list for the relevant type info, which seems a bit iffy)

tests/buffers/buffmt.pyx
haighcam: I'm happy to add some more test cases for doubly nested structs.

I'm not sure the nested struct case warrant's it's own code generator. It might be useful to have a broader buffmt one that covers general mixtures of structs, arrays, strings and different primitives, although that's out of scope of what I'm willing to contribute for this PR.

Cython/Utility/Buffer.c
haighcam: Based on the above comments here are a few potential approaches, all pass the buffmt tests as they are in the branch.

A nested while loop, the empty struct handling is a bit clunky but we need to break out of the loop and continue on the outer loop to advance the field correctly.
```c
    while (1) {
      if (field == &ctx->root) {
        ctx->head = NULL;
        if (ctx->enc_count != 0) {
          __Pyx_BufFmt_RaiseExpected(ctx);
          return -1;
        }
        break; /* breaks both loops as ctx->enc_count == 0 */
      }
      ctx->head->field = ++field;
      if (field->type == NULL) {
        --ctx->head;
        field = ctx->head->field;
        continue;
      } else {
        while (field->type->typegroup == 'S') {
            size_t parent_offset = ctx->head->parent_offset + field->offset;
            if (field->type->fields->type == NULL) break; /* empty struct */
            field = field->type->fields;
            ++ctx->head;
            ctx->head->field = field;
            ctx->head->parent_offset = parent_offset;
        }
        if (field->type->typegroup == 'S' && field->type->fields->type == NULL) continue;
        break;
      }
    }
```

An added flag to skip advancing the field when not needed. This avoids the extra while loop but makes it a bit less clear what the whole loop is doing.
```c
while (1) {
      if (field == &ctx->root) {
        ctx->head = NULL;
        if (ctx->enc_count != 0) {
          __Pyx_BufFmt_RaiseExpected(ctx);
          return -1;
        }
        break; /* breaks both loops as ctx->enc_count == 0 */
      }
      if (nested_struct) {
        nested_struct = false;
      } else {
        ctx->head->field = ++field;
      }
      if (field->type == NULL) {
        --ctx->head;
        field = ctx->head->field;
        continue;
      } else if (field->type->typegroup == 'S') {
        size_t parent_offset = ctx->head->parent_offset + field->offset;
        if (field->type->fields->type == NULL) continue; /* empty struct */
        field = field->type->fields;
        ++ctx->head;
        ctx->head->field = field;
        ctx->head->parent_offset = parent_offset;
        nested_struct = true;
        continue;
      } else {
        break;
      }
    }
```

Pointer stuff, this avoids the extra loop and extra flag. It should be safe since the pointer will always be incremented again before it is accessed, but it might need an extra comment to make it clear what is going on with `ctx->head->field = --field`.
```c
    while (1) {
      if (field == &ctx->root) {
        ctx->head = NULL;
        if (ctx->enc_count != 0) {
          __Pyx_BufFmt_RaiseExpected(ctx);
          return -1;
        }
        break; /* breaks both loops as ctx->enc_count == 0 */
      }
      ctx->head->field = ++field;
      if (field->type == NULL) {
        --ctx->head;
        field = ctx->head->field;
        continue;
      } else if (field->type->typegroup == 'S') {
        size_t parent_offset = ctx->head->parent_offset + field->offset;
        if (field->type->fields->type == NULL) continue; /* empty struct */
        field = field->type->fields;
        ++ctx->head;
        ctx->head->field = --field;
        ctx->head->parent_offset = parent_offset;
        continue;
      } else {
        break;
      }
    }
```

I lean towards either the fist or last option, but would be interested in any feedback.

da-woods: Personally I think I prefer the first. Could also use `goto` instead of `break/continue` to make it clearer where the `continue` is going to.

I'd missed that I'd changed the meaning of `if (field->type->fields->type == NULL) continue; /* empty struct */` in my quick test. Thanks for spotting that.

tests/buffers/buffmt.pyx
da-woods: I think for the sake of this PR some targeted testing of doubly-nested structs is fine and we can add the more comprehensive code generation ourselves if we want to later. It's reasonable for you to have limits to the amount of time you want to put into this.

da-woods:
I also think it might be good if this missed the 3.3.1 release. Just because this isn't a new bug and that release should probably mostly be bugfixes to 3.3.0. (That's a comment for when it gets merged rather than a suggestion that you should do something)

Cython/Utility/Buffer.c
scoder: I prefer the first one, too.
And I second the recommendation for `goto` in the empty struct case for better readability.

EDIT: The "empty struct" condition can move one line up, I think. No need to calculate the `parent_offset` first.

PR #7971, opened 2026-09-05
Make GraalPy use cyfunction Limited API structure

da-woods:
This avoids them having to write accessor functions to reach inside PyCFunctionObject.

Use macros to avoid repeating the `MethodDef` and `Module` access.

### DIFF
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -37,13 +37,18 @@ if (likely(__pyx_CyFunction_init($module_cname) == 0)); else
 #define __Pyx_CyFunction_GetClosure(f) \
     ((__Pyx_as_CyFunctionObject(f))->func_closure)
 
-#if CYTHON_COMPILING_IN_LIMITED_API
+#if CYTHON_COMPILING_IN_LIMITED_API || CYTHON_COMPILING_IN_GRAAL
   #define __Pyx__CyFunction_GetClassObj(f) \
       ((f)->func_classobj)
+  #define __Pyx_CyFunction_MethodDef(cyfunc) ((cyfunc)->func_methoddef)
+  #define __Pyx_CyFunction_Module(cyfunc) ((cyfunc)->func_module)
 #else
   #define __Pyx__CyFunction_GetClassObj(f) \
-      ((PyObject*) ((PyCMethodObject *) (f))->mm_class)
+      (((PyCMethodObject *) (f))->mm_class)
+  #define __Pyx_CyFunction_MethodDef(cyfunc) (((PyCFunctionObject*)(cyfunc))->m_ml)
+  #define __Pyx_CyFunction_Module(cyfunc) (((PyCFunctionObject*)(cyfunc))->m_module)
 #endif
+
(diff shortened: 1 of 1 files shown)


da-woods:
This is my proposed replacement for the CyFunction bits of #7964. My claim is that it's better because it reduces the amount of `#ifdef GRAALPY` code that we need, and also reduces the amount of special Cython-helper functions that they need to support.

It's only really for CPython where we benefit from using the same structure for our functions and Python C functions as an optimization for optimizations. The Limited API version has full functionality so they don't lose anything by using it.

Cython/Utility/CythonFunction.c
```diff
@@ -37,15 +37,20 @@ if (likely(__pyx_CyFunction_init($module_cname) == 0)); else
 #define __Pyx_CyFunction_GetClosure(f) \
     ((__Pyx_as_CyFunctionObject(f))->func_closure)
 
-#if CYTHON_COMPILING_IN_LIMITED_API
+#if CYTHON_COMPILING_IN_LIMITED_API || CYTHON_COMPILING_IN_GRAAL
   #define __Pyx__CyFunction_GetClassObj(f) \
       ((f)->func_classobj)
+  #define __Pyx_CyFunction_MethodDef(cyfunc) ((cyfunc)->func_methoddef)
+  #define __Pyx_CyFunction_Module(cyfunc) ((cyfunc)->func_module)
```
scoder, line 46: Might be worth a comment that these macros are expected to be lvalues (thus the missing `Get` in their name).

scoder (approved):
Simplifies a lot of code. Nice.

da-woods:
I'll wait for confirmation from the GraalPy people that this doesn't cause unexpected problems for them before merging it. I'm pretty sure it is good. But it would be a bit unhelpful to make a change for their benefit that actually makes things worse.

Cython/Utility/CythonFunction.c
```diff
@@ -65,11 +72,13 @@ typedef struct {
...
 #endif
-#if (CYTHON_COMPILING_IN_LIMITED_API || CYTHON_COMPILING_IN_PYPY) && CYTHON_VECTORCALL
+#if (CYTHON_COMPILING_IN_LIMITED_API || CYTHON_COMPILING_IN_PYPY || CYTHON_COMPILING_IN_GRAAL) && CYTHON_VECTORCALL
     __pyx_vectorcallfunc func_vectorcall;
 #endif
-#if CYTHON_COMPILING_IN_LIMITED_API
+#if CYTHON_COMPILING_IN_LIMITED_API || CYTHON_COMPILING_IN_GRAAL
     PyObject *func_weakreflist;
+    // No-args super() class cell
+    PyObject *func_classobj;
```
scoder, line 81: Moving this field up changes much of the struct layout. I know that we only share our extension types per Cython version, but this seems something I wouldn't expect in a point release.

da-woods: Reverted. I think it should be harmless but it also doesn't give us anything except one less pre-processor guard.

msimacek:
Looks good to me

PR #7972, opened 2026-09-05
Fix invalid C names in fused-function API headers

ardiloot:
Exporting a fused function with `cdef api` in Cython 3.3.0 produces an invalid C API header. Cython translation succeeds, but C/C++ code that includes the generated header fails to compile.

This regressed in #7656: type-based dictionary keys such as `identity[int]` are also emitted as C identifiers and macro names. Before that change, the header used valid names such as `__pyx_fuse_0identity`.

Keep the existing C identifiers in the header, but look up functions using the new type-based keys first, falling back to the older keys. This matches normal `cimport` behavior and preserves both the lookup improvements from #7656 and compatibility with older modules.

Includes regression tests for C and C++ header consumers, both export-name formats, preference for the new keys, reordered fused types, and missing or incompatible API entries.

### DIFF
--- a/Cython/Compiler/ModuleNode.py
+++ b/Cython/Compiler/ModuleNode.py
@@ -501,9 +501,10 @@ class ModuleNode(Nodes.Node, Nodes.BlockNode):
             h_code.putln("")
             for entry in api_funcs:
                 type = CPtrType(entry.type)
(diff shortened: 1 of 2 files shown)


scoder:
Thanks. Looks good to me. I'll leave the final word to @da-woods.

da-woods:
Thanks. Yes looks good to me too.

PR #7974, opened 2026-09-06
gate pypy workarounds on python version

mattip:
These work-arounds are not needed on PyPy 3.12. In fact PyPy 3.12 will no longer define macros to mangle limited API functions, they will be supported just as on CPython. This is part of an effort to allow PyPy to import cp12-abi3 wheels. As-is, the use of these work-arounds causes some tests to fail.


da-woods:
Thanks - will set it to merge when things pass

PR #7976, opened 2026-09-07
Avoid leaking function-type references on repeated initialization

cxzhong:
Repeated calls to `__Pyx_InitAfterSharedUtility()` currently acquire additional owned references to the cached CyFunction and FusedFunction types and overwrite the references already held in module state. This patch makes the two type initializers return early once their respective module-state slots are populated. A standalone regression test reports `(10, 10)` extra references before the fix and `(0, 0)` afterward.

### Trigger and ownership problem

This came up while reviewing the Cython 3.3 compatibility work in sagemath/sage#42742, specifically https://github.com/sagemath/sage/pull/42742#discussion_r3948297972.

#7556 (eee7a607684485e2feb0177da4ea621daf9f1d5c) moved these two initializers from `__Pyx_InitGlobals()` into `__Pyx_InitAfterSharedUtility()`, after shared utility imports and extension type initialization. This ordering also applies to builds without a separate shared utility module.

Sage's custom type-ready hook can execute Cython code that creates functions while extension types are being initialized. Its 3.3 workaround therefore calls the helper before invoking the hook. Those early calls can occur for multiple extension types, followed by the generated module initializer's own call.

Each initializer obtains a new owned reference to a shared type, then assigns it to `mstate->__pyx_CyFunctionType` or `mstate->__pyx_FusedFunctionType` without releasing the previous reference. The ordinary implementation fetches the cached type; the shared-utility implementation obtains and validates it from the shared module. Both have the same ownership issue when called again.

Normal generated module initialization calls this helper once and does not have the repeated-call trigger. The observed effect is extra references retained on cached types, rather than allocation of a new type on every call. The initialization timing change exposed this downstream use; the initializers' assumption of a single call predates 3.3.

### Change

Add an early return inside `__pyx_CyFunction_init()` and `__pyx_FusedFunction_init()` when the corresponding module-state slot is non-NULL. Each function has an ordinary implementation and a shared-utility-backed implementation, so there are four guards in total. The common initialization expressions remain unchanged. The ordinary FusedFunction guard runs before allocating its bases tuple.

The initializer itself now owns the repeated-call check, so both helper calls and direct calls preserve the existing owned reference.

The existing error path remains intact. If CyFunction initialization succeeds but FusedFunction initialization fails, another invocation can retry the missing type without reacquiring CyFunction. Module-state cleanup clears these slots, so no process-global once flag is needed.

This is an internal ownership fix for the two function-type initialization steps. It does not establish the private helper as a public API or promise arbitrary initialization ordering or concurrent/reentrant calls.

### Reproduction and validation

The new `tests/run/cyfunction_init.pyx` requires no Sage installation. It holds a normal function and a fused function, records their type reference counts, and checks two paths: invoking the helper ten times and directly invoking each type initializer ten times. Neither path may change either count. It then calls the normal function and both the integer and floating-point fused specializations. Reference-count assertions follow the existing test suite convention of excluding PyPy and GraalVM.

On unpatched master `82dc8be77701dce30cc85c7d388ccded559879b7`, the helper regression fails in both C and C++ with `(10, 10)` instead of `(0, 0)`. As an additional control, the previous revision with checks only in the initialization expressions passes the helper case but fails the new direct-call case with `(10, 10)` in both ordinary and shared-utility builds, for both C and C++. With the checks inside the initializers, both paths return `(0, 0)`.

Validation of the current patch:

- CPython 3.12.13 and 3.14.4: `cyfunction_init`, `cyfunction`, `cyfunction_defaults`, `fused_def`, and `fused_cpdef` pass in C and C++ (138 runner tests per Python version).
- CPython 3.12.13 with `--shared-utility`: both regression paths pass in C and C++ (4 runner tests). The generated code was checked to use `CythonFunctionFromSharedModule` and `FusedFunctionFromSharedModule`.
- `git diff --check` passes.

Commands, run from the checkout with each Python interpreter:

```sh
CFLAGS=-O0 python runtests.py -vv --no-unit --no-doctest --no-pyregr --no-examples \
  'run\.(cyfunction_init|cyfunction|cyfunction_defaults|fused_def|fused_cpdef)$'
CFLAGS=-O0 python runtests.py -vv --shared-utility cyfunction_init
```

The local validation is targeted; the full Cython test suite has not been run.

### DIFF
--- a/Cython/Utility/CythonFunction.c
+++ b/Cython/Utility/CythonFunction.c
@@ -14,8 +14,7 @@ Py_CLEAR(clear_module_state->__pyx_CyFunctionType);
 //@substitute: naming
 //@init_block: init_after_shared_utility
 
-if (likely(CGLOBAL(__pyx_CyFunctionType) != NULL ||
-           __pyx_CyFunction_init($module_cname) == 0)); else
+if (likely(__pyx_CyFunction_init($module_cname) == 0)); else
 
 //////////////////// CythonFunctionPerModule.proto //////////////////////////
 // This section always gets included whether we're using CythonFunction through
@@ -224,6 +223,9 @@ static CYTHON_INLINE void __Pyx_CyFunction_SetAnnotationsDict(PyObject *func, Py
 // Cutdown version that's only used when importing from shared utility code
 static int __pyx_CyFunction_init(PyObject *module) {
     $modulestatetype_cname *mstate = __Pyx_PyModule_GetState(module);
+    if (mstate->__pyx_CyFunctionType != NULL) {
+        return 0;
+    }
 
     PyTypeObject *tp = __Pyx_Get_CyFunction_Type();
     if (!tp) return -1;
@@ -1449,6 +1451,9 @@ static PyType_Spec __pyx_CyFunctionType_spec = {
 
 static int __pyx_CyFunction_init(PyObject *module) {
     $modulestatetype_cname *mstate = __Pyx_PyModule_GetState(module);
+    if (mstate->__pyx_CyFunctionType != NULL) {
+        return 0;
+    }
     mstate->__pyx_CyFunctionType = __Pyx_FetchCommonTypeFromSpec(
         mstate->__pyx_CommonTypesMetaclassType, module, &__pyx_CyFunctionType_spec, NULL);
     if (unlikely(mstate->__pyx_CyFunctionType == NULL)) {
@@ -1536,8 +1541,7 @@ Py_CLEAR(clear_module_state->__pyx_FusedFunctionType);
 //@substitute: naming
 //@init_block: init_after_shared_utility
 
-if (likely(CGLOBAL(__pyx_FusedFunctionType) != NULL ||
-           __pyx_FusedFunction_init($module_cname) == 0)); else
+if (likely(__pyx_FusedFunction_init($module_cname) == 0)); else
 
 //////////////////// FusedFunctionPerModule.proto ////////////////
 //@requires: CythonFunctionPerModule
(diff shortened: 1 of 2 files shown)


Cython/Utility/CythonFunction.c
```diff
@@ -14,7 +14,8 @@ Py_CLEAR(clear_module_state->__pyx_CyFunctionType);
 //@substitute: naming
 //@init_block: init_after_shared_utility
 
-if (likely(__pyx_CyFunction_init($module_cname) == 0)); else
+if (likely(CGLOBAL(__pyx_CyFunctionType) != NULL ||
+           __pyx_CyFunction_init($module_cname) == 0)); else
```
scoder: Why is this better than adding an initialised check inside of `__pyx_CyFunction_init()` ?

da-woods:
This is a relatively simple change so I don't really mind it but I also don't want to commit to handling this kind of repeated initialization correctly. I know there's all kinds of things that I've written in the initialization code that make assumptions based on knowing where they're called from and knowing that they're only called once per module-state or once ever.

I also suspect there might be thread-safety issues here (especially with freethreading but probably with the GIL too). I don't think we should fix them (because it complicates what should be simple code). But it's worth bearing in mind that it may be fragile.

cxzhong:
@da-woods Agreed. I'd like to keep this limited to the reference-ownership fix, without extending the supported initialization contract.

I followed up using free-threaded CPython 3.14.4, with the GIL verified disabled. Concurrent first initialization of the same state can indeed still acquire multiple owned references and overwrite the slot. In an isolated test using the generated initializers and real Cython function types, 32 threads attempted to initialize each of 100 fresh states. The reference-count increases were 103 for CyFunction and 102 for FusedFunction, instead of the expected 100; sequential controls both gave 100. The initializer bodies and reference providers were unchanged, with no counters or delays inserted into them. Counts were measured after all worker threads had exited and GC had run.

I haven't shown that normal Sage imports can reach this condition. I did not observe the same issue in the concurrent-import fixtures I tested, covering both the same module and different modules with early helper calls from a type-ready hook. These were isolated fixtures, not a full Sage build.

The intended scope of this patch is to avoid extra references on repeated sequential calls, without providing a concurrency or reentrancy guarantee. I agree that we shouldn't add synchronization in this PR or turn the private helper into a supported API.

PR #7977, opened 2026-09-08
Shared module: ensure build directory path is stripped from source filename

rgommers:
When used as `--generated-shared=/tmp/builddir/_cyutility.c`, the result was that the build directory ended up in the generated C code, in the `__pyx_f` filename table.

This is a follow-up fix to gh-7723

There is a workaround today, which is to use `cython --working /tmp/builddir --generate-shared=_cyutility.c`. The generated code should be robust to passing the output file as a full path directly though, that's an obvious thing to do.

### DIFF
--- a/Cython/Build/SharedModule.py
+++ b/Cython/Build/SharedModule.py
@@ -139,8 +139,8 @@ def generate_shared_module(options):
     Errors.reset()
 
     dest_c_file = options.shared_c_file_path
-    pyx_file = os.path.splitext(dest_c_file)[0] + '.pyx'
     module_name = os.path.splitext(os.path.basename(dest_c_file))[0]
+    pyx_file = f'{module_name}.pyx'
 
     selected_features = None
     if options.shared_utility_features_enabled or options.shared_utility_features_disabled:
@@ -155,6 +155,7 @@ def generate_shared_module(options):
     source_desc = SharedUtilitySourceDescriptor(pyx_file)
     comp_src = Main.CompilationSource(source_desc, EncodedString(module_name), os.getcwd())
     result = Main.create_default_resultobj(comp_src, options)
+    result.c_file = dest_c_file
 
     pipeline = create_shared_library_pipeline(
         context, scope, options, result,
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,11 @@ Cython Changelog
 Bugs fixed
 ----------
 
+* Generating a shared utility module no longer embeds its destination directory
+  in the generated C code, making the output reproducible across build directories.
+  This is a follow-up fix to the one in 3.3.0, for ``--generate-shared=/path/to/_{sharedname}.c``.
+  (Github issue :issue:`7723`)
+
 * A dangling pointer was fixed when assigning to attributes of extension types.
   (Github issue :issue:`7907`)
 


Cython/Build/SharedModule.py
```diff
@@ -139,8 +139,8 @@ def generate_shared_module(options):
     Errors.reset()
 
     dest_c_file = options.shared_c_file_path
-    pyx_file = os.path.splitext(dest_c_file)[0] + '.pyx'
     module_name = os.path.splitext(os.path.basename(dest_c_file))[0]
+    pyx_file = f'{module_name}.pyx'
```
scoder, line 143: We should have at least the package path included here. If several Python packages all have their own shared Cython module under `mainpkg._shared`, then it matters which `_shared` module just complained about something.

`Utils.find_root_package_dir()` can walk up a package hierarchy, although I'm not sure how well this works for the non-file shared module.

rgommers: Hmm, I see your point, but it's not immediately clear that walking up can help here. The situation that needs handling is:
```bash
$ cython --generate-shared=/path/to/builddir/mypkg/_cyutility.c
```
and `builddir/mypkg` may be completely empty at that point - there's certainly no guarantee that there's an `__init__.py` in it.

Adding the final part of the directory would usually be right, and then it'd be `mypkg._cyutility`. But I can also imagine that there's multiple packages that will use `<pkgname>/utils/_cyutility.c`, and then we're back to non-unique names.

It's just error messages right, there's no runtime behavior like imports attached to this?

There's also `cython --module-name`, but it's rarely used. Maybe `--module-name` should be mandatory if there's no way to disambiguate?

I'll have to think a bit more about this.

rgommers: Other thought: the name is completely misleading anyway, because the file never exists in the user's project. Maybe it should be `__cy_generated_{module_name}.pyx` to not send the user on a wild goose chase for this file?

scoder: > It's just error messages right, there's no runtime behavior like imports attached to this?

The shared module has no package dependencies itself. It only depends on the Cython ABI module, but that is globally shared per interpreter (by ABI version).

Another issue could be the entry in `sys.modules`. That should reflect the FQMN (Python's import system should take care of that), but it seems weird if the module thinks it's something different than what refers to it from `sys.modules`.

I remember arguing in length for making the module aware of its full package name when designing this feature, and IIRC it should work with `cythonize()` in a `setup.py` script. My take was that we'd pass the same `--shared` option to _all_ modules, including the shared module itself, so that it knows as well where it'll reside. `--module-name` would also work, but I actually liked the idea of just always having `--shared pkg.mdl` if you use the shared module.

matusvalo: Just my two cents:

> My take was that we'd pass the same --shared option to all modules, including the shared module itself, so that it knows as well where it'll reside

This would be not much user friendly but if we don't came with better solution I think this is reasonable.

My general take is that passing this kind of configuration is good for manual experiments (quick compilation of files) but inside project this is too much cumbersome. I propose to add support of reading configuration from `pyproject.toml` file. Why:
1. it is standard
2. everyone does it (black, mypy, pylint, ...)
3. it is user friendly.
4. can be used not only by setuptools but also meson and friends (they will basically do not care since the loading of configuration will be done by cython).

This allows project to have one central place for configuration and when `cython` command executed under project, it searches for `pyproject.toml` for configuration (this is what tools do in general). In pyproject.toml can be stored all global configurations for the project and this prevents user to unnecessary repeating passing parameters to commands. This will allow also inteligent selecting of [shared features](https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#selecting-features-to-be-shared) - during compilation cython can search which features are shared and builtin/include ones that are not shared (instead of todays plain crashing when user misses to include some shared feature to shared module)

The pyproject.toml can look like follows:
```toml
[tool.cython]
shared = "module.pyshared"
[tool.cython.options]
...
[tool.cython.directive]
...
```

The possible downside is that toml parser was included in python 3.11 hence, it requries a dependency for cython for lower python versions...

scoder: Reading configuration from `pyproject.toml` seems a good idea (and worth a feature ticket), but:
- it will have to be in addition to the existing interfaces which must keep working
- we will probably not include this in 3.3.x since it might need a few iterations to get right

Thus, we need a solution for the status quo either way to resolve this ticket.

da-woods: I'm broadly in favour of `pyproject.toml` but I wonder if it's other people's job to do it (i.e. the build backend that is typically driving cython). `setuptools` and `hatch-cython` both support `pyproject.toml` as a way of specifying Cython builds. So if we implemented it ourselves then we'd want to be sure not to contradict their own use of it.

rgommers: Reading from `pyproject.toml` definitely won't be a general solution, and we cannot use that in SciPy et al. I'm fairly certain it cannot work at all for Meson, CMake, etc. because they are completely unaware of `pyproject.toml`'s existence. It's a Python-specific file, while `cython` gets invoked from a general-purpose, Python-independent build system. Those build systems may not even have access to a TOML reader. It's not a robust idea, so I'd caution against relying on any static file and away from deterministic CLI invocations.


> This allows project to have one central place for configuration and when `cython` command executed under project, it searches for `pyproject.toml` for configuration (this is what tools do in general).

This is not what _compilers_ and other tools during _build execution_ do. Cython also doesn't have a good history here:

- the path walking for `__init__.py` files to determine module name has heuristics that turn into a surprising unwritten rule and needs explicitly working around for out-of-tree builds.
- there's a reproducibility bug (fix coming soon) for `.pxd` path handling as well.

Explicit and in a single place, `cython <flags> <input-sources> <output-sources>` is more discoverable and easier to maintain than splitting build config across multiple files. 

I'd recommend discarding any idea of `pyproject.toml` usage here. But if you do want to push it forward, please document it explicitly as a `setuptools`-focused convenience option only.

> I actually liked the idea of just always having `--shared pkg.mdl` if you use the shared module.

I like this, seems like a robust solution.

scoder: > it cannot work at all for Meson, CMake, etc. because they are completely unaware of `pyproject.toml`'s existence

But would they need to know about it? As with tools like test runners, coverage.py or linters, Cython would simply read the file if it exists and extract default parameters/directives/etc. from it, which can then be overwritten on the command line or within each source file (as usual). An existing toml file would have to be findable by `cython` when it runs, which can be achieved by starting `cython` from the project base directory or by copying `pyproject.toml` to the build directory (although it would seem strange if build tools were unable to access the project checkout).

scoder: (Either way, this isn't the right ticket to discuss this. I created https://github.com/cython/cython/issues/7978 for reading `pyproject.toml`.)

rgommers: Good idea to open a new issue, I'll answer your last question there.

rgommers: Back to the fix here. Maybe something like this, attempting to get it right while handling backwards compat in case that's not possible:

- Use `--module-name` if provided by the user
- Otherwise, attempt to derive module name from `__init__.py` path walking, similar to regular `.pyx` -> `.c`
- If no `__init__.py` is found, emit a `DeprecationWarning` to tell the user this will break in release 3.N or 4.0, and then use `_cy_generated_{module_name}`. 
    - The deprecation message should say to use `--module-name` (I think `--shared pkg.mdl` is essentially the same as `--module-name`, so no need for a separate new flag AFAICT).

I'll also note that this is no longer only about the reproducibility issue I'm trying to fix here. The latter is only `pyx_file`, which doesn't require improving `module_name`. Doing the latter is an extra bug fix for correctness rather than reproducibility.

rgommers: Any thoughts about the fix I proposed in the comment above?

matusvalo: > Any thoughts about the fix I proposed in the comment above?

Sorry for long wait. I think your proposal makes sense. @scoder do you have any objections? Does it makes sense to document this logic too? (not sure)

PR #7980, opened 2026-09-09
Restore compact-int fast path in __Pyx_PyIndex_AsSsize_t

### LINKED ISSUE #7979: [BUG] __Pyx_PyIndex_AsSsize_t ~9% slower per conversion in 3.3.0 (lost compact-int fast path)
### Describe the bug

GH-7471 (`d5ff728`, "Refactor int conversion into separate functions", first released in 3.3.0) split `__Pyx_PyLong_AsSsize_t` out of `__Pyx_PyIndex_AsSsize_t`. Splitting out the unrolled multi-digit switch is a clear code-size win, but two things regressed on the hot path:

1. **The compact-int fast path is gone from the wrapper.** 3.2.x returned `__Pyx_PyLong_CompactValue(b)` after a single `__Pyx_PyLong_IsCompact(b)` test. `__Pyx_PyLong_AsSsize_t` does not use those macros: for a single-digit int it computes `__Pyx_PyLong_DigitCount`, `__Pyx_PyLong_IsNeg` and `__Pyx_PyLong_Digits`, tests `size == 0`, branches to the size-1 case and applies the sign — where 3.2.x did one tagged read.
2. **`PyLong_CheckExact` became `PyLong_Check`.** `CheckExact` compares `Py_TYPE(b)` to `&PyLong_Type`; `Check` additionally loads `tp_flags` off the type object and tests `Py_TPFLAGS_LONG_SUBCLASS`, i.e. an extra dependent load per conversion. Widening the test is a real improvement for `bool` and int subclasses (they now skip `PyNumber_Index`) and should be kept — it just needn't be paid by exact ints.

This affects every Python-int subscript of a C array or pointer.

```cython
# cython: language_level=3
from libc.stdlib cimport calloc, free

cdef class Buf:
    cdef char* data
    def __cinit__(self):
        self.data = <char*>calloc(8192, 1)
        if self.data is NULL:
            raise MemoryError
    def __dealloc__(self):
        free(self.data)

def py_index(Buf b, int nargs, int reps):
    """One __Pyx_PyIndex_AsSsize_t per iteration."""
    cdef int r, k
    cdef object idx
    for r in range(reps):
        idx = 0
        for k in range(nargs):
            b.data[idx] = 1
            idx += 1

def py_index_two(Buf b, int nargs, int reps):
    """Two conversions per iteration."""
    cdef int r, k
    cdef object idx, data_idx
    for r in range(reps):
        idx = 0
        data_idx = 4096
        for k in range(nargs):
            b.data[idx] = b.data[data_idx]
            idx += 1
            data_idx += 4
```

Controls `c_index` / `c_index_two` are the same two functions with `cdef Py_ssize_t` cursors instead of `object`, so no conversion is emitted.

Built `-O3`, every module rebuilt from scratch with identical flags, minimum of 31 samples, five rounds with the builds interleaved in alternating order. Numbers are ns per inner iteration:

| build | `py_index` | vs 3.2.9 | `py_index_two` | vs 3.2.9 |
|---|---|---|---|---|
| 3.2.9 | 5.327 | — | 11.069 | — |
| 3.3.0 | 5.784 | **1.086** | 12.502 | **1.129** |
| 3.3.0 + compact | 5.524 | 1.037 | 11.320 | 1.023 |
| 3.3.0 + compact + `CheckExact` | 5.293 | **0.994** | 11.210 | **1.013** |

Controls stayed flat throughout (`c_index` 0.163-0.164, `c_index_two` 0.238-0.244), and the regression is directive-independent: `py_index_two` is 1.125-1.129x across `language_level=3` alone and with `binding`, `embedsignature`, `freethreading_compatible` and all three combined.

### Expected behaviour

Conversion of a small Python int should stay as cheap as in 3.2.x.

### Environment

- **OS:** Linux x86-64
- **Python version:** 3.12.3
- **Cython version:** 3.3.0 (regression also present on master)

### Additional context

Suggested patch — keep the delegation for the rare multi-digit case, restore the compact read, and test the exact type first:

```diff
 static CYTHON_INLINE Py_ssize_t __Pyx_PyIndex_AsSsize_t(PyObject* b) {
-    if (likely(PyLong_Check(b))) {
+    // Exact ints dominate real call sites and cost one pointer comparison,
+    // whereas PyLong_Check() needs a dependent load of tp_flags.
+    if (likely(PyLong_CheckExact(b)) || PyLong_Check(b)) {
+#if CYTHON_USE_PYLONG_INTERNALS
+        // Handle the overwhelmingly common single-digit case directly.
+        if (likely(__Pyx_PyLong_IsCompact(b))) {
+            return __Pyx_PyLong_CompactValue(b);
+        }
+#endif
         return __Pyx_PyLong_AsSsize_t(b);
     } else {
```

Measured above: the compact read recovers most of the gap, the `CheckExact` test recovers the rest (0.994x / 1.013x vs 3.2.9).

**Behaviour is unchanged.** 20 cases — `bool`, int subclasses, the `__index__` protocol (including one returning a subclass and one raising), 1/2/3-digit and oversized ints, `Py_ssize_t` overflow, and non-integer types — give identical results and identical exception types under 3.2.9, unpatched 3.3.0, and both patched variants.

juenglin:
Fixes #7979.

GH-7471 (3.3.0) split `__Pyx_PyLong_AsSsize_t` out of `__Pyx_PyIndex_AsSsize_t` and, in doing so, dropped the single-digit compact-int fast path and widened the type test from `PyLong_CheckExact` to `PyLong_Check` on the hot path.

This restores both:

- Test the exact type first (one pointer comparison, versus `PyLong_Check`'s extra `tp_flags` load).
- Take the tagged compact-int read before falling back to the general conversion.
- `PyLong_Check` is kept in the slow branch so int subclasses and `bool` still avoid `PyNumber_Index`.

### Measurements

Fresh builds, `-O3`, CPython 3.12.3, five interleaved rounds, ns per inner iteration of a loop doing one/two Python-int-to-`Py_ssize_t` conversions per iteration (see #7979 for the full repro):

| build | one conversion | vs 3.2.9 | two conversions | vs 3.2.9 |
|---|---|---|---|---|
| 3.2.9 | 5.327 | — | 11.069 | — |
| 3.3.0 | 5.784 | 1.086 | 12.502 | 1.129 |
| this PR | 5.293 | 0.994 | 11.210 | 1.013 |

### Correctness

Verified behaviourally identical to 3.2.9 and unpatched 3.3.0 across 20 cases: `bool`, `int` subclasses, the `__index__` protocol (including a subclass return and a raising `__index__`), 1/2/3-digit and oversized ints, `Py_ssize_t` overflow, and non-integer types. Ran the full `slicing`/`bufaccess`/`memslice`/`indexing`/`numpy_` suite against this change (53 modules x 2 backends) with no failures.

### Testing

`Cython/Compiler/Tests/TestTypeConversion.py` loads the `TypeConversions` utility code directly and asserts the compact-int fast path is present, since the fix is behaviour-preserving and a runtime `.pyx` test can't otherwise detect it. Verified it fails against the pre-fix source and passes against this one.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,11 @@ Cython Changelog
 Bugs fixed
 ----------
 
+* ``__Pyx_PyIndex_AsSsize_t()`` lost its fast path for the common case of
+  converting an in-range Python ``int`` to a C integer, making every
+  Python-object index into a C array or pointer measurably slower.
+  (Github issue :issue:`7979`)
+
 * A dangling pointer was fixed when assigning to attributes of extension types.
   (Github issue :issue:`7907`)
 
--- a/Cython/Compiler/Tests/TestTypeConversion.py
+++ b/Cython/Compiler/Tests/TestTypeConversion.py
@@ -1,0 +1,27 @@
+from Cython.Compiler import Code
+from Cython.TestUtils import TimedTest
+
+
+def _function_body(source, signature):
+    start = source.index(signature)
+    end = source.index("\n}", start)
+    return source[start:end]
+
+
+class TestPyIndexAsSsize_t(TimedTest):
+    """
+    GH-7979: __Pyx_PyIndex_AsSsize_t lost its compact-int fast path in 3.3.0.
+    """
+
+    def test_compact_int_fast_path_present(self):
+        proto, impl = Code.TempitaUtilityCode.load_as_string(
+            "TypeConversions", "TypeConversion.c")
+        source = (proto or "") + (impl or "")
+        body = _function_body(source, "__Pyx_PyIndex_AsSsize_t(PyObject* b) {")
+
+        self.assertIn(
+            "__Pyx_PyLong_CompactValue", body,
+            "missing compact-int fast path (GH-7979): %s" % body)
+        self.assertIn(
+            "PyLong_CheckExact", body,
+            "missing exact-type check before PyLong_Check() (GH-7979): %s" % body)
--- a/Cython/Utility/TypeConversion.c
+++ b/Cython/Utility/TypeConversion.c
@@ -429,7 +429,19 @@ static CYTHON_INLINE Py_ssize_t __Pyx_PyLong_AsSsize_t(PyObject* b) {
 }
 
 static CYTHON_INLINE Py_ssize_t __Pyx_PyIndex_AsSsize_t(PyObject* b) {
-    if (likely(PyLong_Check(b))) {
+    // Test the exact type first: it is a single pointer comparison, whereas
+    // PyLong_Check() needs an additional dependent load of tp_flags on the
+    // type object. Exact ints dominate real call sites (loop cursors and
+    // container indices), so pay the cheaper test for them and keep
+    // PyLong_Check() so subclasses (and bool) still avoid PyNumber_Index().
+    if (likely(PyLong_CheckExact(b)) || PyLong_Check(b)) {
+#if CYTHON_USE_PYLONG_INTERNALS
+        // Handle the overwhelmingly common single-digit case with a direct
+        // tagged read, before falling back to the general conversion.
+        if (likely(__Pyx_PyLong_IsCompact(b))) {
+            return __Pyx_PyLong_CompactValue(b);
+        }
+#endif
         return __Pyx_PyLong_AsSsize_t(b);
     } else {
         Py_ssize_t ival;


scoder:
Thanks for the PR, and for bringing this up in the first place. Hearing about visible regressions in actual user code is always helpful for us.

Rather than adding a test for the C code, I'd prefer adding a microbenchmark that shows the regression. Your ticket seems to have a good one, maybe you can integrate that into `Demos/benchmarks/bm_matrix.py` ? That suite of benchmarks currently tests the opposite, C int indexing into different Python containers (and doing optimised object operations on the results). A benchmark for indexing into a C container with Python indices would be a good addition. I suggest to call them something like `bm_matrix_carray_pyint` and the like.

It might also be worth adding the exact pylong type check back into the `FROM_PY_FUNCTION` code at the end of `TypeConversion.c`.

PR #7982, opened 2026-09-10
Shared module: sort exports for reproducible builds

rgommers:
Directory enumeration order changes _cyutility.c and compiled symbol order. Excerpt from `diffoscope` on SciPy wheels built in CI:

```diff
--- compare_output/cyutility-reference/_cyutility.c
+++ compare_output/cyutility-rebuilt/_cyutility.c
@@ -2914,2 +2914,2 @@
-/* CheckUnpickleChecksumError.export */
-static void __Pyx_RaiseUnpickleChecksumError(long checksum, long checksum1, long checksum2, long checksum3, const char *members);
+/* IsLittleEndian.proto */
+static CYTHON_INLINE int __Pyx_Is_Little_Endian(void);
@@ -2933,2 +2933,2 @@
-/* IsLittleEndian.proto */
-static CYTHON_INLINE int __Pyx_Is_Little_Endian(void);
+/* CheckUnpickleChecksumError.export */
+static void __Pyx_RaiseUnpickleChecksumError(long checksum, long checksum1, long checksum2, long checksum3, const char *members);
```

### DIFF
--- a/Cython/Build/SharedModule.py
+++ b/Cython/Build/SharedModule.py
@@ -46,7 +46,7 @@ def _iter_exports(selected_features=None):
     UtilityCode = Code.UtilityCode
     match_special = UtilityCode.get_special_comment_matcher('/')
 
-    for c_utility_file in os.listdir(Code.get_utility_dir()):
+    for c_utility_file in sorted(os.listdir(Code.get_utility_dir())):
         if not c_utility_file.endswith('.c'):
             continue
 


rgommers:
In case this needs a regression test, happy to add one of course.

scoder:
Thanks. Right, that's a source of difference. The utility code sections inside of the files should be ordered by Python dictionary order and dependency ordering in the code, but the ordering of the file system is out of our control.

Rather than having a regression test specifically for this, it would seem even more helpful to have a broader reproducibility check in our CI setup. I already started generating build diffs, although mostly to allow looking at differences between CI runs and branches. We could use this also to spot differences between the files generated for different architectures and/or Python versions of the same run (and there shouldn't be any, except for very few excluded tests, i.e. missing files).

https://github.com/cython/cython/blob/74dc927da90e4380923644623ff1789c0d6d1d87/.github/workflows/ci-job.yml#L179-L202

rgommers:
That kind of higher-level testing sounds like a good idea to me. That's how I'm finding these anyway, by testing on SciPy. I'm up to four different issues in Cython now, and I think that's all there is - at least with how SciPy uses Cython, but that's a pretty good integration test.

How about I submit PRs for each of those issues, and then after that an integration test with a small Python package with a diffing method for the generated C/C++ code that would have caught all of those issues?  I suspect that that will be complementary to the CI job you linked above. It will also be easier to understand probably; what I have in mind is something like:

1. `cd testpkg && python -m build --wheel -Cbuild-dir=builddir1`  # say on Linux with python 3.10
2. `cd testpkg && python -m build --wheel -Cbuild-dir=nested/builddir2`  # say on Windows with python 3.15
3. Grab C and C++ sources from `builddir1` and `nested/builddir2` and diff them

The kind of issues you get with, say, a generated `.pyx` that uses `--generate-shared`, does `include somefile.pxi` and uses `cimport numpy` are probably not exercised well by the current test suite.

PR #7986, opened 2026-09-11
fix: Make Cache directory creation multi-process safe

### LINKED ISSUE #7985: [BUG] Compiler-cache creation not multi-process safe
### Describe the bug

This is a related issue to the one in #7983 (observed as part of the same set of CI changes), but hopefully easier to solve.

When constructing a cache, the cache directory creation can race, because two processes can observe that the cache does not exist, and then both try and make it.

https://github.com/cython/cython/blob/7ad0fffd33a44d93e23779757ad5ed37aa0d5e0e/Cython/Build/Cache.py#L96

_No response_

### Expected behaviour

This should probably work. I think just passing `exist_ok=True` to `os.makedirs` should do the trick

### OS

_No response_

### Python version

_No response_

### Cython version

_No response_

### Additional context

_No response_

wence-:
- Closes #7985

### DIFF
--- a/Cython/Build/Cache.py
+++ b/Cython/Build/Cache.py
@@ -92,8 +92,8 @@ class Cache:
         else:
             self.path = path
         self.cache_size = cache_size if cache_size is not None else MAX_CACHE_SIZE
-        if not os.path.exists(self.path):
-            os.makedirs(self.path)
+        # Make sure the cache directory exists (and allow for concurrent creation).
+        os.makedirs(self.path, exist_ok=True)
 
     def transitive_fingerprint(
         self, filename, dependencies, compilation_options, flags=FingerprintFlags()


Cython/Build/Cache.py
```diff
@@ -93,7 +93,8 @@ def __init__(self, path, cache_size=None):
             self.path = path
         self.cache_size = cache_size if cache_size is not None else MAX_CACHE_SIZE
         if not os.path.exists(self.path):
-            os.makedirs(self.path)
+            # Guard against multiple processes racing to make the cache.
+            os.makedirs(self.path, exist_ok=True)
```
scoder: If we allow the directory to exist, there is no reason to check _if_ it exists upfront, because that is the pretty much first thing that `os.makedirs()` will do.
```suggestion
        # Make sure the cache directory exists (and allow for concurrent creation).
        os.makedirs(self.path, exist_ok=True)
```

wence-: Done, thanks

scoder:
Thanks

PR #7987, opened 2026-09-12
Rename some declarations used in Cython utility code to prevent collisions with user defined/cimported names

### LINKED ISSUE #7939: [BUG] list[T] class attribute no longer compiles with 3.3 due to auto_pickle
### Describe the bug

I'm reporting [this issue from Cartopy](https://github.com/SciTools/cartopy/issues/2722) upstream as I can't see what exactly is going wrong. The attached code is a distilled version of https://github.com/SciTools/cartopy/blob/main/lib/cartopy/trace.pyx When I run `cythonize` on the code, it results in the following compile error somewhere in Cython internals:
```
$ CYTHON_FORCE_REGEN=1 ./cythonize.py trace.pyx 
[1/1] Cythonizing /home/elliott/code/cython/trace.pyx

Error compiling Cython file:
...
    cdef cpp_list[X].const_iterator iter = v.begin()

    while iter != v.end():
        item = cython.operator.dereference(iter)
        Py_INCREF(item)
        __Pyx_PyList_SET_ITEM(o, i, item)
                              ^
list.to_py:145:30: Cannot convert 'list[T]' to Python object

Error compiling Cython file:
...
        Py_INCREF(item)
        __Pyx_PyList_SET_ITEM(o, i, item)
        cython.operator.preincrement(iter)
        i += 1

    return o
           ^
list.to_py:149:11: Cannot convert 'list[T]' to Python object
  File "/home/elliott/code/cython/./cythonize.py", line 9, in <module>
    main()
    ~~~~^^
  File "/home/elliott/code/cython/Cython/Build/Cythonize.py", line 324, in main
    ext_modules_by_basedir = _cython_compile_files(all_paths, options)
  File "/home/elliott/code/cython/Cython/Build/Cythonize.py", line 51, in _cython_compile_files
    ext_modules = cythonize(
        paths,
    ...<8 lines>...
        language=options.language,
        **options.options)
  File "/home/elliott/code/cython/Cython/Build/Dependencies.py", line 1153, in cythonize
    cythonize_one(*args)
    ~~~~~~~~~~~~~^^^^^^^
  File "/home/elliott/code/cython/Cython/Build/Dependencies.py", line 1264, in cythonize_one
    raise CompileError(None, pyx_file)
Cython.Compiler.Errors.CompileError: /home/elliott/code/cython/trace.pyx
```

The crash does not point to anything in the original code, but if I remove the class attribute, then it builds fine.

If I build with 3.2.9, then it also is fine. AFAICT, with that version, the `lines` attribute is not exposed in any way to Python, so I'm not sure what is causing a `PyList_SET_ITEM` call:
```python
>>> from cartopy.trace import LineAccumulator
>>> dir(LineAccumulator)
['__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__ne__', '__new__', '__pyx_vtable__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setstate__', '__sizeof__', '__str__', '__subclasshook__']
>>> la = LineAccumulator()
>>> dir(la)
['__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__ne__', '__new__', '__pyx_vtable__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setstate__', '__sizeof__', '__str__', '__subclasshook__']
```

```cython
# distutils: language = c++

from libcpp.list cimport list

cdef class LineAccumulator:
    cdef list[double] lines
```

### Expected behaviour

Compile works, as it does with 3.2.9.

### OS

Fedora 44

### Python version

3.14.6

### Cython version

6d669bf6e19bc497508f092fd3129d17a1573f07

### Additional context

A bisect from 3.2.9 to 3.3.0 points to 6d669bf6e19bc497508f092fd3129d17a1573f07 but I'm uncertain how that caused the problem.

scoder:
This workaround is needed because user cimports currently leak into the utility code, which really shouldn't be the case.

Closes https://github.com/cython/cython/issues/7939

### DIFF
--- a/Cython/Utility/CppConvert.pyx
+++ b/Cython/Utility/CppConvert.pyx
@@ -4,25 +4,25 @@
 #################### string.from_py ####################
 
 cdef extern from *:
-    cdef cppclass string "{{type}}":
-        string() except +
-        string(char* c_str, size_t size) except +
+    cdef cppclass __pyx_cpp_string "{{type}}":
+        __pyx_cpp_string() except +
+        __pyx_cpp_string(char* c_str, size_t size) except +
     cdef const char* __Pyx_PyObject_AsStringAndSize(object, Py_ssize_t*) except NULL
 
 @cname("{{cname}}")
-cdef string {{cname}}(object o) except *:
+cdef __pyx_cpp_string {{cname}}(object o) except *:
     cdef Py_ssize_t length = 0
     cdef const char* data = __Pyx_PyObject_AsStringAndSize(o, &length)
-    return string(data, <size_t> length)
+    return __pyx_cpp_string(data, <size_t> length)
 
 
 #################### string.to_py ####################
 
 #cimport cython
-#from libcpp.string cimport string
+#from libcpp.string cimport string as __pyx_cpp_string
 cdef extern from *:
     const Py_ssize_t PY_SSIZE_T_MAX
-    cdef cppclass string "{{type}}":
+    cdef cppclass __pyx_cpp_string "{{type}}":
         char* data()
         size_t size()
 
@@ -31,7 +31,7 @@ cdef extern from *:
     cdef object __Pyx_{{py_type}}_FromStringAndSize(const char*, size_t)
 
 @cname("{{cname.replace("PyObject", py_type, 1)}}")
-cdef inline object {{cname.replace("PyObject", py_type, 1)}}(const string& s):
+cdef inline object {{cname.replace("PyObject", py_type, 1)}}(const __pyx_cpp_string& s):
     if s.size() > <size_t> PY_SSIZE_T_MAX:
         raise MemoryError()
     return __Pyx_{{py_type}}_FromStringAndSize(s.data(), <Py_ssize_t> s.size())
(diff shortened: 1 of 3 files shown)


Cython/Utility/CppConvert.pyx
```diff
@@ -256,27 +254,31 @@ cdef object {{cname}}(const map[X,Y]& s):
 
 #################### complex.from_py ####################
 
+cimport cython
```
da-woods, line 257: Potentially this should be `cimport cython as __pyx_cython` if we're really worrying about it?

Although this would be getting into pretty unlikely conflicts. Maybe it's not worth going this far for the short-term fix

da-woods (commented):
This looks good to me. I can't think of any other bits of utility code that are obviously susceptible.

da-woods:
Looking again, I'd got it the wrong way around on my comment on the issue and the problem is (mainly?) user-defined names leaking into utility code rather than utility code names leaking out. 

The except_star utility code might be at risk because it uses `list` and `set` (which are c++ types). Although practically I doubt anyone's using it yet. A few other bits (e.g. fused function) are theoretically dodgy, although they use `tuple` and `dict` which shouldn't overlap with the bits of c++ we export (`std::tuple` exists but isn't easily wrapped with Cython).

Probably still worth going with this as the short-term fix and not worrying about except_star.

scoder:
The `except_star` code is tricky because safe use of `list()` would require access to `builtins.list` etc. as fully qualified type names, which we don't currently have. And I'd rather fix the scope lookups than add those types, so I'll not add the work-around there.

PR #7990, opened 2026-09-12
Avoid assuming Python integer powers return integers

### LINKED ISSUE #7989: [BUG] Comparing powers with negative exponents gives wrong results or segfaults
### Describe the bug

Using `min()`/`max()` on `2 ** int(exponent)` gives the wrong result when the exponent is negative. 
Comparing two of these powers with `==` also segfaults.

Both examples below work with 3.2.9 and fail with 3.3.0 and `master`.

```python
def clamp(exponent, maximum=16):
    return min(max(2 ** int(exponent), 1), maximum)


def equal_powers(left, right):
    return 2 ** int(left) == 2 ** int(right)
```

```console
$ pixi run python -m Cython.Build.Cythonize -i reproducer.py
$ pixi run python -c 'import reproducer; print(reproducer.clamp(-4))'
16
```

The comparison segfaults:

```bash
pixi run python -c 'import reproducer; print(reproducer.equal_powers(-4, -4))'
```

### Expected behaviour

In Python, `clamp(-4)` returns `1` and `equal_powers(-4, -4)` returns `True`.

### OS

Linux x86_64

### Python version

3.12.14

### Cython version

3.3.0

### Additional context

_No response_

claydugo:
Closes https://github.com/cython/cython/issues/7989

### DIFF
--- a/tests/run/powop.pyx
+++ b/tests/run/powop.pyx
@@ -175,3 +175,19 @@ def optimised_pow2_inplace(n):
     x = 2
     x **= n
     return x
+
+
+def clamp(exponent, maximum=16):
+    """
+    >>> clamp(-4)
+    1
+    """
+    return min(max(2 ** int(exponent), 1), maximum)
+
+
+def equal_powers(left, right):
+    """
+    >>> equal_powers(-4, -4)
+    True
+    """
+    return 2 ** int(left) == 2 ** int(right)
--- a/Cython/Compiler/ExprNodes.py
+++ b/Cython/Compiler/ExprNodes.py
@@ -12905,7 +12905,7 @@ class BinopNode(ExprNode):
             elif type1.is_pyunicode_ptr:
                 type1 = Builtin.unicode_type
             if type1.is_builtin_type or type2.is_builtin_type:
-                if type1 is type2 and type1 is not type_type and self.operator in '**%+|&^':
+                if type1 is type2 and type1 is not type_type and self.operator in '*%+|&^':
                     # FIXME: at least these operators should be safe - others?
                     return type1
                 result_type = self.infer_builtin_types_operation(type1, type2)


tests/run/powop.pyx
```diff
@@ -175,3 +175,19 @@ def optimised_pow2_inplace(n):
...
+    """
+    return min(max(2 ** int(exponent), 1), maximum)
+
+
+def equal_powers(left, right):
+    """
+    >>> equal_powers(-4, -4)
+    True
+    """
+    return 2 ** int(left) == 2 ** int(right)
```
scoder: It's usually not enough to test a single case since it cannot validate that the code actually does something (other than returning the expected value).
Also, to allow for type inference, it's best to assign the result to an untyped variable first.
```suggestion
def clamp(exponent, maximum=16):
    """
    >>> clamp(-4)
    1
    >>> clamp(4)
    16
    >>> clamp('3')
    8
    """
    result = min(max(2 ** int(exponent), 1), maximum)
    return result


def equal_powers(left, right):
    """
    >>> equal_powers(-4, -4)
    True
    >>> equal_powers(-4, 4)
    False
    >>> equal_powers(3, 3)
    True
    """
    result = 2 ** int(left) == 2 ** int(right)
    return result
```

scoder (approved):

scoder:
Thanks

da-woods:
Drive by comment on my phone without having looked at the PR: it seems plausible we might have made the mistake for `float` (which can give `complex`). If so it might be good to cover that too. If not then don't worry, sorry for the noise.

da-woods:
Looking at the PR: I suspect we did have the same problem with floats. I suspect you've fixed it. It would be nice to include in the test.

scoder:
I'm about to put back the obvious bit of type inference for the "one is int or float, the other is a float" case separately, adding a test as soon as this is merged. Operations that involve complex numbers are probably fine to infer complex, too. Can't think of anything else right away.

da-woods:
I was more just thinking that `pow(float, float)` can be `float` or `complex` in the same way that `pow(int, int)` can be `int` or `float` and that it might be good to make sure that was handled right too 

Edit: I mean `float**float` rather than `pow`...

da-woods:
I don't think the "sanitizers" failures are the fault of this PR (in that we are seeing something similar in other runs recently). I'm away from a keyboard for a the next week though so can't really investigate

PR #7992, opened 2026-09-13
Normalize source paths for included .pxi files

rgommers:
Derive include descriptions from the including source or include search root. This avoids build directory paths in generated code. Honor explicit logical names for sources inside the working directory as well.

Only the source file name that ends up in a built extension module is changed here; paths for file access and diagnostics are kept the same.

Diffoscope output comparing strings in an extension module in SciPy from two wheel builds with different build-directory depths:

```diff
-../scipy/cluster/hierarchy/_structures.pxi
+../../../scipy/cluster/hierarchy/_structures.pxi
```

After this change, both now use `scipy/cluster/hierarchy/_structures.pxi`.

***

AI coding tool usage declaration: I used Codex with GPT 6 to review this locally and ensure the fix is complete (I had missed the Scanning.py change, it proposed a change there that I had to whittle down to a minimal diff). 

Change validated in a scipy-release CI job. Integration test for the whole set of reproducibility fixes coming as a follow-up, as discussed in gh-7982.


rgommers:
Okay, this is incomplete - changing to Draft and debugging locally.

rgommers:
Okay, now green both locally and in CI on my fork. The first attempt also changed filename handling for `.pyx` files, causing coverage failures where the new paths dropped a directory prefix such as `src/`. The fixup commit separates include-file path handling, preserving the existing behavior for `.pyx` and `.pxd` files, while still fixing the `.pxi` issue.

rgommers:
The one failure on a single Windows job for Limited API isolated tests seems unrelated.

PR #7993, opened 2026-09-13
Set `PYTHON_GIL=0` in sanitizers tests

da-woods:
This affects the thread sanitizer test that we run in free-threaded mode. It ends up re-enabling the GIL without this flag (because we don't mark extension modules as freethreading-compatible)

For the other tests, it get gets set conditionally in ci-run.sh depending on the value of `PYTHON_VERSION`. Just setting it unconditionally seems easier to me than fighting with GitHub actions to make it conditional. In GIL builds the environment variable will just be ignored so it's harmless.


da-woods:
I'm expecting this to fail because the sanitizers tests generally seem to be failing right now. But just spotted that the to thread sanitizer test was re-enabling the GIL while looking at that

da-woods:
There looks to be specific new bugs that this throws up. Which was kind of the point of doing this, but needs investigating properly...

PR #7996, opened 2026-09-13
Allow None testing for builtin methods

### LINKED ISSUE #7991: [BUG] 3.3.0 regression: dict[str, str].get(missing_key) is typed as str, so the returned None takes str fast paths (bool/len/in give wrong results)
Since 3.3.0, the declared item type of a `dict[str, str]` is used to type the result of `.get(key)` as `str`. When the key is missing, `.get()` returns `None`, but the generated code applies the `str` fast paths to it without a `None` check. Truthiness silently returns the wrong answer, and `len()` reads out of bounds.

## Reproducer

```python
# g.py
def truth(d: dict[str, str]):
    v = d.get('missing')
    return bool(v)

  def length(d: dict[str, str]):
      v = d.get('missing')
      return len(v)

def contains(d: dict[str, str]):
    v = d.get('missing')
    return 'a' in v
```

> [!IMPORTANT]
> The bug affects not only missing keys but also `None` values:
> ```python
> def length(d: dict[str, str]):
>    v = d.get('none')
>    return len(v)  
> length({'none': None})
> ```

```python
# setup.py

setup(name='g', ext_modules=cythonize([Extension("g", ["g.py"])],
      compiler_directives={"language_level": 3}))
```

Called as `f({'a': 'b'})`, so the key is missing and `.get()` returns `None`:

| | CPython | Cython 3.2.5 | Cython 3.3.0 |
|---|---|---|---|
| `bool(v)` | `False` | `False` | **`True`** |
| `len(v)` | `TypeError` | `TypeError` | **`4294967295`** |
| `'a' in v` | `TypeError` | `TypeError` | **`True`** |

`v == 'a'` and `v + 'a'` behave correctly.

## Generated C

3.3.0, for `bool(v)` — no `None` check:

```c
Py_ssize_t __pyx_temp = __Pyx_PyUnicode_IS_TRUE(__pyx_v_v);
```

3.2.5, same line:

```c
if (__pyx_v_v == Py_None) __pyx_t_1 = 0;
else {
  Py_ssize_t __pyx_temp = __Pyx_PyUnicode_IS_TRUE(__pyx_v_v);
  ...
}
```

The assignment itself explicitly permits `None`:

```c
if (!(likely(PyUnicode_CheckExact(__pyx_t_1))||((__pyx_t_1) == Py_None) || __Pyx_RaiseUnexpectedTypeError("str", __pyx_t_1))) ...
```

so nothing raises there — only the later fast paths assume a real `str`.

## Notes

* Independent of `infer_types`; reproduces with and without it.
* Annotating `v: str | None = d.get('missing')` does not help, and neither does passing the default explicitly as `d.get('missing', None)`.
* Dropping the item type (`d: dict`) avoids it, as does an explicit `v is not None and v`.
* Presumably comes from "Declared container item types are now used by the type system" (#7288, #7798).

## Environment

Cython 3.3.0 (vs 3.2.5), CPython 3.12.3, Linux x86_64 (also observed on aarch64), gcc 15.

This hit us in production: a value read with `.get()` from a `dict[str, str] | None` passed an `if v` guard while being `None`, so `None` flowed into code that could not receive it.

matusvalo:
Fixes #7991

> [!NOTE]
> Test are missing. Not sure when I will have time adding one. If someone has spare time, help is welcome.


scoder:
This might be a rather heavy regression. It now assumes None for all Python objects rather than just the few cases that really may return None. Most methods don't.

matusvalo:
> This might be a rather heavy regression

My assumption was that test `__pyx_v_v == Py_None` is pretty cheap. OK let's see how difficult it will be to allow cython to understand `Optional[]` here.

matusvalo:
> This might be a rather heavy regression.

And also how we will manage case when key exist but contains `None` as value:
```python
def main(d: dict[str,str]):
    v = d.get('none')
    return len(v)

main({})  # returns 22285496 - reported bug
main({'none': None})  # returns 22285496 too instead of `TypeError`
```

Cython currently, expects `None` value for types not annotated as `Optional`. Hence, to handle `None` properly, we need to have full strict support of `Optional[]` which is kind of there but not everywhere supported:

```python
def main(d: dict):
    v: str = d.get('none') # v variable in cython is str | None in reality
    return len(v)
    
>>> import d
    import d
  File "d.pyx", line 7, in init d
    main({'none': None})
  File "d.pyx", line 5, in d.main
    return len(v)
TypeError: object of type 'NoneType' has no len(
```

Cython/Compiler/ExprNodes.py
```diff
@@ -6249,7 +6249,7 @@ def infer_type(self, env):
             if function.is_attribute:
                 method_obj_type = function.obj.infer_type(env)
                 if method_obj_type.is_builtin_type:
-                    result_type = Builtin.find_return_type_of_builtin_method(self.pos, env, method_obj_type, function.attribute)
+                    _, result_type = Builtin.find_return_type_of_builtin_method(self.pos, env, method_obj_type, function.attribute)
```
matusvalo, line 6252: Here I am not 100% sure what to do with allows_none flag. I did not went deep what to do here so it is possible that there is some space for improvements...

matusvalo:
TestBuiltin.py tests are failing, needs fixing...

Cython/Compiler/Builtin.py
```diff
@@ -658,14 +658,14 @@ def _generate_divmod_function(scope, argument_types):
     ),
     'set': dict(
         # Inherited from 'frozenset' above.
-        pop='I',
+        pop='Optional[I]',
```
scoder, line 661: This isn't optional, it returns the value or raises an exception.

Cython/Compiler/Builtin.py
```diff
@@ -537,7 +537,7 @@ def _generate_divmod_function(scope, argument_types):
         copy='T',
         count='Py_ssize_t',
         index='Py_ssize_t',
-        pop='I',
+        pop='Optional[I]',
```
scoder, line 540:
```suggestion
        pop='I',
```

Cython/Compiler/Builtin.py
```diff
@@ -658,14 +658,14 @@ def _generate_divmod_function(scope, argument_types):
...
     ),
     'dict': dict(
         copy='T',
         fromkeys='T',  # classmethod
         popitem='tuple[K,I]',
-        pop='I',
-        get='I',
+        pop='Optional[I]',
```
scoder, line 667: `.pop()` isn't optional. But it has two variants that allow it to return arbitrary values (when given a default value). If we can't distinguish between the two variants here, then the return value must be `object`. (We might be able to distinguish the two programmatically elsewhere but probably not here.)

Cython/Compiler/Builtin.py
```diff
@@ -701,33 +701,35 @@ def _parse_atomic_signature(builtin_type, sig: str) -> PyrexTypes.PyrexType:
...
-def _parse_signature(pos, env, builtin_type, return_signature: str) -> PyrexTypes.PyrexType:
+def _parse_signature(pos, env, builtin_type, return_signature: str) -> tuple[bool, PyrexTypes.PyrexType]:
     if '[' in return_signature:
         container_name, _, subscript_signature = return_signature[:-1].partition('[')
-        container_entry = builtin_scope.lookup(container_name)
-        container_type = container_entry.type if container_entry is not None else builtin_types[container_name]
 
-        if container_name in ('tuple', 'dict', 'frozendict'):
+        if container_name == 'Optional':
+            return True, _parse_signature(pos, env, builtin_type, subscript_signature)[1]
```
scoder, line 709: We have a type for `Optional[...]`. Can't we return that here, rather than adding a flag to the return value?

Cython/Compiler/Builtin.py
matusvalo: Unfortunatelly, no otherwise following code will break in the same way as reported in bug - e.g.:

```python
import cython

def main(d: dict[str,str]):
    v = d.pop('none')       # <------------ key found but still None returned, hence v needs to handle None value
    return len(v)

main({'none': None})
```

Cython/Compiler/Builtin.py
matusvalo: I was investigating that case. I suppose you reference `PyrexTypes.SpecialPythonTypeConstructor`. The issue I found that this type is not containing of "subscripted" type hence I can return information that type is optional but how to return information about the type itself? I didn't want to introduce complex changes for this bugfix hence, I chose "safe" and stupid approach. Or am I missing something?

Cython/Compiler/Builtin.py
matusvalo: See my comment above

Cython/Compiler/Builtin.py
matusvalo: See my comment above

Cython/Compiler/Builtin.py
scoder: That's a user (typing) error, and is in no way specific to `None`. `main({'none': 123})` exposes the same problem without using `None`. The problem is that we cannot rely on the return type of the method since it is user provided, so we should inject a type check for safety. But a type check and not a `None` check.

scoder:
IIUC, we only have an issue with a few dict methods. Here's  a proposal:

```patch
diff --git a/Cython/Compiler/Builtin.py b/Cython/Compiler/Builtin.py
index 93cde7b0b..193366628 100644
--- a/Cython/Compiler/Builtin.py
+++ b/Cython/Compiler/Builtin.py
@@ -718,15 +718,22 @@ def _parse_signature(pos, env, builtin_type, return_signature: str) -> PyrexType
     return _parse_atomic_signature(builtin_type, return_signature)
 
 
-def find_return_type_of_builtin_method(pos, env, builtin_type, method_name) -> PyrexTypes.PyrexType:
+def find_return_type_of_builtin_method(pos, env, builtin_type, method_name, args=None) -> PyrexTypes.PyrexType:
         if method_name in methods:
             return_type_signature: str = methods[method_name]
-            container_type = _parse_signature(pos, env, builtin_type, return_type_signature)
-            return container_type
+            return_type = _parse_signature(pos, env, builtin_type, return_type_signature)
+            if type_name == 'dict' and method_name in ('pop', 'get', 'setdefault'):
+                # These are special in that they accept a second 'default' argument that
+                # may determine the return type.
+                if args is None or not 1 <= len(args) <= 2:
+                    return_type = PyrexTypes.py_object_type  # unusual call
+                elif len(args) == 2:
+                    return_type = PyrexTypes.independent_spanning_type(return_type, args[1].type)
+            return return_type
     return PyrexTypes.py_object_type
 
 
index 1c4e772bf..964b9b784 100644
@@ -6247,7 +6247,9 @@ class CallNode(ExprNode):
             if function.is_attribute:
                 method_obj_type = function.obj.infer_type(env)
                 if method_obj_type.is_builtin_type:
-                    result_type = Builtin.find_return_type_of_builtin_method(self.pos, env, method_obj_type, function.attribute)
+                    method_posargs = getattr(self, 'args', None)
+                    result_type = Builtin.find_return_type_of_builtin_method(
+                        self.pos, env, method_obj_type, function.attribute, method_posargs)
                     if result_type is not py_object_type:
                         return result_type
             entry = getattr(function, 'entry', None)
@@ -6334,7 +6336,9 @@ class CallNode(ExprNode):
                 self.may_return_none = False
         elif function.is_attribute and function.obj.type.is_builtin_type:
             method_obj_type = function.obj.type
-            result_type = Builtin.find_return_type_of_builtin_method(self.pos, env, method_obj_type, function.attribute)
+            method_posargs = getattr(self, 'args', None)
+            result_type = Builtin.find_return_type_of_builtin_method(
+                self.pos, env, method_obj_type, function.attribute, method_posargs)
             self.may_return_none = result_type is py_object_type
             if result_type != self.type:
                 if not result_type.is_pyobject and result_type.equivalent_type:
```

matusvalo:
> IIUC, we only have an issue with a few dict methods. Here's a proposal:

Thanks! It seems to fix the issue. I added tests from this PR and created a PR for that: #8003

PR #7998, opened 2026-09-14
Respect annotation_typing for local variable annotations

### LINKED ISSUE #7994: [BUG] annotation_typing=False ignored for annotated local dict assigned a defaultdict
### Describe the bug

A local `dict[...]` annotation is enforced as an exact `dict` at runtime even when `annotation_typing=False`. Assigning a `collections.defaultdict` works in ordinary Python, but the compiled module raises `TypeError` when a comprehension accesses the value. I reproduced it with the global compiler directive, the file-level directive, and both directives together.

Python and mypy accept defaultdict -> dict casting because defaultdict subclasses dict. Cython appears to have an optimized dict implementation and when the code reaches .items(), it checks for an exact dict, raising the exception.

`repro.py`:

```python
# cython: annotation_typing=False

from collections import defaultdict


def reproduce():
    values: dict[str, dict[str, int]] = defaultdict(lambda: defaultdict(int))
    values["target"]["severity"] += 1
    return [counts for _, counts in values.items()]
```

`setup.py`:

```python


setup(
    ext_modules=cythonize(
        [Extension("repro", ["repro.py"])],
        compiler_directives={"annotation_typing": False, "language_level": "3"},
    )
)
```

Build and run:

```console
$ python setup.py build_ext --inplace
$ python -c 'import repro; print(repro.reproduce())'
  File "repro.py", line 9, in repro.reproduce
    return [counts for _, counts in values.items()]
TypeError: Expected dict, got collections.defaultdict
```

Running the source as ordinary Python succeeds:

```console
$ python -c 'import runpy; namespace = runpy.run_path("repro.py"); print(namespace["reproduce"]())'
[defaultdict(<class 'int'>, {'severity': 1})]
```

The issue is still present on master as of `dbbbdbb97f116c46495b5ab88ba0195a277016f9`

### Expected behaviour

With `annotation_typing=False`, Cython should ignore the local `dict[...]` annotation and preserve Python's acceptance of the `defaultdict` subclass. The compiled function should return the same value as ordinary Python.

### OS

Ubuntu 25.10

### Python version

Python 3.14.3

### Cython version

Cython 3.3.0

### Additional context

Cython documents exact built-in types, including rejecting subclasses of an annotated `dict`, when annotation typing is enabled. The unexpected part is that the exact-type behavior remains when annotation typing is explicitly disabled.

This is superficially similar to #5246, but that issue involved Cython 0.29.33 with `distutils` and allowed the file-level directive as a workaround, whereas this reproduces with Cython 3.3.0 using `setuptools`/`cythonize` and fails with either the file-level or global directive alone.

GruffElixir:
Fixes #7994. Type inference now honors annotation_typing=False for local annotations, so runtime values such as defaultdict are not incorrectly coerced to dict. Tests: python runtests.py -vv annotation_typing annotation_typing_defaultdict.


GruffElixir:
Quick CI note: the normal compiler, docs, lint, and feature-test paths are green. The remaining reds are confined to runner/runtime jobs: pydebug couldn’t install NumPy and then hit an unrelated ufaccess.check_object_nulled_1d segfault, while the thread sanitizer aborts with 
ested bug in the same thread; the sanitizer aggregator is red as a consequence. The new annotation_typing regression passes locally. Could someone rerun those jobs on the current head? I’ll dig in if they reproduce.

PR #7999, opened 2026-09-18
Remove deferred imports from Shadow.PointerType

vfazio:
Previously, Shadow.PointerType performed a deferred import of the `Shadow.cast` function in `__init__` and `__setitem__`.

The `cast` function is defined at the module level and there are no conflicts with another `cast` function (like `typing.cast`) that would require shadowing the function.

The imports also prevented users from copying the Shadow module [0] out with a different name.

[0]: https://cython.readthedocs.io/en/latest/src/tutorial/pure.html#magic-attributes


da-woods:
Looks sensible to me. I suspect it's a remnant of something else, but it's clearly no longer useful.

vfazio:
> Looks sensible to me. I suspect it's a remnant of something else, but it's clearly no longer useful.

I couldn't tease out why it was added in https://github.com/cython/cython/commit/6e31b41f6353bd623d98aa4de40d3645965e47ba since it wasn't in the pyi file... maybe the LLM thought it was necessary for some reason but i couldn't divine why

We do use a full copy of the shadow file for some of our projects so merging this will allow us to copy it verbatim without edits.

vfazio:
lmk if there's anything else I can help with on this PR.

scoder:
Thanks

PR #8003, opened 2026-09-20
Fix handling default values in dictionary

### LINKED ISSUE #7991: [BUG] 3.3.0 regression: dict[str, str].get(missing_key) is typed as str, so the returned None takes str fast paths (bool/len/in give wrong results)

matusvalo:
Alternative to #7996

Fixes #7991

Co-authored-by: scoder <stefan_ml@behnel.de>


tests/run/type_inference_generics.py
```diff
@@ -62,6 +62,54 @@ def test_builtin_methods():
...
 
+def test_dict_handle_none():
+    """
+    >>> test_dict_handle_none()
+    """
+
+    none_dict: dict[str, str] = {'none': None}
+
+    assert 'none' in none_dict
+    none_value = none_dict.get('none')
```
scoder, line 75: Would be interesting to see if the inferred types are as expected. (etc.)
```suggestion
    none_value = none_dict.get('none')
    print(cython.typeof(none_value))
```

Cython/Compiler/Builtin.py
```diff
@@ -718,15 +718,22 @@ def _parse_signature(pos, env, builtin_type, return_signature: str) -> PyrexType
...
     container_type = builtin_type.get_container_type()
     type_name = container_type.name if container_type else builtin_type.name
     if type_name in inferred_method_return_types:
         methods = inferred_method_return_types[type_name]
         if method_name in methods:
             return_type_signature: str = methods[method_name]
-            container_type = _parse_signature(pos, env, builtin_type, return_type_signature)
-            return container_type
+            return_type = _parse_signature(pos, env, builtin_type, return_type_signature)
+            if type_name == 'dict' and method_name in ('pop', 'get', 'setdefault'):
```
matusvalo, line 730: `setdefault` miss tests... To be implemented.

Cython/Compiler/Builtin.py
```diff
@@ -718,15 +718,22 @@ def _parse_signature(pos, env, builtin_type, return_signature: str) -> PyrexType
...
+                # These are special in that they accept a second 'default' argument that
+                # may determine the return type.
+                if args is None or not 1 <= len(args) <= 2:
```
matusvalo: Is it possible to have `len(args) == 0` or `len(args) > 2`?

scoder: Probably. But then you'd normally get a runtime error. Since the only special case is for length 2 here (and for `args is None`, when a keyword argument is passed for `default`), I found it reasonable to exclude all unusual cases and not just those, but the decision may go either way.

Cython/Compiler/ExprNodes.py
```diff
@@ -6334,7 +6336,9 @@ def coerce_to_result_type(self, env, function, func_type=None):
                 self.may_return_none = False
         elif function.is_attribute and function.obj.type.is_builtin_type:
             method_obj_type = function.obj.type
-            result_type = Builtin.find_return_type_of_builtin_method(self.pos, env, method_obj_type, function.attribute)
+            method_posargs = getattr(self, 'args', None)
+            result_type = Builtin.find_return_type_of_builtin_method(
+                self.pos, env, method_obj_type, function.attribute, method_posargs)
             self.may_return_none = result_type is py_object_type
```
scoder: Hmm. This flag should still be adapted I think.

matusvalo:
OK I needed to slightly adjust the approach but the core should be there. Please be gentle with this PR :-)...

Cython/Compiler/ExprNodes.py
matusvalo: I changed return type to tuple. It is the only reasonable way I am aware of how to do that properly...

tests/run/type_inference_generics.py
```diff
@@ -62,6 +62,118 @@ def test_builtin_methods():
...
+    # get() with missing key and default as None
+    popped_missing = empty_dict.pop('missing', None)
+    try:
+        len(popped_missing)
+    except TypeError:
+        pass
+    else:
+        assert False, len(popped_missing)
+    # get() with missing key and default as None without assignment
+    print(cython.typeof(empty_dict.get('missing')) if cython.compiled else 'Python object')
```
matusvalo, line 145: This is producing Python object returning type all the time. I suppose that this is a bug but if confirmed I would prefer to deffer the fix to separate PR...

matusvalo: Also I suspect the issue is not in this PR. The issue is that the `find_return_type_of_builtin_method()` with `args` set to `None` all the time which should be wrong I suppose...

Cython/Compiler/Builtin.py
```diff
@@ -718,16 +719,35 @@ def _parse_signature(pos, env, builtin_type, return_signature: str) -> PyrexType
...
+                    return PyrexTypes.py_object_type, True
+                if len(args) == 1:
+                    # Method called without second parameter - default value is None
+                    return return_type, True
+                if len(args) == 2 and args[1].is_none:
+                    # Method called with second parameter set to None
+                    return return_type, True
+                if len(args) == 2 and args[1].type is None:
+                    # Method called with second parameter set to None
+                    return return_type, True
```
da-woods: I'm a bit suspicious of `args[1].type is None`. I think that means "second parameter type hasn't been evaluated yet" rather than "second parameter type is None"

Which probably means `py_object_type, True` because there just isn't information to tell otherwise?

tests/run/type_inference_generics.py
```diff
@@ -62,6 +62,118 @@ def test_builtin_methods():
...
+    print(cython.typeof(inserted_default_none) if cython.compiled else 'str object')
+    try:
+        len(inserted_default_none)
+    except TypeError:
+        pass
+    else:
+        assert False, len(inserted_default_none)
+    assert inserted_default_none is None, inserted_default_none
+    assert empty_dict['none_missing'] is None, empty_dict
+
```
da-woods: A couple of cases that I don't think are covered by the tests (and which I'm not immediately sure if they'll work)

```
def pop_key(d[str, str | None], default: str):
    return d.pop('key', default)

def pop_key2(d[str, str], default: str | None):
    return d.pop('key', default)
```

Does the may be None get populated from an argument type?

Cython/Compiler/Builtin.py
```diff
@@ -718,16 +719,35 @@ def _parse_signature(pos, env, builtin_type, return_signature: str) -> PyrexType
     return _parse_atomic_signature(builtin_type, return_signature)
 
 
-def find_return_type_of_builtin_method(pos, env, builtin_type, method_name) -> PyrexTypes.PyrexType:
+def find_return_type_of_builtin_method(pos, env, builtin_type, method_name, args=None) -> tuple[PyrexTypes.PyrexType, bool]:
```
```suggestion
def find_return_type_of_builtin_method(pos, env, builtin_type, method_name, args=None) -> tuple[PyrexTypes.PyrexType, bool]:
    # Returns a tuple: (return type, may be None)
```
Alternatively, we could use a named enum for the flag to make it readable, but frankly, it's sad that there is no type that explicitly says "builtin type or None". I'd keep it to a comment for now.

Cython/Compiler/Builtin.py
```diff
@@ -718,16 +719,35 @@ def _parse_signature(pos, env, builtin_type, return_signature: str) -> PyrexType
...
+                if len(args) == 2 and args[1].is_none:
+                    # Method called with second parameter set to None
+                    return return_type, True
+                if len(args) == 2 and args[1].type is None:
+                    # Method called with second parameter set to None
+                    return return_type, True
+                if len(args) == 2:
+                    # Method called with second parameter
+                    return_type = PyrexTypes.independent_spanning_type(return_type, args[1].type)
+                    return return_type, False
```
scoder: If the method is called with an unexpected number of arguments, I don't think we should make assumptions about the return type.
```suggestion
                if len(args) > 2:
                    # Unusual call, possibly a runtime error.
                    return PyrexTypes.py_object_type, True
                # Method called with 'default' parameter.
                default_type = args[1].type or args[1].infer_type(env)
                return_type = PyrexTypes.independent_spanning_type(return_type, default_type)
                return return_type, return_type is PyrexTypes.py_object_type
```

